PR #21, opened 2020-03-20
Make sure torch inputs contiguous before passing data pointer to TRT

bddppq:


core/execution/register_trt_op.cpp
```diff
@@ -11,13 +11,16 @@ namespace execution {
...
+    std::vector<at::Tensor> contig_inputs{};
+    contig_inputs.reserve(inputs.size());
     for (size_t i = 0; i < inputs.size(); i++) {
-        auto in_gpu = inputs[i].to(at::kCUDA);
-        auto shape = core::util::toDimsPad(in_gpu.sizes(), 4);
-        //in_gpu = inputs[i].reshape(core::util::toVec(shape));
-        //LOG_DEBUG("In shape:" << in_gpu.sizes() );
-        ctx->setBindingDimensions(i, shape);
-        gpu_handles.push_back(in_gpu.data_ptr());
+        auto dims = core::util::toDimsPad(inputs[i].sizes(), 4);
```
bddppq: Should here be 1 as in the test RunGraphEngine? (I assume it's because TRT doesn't have zero-dim Scalar as in PyTorch)

narendasan: It should probably be 1

narendasan: The main goal should be user facing behavior that is no different than pytorch, so if it works in pytorch it should work here. I think my main concern when i first wrote this what things like TensorRT FullyConnected which expects a 4D input but the test works fine with pad 1

core/execution/register_trt_op.cpp
```diff
@@ -11,13 +11,16 @@ namespace execution {
...
+        auto shape = core::util::toVec(dims);
+        contig_inputs.push_back(inputs[i].to(at::kCUDA).view(shape).contiguous());
```
narendasan: Good catch!

core/execution/register_trt_op.cpp
bddppq: Thanks for clarification. Yeah if there are particular layers expecting different input dimensions between TRT and PyTorch, it's better to insert a padding op in the corresponding converter (e.g. pytorch linear -> trt pad 4 + trt FC). Unconditionally padding the input will break ops that have the same input dimensions between PyTorch and TRT.

Issue #16, opened 2020-03-18
Start to handle branching in simple cases


narendasan:
The system works pretty well for traced models, but not much work has been done with torch script models that have branching. I noticed some common cases that we should be able to handle include branching for none arguments such as graphs like this:

```
  %50 : Function = prim::Constant[name="linear"]()
  %53 : bool = prim::Constant[value=0]() # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1368:7
  %54 : None = prim::Constant() # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1368:40
  %55 : int = prim::Constant[value=2]() # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1368:22
  %56 : int = prim::Constant[value=1]() # :0:0
  %57 : int = aten::dim(%input1.1) # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1368:7
  %58 : bool = aten::eq(%57, %55) # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1368:7
  %59 : bool = prim::If(%58) # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1368:7
    block0():
      %60 : bool = aten::__isnot__(%94, %54) # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1368:28
      -> (%60)
    block1():
      -> (%53)
  %input2.1 : Tensor = prim::If(%59) # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1368:4
    block0():
      %bias0.4 : Tensor = prim::unchecked_cast(%94)
      %101 : Tensor = aten::linear(%input1.1, %95, %bias0.4)
      -> (%101)
    block1():
      %106 : Tensor? = prim::Constant()
      %107 : Tensor = aten::linear(%input1.1, %95, %106)
      %67 : bool = aten::__isnot__(%94, %54) # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1373:11
      %output0.6 : Tensor = prim::If(%67) # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1373:8
        block0():
          %bias1.4 : Tensor = prim::unchecked_cast(%94)
          %output0.7 : Tensor = aten::add_(%107, %bias1.4, %56) # /usr/local/lib/python3.6/dist-packages/torch/nn/functional.py:1374:12
[...]

Issue #1, opened 2020-03-11
Module tests fail because of segfault in cuDNN destructor


narendasan:
The tests pass functionality wise, but in the clean up, the test segfaults. It seems like this is a issue others have seen in libtorch https://github.com/pytorch/pytorch/issues/17658

narendasan:
Recent issues may be fixed in PyTorch next with this https://github.com/pytorch/pytorch/pull/36416

PR #48, opened 2020-04-29
[WIP] Upgrading to LibTorch 1.5.0 (CUDA 10.2, cuDNN 7.6.5, TensorRT 7.0.0)

### LINKED ISSUE #14: Grab dependencies automatically
We should either look at containerizing the build, or find someway to pull TensorRT, CUDA etc. with bazel. We already do this with libtorch. This should make it less likely builds will fail because of peoples environment

### LINKED ISSUE #32: Run standard passes during lowering
Lowering should run standard jit passes like dead code elimination

### LINKED ISSUE #42: Version Bumps
With Libtorch 1.5.0 out we are going to be updating soon: 
- Libtorch 1.5.0
- CUDA 10.2
- cuDNN 7.6
- TensorRT 7.0

narendasan:
- Closes #42
- Closes #32
- Closes #14 : The default build now requires users to grab the tarballs
from the NVIDIA website to support hermetic builds, may look at some
methods to smooth this out later. The old method is still available

PR also contains a good amount of the work required to support scripting (#16).

Discovered that using adaptive pooling with dynamic shape does not work. This will have to be a limitation of the system until TensorRT has the ability to configure pooling window sizes at runtime. 

#1 is back because of a bug in PyTorch, fix is already in pytorch master, but all tests have been verified.


alanzhai219:
nice upgrade. so when will your team merge the 1.5 into master? I think pytorch1.5 jit coding-style is much good.

narendasan:
It will probably be a couple days, we need to address a bunch of changes that PyTorch has added to how they generate the IR.

alanzhai219:
FB released 1.5 version for a couple days and I just skim through the jit code, not looking into it deeply. I think the jit mechanism isn't changed much and jit codes have been reconstructed
 better. I remember a Nvidia engineer wrote a blog explaining how jit works in pytorch-1.3, which may be helpful.

PR #97, opened 2020-06-12
C++ API Refactor

narendasan:
# Description

This PR decouples the header files so now you need to include specific headers for logging and post training quantization. This lets the main header keep being independent of dependencies on Torch or TensorRT headers but in the case of PTQ it pulls it in. This seems like a good compromise with minimal impact on the user workflow since you will need to include these headers anyway to use PTQ but you wouldn't in the case of the main header. 

Fixes # (issue)

## Type of change

# Checklist:


narendasan:
This PR will hold until #92 is merged since we will likely bump the version after merging this PR considering the breaking change in the API

PR #147, opened 2020-07-17
(//core): Add support for DLA runs on NVIDIA AGX platform

andi4191:

# Description
Added the support for DLA on the NVIDIA Jetson AGX platform.

Fixes # (issue)

## Type of change

# Checklist:

- [x] I have made corresponding changes to the documentation and have regenerated the documentation (`make html` in docsrc)


cpp/api/include/trtorch/trtorch.h
```diff
@@ -306,16 +307,45 @@ struct TRTORCH_API ExtraInfo {
...
+
+	/*
+	 * Target gpu id
+	 */
+	uint64_t gpu_id;
+
+        /*
+         * When using DLA core on NVIDIA AGX platforms gpu_id should be set as Xavier device
+         */
+	int dla_core;
```
narendasan, line 329: Spacing alignment and do we want to call it dla_id so its the same as gpu_id? or is dla_core standard nomenclature?

cpp/api/include/trtorch/trtorch.h
```diff
@@ -306,16 +307,45 @@ struct TRTORCH_API ExtraInfo {
...
+        /**
+         * (Only used when targeting DLA (device))
+         * Lets engine run layers on GPU if they are not supported on DLA
+         */
+	bool allow_gpu_fallback;
+
+	Device() :
+		device_type(DeviceType::kGPU),
+		gpu_id(0),
+		dla_core(-1),
```
narendasan, line 340: Can we just set default DLA core to 0 instead of -1 and set the type to uint64_t since dla_core wont be set unless device type is DLA right?

py/trtorch/csrc/trtorch_py.cpp
```diff
@@ -196,8 +196,10 @@ PYBIND11_MODULE(_C, m) {
     .def_readwrite("refit",                &ExtraInfo::refit)
     .def_readwrite("debug",                &ExtraInfo::debug)
     .def_readwrite("strict_types",         &ExtraInfo::strict_types)
-    .def_readwrite("allow_gpu_fallback",   &ExtraInfo::allow_gpu_fallback)
-    .def_readwrite("device",               &ExtraInfo::device)
+    .def_readwrite("allow_gpu_fallback",   &ExtraInfo::device::allow_gpu_fallback)
```
narendasan, line 199: Is there  Python object that we use here to create a device? what is the API look like for the user?

cpp/trtorchc/main.cpp
```diff
@@ -240,10 +243,17 @@ int main(int argc, char** argv) {
     if (device_type) {
         auto device = args::get(device_type);
         std::transform(device.begin(), device.end(), device.begin(), [](unsigned char c){ return std::tolower(c); });
+	if (gpu_id) {
+	    compile_settings.device.gpu_id = args::get(gpu_id);
+	    cudaSetDevice(compile_settings.device.gpu_id);
```
narendasan, line 248: Is it possible to do this in ConversionCtx instead of here since it will not cover any other interface to the compiler?

cpp/api/include/trtorch/trtorch.h
andi4191: I tried aligning it with TensorRT's DLA nomenclature.

cpp/api/include/trtorch/trtorch.h
andi4191: yes, we can do that. I will update it

py/trtorch/csrc/trtorch_py.cpp
andi4191: I added a struct to ExtraInfo
https://github.com/andi4191/TRTorch/blob/anuragd/dla_aarch64_support/py/trtorch/csrc/trtorch_py.cpp#L104

narendasan: How would this get used though? Can you add a test in the python_api_tests to demonstrate?

cpp/trtorchc/main.cpp
andi4191: I aligned it with TensorRT's trtexec sample. My concern is it should be handled by the sample application rather than the internal core library. I can still change it otherwise

narendasan: I think in this case we need to handle it since it seems like we provide the bulk of application logic otherwise  the gpu_id doesnt do anything then. If we can do it we can also create an interface that allows people to use pytorch device specifiers to set the device used for compilation. We do need to consider how this will work though. For instance we should probably store the default device so post compilation we return the system back to normal for the user. We would also need to consider how this might work during runtime, if I have a system with 2 different GPUs and I have a program compiled for say TITAN V, what is the user workflow, how would they make sure that the system works properly, I am hesitant to just tell people to use the CUDA API directly. What would be ideal is if we could either configure the system correctly at runtime for the user or give them the options to do that for themselves using TRTorch APIs.

narendasan: Like if I load a TRTorch program maybe in the serialization we store device information so as it is deserialized we search for if there is an applicable GPU and set the right cuda device?

py/trtorch/csrc/trtorch_py.cpp
andi4191: ok

cpp/trtorchc/main.cpp
andi4191: I agree we need to handle it internally in that case, If a TRTorch program performs deserialization then it should have the required information about the device as well otherwise, it will be a hassle for end-user. Let me look closer to the serialization/deserialization workflow and try to add device information as well.

narendasan: Do we have a system with 2 different GPUs that we can experiment with? Maybe a DDPX if there isnt an open workstation?

andi4191: I thought you have 2 cards on your ws. We can utilize a DDPX for this. Eventually, it will help us extend the support on Drive AGX as well.

narendasan: Both my cards are TITAN V

andi4191:
Closing this PR.
Please refer to #160 for the merge request.

PR #155, opened 2020-07-23
ITensorOrFreeze() to pull ITensor from Var

abhi-iyer:
# Description
ITensorOrFreeze(ConversionCtx* ctx, const torch::jit::Node* n) can now be used to safely pull an ITensor from a Var. This replaces the previous usage of ".ITensor()". 

Created Weights.h, which isolates struct Weight away from converters.h. Made necessary changes to BUILD files to reflect this.

Can now also create a Weight given an INT32 integer directly. 
 

## Type of change

# Checklist:


core/conversion/converters/Weights.cpp
```diff
@@ -18,16 +17,33 @@ Weights::Weights() {
...
+    this->num_output_maps = 1;
+
+    this->data.type = nvinfer1::DataType::kINT32;
+    int32_t* buf = reinterpret_cast<int32_t*>(malloc(1 * sizeof(int32_t)));
+    buf[0] = val;
+    this->data.values = buf;
+    this->data.count = 1;
+    ctx->builder_resources.push_back(buf);
+
+    this->shape.nbDims = 0;
```
narendasan, line 43: I think we should set kernel shape too just to be thorough

core/conversion/converters/Weights.cpp
```diff
@@ -18,16 +17,33 @@ Weights::Weights() {
...
+
     this->data.type = nvinfer1::DataType::kFLOAT;
     float* buf = reinterpret_cast<float*>(malloc(1 * sizeof(float)));
     buf[0] = val;
     this->data.values = buf;
     this->data.count = 1;
     ctx->builder_resources.push_back(buf);
+    
     this->kernel_shape.nbDims = 1;
     this->kernel_shape.d[0] = 1;
```
narendasan, line 29: We need to add the shape initialization here too, also lets make this a scalar vs a 1d tensor

core/conversion/converters/Weights.h
```diff
@@ -0,0 +1,34 @@
...
+#include "core/util/prelude.h"
+#include "core/conversion/conversionctx/ConversionCtx.h"
+
+namespace trtorch {
+namespace core {    
+namespace conversion {
+namespace converters {
+
+struct Weights {
+    //TODO: Rebuild this in a way that makes sense for more than just conv2/3D and linear
```
narendasan, line 12: We can remove this comment

core/conversion/converters/Weights.h
```diff
@@ -0,0 +1,34 @@
...
+    nvinfer1::Weights data;
+    nvinfer1::Dims kernel_shape;
+    nvinfer1::Dims shape;
+    int64_t num_input_maps;
+    int64_t num_output_maps;
+
+    Weights();
+    Weights(ConversionCtx* ctx, at::Tensor t);
+    Weights(ConversionCtx* ctx, float val);
+    Weights(ConversionCtx* ctx, int32_t val);
```
narendasan, line 22: Do we need on for bool too?

core/conversion/converters/Weights.h
```diff
@@ -0,0 +1,34 @@
...
+    friend std::ostream& operator<<(std::ostream& os, const Weights& w);
+};
+
+inline nvinfer1::ITensor* tensor_to_const(ConversionCtx* ctx, at::Tensor t) {
```
narendasan, line 26: This and its uses can be replaced now I think since we have `ITensorOrFreeze`

core/conversion/converters/converters.h
```diff
@@ -9,9 +9,10 @@
 #include "core/util/prelude.h"
 #include "core/conversion/var/Var.h"
 #include "core/conversion/conversionctx/ConversionCtx.h"
+#include "core/conversion/converters/Weights.h"
 
 namespace trtorch {
-namespace core {
+namespace core {
```
narendasan, line 15: Unnecessary whitespace change

core/conversion/converters/impl/batch_norm.cpp
```diff
@@ -15,7 +15,7 @@ auto batch_norm_registrations TRTORCH_UNUSED = RegisterNodeConversionPatterns()
                             Tensor? mean, Tensor? var,
                             bool training, float momentum, float eps, bool cudnn_enabled) -> (Tensor))SIG",
     [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
-      auto input = args[0].ITensor();
+      auto input = args[0].ITensorOrFreeze(ctx, n);
```
narendasan, line 18: I dont think there would be a case where there is a static tensor we are going to run through batch norm, can you think of one? On the other hand is there a downside to having this?

core/conversion/converters/impl/conv_deconv.cpp
```diff
@@ -17,7 +17,7 @@ auto conv_registrations TRTORCH_UNUSED = RegisterNodeConversionPatterns()
                                  int[] output_padding, int groups, bool benchmark,
                                  bool deterministic, bool cudnn_enabled) -> (Tensor))SIG",
         [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
-            auto in = args[0].ITensor();
+            auto in = args[0].ITensorOrFreeze(ctx, n);
```
narendasan, line 20: Same comment here

core/conversion/var/Var.cpp
```diff
@@ -85,6 +85,27 @@ std::string Var::type_name() const {
   }
 }
 
+nvinfer1::ITensor* Var::ITensorOrFreeze(ConversionCtx* ctx, const torch::jit::Node* n) {
+  TRTORCH_CHECK(isITensor() || isIValue(), "Requested either IValue or ITensor, however Var type is " << type_name());
```
narendasan, line 89: The check should be `isITensor() || isIValue() && IValue().isTensor()`. It either needs to be a ITensor or a IValue that contains a Tensor. Or are you intending to use this in the scalar case too?

core/conversion/var/Var.cpp
```diff
@@ -85,6 +85,27 @@ std::string Var::type_name() const {
   }
 }
 
+nvinfer1::ITensor* Var::ITensorOrFreeze(ConversionCtx* ctx, const torch::jit::Node* n) {
```
narendasan, line 88: Why does this take a node as input?

core/conversion/var/Var.cpp
```diff
@@ -85,6 +85,27 @@ std::string Var::type_name() const {
...
+  TRTORCH_CHECK(isITensor() || isIValue(), "Requested either IValue or ITensor, however Var type is " << type_name());
+  nvinfer1::ITensor* out;
+
+  if (isIValue()) {
+    auto weights = converters::Weights(ctx, ptr_.ivalue->toTensor());
+
+    auto const_layer = ctx->net->addConstant(weights.shape, weights.data);
+    TRTORCH_CHECK(const_layer, "Unable to freeze tensor for node: " << *n);
+    const_layer->setName((util::node_info(n) + " [Freeze Tensor]").c_str());
```
narendasan, line 97: I am okay dropping this to simplify the function declaration, we could create a random id (a good way is to just cast the pointer to an int) and then print that id and set it as the name so it would be like in the debug messages 
"Freezing tensor as const layer id: 0xDEADBEAF" and the layer name is "0xDEADBEAF [Freeze Tensor]"

core/conversion/converters/Weights.h
abhi-iyer: I removed it because a IConstantLayer doesn't accept a bool Weight/Tensor anyways. The correct-use case for creating a bool Tensor is to pass the IConstantLayer through an identity layer, and manually set the output to be of DataType::kBOOL. 

Let me know what you think.

core/conversion/converters/Weights.h
abhi-iyer: This function doesn't seem to have any uses, so I'll remove it.

core/conversion/var/Var.cpp
abhi-iyer: No, you're correct. I'll add it in. Thanks!

core/conversion/converters/impl/batch_norm.cpp
abhi-iyer: Yes, unlikely that a static tensor is passed in. Only case I can think of is if someone creates a non-random tensor in PyTorch and applies batch norm on it (unlikely, though). 

I added this in because for consistency's sake, there are no downsides to having this in. Shall I remove or keep it?

core/conversion/converters/impl/conv_deconv.cpp
abhi-iyer: See response for aten::batch_norm

core/conversion/var/Var.cpp
abhi-iyer: Used for assigning the constant layer's name

core/conversion/var/Var.cpp
abhi-iyer: done

core/conversion/var/Var.cpp
narendasan: Id rather leave the node out honestly even if it makes the error slightly less helpful (hopefully context can cover for us)

core/conversion/converters/impl/batch_norm.cpp
narendasan: I think the way I see it if they pass us a static tensor we should just do the full batch norm calculation right there since it would improve performance, I kind of want to leave it as is for layers that should not be taking in static tensors and if it is the case that we get bug reports that some are getting passed in then we should go ahead and do the computation at compile time. Maybe leave a comment that says if static tensors are getting passed in do the full calculation in the converter or something to that effect.

core/conversion/converters/Weights.h
abhi-iyer: I'm going to keep it out for now. We can add one in if we need to have a Weight (but not a Tensor) with a single bool.

core/conversion/converters/impl/unary.cpp
```diff
@@ -14,7 +14,7 @@ namespace {
           {"aten::" #unary "(Tensor self) -> Tensor",                           \
            [](ConversionCtx *ctx, const torch::jit::Node *n,                    \
               args &args) -> bool {                                             \
-             auto in = args[0].ITensor();                                       \
+             auto in = args[0].ITensorOrFreeze(ctx);                                       \
```
narendasan, line 17: nit: can you align the \'s, also this might be another case of if the tensor is static we should be doing the computation outright instead of adding a layer so we might not need this

core/conversion/converters/Weights.h
narendasan: ok

core/conversion/converters/impl/unary.cpp
abhi-iyer: done

core/conversion/var/Var.cpp
```diff
@@ -85,6 +87,32 @@ std::string Var::type_name() const {
   }
 }
 
+nvinfer1::ITensor* Var::ITensorOrFreeze(ConversionCtx* ctx) {
+  TRTORCH_CHECK(isITensor() || (isIValue() && ptr_.ivalue->isTensor()), "Requested either IValue containing a Tensor, or ITensor, however Var type is " << type_name());
```
narendasan, line 91: We might as well tell people the contained type too since the fail case is if it is an IValue but not containing a Tensor or the var is none. maybe something like
```c++
 ".. However Var type is " << type_name << (isIValue() ? "containing " << *(ptr_.ivalue->type()) : "");
```

abhi-iyer: ptr_.ivalue->type() is a c10::TypePtr, it doesn't convert to a string easily without some proper checks. 

I think I would need to have something like:

``````

Comment (narendasan) [reply core/conversion/var/Var.cpp:91]: I think that when you dereference the TypePtr it gives you a string we use this in the Var unwrapping functions too iirc

Comment (narendasan) [reply core/conversion/var/Var.cpp:91]: https://github.com/NVIDIA/TRTorch/blob/fe06d0993d1ebf9083eada67a42e4fd46c4bee92/core/conversion/var/Var_inl.h#L12

Comment (abhi-iyer) [reply core/conversion/var/Var.cpp:94]: I think printing out the contents of the IValue (if it is an IValue) every time might be useful from a debugging standpoint. I've changed this accordingly

**Comment by abhi-iyer** [comment]
> Also noticed that in the Weights constructor for Tensor we need to replace the LOG_INTERNAL_ERROR with TRTORCH_THROW_ERROR, can you add that change as well?

added

File: core/conversion/converters/Weights.cpp [hunk 14]
```diff
@@ -18,20 +17,37 @@ Weights::Weights() {
...
+    this->data.count = 1;
+    ctx->builder_resources.push_back(buf);
+
+    this->shape.nbDims = 0;
+    this->kernel_shape.nbDims = 0;
 }
 
 Weights::Weights(ConversionCtx* ctx, at::Tensor t) {
     if (t.sizes().size() > nvinfer1::Dims::MAX_DIMS) {
         //TODO: Handle this with exceptions or whatever
```
narendasan, line 49: nit remove these comments as well

abhi-iyer: done

abhi-iyer:
> Are there pending changes on this PR?

No other changes from my side

core/conversion/converters/converters.h.orig
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
narendasan, line 1: Do we still need this file?

core/conversion/converters/impl/activation.cpp
```diff
@@ -10,7 +10,7 @@ namespace {
 
 #define convert(act, trt_type)                                                 \
   bool act(ConversionCtx* ctx, const torch::jit::Node* n, args& args) {        \
-    auto in = args[0].ITensor();                                               \
+    auto in = args[0].ITensorOrFreeze(ctx);                                               \
```
narendasan, line 13: nit whitespace

core/conversion/converters/converters.h.orig
abhi-iyer: odd, not sure how this file got into the commit - I made sure to leave it out. we can delete it, I'll do that

core/conversion/converters/impl/activation.cpp
abhi-iyer: done

narendasan:
Last two commits werent signed properly, you can amend them with `git rebase -i` rewording the last two commit messages

abhi-iyer:
> Last two commits werent signed properly, you can amend them with `git rebase -i` rewording the last two commit messages

Yep, sorry, done

narendasan:
Last commit also was not signed properly

narendasan:
I would rebase and force push on the branch instead of merging

abhi-iyer:
Should be good now. Let me know if I need to do anything else.

narendasan:
Im going to squash this PR this seems like theres a bunch of stuff to fix

narendasan:
Branch was merged in manually

PR #160, opened 2020-07-30
(core//): Added DLA support for NVIDIA platforms

andi4191:

# Description

Changes include:
1. Device selection API introduced (For multi-GPU systems)
2. DLA support for NVIDIA Jetson AGX platform
3. Test script for API validation on DLA in half-precision
4. An abstraction of device selection based on TensorRT engine

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/cpp/api/include/trtorch/trtorch.h
+++ b/cpp/api/include/trtorch/trtorch.h
@@ -11,6 +11,7 @@
 #include <string>
 #include <vector>
 #include <memory>
+#include <cuda_runtime.h>
 
 // Just include the .h?
 #ifndef DOXYGEN_SHOULD_SKIP_THIS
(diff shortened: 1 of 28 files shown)


cpp/api/include/trtorch/trtorch.h
```diff
@@ -404,4 +433,16 @@ TRTORCH_API torch::jit::Module CompileGraph(const torch::jit::Module& module, Ex
...
+/**
+ * @brief Set gpu device id
+ *
+ * @param gpu_id
+ *
+ * Sets gpu id using cudaSetDevice
+ *
+ * @return: cudaError_t return status
+ */
+TRTORCH_API cudaError_t set_device(const int gpu_id);
```
narendasan, line 446: We should use the device struct to set device here I think

narendasan: Also we should not return the cudaError_t, maybe we could do something else here or handle the cuda error for the user?

andi4191: This is intended for runtime. Using other Device info such as dla_core, allow_gpu_fallback won't be applicable here. These configurable parameters are required during building time only. Hence removed from here.

Device info attributes are already added in ExtraConfig during build time.

Return status would help end-users to debug errors (such as incorrect device ordinal) first hand. cudaError_t return type API is already released from CUDA

third_party/cuda/BUILD
```diff
@@ -19,7 +19,7 @@ cc_library(
     name = "cudart",
     srcs = select({
         ":aarch64_linux": [
-            "targets/aarch64-linux/lib/libcudart.so",
+            "targets/aarch64-linux-gnu/lib/libcudart.so",
```
narendasan, line 22: This is the wrong path on stock JetPack

narendasan (changes requested):

third_party/cuda/BUILD
andi4191: Jetpack 4.4 seems to fail without this suffix. Should we tag this release and then change for future JP releases?

andi4191: Fixed it

cpp/api/include/trtorch/trtorch.h
narendasan: We should still make this void or something or develop a TRTorch Error Type

tests/py/test_api_dla.py
```diff
@@ -0,0 +1,72 @@
+import unittest
```
narendasan, line 1: I am getting this error right now `ImportError: /media/nvidia/419bace3-a6c8-49a0-b8f7-20266100776e/naren/TRTorch/py/trtorch/lib/libtrtorch.so: undefined symbol: NvMediaDeviceDestroy
`

Can you build a C++ version of the test that is added to the test suite conditionally based on the architecture? It will be easier to figure out what is going on. (Having this test is good too)

narendasan (changes requested):

tests/py/test_api_dla.py
narendasan: All libraries are showing as linked when i ldd the libraries

andi4191:
Closing this PR. The device metadata serialization part seems to be merged here as well somehow. Splitting this PR into:
1. DLA support only  + Review comments from here
2. Device meta data serialization/deserialization

PR #164, opened 2020-08-01
Loop Conversion

abhi-iyer:
# Description

Dynamic loop conversion from PyTorch to TRT. A few limitations with this approach, as of now:

- all recurrent values inside loop can be Tensors only
- works with for loops, albeit with a comparative threshold (i.e. TRT vs JIT) that is greater than the set limit of 2e-5
- breaks converter contract: all primitive types (e.g. int, float, etc.) could potentially be Tensors when processed inside the converter

## Type of change

# Checklist:

- [ ] My code follows the style guidelines of this project

### DIFF
--- a/core/lowering/lowering.cpp
+++ b/core/lowering/lowering.cpp
@@ -34,7 +34,7 @@ void LowerGraph(std::shared_ptr<torch::jit::Graph>& g) {
     passes::Conv2DToConvolution(g);
     passes::FuseAddMMBranches(g);
     torch::jit::EliminateCommonSubexpression(g);
-    torch::jit::UnrollLoops(g);
+    //torch::jit::UnrollLoops(g);
     torch::jit::EliminateCommonSubexpression(g);
     passes::UnpackAddMM(g);
     //passes::UnpackBatchNorm(g);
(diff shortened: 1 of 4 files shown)


abhi-iyer:
For more information on Loop Conversion, please see:

[TensorRT Loops - A Rundown.pdf](https://github.com/NVIDIA/TRTorch/files/5009794/TensorRT.Loops.-.A.Rundown.pdf)

core/lowering/lowering.cpp
```diff
@@ -35,7 +35,7 @@ void LowerGraph(std::shared_ptr<torch::jit::Graph>& g) {
     passes::Conv2DToConvolution(g);
     passes::FuseAddMMBranches(g);
     torch::jit::EliminateCommonSubexpression(g);
-    torch::jit::UnrollLoops(g);
+    //torch::jit::UnrollLoops(g);
```
narendasan, line 38: Does the unroll loops behavior still not work even with the non compliant converters?

abhi-iyer:
TRT has no guarantee about output binding order.

e.g. an RNN loop that outputs hidden state `hx` and cell state `cx` might be reversed when outputs are processed, even if they were marked as inputs in the right order

PR #224, opened 2020-11-17
Explicit pattern matching for add + MM fusion

peri044:
# Description

Change the condition to find ADDMM variant nodes to be more explicit. The following issue happens even though it is not a add + matmul variant
```
 Found that node %scale_factor0.9 : float?, %size3.3 : int? = prim::If(%16230) # /opt/conda/lib/python3.6/site-packages/torch/nn/functional.py:2969:4
  block0():
     = prim::RaiseException(%16215) # /opt/conda/lib/python3.6/site-packages/torch/nn/functional.py:2970:8
    -> (%16220, %16221)
  block1():
    -> (%652, %size1.3)
 is an AddMM variants node (FuseAddMMBranches)
```

## Type of change

# Checklist:

### DIFF
--- a/core/lowering/passes/fuse_addmm_branches.cpp
+++ b/core/lowering/passes/fuse_addmm_branches.cpp
@@ -44,21 +44,18 @@ struct AddMMBranchFusion {
     auto arm2 = n->blocks()[1];
 
     auto arm1_start = arm1->nodes().begin();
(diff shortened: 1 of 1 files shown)


narendasan (changes requested):
Can you add a test case that verifies the change? Like a case that would fail before but not after?

narendasan (approved):

Issue #267, opened 2021-01-08
prim::ListUnpack unable to get schema


inocsin:
When I try to complie a model, I got such error
```
[1;35mDEBUG: [0mUnable to get schema for Node %b.1 : int, %nframe.1 : int, %c : int, %h.1 : int, %w.1 : int = prim::ListUnpack(%15) (NodeConverterRegistry.Convertable)
terminate called after throwing an instance of 'trtorch::Error'
  what():  [enforce fail at core/conversion/conversion.cpp:392] Expected schema to be true but got false
Unable to get schema for Node %b.1 : int, %nframe.1 : int, %c : int, %h.1 : int, %w.1 : int = prim::ListUnpack(%15) (conversion.VerifyCoverterSupportForBlock)
```
and the related graph definition is this
```
  %15 : int[] = aten::size(%images.1) # <string>:7:9
  %b.1 : int, %nframe.1 : int, %c : int, %h.1 : int, %w.1 : int = prim::ListUnpack(%15)
```
Input shape is (1,1,3,672,672)

detailed log is here 
[listunpack.txt](https://github.com/NVIDIA/TRTorch/files/5786336/listunpack.txt)
GDB backtrace
```
#0  0x00007fff63987438 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007fff6398903a in __GI_abort () at abort.c:89
#2  0x00007ffff7a8ddde in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff7a99896 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7a99901 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7a99b55 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x000000000047b116 in trtorch::core::conversion::GetUnsupportedOpsInBlock[abi:cxx11](torch::jit::Block const*) (b=0x5d4b9d50) at core/conversion/conversion.cpp:390
#7  0x000000000047b3a7 in trtorch::core::conversion::VerifyConverterSupportForBlock (b=0x5d4b9d50) at core/conversion/conversion.cpp:406
[...]

PR #283, opened 2021-01-13
feat(//core/): Add support for Split converter and unpack evaluator

peri044:
# Description

Support aten::split and prim::ListUnpack

Fixes https://github.com/NVIDIA/TRTorch/issues/267 and https://github.com/NVIDIA/TRTorch/issues/259

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/conversion.cpp
+++ b/core/conversion/conversion.cpp
@@ -180,22 +180,7 @@ void MarkOutputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> outp
     if (it == ctx->value_tensor_map.end()) {
       if (ctx->evaluated_value_map.find(out) != ctx->evaluated_value_map.end()) {
         auto out_ivalue = ctx->evaluated_value_map[out];
-        if (out_ivalue.isList()) {
-          auto output_list = out_ivalue.toList();
-          LOG_DEBUG("One of the outputs is a TensorList. output_list size: " << output_list.size());
-
-          for (int i = 0; i < output_list.size(); i++) {
-            std::string name = std::string("output_") + std::to_string(ctx->num_outputs);
-            auto output_container = output_list.get(i).toCustomClass<TensorContainer>();
-            nvinfer1::ITensor* out_tensor = output_container.get()->tensor();
-            out_tensor->setName(name.c_str());
-            ctx->net->markOutput(*out_tensor);
-            LOG_INFO(
-                ctx->logger,
-                "Marking Output " << out->debugName() << " named " << name << " in engine (ctx.MarkOutput)");
-            ctx->num_outputs += 1;
-          }
-        } else if (out_ivalue.isCustomClass()) {
+        if (out_ivalue.isCustomClass()) {
           std::string name = std::string("output_") + std::to_string(ctx->num_outputs);
           auto output_container = out_ivalue.toCustomClass<TensorContainer>();
           nvinfer1::ITensor* out_tensor = output_container.get()->tensor();
@@ -374,8 +359,8 @@ void ConvertBlockToNetDef(
             auto eval_list = eval.value().toTuple();
             TRTORCH_CHECK(
                 eval_list->elements().size() == n->outputs().size(),
-                "Size of evaluated results: " << eval_list->elements().size() << " and node outputs size: " << n->outputs().size()
-                                              << " must match.");
+                "Size of evaluated results: " << eval_list->elements().size()
+                                              << " and node outputs size: " << n->outputs().size() << " must match.");
             for (int i = 0; i < eval_list->elements().size(); i++) {
               auto eval_output = eval_list.get()->elements()[i];
               LOG_DEBUG(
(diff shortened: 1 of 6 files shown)


peri044:
One part of this issue #267 should be resolved by this PR. I need to look into other details in the issue.

core/conversion/evaluators/prim.cpp
```diff
@@ -32,6 +32,12 @@ auto prim_registrations =
                     [](const torch::jit::Node* n, kwargs& args) -> c10::optional<torch::jit::IValue> {
                       return at::scalar_to_tensor(args.at(n->output(0)).IValue()->toScalar());
                     }})
+        .evaluator({torch::jit::prim::ListUnpack,
```
narendasan, line 35: Can we do a list unpack test?

core/conversion/conversion.cpp
```diff
@@ -337,12 +369,28 @@ void ConvertBlockToNetDef(
     } else if (to_eval) {
       auto eval = EvaluateNode(ctx, n);
       if (eval) {
-        if (!eval.value().isTensor()) {
+        if (n->outputs().size() > 1) { // For ListUnpack scenario
+          if (eval.value().isList()) {
+            LOG_DEBUG(
+                ctx->logger,
+                "Found the evaluated value to be a list" << eval.value() << " for node: " << util::node_info(n));
```
narendasan, line 376: This is a misleading message i feel as this case is when there are multiple outputs from an evaluator. It makes seem like the object returned is a singular list and not a list of results.

core/conversion/conversion.cpp
```diff
@@ -337,12 +369,28 @@ void ConvertBlockToNetDef(
     } else if (to_eval) {
       auto eval = EvaluateNode(ctx, n);
       if (eval) {
-        if (!eval.value().isTensor()) {
+        if (n->outputs().size() > 1) { // For ListUnpack scenario
+          if (eval.value().isList()) {
```
narendasan, line 373: I feel like tuples feel like the better type to use here. I think functionally they operate the same but cognitively they are more in line with how people think about multiple returns from functions.

core/conversion/conversion.cpp
```diff
@@ -337,12 +369,28 @@ void ConvertBlockToNetDef(
...
+                ctx->logger,
+                "Found the evaluated value to be a list" << eval.value() << " for node: " << util::node_info(n));
+            auto eval_list = eval.value().toList();
+            TRTORCH_CHECK(
+                eval_list.size() == n->outputs().size(),
+                "Size of evaluated results: " << eval_list.size() << " and node outputs size: " << n->outputs().size()
+                                              << " must match.");
+            for (int i = 0; i < eval_list.size(); i++) {
+              auto eval_output = eval_list.get(i);
+              ctx->AssociateValueAndIValue(n->output(i), eval_output);
```
narendasan, line 384: We should move the debug statement above to here and list out all found results.

core/conversion/evaluators/prim.cpp
```diff
@@ -32,6 +32,12 @@ auto prim_registrations =
...
+                    [](const torch::jit::Node* n, kwargs& args) -> c10::optional<torch::jit::IValue> {
+                      // Outputs is an IValue which has list of tensors which can be found in ctx->evaluated_value_map
+                      const torch::jit::IValue* outputs = args.at(n->input()).IValue();
+                      return *std::move(outputs);
```
narendasan, line 39: You might be able to use something like this to quickly switch to a tuple https://github.com/pytorch/pytorch/blob/22a34bcf4e5eaa348f0117c414c3dd760ec64b13/test/cpp/api/jit.cpp#L91

peri044: Done

core/conversion/conversion.cpp
peri044: Done

core/conversion/evaluators/prim.cpp
peri044: Added a test for list unpack.

core/conversion/conversion.cpp
```diff
@@ -173,18 +176,47 @@ void AddInputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> inputs
...
-        "No corresponding output TRT Tensor found for TorchScript output: " << out->debugName());
-    auto out_tensor = it->second;
-    out_tensor->setName(name.c_str());
-    ctx->net->markOutput(*out_tensor);
-    LOG_INFO(ctx->logger, "Marking Output " << out->debugName() << " named " << name << " in engine (ctx.MarkOutput)");
-    ctx->num_outputs += 1;
+    if (it == ctx->value_tensor_map.end()) {
+      if (ctx->evaluated_value_map.find(out) != ctx->evaluated_value_map.end()) {
+        auto out_ivalue = ctx->evaluated_value_map[out];
+        if (out_ivalue.isList()) {
```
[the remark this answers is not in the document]
peri044: Yes. That's right. Initially I've been trying various split graphs which is when this was needed before our discussion. I've been testing it using trtorchexec and it worked with that. But the test cases use a different API to construct/run the graph and it would be a problem. Removed the code. Anyways the split graph testcases have unpack anyway. We shall look into better way of handling different output types later. (Related issue opened : [https://github.com/NVIDIA/TRTorch/issues/284](url))

narendasan (approved):
LGTM

PR #285, opened 2021-01-14
feat(//core/): Add support for tensorrt plugin and implement instance…

### LINKED ISSUE #210: ✨[Feature] Please support aten::instance_norm
I'm struggling to convert a PyTorch model.

To have this operator supported.

To implement it myself, however I'm yet there!

**Additional context**
Thanks for considering!

ChaoFang-TRI:
… norm layer

Fix #210 

Signed-off-by: Chao Fang <chao.fang@tri.global>

# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/conversionctx/ConversionCtx.cpp
+++ b/core/conversion/conversionctx/ConversionCtx.cpp
@@ -2,6 +2,9 @@
 #include <sstream>
 #include <utility>
 
+#include "NvInferPlugin.h"
+#include "NvInferPluginUtils.h"
+
 #include "core/conversion/conversionctx/ConversionCtx.h"
 
 namespace trtorch {
@@ -45,6 +48,25 @@ ConversionCtx::ConversionCtx(BuilderSettings build_settings)
           "[TRTorch Conversion Context] - ",
           util::logging::get_logger().get_reportable_severity(),
           util::logging::get_logger().get_is_colored_output_on()) {
+    // Get list of all available plugin creators
+
+    initLibNvInferPlugins(&logger, "");
+
+    int numCreators = 0;
+    auto tmpList = getPluginRegistry()->getPluginCreatorList(&numCreators);
+    for (int k = 0; k < numCreators; ++k)
+    {
+        if (!tmpList[k])
+        {
+            std::cout << "Plugin Creator for plugin " << k << " is a nullptr." << std::endl;
+            continue;
+        }
+        std::string pluginName = tmpList[k]->getPluginName();
+        mPluginRegistry[pluginName] = tmpList[k];
+        LOG_DEBUG("Register plugin: " << pluginName);
+    }
+    LOG_DEBUG("Number of plugin: " << mPluginRegistry.size());
+
   // TODO: Support FP16 and FP32 from JIT information
   builder = nvinfer1::createInferBuilder(logger);
   net = builder->createNetworkV2(1U << static_cast<uint32_t>(nvinfer1::NetworkDefinitionCreationFlag::kEXPLICIT_BATCH));
--- a/core/conversion/conversionctx/ConversionCtx.h
+++ b/core/conversion/conversionctx/ConversionCtx.h
@@ -66,6 +66,9 @@ struct ConversionCtx {
   // copy of the values
   std::vector<void*> builder_resources;
 
+  //Registry of official tensorrt plugin layers.
+  std::unordered_map<std::string, nvinfer1::IPluginCreator*> mPluginRegistry;
+
   std::unordered_map<const torch::jit::Value*, nvinfer1::ITensor*> value_tensor_map;
   std::unordered_map<const torch::jit::Value*, torch::jit::IValue> evaluated_value_map;
 };
--- a/core/conversion/converters/impl/batch_norm.cpp
+++ b/core/conversion/converters/impl/batch_norm.cpp
@@ -9,6 +9,13 @@ namespace converters {
 namespace impl {
 namespace {
 
+static inline at::Tensor repeat_if_defined(const at::Tensor& t, int64_t repeat) {
+    if (t.defined()) {
+        return t.repeat(repeat);
+    }
+    return t;
+}
+
 auto batch_norm_registrations TRTORCH_UNUSED = RegisterNodeConversionPatterns().pattern({
     R"SIG(aten::batch_norm(Tensor input, Tensor? gamma, Tensor? beta,
                             Tensor? mean, Tensor? var,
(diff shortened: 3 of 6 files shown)


ChaoFang-TRI:
@narendasan The tensorRT plugin support need more tests. It would be nice if any NVIDIA folks could help.

ChaoFang-TRI:
I keep getting linter erro:
ERROR: Skipping '//tools/linter:cpplint_diff': error loading package 'tools/linter': Unable to find package for @pylinter_deps//:requirements.bzl: The repository '@pylinter_deps' could not be resolved.

narendasan:
> I keep getting linter erro:
> ERROR: Skipping '//tools/linter:cpplint_diff': error loading package 'tools/linter': Unable to find package for @pylinter_deps//:requirements.bzl: The repository '@pylinter_deps' could not be resolved.

Do you have the Python dependencies in the `WORKSPACE` enabled? It should just be picking up the `requirements.txt` file in `//tools/linter`

https://github.com/NVIDIA/TRTorch/blob/72bf74b2a2e425e6c83542f99c92b9e551148149/WORKSPACE#L142

narendasan:
> @narendasan The tensorRT plugin support need more tests. It would be nice if any NVIDIA folks could help.

Yeah for sure we can take a look here. Next test that would be good to see is if a TRT plugin based model is portable. So compile a module, save to disk, start a new process, load the module and see if it runs or something similar

core/conversion/conversionctx/ConversionCtx.cpp
```diff
@@ -45,6 +48,25 @@ ConversionCtx::ConversionCtx(BuilderSettings build_settings)
           "[TRTorch Conversion Context] - ",
           util::logging::get_logger().get_reportable_severity(),
           util::logging::get_logger().get_is_colored_output_on()) {
+    // Get list of all available plugin creators
+
+    initLibNvInferPlugins(&logger, "");
```
narendasan, line 53: I suspect that this will need to occur in a more global manner since it will need to cover cases were we only use the runtime and the conversion context is therefore never created. We also probably should use the global logger in that case as well instead of the conversion logger.

narendasan: Ideally this would be located somewhere in the runtime module and run on library load since we would also want this for libtrtorchrt.so as well.

core/conversion/conversionctx/ConversionCtx.h
```diff
@@ -66,6 +66,9 @@ struct ConversionCtx {
   // copy of the values
   std::vector<void*> builder_resources;
 
+  //Registry of official tensorrt plugin layers.
+  std::unordered_map<std::string, nvinfer1::IPluginCreator*> mPluginRegistry;
```
narendasan, line 70: Considering we are starting to ship our own TRTorch specific plugins as well maybe we could unify the various registries in a separate module which is a dep of conversion and runtime. It might solve the global init issue as well. @peri044  thoughts?

tests/core/conversion/converters/test_batch_norm.cpp
```diff
@@ -12,25 +12,68 @@ TEST(Converters, ATenBatchNormConvertsCorrectly) {
...
+  auto params = trtorch::core::conversion::get_named_params(g->inputs(), {gamma, beta, mean, var});
+  auto jit_results = trtorch::tests::util::RunGraph(g, params, {in});
+
+  params = trtorch::core::conversion::get_named_params(g->inputs(), {trt_gamma, trt_beta, trt_mean, trt_var});
+  auto trt_results = trtorch::tests::util::RunGraphEngine(g, params, {trt_in});
+
+  ASSERT_TRUE(trtorch::tests::util::almostEqual(jit_results[0], trt_results[0].reshape_as(jit_results[0]), 2e-6));
+}
+
+TEST(Converters, ATenInstanceNormWithConvertsCorrectly) {
```
narendasan, line 44: Another good test would be to make sure we are hitting both paths of the instance norm converter

narendasan (commented):
Thanks! This is a great addition. Left some preliminary comments. Would also like @peri044 to take a look. I think we need to think a bit about how to make sure plugins are available both at conversion and runtime.

core/conversion/converters/impl/batch_norm.cpp
```diff
@@ -75,8 +82,147 @@ auto batch_norm_registrations TRTORCH_UNUSED = RegisterNodeConversionPatterns().
...
+         std::vector<nvinfer1::PluginField> f;
+         f.emplace_back(nvinfer1::PluginField("epsilon", &epsilon, nvinfer1::PluginFieldType::kFLOAT32, 1));
+         f.emplace_back(nvinfer1::PluginField("scales", gamma.data.values, nvinfer1::PluginFieldType::kFLOAT32, gamma.data.count));
+         f.emplace_back(nvinfer1::PluginField("bias", beta.data.values, nvinfer1::PluginFieldType::kFLOAT32, beta.data.count));
+
+         fc.nbFields = f.size();
+         fc.fields = f.data();
+         nvinfer1::IPluginV2* pluginV2 = ctx->mPluginRegistry.at(pluginName)->createPlugin("instancenorm", &fc);
+
+         TRTORCH_CHECK(pluginV2, "Unable to create interpolation plugin from node" << *n);
```
peri044, line 141: interpolation plugin name -> instance norm plugin. To distinguish from line 143 check, this message can be changed to something like "Unable to create instance norm plugin from TensorRT plugin registry"

core/conversion/conversionctx/ConversionCtx.h
peri044: That's a nice idea. We can have a single registry and our own version of initLibNvInferPlugins which can initialize TRT and TRTorch plugins.
But here are some questions in mind 
Do we plan on integrating custom plugins with libtrtorch.so as well ? or do we have something like libtrtorch_plugins.so ? 
- For the former case, If the extracted TRT engine has a plugin that doesn't depend on aten:calls and uses custom cuda kernels (as for general plugins), this sample needs to load libnvinfer.so, libtrtorch.so which depends on libtorch.so ? 
- In the later case, the sample can load libnvinfer.so and libtrtorch_plugins.so. Maybe this would be ideal from a deployment perspective ?

peri044 (changes requested):
The testcase which hits one path of instance norm works on my local.

peri044:
@ChaoFang-TRI Thanks for the PR. Any thoughts on the above review comments ? Can you address some of them ?

ChaoFang-TRI:
> @ChaoFang-TRI Thanks for the PR. Any thoughts on the above review comments ? Can you address some of them ?

Hi Peri044, I was occupied for family issue in the past month, sorry for the delayed response. Please feel free to jump in on addressing comments! I will push some commits as well to address some commits.

core/conversion/conversionctx/ConversionCtx.cpp
ChaoFang-TRI: I am not sure where is the best place to initialize the plugin. Can you suggest a specific place?

core/conversion/conversionctx/ConversionCtx.h
ChaoFang-TRI: Seems this is touching the core requirement of TRTorch. I like the second idea to load load libnvinfer.so and libtrtorch_plugins.so.

peri044:
I tried to add this instance norm to the [plugin suite PR](https://github.com/NVIDIA/TRTorch/pull/425) . The testcase provided in this PR does not use TensorRT plugin but rather hits the else section (using native TRT layers). 
The following testcase  graph uses plugin block in the instance norm code in this PR; but upon testing this, I face seg fault issue. So I'm holding off this plugin for now and later add this once the issues are fixed.  
```
const auto graph = R"IR(
      graph(%0 : Tensor,
            %1: Float(5, strides=[1]),
            %2: Float(5, strides=[1]),
            %3: Float(5, strides=[1]),
            %4: Float(5, strides=[1])):
        %9 : Tensor? = prim::Constant()
        %5 : bool = prim::Constant[value=1]()
        %6 : float = prim::Constant[value=0.10000000000000001]()
        %7 : float = prim::Constant[value=1.0000000000000001e-05]()
        %8 : Tensor = aten::instance_norm(%0, %1, %2, %9, %9, %5, %6, %7, %5)
        return (%8))IR";
```

narendasan:
We should be merging in support for nvinfer plugin in #425, Can we reduce the scope of this PR to only include the additional converters?

narendasan:
Closing in favor of #573

PR #293, opened 2021-01-26
Add clamp conversion functionality

peri044:
# Description

Add clamp conversion

Fixes https://github.com/NVIDIA/TRTorch/issues/274

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/element_wise.cpp
+++ b/core/conversion/converters/impl/element_wise.cpp
@@ -144,6 +144,41 @@ auto element_wise_registrations TRTORCH_UNUSED =
                     LOG_DEBUG("Output tensor shape: " << out->getDimensions());
                     return true;
                   }})
+        .pattern({"aten::clamp(Tensor self, Scalar? min=None, Scalar? max=None) -> (Tensor)",
+                  [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
+                    // Compute min(max(min_threshold, input), max_threshold)
+                    auto self = args[0].ITensorOrFreeze(ctx);
+                    auto clamp_layer_out = self;
+                    if (args[1].isIValue() && args[1].IValue()->isScalar()) {
+                      auto minScalar = args[1].unwrapToScalar().to<float>();
+                      auto minTensor = tensor_to_const(ctx, torch::tensor({minScalar}));
+                      auto max_layer = add_elementwise(
+                          ctx,
+                          nvinfer1::ElementWiseOperation::kMAX,
+                          clamp_layer_out,
+                          minTensor,
+                          util::node_info(n) + std::string("_max"));
+                      TRTORCH_CHECK(max_layer, "Unable to create elementwise max layer for node: " << *n);
+                      clamp_layer_out = max_layer->getOutput(0);
+                    }
+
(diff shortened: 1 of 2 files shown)


core/conversion/converters/impl/element_wise.cpp
```diff
@@ -144,6 +144,41 @@ auto element_wise_registrations TRTORCH_UNUSED =
                     LOG_DEBUG("Output tensor shape: " << out->getDimensions());
                     return true;
                   }})
+        .pattern({"aten::clamp(Tensor self, Scalar? min=None, Scalar? max=None) -> (Tensor)",
```
narendasan, line 147: Could we possibly lower to hardtanh or is the functionality different?

core/conversion/converters/impl/element_wise.cpp
```diff
@@ -144,6 +144,41 @@ auto element_wise_registrations TRTORCH_UNUSED =
...
+                  [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
+                    // Compute min(max(min_threshold, input), max_threshold)
+                    auto self = args[0].ITensorOrFreeze(ctx);
+                    auto clamp_layer_out = self;
+                    if (args[1].isIValue() && args[1].IValue()->isScalar()) {
```
narendasan, line 152: I think we might be able to do !isNone on the IValue instead of two checks

peri044: If I remember correctly, isNone() didn't work for me. I will double check again.

peri044: The op does basically this `min(max(min_threshold, input), max_threshold)`
In the case when min_threshold or max_threshold is not given, args[1] or args[2] will be registered as an IValue. 
The isNone() check will check if the type is one of { kITensor, kIValue, kNone }. The type in this case would be IValue so isNone() will not return true.

core/conversion/converters/impl/element_wise.cpp
peri044: hardtanh looks a bit different [https://pytorch.org/docs/stable/generated/torch.nn.Hardtanh.html](url)
hardtanh compares the elements to -1 and 1 and replaces with default values (thresholds) where as clamp compares the input elements to thresholds directly. They both probably can match in certain scenarios

core/conversion/converters/impl/element_wise.cpp
narendasan: I see

core/conversion/converters/impl/element_wise.cpp
narendasan: hardtanh lets you configure the min and max values. really my point is we should be able to lower clamp and hardtanh to the same clip activation converter. This is something we could leave for later though

narendasan: Ideally our converter library should be as small as possible and we should do as much as we can through lowering

narendasan (approved):

narendasan:
@peri044 can you resolve merge conflicts? Ill merge after that.

peri044:
Resolved conflicts

PR #258, opened 2020-12-20
support dynamic shape input and scale_factor in interpolate layer

uni19:
# Description
support dynamic shape input and scale_factor in interpolate layer

## Type of change
- New feature (non-breaking change which adds functionality)
# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests to verify my fix or my feature
- [x] New and existing unit tests pass locally with my changes

PR #295, opened 2021-01-27
Extending @uni19's work on support dynamic shape input and scale_factor in interpolate layer

### LINKED ISSUE #290: ↔ [Converter] Add support for aten::upsample_nearest2d.vec in TRTorch
## aten::upsample_nearest2d

- **Function Schema**:
`Schema: aten::upsample_nearest2d.vec(Tensor input, int[]? output_size, float[]? scale_factors) -> (Tensor)`

- **Original PyTorch API**:
Not sure where it is

- **Relevant TensorRT Documentation**:

## Alternatives
I think there is already a similar implementation 
`aten::upsample_nearest2d(Tensor self, int[2] output_size, float? scales_h=None, float? scales_w=None) -> (Tensor)`

maybe a small modification can solve this

## Additional context
I try to convert a yolov5 with TRTorch
[https://github.com/ultralytics/yolov5](https://github.com/ultralytics/yolov5)
I know there is another solution
[https://github.com/wang-xinyu/tensorrtx](https://github.com/wang-xinyu/tensorrtx)
but I want to practice TRTorch and further convert more complicated model...

I use the `models/export.py` provided by ultralytics's repo to convert yolo5s to `yolo5s.torchscript.pt` and then convert it as following:
```python
# load model
traced_model = torch.jit.load("yolov5s.torchscript.pt", map_location=torch.device("cpu"))
traced_model = traced_model.to(device)
traced_model = traced_model.eval()

# convert to TRTorch
trt_model_fp32 = trtorch.compile(traced_model, {
    "input_shapes": [(batch_size, 3, 416, 416)],
    "op_precision": torch.float32,  # Run with FP32
    "workspace_size": 1 << 20
})
```

The error message:
```bash
RuntimeError: [enforce fail at core/conversion/conversion.cpp:110] Expected converter to be true but got false
Unable to convert node: %282 : Tensor = aten::upsample_nearest2d(%input.31, %105, %106) # /home/xxxxxxxxxx/venv/lib/python3.6/site-packages/torch/nn/functional.py:3132:0 (conversion.AddLayer)
Schema: aten::upsample_nearest2d.vec(Tensor input, int[]? output_size, float[]? scale_factors) -> (Tensor)
Converter for aten::upsample_nearest2d requested, but no such converter was found.
If you need a converter for this operator, you can try implementing one yourself
or request a converter: https://www.github.com/NVIDIA/TRTorch/issues
```

narendasan:
# Description

This PR includes the changes in #258, does some refactoring of the test suite for interpolate and should expand coverage of the interpolate converter for the two variants of each of the currently supported resize functions (self, and vec). It should mean that we support interpolate for nearest and linear variants for all cases except when users use a type of linear upsampling, use scale factor for setting the size and set `align_corners` to true. It seems like @uni19  found that TensorRT  and PyTorch have different behavior in this case. I will check with the TRT team to see if they have observed this. Hopefully it is not a massive issue as PyTorch since 1.5 has switched to `align_corners` being false by default. As a stop gap this PR extends the interpolate plugin to handle this case, however it can only do so for static TRT engines. Currently the build time dimension calculation can only be done with `int`s where as scale factors for PyTorch are `float`s. Therefore in the case of dynamic engine for a model using [Bi/Tri]Linear interpolate with scale factor and `align_corners`=True, TRTorch will error out. Hopefully this is niche enough of a use case to not cause issues until we get a response from TensorRT. 

Fixes #290 

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/interpolate.cpp
+++ b/core/conversion/converters/impl/interpolate.cpp
@@ -15,7 +15,6 @@ namespace {
 /*
  * Helper functions
  */
-#if NV_TENSORRT_MAJOR < 7 || (NV_TENSORRT_MAJOR == 7 && NV_TENSORRT_MINOR < 1)
 void create_plugin(
     ConversionCtx* ctx,
     const torch::jit::Node* n,
@@ -24,11 +23,14 @@ void create_plugin(
     std::vector<int64_t> in_shape,
     std::vector<int64_t> out_shape,
     std::vector<int64_t> out_size,
-    std::string mode) {
+    std::vector<double> scales,
+    std::string mode,
+    bool align_corners,
+    bool use_scales=false) {
   LOG_WARNING("Interpolation layer will be run through ATen, not TensorRT. Performance may be lower than expected");
 
   auto creator = new plugins::InterpolatePluginCreator();
-  auto plugin = creator->createPlugin(name, in_shape, out_shape, out_size, mode, false);
+  auto plugin = creator->createPlugin(name, in_shape, out_shape, out_size, scales, mode, align_corners, use_scales);
 
   auto resize_layer = ctx->net->addPluginV2(reinterpret_cast<nvinfer1::ITensor* const*>(&in), 1, *plugin);
   TRTORCH_CHECK(resize_layer, "Unable to create interpolation plugin from node" << *n);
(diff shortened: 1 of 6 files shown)


peri044 (approved):
LGTM. Verified testcases on my local with TRT 7.2.2.3

PR #299, opened 2021-01-29
support leaky_relu converter/test_case

### LINKED ISSUE #260: ↔ [Converter] Add support for aten::leaky_relu in TRTorch
##

- **Function Schema**: aten::leaky_relu(Tensor self, Scalar negative_slope=0.01) -> (Tensor)

- **Original PyTorch API**: https://pytorch.org/docs/stable/generated/torch.nn.LeakyReLU.html

narendasan:
# Description

Adds support for leaky relu 

Fixes #260 

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/activation.cpp
+++ b/core/conversion/converters/impl/activation.cpp
@@ -124,7 +124,23 @@ auto acthardtanh TRTORCH_UNUSED =
                out_tensor = ctx->AssociateValueAndTensor(n->outputs()[0], out_tensor);
                LOG_DEBUG("Output shape: " << out_tensor->getDimensions());
                return true;
+             }})
+       .pattern(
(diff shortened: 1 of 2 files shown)


peri044 (approved):
LGTM

rafale77:
Just tested leaky_relu and found that my model also had a layer named "leaky_relu_" which is likely the one in the functional library? Might want to consider supporting that one as well by adding to it.

peri044:
aten::leaky_relu_ is an inplace variant [.](https://pytorch.org/docs/stable/nn.functional.html#torch.nn.functional.leaky_relu_) Seems like the same op. We should handle it via lowering.

narendasan:
Hmm, would have expected the functionalization pass to handle this but we can easily alias the converter

PR #302, opened 2021-01-29
Adding support for aten::topk

narendasan:
# Description

Adds support for `aten::topk`

## Type of change

# Checklist:


core/conversion/converters/impl/topk.cpp
```diff
@@ -0,0 +1,59 @@
...
+namespace {
+
+auto topk_registrations TRTORCH_UNUSED = RegisterNodeConversionPatterns().pattern(
+    {"aten::topk(Tensor self, int k, int dim=-1, bool largest=True, bool sorted=True) -> (Tensor values, Tensor indices)",
+     [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
+       auto self = args[0].ITensorOrFreeze(ctx);
+       auto k = args[1].unwrapToInt();
+       auto dim = args[2].unwrapToInt();
+       auto largest = args[3].unwrapToBool();
+       // auto sorted = args[4].unwrapToBool(); # Currently unused
```
narendasan, line 24: @inocsin What is the role of this argument and why don't we use it? I commented it out because it was throwing warnings

peri044: torch.topk has an argument sorted = True which returns the top k elements in a sorted order (decreasing). By default it is true. Setting it False will result in random order of output top k elements which we can't support in TensorRT. TensorRT always returns the top "k" max or min elements depending on "largest" argument. Maybe we can add this detail as a comment to explain why we aren't using args[4]

peri044 (approved):

core/conversion/converters/impl/topk.cpp
narendasan: will this cause an issue to users? if they expect random does always giving them ordered matter? At the very least yeah there should be a comment/

peri044: This won't cause an issue to users in my opinion. So far all usecases for this op have been gathering either top k largest elements or top k smallest elements, both of which are supported. We can revisit this if anyone requests such usecase.

PR #307, opened 2021-01-29
Adding aten::is_floating_point evaluator

narendasan:
# Description

This adds support for the `aten::is_floating_point` operator. Also adds a `aten::Int` evaluator that needs to be discussed

## Type of change

# Checklist:


core/conversion/evaluators/aten.cpp
```diff
@@ -464,7 +464,38 @@ auto aten_registrations TRTORCH_UNUSED =
...
+                      } else {
+                        auto tensor = tensor_var.unwrapToTensor();
+                        auto t = tensor.scalar_type();
+                        return at::isFloatingType(t);
+                      }
+                    },
+                    EvalOptions().validSchemas({
+                        "aten::is_floating_point(Tensor self) -> (bool)",
+                    })})
+        .evaluator({c10::Symbol::fromQualString("aten::Int"),
```
narendasan, line 484: Is there a better way to handle this? Type casts dont have meaning in TensorRT. Are what are the possible use cases of this op and how do we handle them?

core/conversion/evaluators/aten.cpp
```diff
@@ -464,7 +464,38 @@ auto aten_registrations TRTORCH_UNUSED =
...
+                    [](const torch::jit::Node* n, kwargs& args) -> c10::optional<torch::jit::IValue> {
+                      auto tensor_var = args.at(n->input(0));
+                      if (tensor_var.isITensor()) {
+                        TRTORCH_ASSERT(!tensor_var.isITensor(), "Unable evaluate ITensor to int");
```
peri044, line 488: "Unable to evaluate ITensor to int in aten::int evaluator" maybe ?

core/conversion/evaluators/aten.cpp
peri044: Tagging @inocsin for viz. I have similar question as well. Can anyone let me know the usage of this op ? 
In line 487, there is a check if node input is an ITensor or not. ITensor() true case happens when  node input is in ctx->value_tensor_map right ? What is the layer preceeding this layer usually ? Can you add test case if there are no blockers ?

peri044 (changes requested):

core/conversion/evaluators/aten.cpp
inocsin: Yeah, do you have any idea how to get the ITensor value in the evalutor?

core/conversion/evaluators/aten.cpp
inocsin:
```
  %2270 : bool = aten::is_floating_point(%input174.1) # <string>:3:6
  %2271 : bool = prim::If(%2270) # <string>:3:6
```
It was used as in If-Else

inocsin: aten::Int, aten::is_floating_point test case is here https://github.com/NVIDIA/TRTorch/pull/296 , the evaluator test framework was not ready at that time, so I submit anothor commit with test case.

core/conversion/evaluators/aten.cpp
narendasan: ITensors have no definite value at compile time. If an ITensor hits an evaluator compilation is going to fail.

narendasan:
@peri044 is looking at casting so we might need to revisit this

PR #309, opened 2021-01-30
Adding support for rsub, min, max, floor_divide

narendasan:
# Description

Adds converters for the following ops:

- aten::rsub
- aten::min
- aten::max
- aten::floor_divide

## Type of change

# Checklist:


core/conversion/converters/impl/element_wise.cpp
```diff
@@ -200,6 +200,61 @@ auto element_wise_registrations TRTORCH_UNUSED =
...
                   }})
+        .pattern({"aten::rsub.Scalar(Tensor self, Scalar other, Scalar alpha=1) -> (Tensor)",
+                  [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
+                    // Should implement other - alpha * self
+                    auto self = args[0].ITensorOrFreeze(ctx);
+                    auto otherScalar = args[1].unwrapToScalar().to<float>();
+                    auto other = tensor_to_const(ctx, torch::tensor({otherScalar}));
+                    auto scalar = args[2].unwrapToScalar().to<float>();
+
+                    if (1 != scalar) {
```
peri044, line 211: I think we should replace scale layer with element wise PROD layer eg: [https://github.com/NVIDIA/TRTorch/blob/9bdb5c81a0ea9be9047b02672cc0e3addf73d67c/core/conversion/converters/impl/element_wise.cpp#L110-L117](url)
This is because scale layer needs input to be atleast 3 dimensions. Related issue and fix:  https://github.com/NVIDIA/TRTorch/pull/249

core/conversion/converters/impl/element_wise.cpp
```diff
@@ -200,6 +200,61 @@ auto element_wise_registrations TRTORCH_UNUSED =
...
+                    return true;
+                  }})
+        .pattern({"aten::rsub.Tensor(Tensor self, Tensor other, Scalar alpha=1) -> (Tensor)",
+                  [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
+                    // Should implement other - alpha * self
+                    auto self = args[0].ITensorOrFreeze(ctx);
+                    auto other = args[1].ITensorOrFreeze(ctx);
+                    auto scalar = args[2].unwrapToScalar().to<float>();
+
+                    if (1 != scalar) {
```
peri044, line 238: Same here as well. Replacing scale layer ?

peri044 (changes requested):

core/conversion/converters/impl/element_wise.cpp
narendasan: @peri044 are you able to add these changes? the branch should be hosted in our repo so we should have access, I am not entirely sure what you mean in this case.

peri044: Replaced instances of scale with elementwise prod at other places as well. Added additional testcases.

PR #310, opened 2021-02-01
support div.Scalar(Tensor self, Scalar other)

inocsin:
- New feature
Support aten::div.Scalar(Tensor self, Scalar other), fix the conflict of aten::div in evaluator
solved this issue https://github.com/NVIDIA/TRTorch/issues/303

# Checklist:

### DIFF
--- a/core/conversion/evaluators/aten.cpp
+++ b/core/conversion/evaluators/aten.cpp
@@ -426,7 +426,8 @@ auto aten_registrations TRTORCH_UNUSED =
                       }
                     },
                     EvalOptions().validSchemas({
-                        "aten::div.Scalar(Scalar a, Scalar b) -> (float)",
+                        "aten::div.float(float a, float b) -> (float)",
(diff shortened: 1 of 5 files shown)


core/conversion/evaluators/aten.cpp
```diff
@@ -426,7 +426,8 @@ auto aten_registrations TRTORCH_UNUSED =
                       }
                     },
                     EvalOptions().validSchemas({
-                        "aten::div.Scalar(Scalar a, Scalar b) -> (float)",
```
narendasan, line 429: Did the schemas here change? are these verified function signatures?

narendasan: We should add evaluator tests to make sure these still work correctly

inocsin: Test case is updated

narendasan (approved):
LGTM

PR #316, opened 2021-02-03
support index=-1, and fix bug when nDims=2 or 3 in softmax

narendasan:
# Description

Lets softmax handle index -1

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/softmax.cpp
+++ b/core/conversion/converters/impl/softmax.cpp
@@ -26,12 +26,17 @@ static auto softmax_registrations TRTORCH_UNUSED = RegisterNodeConversionPattern
        }
 
        int64_t dim = args[1].IValue()->toInt();
+       LOG_DEBUG("Softmax original dim " << dim);
+       if (dim == -1) {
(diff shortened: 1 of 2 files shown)


core/conversion/converters/impl/softmax.cpp
```diff
@@ -26,12 +26,17 @@ static auto softmax_registrations TRTORCH_UNUSED = RegisterNodeConversionPattern
        }
 
        int64_t dim = args[1].IValue()->toInt();
+       LOG_DEBUG("Softmax original dim " << dim);
+       if (dim == -1) {
+         dim = shape.size() - 1;
```
peri044, line 31: can we extend it to handle all negative indices ? `dim = shape.size()  + dim` for dim < 0 or will that case never happens ?

peri044 (changes requested):
Ran it on my local. Works fine. LGTM

core/conversion/converters/impl/softmax.cpp
narendasan: Yeah good change

Issue #306, opened 2021-01-29
✨[Feature] Add support for converter aliasing


narendasan:
**Is your feature request related to a problem? Please describe.**

Right now we require a specific converter definition for each operator we support. In the evaluator library 
we support the ability to handle multiple operator variants with the same evaluator. It would be nice to support this for converters as well.

**Describe the solution you'd like**

We should be able to set more than one operator to be handled by a converter implementation when we register. Something like the following syntax would be helpful. 

```c++
.pattern({{
    "aten::add.Scalar(Tensor self, Scalar other, Scalar alpha=1) -> (Tensor)",
    "aten::add_.Scalar(Tensor self, Scalar other, Scalar alpha=1) -> (Tensor)"},
      [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
        // Should implement self + alpha * other
        auto self = args[0].ITensorOrFreeze(ctx);
        auto otherScalar = args[2].unwrapToScalar().to<float>() * args[1].unwrapToScalar().to<float>();
        auto other = tensor_to_const(ctx, torch::tensor({otherScalar}));
  
        auto add =
            add_elementwise(ctx, nvinfer1::ElementWiseOperation::kSUM, self, other, util::node_info(n));
        TRTORCH_CHECK(add, "Unable to create add layer from node: " << *n);
  
        add->setName(util::node_info(n).c_str());
        auto out = ctx->AssociateValueAndTensor(n->outputs()[0], add->getOutput(0));
        LOG_DEBUG("Output tensor shape: " << out->getDimensions());
        return true;
      }})
```

**Describe alternatives you've considered**

This could also be done via some structure in the lowering passes, we might want to make some infrastructure to make these aliases simple to define. 

**Additional context**

PR #318, opened 2021-02-04
True divide lowering

narendasan:
# Description

Add a pass to lower from true divide to div

## Type of change

# Checklist:


core/lowering/passes/op_aliasing.cpp
```diff
@@ -0,0 +1,33 @@
+#include <torch/csrc/jit/passes/subgraph_rewrite.h>
+
+#include "core/util/prelude.h"
+
+namespace trtorch {
+namespace core {
+namespace lowering {
+namespace passes {
+
+void AliasOperators(std::shared_ptr<torch::jit::Graph>& graph) {
```
peri044, line 10: @narendasan All the alias op patterns (if there are any in future), should make it here in one place right ? If not, we should probably rename the AliasOperators to TrueDivideLowering maybe ?

peri044 (approved):

core/lowering/passes/op_aliasing.cpp
narendasan: Yeah the idea is that we'd list all direct 1:1 aliases in this pass. Id like to build some infra around it so it can be defined by a map or something (#306)

peri044: Sounds good

PR #321, opened 2021-02-04
Add support for bool in elementwise ops 

narendasan:
# Description

Adds support for bool tensors in element wise ops

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/element_wise.cpp
+++ b/core/conversion/converters/impl/element_wise.cpp
@@ -359,8 +359,22 @@ auto element_wise_registrations TRTORCH_UNUSED =
                     // Should implement self * other
                     auto self = args[0].ITensorOrFreeze(ctx);
                     auto other = args[1].ITensorOrFreeze(ctx);
-                    auto mul =
+                    nvinfer1::ILayer* mul = nullptr;
+                    if (self->getType() ==nvinfer1::DataType::kBOOL || other->getType() == nvinfer1::DataType::kBOOL) {
+                      auto self_id = ctx->net->addIdentity(*self);
+                      auto other_id = ctx->net->addIdentity(*other);
+                      if (self->getType() == nvinfer1::DataType::kBOOL) {
+                        self_id->getOutput(0)->setType(nvinfer1::DataType::kINT32);
+                      }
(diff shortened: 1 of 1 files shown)


core/conversion/converters/impl/element_wise.cpp
```diff
@@ -359,8 +359,22 @@ auto element_wise_registrations TRTORCH_UNUSED =
                     // Should implement self * other
                     auto self = args[0].ITensorOrFreeze(ctx);
                     auto other = args[1].ITensorOrFreeze(ctx);
-                    auto mul =
+                    nvinfer1::ILayer* mul = nullptr;
```
narendasan, line 362: @inocsin Are these changes that would only be relevant to mul or other ops in general? Also did you write tests for this? I didnt see them in the commits

inocsin: Currently, I only find the Bool * Int operation in my models. I think other operation like Int / Bool or Int +/- Bool doesn't make  any sense. I will add test case later.

inocsin: The input type of trtorch is set to float in default, I have to modify the TRTorch/core/conversion/conversion.cpp:150 
```
auto trt_in = ctx->net->addInput(name.c_str(), ctx->input_type, dims.input_shape);
```
to support input type of Bool

inocsin: we can try this https://github.com/NVIDIA/TRTorch/pull/327

narendasan: What does it mean to do bool * int? should that give you a bool out? like what does False * 8 mean?

inocsin: [True, False] * [8, 8] = [8, 0], this can be a mask operation, this demo graph here https://github.com/NVIDIA/TRTorch/pull/327

inocsin: we can close this PR if we merge https://github.com/NVIDIA/TRTorch/pull/327

peri044:
We support aten::to operator now which should handle this. Check out https://github.com/NVIDIA/TRTorch/blob/master/tests/core/conversion/converters/test_cast.cpp for usage.

PR #322, opened 2021-02-04
support Long/Double type IValue

narendasan:
# Description

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/var/Var.cpp
+++ b/core/conversion/var/Var.cpp
@@ -94,9 +94,16 @@ nvinfer1::ITensor* Var::ITensorOrFreeze(ConversionCtx* ctx) {
       "Requested either IValue containing a Tensor, or ITensor, however Var type is " << type_name());
 
   nvinfer1::ITensor* out;
-
+  auto weights = converters::Weights();
   if (isIValue()) {
-    auto weights = converters::Weights(ctx, ptr_.ivalue->toTensor());
+    auto tensor = ptr_.ivalue->toTensor();
+    if (tensor.scalar_type() == at::kLong) {
+      weights = converters::Weights(ctx, tensor.toType(at::kInt));
+    } else if (tensor.scalar_type() == at::kDouble) {
+      weights = converters::Weights(ctx, tensor.toType(at::kFloat));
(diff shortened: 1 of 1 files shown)


narendasan:
@inocsin I dont really understand the point of this change? Seems like its casting the input Tensor based on the already existing type of the Tensor and nothing else really changes.

inocsin:
> @inocsin I dont really understand the point of this change? Seems like its casting the input Tensor based on the already existing type of the Tensor and nothing else really changes.

Some values in torchscript are stored as at::kLong and at::kDouble, but the tensorRT doesn't support those two type. So in conversion time, it will throw an error

narendasan:
Isnt this potentially dangerous? We would be truncating values in double precision to single precision and in this case it would be done silently

inocsin:
> Isnt this potentially dangerous? We would be truncating values in double precision to single precision and in this case it would be done silently

Yeah, maybe we can throw a warning? But if we doesn't support the type conversion, the module cann't be compiled by trt

narendasan:
Are tensors from PyTorch in your target module actually double precision? I think its better in that case that we catch this and fully error out telling users to put in a type cast to floating point so they make the choice explicitly.

inocsin:
> Are tensors from PyTorch in your target module actually double precision? I think its better in that case that we catch this and fully error out telling users to put in a type cast to floating point so they make the choice explicitly.

Yeah, the original values are double precision. In my situation, some users may not want to convert precision explicitly, so that's why I think we should do this for them.

narendasan:
We can add a setting to the `CompileSpec` that is something like 

```c++
CompileSpec {
    ... 
    bool truncate_long_and_double = False;
}
```

So while by default you will error out, if you explicitly opt in, weights will be truncated for you. Then we can tell if say an accuracy issue may be the result of truncating or some other issue.

narendasan:
I can PR into your branch the API to do this and the checks if you think this would be sufficient

narendasan:
@peri044 says we should also put a warning in so even if someone opts in you get a message if it actually truncates.

PR #324, opened 2021-02-05
Fix cuda stream initialization when there are no inputs in the graph

peri044:
# Description

When there are no inputs in the graph, cuda stream initialization will fail in the runtime while running TRT engine.
Use case: aten::arange doesn't have input tensor. Test case will be added when @inocsin pushes a converter for arange.

## Type of change

# Checklist:

### DIFF
--- a/core/runtime/register_trt_op.cpp
+++ b/core/runtime/register_trt_op.cpp
@@ -48,7 +48,16 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
     gpu_handles.push_back(outputs[pyt_idx].data_ptr());
   }
 
-  c10::cuda::CUDAStream stream = c10::cuda::getCurrentCUDAStream(inputs[0].device().index());
+  LOG_DEBUG("Initializing CUDA stream");
+  c10::cuda::CUDAStream stream = at::cuda::getStreamFromPool();
+  if (inputs.size() > 0) {
+    stream = c10::cuda::getCurrentCUDAStream(inputs[0].device().index());
+  } else if (outputs.size() > 0) {
+    stream = c10::cuda::getCurrentCUDAStream(outputs[0].device().index());
+  } else {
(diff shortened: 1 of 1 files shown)


core/runtime/register_trt_op.cpp
```diff
@@ -48,7 +48,16 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
     gpu_handles.push_back(outputs[pyt_idx].data_ptr());
   }
 
-  c10::cuda::CUDAStream stream = c10::cuda::getCurrentCUDAStream(inputs[0].device().index());
+  LOG_DEBUG("Initializing CUDA stream");
+  c10::cuda::CUDAStream stream = at::cuda::getStreamFromPool();
```
narendasan, line 52: Can you explain the logic here? Is it just picking a stream from the device that an input tensor is on? Why would there be a case with only outputs? Shouldnt output tensors be on the same stream as input tensors?

peri044: So previous code used to pick a stream from a device that an input tensor is on. But in case when there are no inputs to the graph (eg: [https://github.com/NVIDIA/TRTorch/pull/325/files#diff-6110deeedfe275a7e94ffae57aca214c81894a8920c2c4963fff2cb30f5d12dcR28](url) ) there are no inputs and I try picking the stream from the outputs. In this arange case, inputs would be empty and receive a seg fault from the c10 call. 
In real examples, most likely we would have inputs as at::arange would be a part of some larger graph. But we should handle all cases usually.

narendasan: what is a case where there are no inputs to the graph?

peri044: Looks like the link got expired. Test cases in this pull request : [https://github.com/NVIDIA/TRTorch/pull/325](url)

peri044:
We shall resolve this PR once an evaluator implementation is done for arange [https://github.com/NVIDIA/TRTorch/pull/325](url)

PR #327, opened 2021-02-05
Convert the output type of bool operation to int

inocsin:
# Description
There is a case 
```
code
  x1 = torch.to(torch.ge(x0, 0.5), torch.device("cuda:0"), 3, False, False)
  _136 = torch.sum(torch.mul(x1, CONSTANTS.c0), [1], False)

graph
  %1284 : Tensor = prim::Constant[value=<Tensor>]()
  %1275 : Tensor = aten::ge(%x0.1, %1274)
  %x1.1 : Tensor = aten::to(%1275, %1315, %1278, %484, %484, %486)
  %1285 : Tensor = aten::mul(%x1.1, %1284)
```
The output of aten::ge is Bool type, and there are three way to support the above graph
1. implement aten::to, to convert datatype from Bool to Int
2. convert datatype in aten::mul (see https://github.com/NVIDIA/TRTorch/pull/321), but I found the trtorch framework doesn't support specifying input data type, so I cann't test this method
3. convert the output type of Boolean Operation (that's what this PR does)

## Type of change

Convert ouput type from bool to int

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/element_wise.cpp
+++ b/core/conversion/converters/impl/element_wise.cpp
@@ -9,6 +9,14 @@ namespace converters {
 namespace impl {
 namespace {
 
+nvinfer1::IIdentityLayer* convert_to_int(nvinfer1::ILayer* op, ConversionCtx* ctx, const torch::jit::Node* n) {
+  auto identity = ctx->net->addIdentity(*op->getOutput(0));
+  identity->setOutputType(0, nvinfer1::DataType::kINT32);
+  TRTORCH_CHECK(identity, "Unable to create Identity layer from node: " << *n);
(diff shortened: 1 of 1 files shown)


narendasan:
Can you write test cases for this?

inocsin:
> Can you write test cases for this?

I think we already have test cases for boolean operation

developer0hye:
Does this code only works when we set precision to INT?

When I update my code with your code, 
image

developer0hye:
Please check my pull #350

peri044:
We now support aten::to operator which handles such masked multiplications. Closing this PR

PR #328, opened 2021-02-05
support adaptive_max_pooling, fix adpative_avg_pooling bug

inocsin:
# Description

## Type of change
Fix aten::adaptive_avg_pool2d bug when input size is not integer multiple of output size, update the test case
Support aten::adaptive_max_pool2d with test case
-Bug fix
Remove "#if NV_TENSORRT_MAJOR < 7 || (NV_TENSORRT_MAJOR == 7 && NV_TENSORRT_MINOR < 1)" part.
Since the TensorRT dependency is updated, the original GPU part will never be used, all the operator will be forced to run on cpu. So I remove the TensorRT flag. But the GPU part failed with unknow reason after the flag was removed, so I refactored the gpu part, now it can run on GPU correctly.

# Checklist:


narendasan:
Can you apply the linting?

inocsin:
> Can you apply the linting?

applied linting

peri044:
The tests are running fine. It looks like the existing Interpolate plugin already calls `at::adaptive_avg_pool2d`  by using the mode `adaptive_pool2d` [https://github.com/NVIDIA/TRTorch/blob/master/core/conversion/converters/impl/plugins/interpolate_plugin.cpp#L329](url)
Can we make use of the mode flag to  support adaptive_max_pool2d as well ? Does this require a separate plugin in this PR?

inocsin:
> The tests are running fine. It looks like the existing Interpolate plugin already calls `at::adaptive_avg_pool2d` by using the mode `adaptive_pool2d` [https://github.com/NVIDIA/TRTorch/blob/master/core/conversion/converters/impl/plugins/interpolate_plugin.cpp#L329](url)
> Can we make use of the mode flag to support adaptive_max_pool2d as well ? Does this require a separate plugin in this PR?

The output number, datatype and parameter number of adaptive_max_pool are different from interpolate_plugin, in that case should we use several if-else to merge these two plugin into one?

inocsin:
@narendasan @peri044 please review the update, I have fixed some bugs

peri044:
Thanks @inocsin Merged these changes into the new plugins interface. Tried to cherry pick your commits but there are lot of changes in the new API.

PR #329, opened 2021-02-08
add clamp_min/clamp_max converter

inocsin:
# Description

Fixes https://github.com/NVIDIA/TRTorch/issues/274

## Type of change

Support aten::clamp_min, aten::clamp_max converters

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/element_wise.cpp
+++ b/core/conversion/converters/impl/element_wise.cpp
@@ -164,12 +164,20 @@ auto element_wise_registrations TRTORCH_UNUSED =
                     // Compute min(max(min_threshold, input), max_threshold)
                     auto self = args[0].ITensorOrFreeze(ctx);
                     auto clamp_layer_out = self;
-                    if (args[1].isIValue() && args[1].IValue()->isScalar()) {
+
+                    if (args[1].isIValue() && args[1].IValue()->isScalar() && args[2].isIValue() &&
+                        args[2].IValue()->isScalar()) {
+                      auto alpha = args[1].unwrapToScalar().to<float>();
+                      auto beta = args[2].unwrapToScalar().to<float>();
+                      auto clip_layer = ctx->net->addActivation(*self, nvinfer1::ActivationType::kCLIP);
+                      TRTORCH_CHECK(clip_layer, "Unable to create clip layer for node: " << *n);
+                      clip_layer->setAlpha(alpha);
+                      clip_layer->setBeta(beta);
+                      clamp_layer_out = clip_layer->getOutput(0);
+                    } else if (args[1].isIValue() && args[1].IValue()->isScalar()) {
                       auto limit = args[1].unwrapToScalar().to<float>();
                       clamp_layer_out = clamp_util(ctx, n, self, limit, nvinfer1::ElementWiseOperation::kMAX, "_max");
-                    }
-
-                    if (args[2].isIValue() && args[2].IValue()->isScalar()) {
+                    } else if (args[2].isIValue() && args[2].IValue()->isScalar()) {
                       auto limit = args[2].unwrapToScalar().to<float>();
                       clamp_layer_out = clamp_util(ctx, n, self, limit, nvinfer1::ElementWiseOperation::kMIN, "_min");
                     }
--- a/core/conversion/converters/impl/element_wise.cpp
+++ b/core/conversion/converters/impl/element_wise.cpp
@@ -179,6 +179,51 @@ auto element_wise_registrations TRTORCH_UNUSED =
                     LOG_DEBUG("Clamp layer output tensor shape: " << out->getDimensions());
                     return true;
                   }})
+        .pattern({"aten::clamp_min(Tensor self, Scalar min) -> (Tensor)",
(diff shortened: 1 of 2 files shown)


peri044 (changes requested):

core/conversion/converters/impl/element_wise.cpp
```diff
@@ -179,6 +179,51 @@ auto element_wise_registrations TRTORCH_UNUSED =
                     LOG_DEBUG("Clamp layer output tensor shape: " << out->getDimensions());
                     return true;
                   }})
+        .pattern({"aten::clamp_min(Tensor self, Scalar min) -> (Tensor)",
```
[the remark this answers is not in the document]
inocsin: Yeah, i will try this

peri044 (approved):
LGTM

narendasan:
Should we be using CLIP here? Is there a performance difference? https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/namespacenvinfer1.html#a4806465e7c2bba8afc3f0388f538d2da.

peri044:
> Should we be using CLIP here? Is there a performance difference? https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/namespacenvinfer1.html#a4806465e7c2bba8afc3f0388f538d2da.

Good idea. Forgot we had such activation. We should use clip. I think both min and max operations would be fused that way compared to separate invocations which can lead to better performance. 
for aten::clamp(min, max), it can improve perf
for aten::clamp_max and aten::clamp_min, the current code and using CLIP might be more or less the same. 1 kernel invocation in both but one extra comparison/per element in CLIP. Not sure how much this will affect performance.

peri044 (approved):
LGTM

tests/core/conversion/converters/test_element_wise.cpp
```diff
@@ -241,7 +241,7 @@ TEST(Converters, ATenRsubWithScalarConvertsCorrectly) {
 TEST(Converters, ATenClampMinConvertsCorrectly) {
   const auto graph = R"IR(
   graph(%x.1 : Tensor):
-          %2 : int = prim::Constant[value=-2]()
+          %2 : int = prim::Constant[value=1.5]()
```
narendasan, line 244: should this be float or int?

tests/core/conversion/converters/test_element_wise.cpp
```diff
@@ -251,7 +251,7 @@ TEST(Converters, ATenClampMinConvertsCorrectly) {
 TEST(Converters, ATenClampMaxConvertsCorrectly) {
   const auto graph = R"IR(
   graph(%x.1 : Tensor):
-          %2 : int = prim::Constant[value=3]()
+          %2 : int = prim::Constant[value=3.5]()
```
narendasan, line 254: same here

tests/core/conversion/converters/test_element_wise.cpp
```diff
@@ -261,13 +261,31 @@ TEST(Converters, ATenClampMaxConvertsCorrectly) {
 TEST(Converters, ATenClampMinMaxConvertsCorrectly) {
   const auto graph = R"IR(
   graph(%x.1 : Tensor):
-          %2 : int = prim::Constant[value=3]()
-          %3 : int = prim::Constant[value=-2]()
+          %2 : int = prim::Constant[value=3.5]()
+          %3 : int = prim::Constant[value=1.5]()
```
narendasan, line 265: These too as well

tests/core/conversion/converters/test_element_wise.cpp
```diff
@@ -261,13 +261,31 @@ TEST(Converters, ATenClampMaxConvertsCorrectly) {
 TEST(Converters, ATenClampMinMaxConvertsCorrectly) {
   const auto graph = R"IR(
   graph(%x.1 : Tensor):
-          %2 : int = prim::Constant[value=3]()
-          %3 : int = prim::Constant[value=-2]()
```
narendasan, line 265: Do we want a new negative number case?

narendasan (changes requested):

tests/core/conversion/converters/test_element_wise.cpp
inocsin: input range is [1,5), in that case, -2 is never used

tests/core/conversion/converters/test_element_wise.cpp
inocsin: I don't think so, let the input range larger([1,5])than min(1.5)/max(3.5) will cover all situation

tests/core/conversion/converters/test_element_wise.cpp
narendasan: sure but the output type should be float if the value is 1.5 right?

tests/core/conversion/converters/test_element_wise.cpp
narendasan: ok

tests/core/conversion/converters/test_element_wise.cpp
inocsin: updated

peri044 (approved):

narendasan (approved):
lgtm

PR #349, opened 2021-02-18
Add an example to compile converter into .so and run it in Python

bowang95:
# Description

Create ELU converter in C++, compile it to .so library and load it in Python

## Type of change

# Checklist:


peri044 (changes requested):

examples/README.md
```diff
@@ -0,0 +1,177 @@
...
+We can register a converter for this operator in our application. You can find more 
+information on all the details of writing converters in the contributors documentation
+([Writing Converters](https://nvidia.github.io/TRTorch/contributors/writing_converters.html)).
+Once we are clear about these rules and writing patterns, we can create a seperate new C++ source file as:
+
+```c++
+#include "core/conversion/converters/converters.h"
+#include "core/util/prelude.h"
+
+namespace trtorch {
```
narendasan, line 25: Lets use a separate namespace something like 

```c++
namespace my_custom_converters {
```

examples/README.md
```diff
@@ -0,0 +1,177 @@
...
+namespace core {
+namespace conversion {
+namespace converters {
+namespace impl {
+namespace {
+
+auto acthardtanh TRTORCH_UNUSED = RegisterNodeConversionPatterns().pattern(
```
narendasan, line 32: probably use something like `auto actelu` here, you probably shouldnt use trtorch macros in this custom extention

examples/README.md
```diff
@@ -0,0 +1,177 @@
...
+## Load `.so` in Python Application
+With the new generated library, TRTorch now support the new developed converter. 
+We use `torch.ops.load_library` to load `.so`. For example, we could load the ELU 
+converter and use it in our application:
+```python
+import torch
+import trtorch
+
+# After "python3 setup install", you should find this .so file under generated "build" directory
+torch.ops.load_library('./build/lib.linux-x86_64-3.6/elu_converter.cpython-36m-x86_64-linux-gnu.so')
```
narendasan, line 110: I think you should also mention that you can also import the .so with `import elu_converter` (verify this works first)

examples/README.md
```diff
@@ -0,0 +1,177 @@
...
+    print('TRTorch output: \n', trtorch_out[0, :, :, 0])
+    MaxDiff(pytorch_out, trtorch_out)
+
+
+if __name__ == "__main__":
+    main()
+
+```
+Run this script, we can get the Tensor before and after ELU operator.
+### Example Output
```
narendasan, line 159: We dont need this section i think, we can just say it works

examples/elu_converter/elu_converter.cpp
```diff
@@ -0,0 +1,33 @@
+#include "core/conversion/converters/converters.h"
+#include "core/util/prelude.h"
+
+namespace trtorch {
```
narendasan, line 4: Same comments as above, use a separate namespace

examples/elu_converter/elu_converter.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "core/util/prelude.h"
+
+namespace trtorch {
+namespace core {
+namespace conversion {
+namespace converters {
+namespace impl {
+namespace {
+
+auto acthardtanh TRTORCH_UNUSED = RegisterNodeConversionPatterns().pattern(
```
narendasan, line 11: I dont think we should use the TRTorch macros in this

narendasan, line 11: use something like `auto actelu`

examples/elu_converter/elu_converter.cpp
```diff
@@ -0,0 +1,33 @@
...
+    {"aten::elu(Tensor self, Scalar alpha=1, Scalar scale=1, Scalar input_scale=1) -> (Tensor)",
+     [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
+       auto in = args[0].ITensorOrFreeze(ctx);
+       auto alpha = args[1].unwrapToDouble();
+
+       auto new_layer = ctx->net->addActivation(*in, nvinfer1::ActivationType::kELU);
+       TRTORCH_CHECK(new_layer, "Unable to create layer for aten::elu");
```
narendasan, line 18: Do custom error checking here instead of using macros

examples/elu_converter/elu_converter.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+       new_layer->setAlpha(alpha);
+       new_layer->setName(util::node_info(n).c_str());
+       auto out_tensor = ctx->AssociateValueAndTensor(n->outputs()[0], new_layer->getOutput(0));
+
+       LOG_DEBUG("Output shape: " << out_tensor->getDimensions());
```
narendasan, line 24: You dont really need this

examples/README.md
```diff
@@ -0,0 +1,177 @@
+# Create a new op in C++, compile it to .so library and load it in Python
```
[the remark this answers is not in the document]
bowang95: Do we need to move this new created "custom_converter" folder in some other directories?  like "notebooks" ? Since "examples" directory is new created for this task.

examples/README.md
bowang95: It seems that `import elu_converter` is not supported by this setting, it shows an error as: `ImportError: dynamic module does not define module export function (PyInit_elu_converter)`. I googled this bug, found that this happens often because of python2/3 linking problem. I tried to solve this by different methods but got no progress..

examples/README.md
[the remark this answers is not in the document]
narendasan: I think the folder structure should be something like 

```
//examples 
    - custom_converters/
        - README.md
        - elu_model.py (your elu_converter_test.py)
        - elu_converter/
            - setup.py
            - elu_converter/
                - csrc/
                    - elu_converter.cpp 
```

examples/README.md
narendasan: No problem, its not a big deal

examples/custom_converters/elu_model.py
```diff
@@ -0,0 +1,50 @@
...
+    }
+    trt_ts_module = trtorch.compile(scripted_model, compile_settings)
+    input_data = torch.randn((1024, 1, 32, 32))
+    input_data = input_data.half().to("cuda")
+    pytorch_out = model.forward(input_data)
+
+    trtorch_out = trt_ts_module(input_data)
+    print('PyTorch output: \n', pytorch_out[0, :, :, 0])
+    print('TRTorch output: \n', trtorch_out[0, :, :, 0])
+    MaxDiff(pytorch_out, trtorch_out)
```
peri044, line 46: MaxDiff -> max_diff  maybe? This is just a function. I don't know if python linting will correct this. Does the existing linter applies linting on this example ? If it's not too much work, it would be nice to setup linter to apply to this directory as well.

peri044 (approved):

peri044:
@narendasan What are your thoughts on adding this elu converter to our core library as well? This standalone converter will serve as a demo for users to try for other unsupported ops.

examples/custom_converters/elu_converter/setup.py
```diff
@@ -0,0 +1,24 @@
...
+from setuptools import setup, Extension
+from torch.utils import cpp_extension
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+# library_dirs should point to the libtrtorch.so, include_dirs should point to the dir that include the headers
+# 1) download the latest package from https://github.com/NVIDIA/TRTorch/releases/
+# 2) Extract the file from downloaded package, we will get the "trtorch" directory
+# 3) Set trtorch_path to that directory
+trtorch_path = os.path.abspath("/home/bowa/Downloads/trtorch")
```
narendasan, line 11: We should set this to some invalid string so people have to change it, or have the code check to see its changed from default instead of your particular downloads dir

bowang95: sorry I missed that, I think it's ok now

PR #370, opened 2021-02-24
more safe and easily readable use to get raw pointer of shared-pointer

developer0hye:
Signed-off-by: Yonghye Kwon <developer.0hye@gmail.com>

# Description
## before
```cpp
  auto g = std::make_shared<torch::jit::Graph>();
  torch::jit::parseIR(graph, &*g);
```

## after
```cpp
  auto g = std::make_shared<torch::jit::Graph>();
  torch::jit::parseIR(graph, g.get());
```

Refer to [this thread](https://stackoverflow.com/questions/505143/getting-a-normal-ptr-from-shared-ptr) uploaded on Stack Overflow.

## Type of change

# Checklist:

### DIFF
--- a/tests/core/conversion/converters/test_activation.cpp
+++ b/tests/core/conversion/converters/test_activation.cpp
@@ -11,7 +11,7 @@ TEST(Converters, ATenReLUConvertsCorrectly) {
         return (%3))IR";
 
   auto g = std::make_shared<torch::jit::Graph>();
-  torch::jit::parseIR(graph, &*g);
+  torch::jit::parseIR(graph, g.get());
 
   auto in = at::randint(-5, 5, {5}, {at::kCUDA});
   auto params = trtorch::core::conversion::get_named_params(g->inputs(), {});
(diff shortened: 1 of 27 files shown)


developer0hye:
@narendasan Is there any reason using '&*' instead  of '.get())?

narendasan:
The structure of tests has its roots in how PyTorch does JIT tests, likely they did the same thing and we copied it. Places like this https://github.com/pytorch/pytorch/blob/c7a70eec1b1a8aaac1f5659f688cd625c4a67831/test/cpp/tensorexpr/test_kernel.cpp#L42. Though I now see both in the PyTorch codebase. Im not highly opinionated on what way we get the raw pointer and seems reasonable if its safer

narendasan:
Is this really a breaking change?

developer0hye:
@narendasan 
I was not thinking straight.
I was not familiar with `shared pointer`. I studied this recently.
It's not a breaking change.

PR #371, opened 2021-02-24
Anuragd/multi-gpu (//py): Fixed multi-gpu scenario with Python set_device API support

### LINKED ISSUE #356: 🐛 [Bug] An error occurs in CompileGraph when gpu_id == 1
When I tried to Complie on the second GPU in a multi-GPU environment, an error occurred. The code sample used is as follows.

```cpp
void load(const std::string& model_path, int64_t gpu_id, int64_t opt_batch_size) {
    torch::jit::Module module = torch::jit::load(model_path);
    torch::Device device = (torch::cuda::is_available() ? torch::Device(torch::kCUDA, gpu_id) : torch::Device(torch::kCPU));
    module.to(device, torch::kHalf);
    module.eval();

    std::vector<int64_t> in_opt = { opt_batch_size, INPUT_CHANNEL_NUM, BOARD_WIDTH, BOARD_WIDTH };

    trtorch::CompileSpec::InputRange range(in_opt);
    trtorch::CompileSpec info({ range });
    info.op_precision = torch::kHalf;
    info.device.gpu_id = gpu_id;
    module = trtorch::CompileGraph(module, info);
}
```

#### Error1
I called this function with gpu_id = 1. I got the following error:

```
terminate called after throwing an instance of 'trtorch::Error'
  what():  [enforce fail at core/conversion/conversionctx/ConversionCtx.cpp:107] Expected cudaSetDevice(settings.device.gpu_id) to be true but got false
Unable to set gpu id: 1
```

I think this line is the cause.
https://github.com/NVIDIA/TRTorch/blob/1d4b967a28e36beee048703f5645ee6fcc95793d/core/conversion/conversionctx/ConversionCtx.cpp#L112

`cudaSetDevice` returns `cudaSuccess` (= 0) on success. However, `TRTORCH_CHECK` judges success or failure as a Boolean type.

I fixed it as follows and rebuilt it so that this error disappeared.

```diff
diff --git a/core/conversion/conversionctx/ConversionCtx.cpp b/core/conversion/conversionctx/ConversionCtx.cpp
index ff23692..bc5bf68 100644
--- a/core/conversion/conversionctx/ConversionCtx.cpp
+++ b/core/conversion/conversionctx/ConversionCtx.cpp
@@ -109,7 +109,7 @@ ConversionCtx::ConversionCtx(BuilderSettings build_settings)
   cfg->setEngineCapability(settings.capability);
 
   if (settings.device.gpu_id) {
-    TRTORCH_CHECK(cudaSetDevice(settings.device.gpu_id), "Unable to set gpu id: " << settings.device.gpu_id);
+    TRTORCH_CHECK(cudaSetDevice(settings.device.gpu_id) == cudaSuccess, "Unable to set gpu id: " << settings.device.gpu_id);
   }
 
   if (settings.device.device_type == nvinfer1::DeviceType::kDLA) {
```

You may also use `set_device`.

https://github.com/NVIDIA/TRTorch/blob/1d4b967a28e36beee048703f5645ee6fcc95793d/core/compiler.cpp#L176-L178

#### Error2
After making the above fix, I get the following error:

```
ERROR: [TRTorch Conversion Context] - Builder was created on device different than current device.
```

I changed `cudaSetDevice` to do it at the beginning of the function and it worked fine.

```diff
diff --git a/core/conversion/conversionctx/ConversionCtx.cpp b/core/conversion/conversionctx/ConversionCtx.cpp
index ff23692..09a419c 100644
--- a/core/conversion/conversionctx/ConversionCtx.cpp
+++ b/core/conversion/conversionctx/ConversionCtx.cpp
@@ -47,6 +47,10 @@ ConversionCtx::ConversionCtx(BuilderSettings build_settings)
           util::logging::get_logger().get_reportable_severity(),
           util::logging::get_logger().get_is_colored_output_on()) {
   // TODO: Support FP16 and FP32 from JIT information
+  if (settings.device.gpu_id) {
+    TRTORCH_CHECK(cudaSetDevice(settings.device.gpu_id) == cudaSuccess, "Unable to set gpu id: " << settings.device.gpu_id);
+  }
+
   builder = nvinfer1::createInferBuilder(logger);
   net = builder->createNetworkV2(1U << static_cast<uint32_t>(nvinfer1::NetworkDefinitionCreationFlag::kEXPLICIT_BATCH));
 
@@ -108,10 +112,6 @@ ConversionCtx::ConversionCtx(BuilderSettings build_settings)
   cfg->setDefaultDeviceType(settings.device.device_type);
   cfg->setEngineCapability(settings.capability);
 
-  if (settings.device.gpu_id) {
-    TRTORCH_CHECK(cudaSetDevice(settings.device.gpu_id), "Unable to set gpu id: " << settings.device.gpu_id);
-  }
-
   if (settings.device.device_type == nvinfer1::DeviceType::kDLA) {
     auto nbDLACores = builder->getNbDLACores();
     TRTORCH_CHECK(
```

It's working, but I'm not sure if this is a good fix as there may be other side effects as well.
I would appreciate it if you could respond appropriately.

andi4191:
# Description

Fix for multi-gpu configuration where the compilation of graph fails.
Root cause: Implicit typecasting of cudaSucess as boolean. Incorrect cuda context before setting the device ID (ConversionCtx)

Fixes #356

## Type of change
- Bug fix (non-breaking))

# Checklist:


tests/py/multi_gpu_test_case.py
```diff
@@ -0,0 +1,21 @@
+import unittest
```
narendasan, line 1: Does this need to be duplicated? can we just use ModelTestCase?

tests/py/test_api_multi_gpu.py
```diff
@@ -0,0 +1,66 @@
...
+import torch
+import torchvision.models as models
+
+from multi_gpu_test_case import MultiGpuTestCase
+
+gpu_id = 1
+class TestCompile(MultiGpuTestCase):
+
+    def setUp(self):
+        self.input = torch.randn((1, 3, 224, 224)).to("cuda")
```
narendasan, line 12: Id say just put gpu_id here like `self.gpu_id = 1`

narendasan (commented):
Minor issues otherwise lgtm

tests/py/test_api_multi_gpu.py
andi4191: I also need the trtorch.set_device(gpu_id) outside the class. Hence, re-used the global var. Single source of truth

tests/py/multi_gpu_test_case.py
andi4191: More descriptive

narendasan: This part doesnt need to be more descriptive, since its a generic test case framework. the actual test class (TestCompile) needs to be more discriptive

tests/py/test_api_multi_gpu.py
narendasan: You should set device during setup. Also shouldnt this test that the device switches properly?

PR #372, opened 2021-02-24
support cumsum converter

### LINKED ISSUE #269: Converters request
```
aten::bmm(Tensor   self, Tensor mat2) -> (Tensor)
aten::pixel_shuffle(Tensor   self, int upscale_factor) -> (Tensor)
aten::replication_pad2d(Tensor   self, int[4] padding) -> (Tensor)
aten::feature_dropout(Tensor   input, float p, bool train) -> (Tensor)
aten::clamp_min(Tensor   self, Scalar min) -> (Tensor)
aten::expand_as(Tensor(a)   self, Tensor other) -> (Tensor(a))
aten::transpose.int(Tensor(a)   self, int dim0, int dim1) -> (Tensor(a))
aten::_cast_Float(Tensor self, bool   non_blocking=False) -> (Tensor)
aten::einsum(str equation, Tensor[]   tensors) -> (Tensor)
aten::instance_norm(Tensor   input, Tensor? weight, Tensor? bias, Tensor? running_mean, Tensor?   running_var, bool use_input_stats, float momentum, float eps, bool   cudnn_enabled) -> (Tensor)
aten::arange(Scalar end, *, int?   dtype=None, int? layout=None, Device? device=None, bool? pin_memory=None)   -> (Tensor)
aten::expand(Tensor(a) self, int[] size,   *, bool implicit=False) -> (Tensor(a))
aten::gelu(Tensor self) -> (Tensor)
aten::layer_norm(Tensor input, int[]   normalized_shape, Tensor? weight=None, Tensor? bias=None, float   eps=1.0e-05, bool cudnn_enable=True) -> (Tensor)
aten::ne.Scalar(Tensor self,   Scalar other) -> (Tensor)
aten::transpose.int(Tensor(a) self, int   dim0, int dim1) -> (Tensor(a))
```

guoruoqian:

# Description

Support cumsum converter.  Cumsum layer will be run through ATen, not TensorRT.

Fixes #269 

## Type of change

# Checklist:


peri044:
We plan to revisit plugins due to some design changes planned. But cumsum operation is reducing sum across a dimension right? 
Won't this https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_reduce_layer.html#details layer in TensorRT work for this converter or do we need a plugin here ?  If we can express this op as a composition of existing ops that TRTorch supports, we probably should do that I think.

guoruoqian:
> We plan to revisit plugins due to some design changes planned. But cumsum operation is reducing sum across a dimension right?
> Won't this https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_reduce_layer.html#details layer in TensorRT work for this converter or do we need a plugin here ? If we can express this op as a composition of existing ops that TRTorch supports, we probably should do that I think.

The cumsum operation is a Prefix Sum (Scan) . As far as i know, it is already well optimized in many libraries, such as [thrust::inclusive_scan](https://thrust.github.io/doc/group__prefixsums_gafb24ad76101263038b0acaddc094d70a.html#gafb24ad76101263038b0acaddc094d70a).  I don't think it's more effieient to implement it directly with `nvinfer1::IReduceLayer` than with `aten::cumsum`.

narendasan:
We should implement support in TRTorch with native TRT if it its possible vs making custom plugins

guoruoqian:
> We should implement support in TRTorch with native TRT if it its possible vs making custom plugins

Yeah, i have implemented cumsum converter with native TRT. Now, it doesn't need plugins.

PR #378, opened 2021-02-25
fix(//core/conversion/converters/Weights): Fix buffer allocation for weights data 

### LINKED ISSUE #326: ❓ [Question] Is there a way to do multithreaded half-precision compilation?
## ❓ Question

I want to compile a Torch script in a different thread than the main thread in a C++ program. However, doing so with half precision for large networks will result in a Segmentation fault.

Here's a program that extracts what I want to do:
https://github.com/SakodaShintaro/trtorch-test/blob/master/main.cpp

```cpp
#include <torch/script.h>
#include <trtorch/trtorch.h>
using namespace std;

void compile(bool fp16) {
  constexpr int64_t INPUT_CHANNEL_NUM = 256;
  constexpr int64_t WIDTH = 32;
  torch::jit::Module module = torch::jit::load("model.ts");
  if (fp16) {
    module.to(torch::kCUDA, torch::kHalf);
  } else {
    module.to(torch::kCUDA);
  }
  module.eval();

  std::vector<int64_t> in_sizes = {1, INPUT_CHANNEL_NUM, WIDTH, WIDTH};
  trtorch::CompileSpec::InputRange range(in_sizes);
  trtorch::CompileSpec info({range});
  if (fp16) {
    info.op_precision = torch::kHalf;
  }
  module = trtorch::CompileGraph(module, info);
}

int main() {
  // fp32, this thread -> OK
  compile(false);
  cout << "fp32, this thread -> finish" << endl;

  // fp32, another thread -> OK
  std::thread thread0([]() { compile(false); });
  thread0.join();
  cout << "fp32, another thread -> finish" << endl;

  // fp16, this thread -> OK
  compile(true);
  cout << "fp16, this thread -> finish" << endl;

  // fp16, another thread -> NG
  std::thread thread1([]() { compile(true); });
  thread1.join();
  cout << "fp16, another thread -> finish" << endl;
}
```

 result

```
fp32, this thread -> finish
fp32, another thread -> finish
fp16, this thread -> finish
Segmentation fault (core dumped)
```

 Is there anything wrong with my code?

## Environment
 I used a Dockerfile I made.
https://github.com/SakodaShintaro/trtorch-test/blob/master/docker/Dockerfile

If I create a container with this image and execute `./Test`, a Segmentation fault will occur on the 4th line.

In `trtorch-test/docker`,

```
docker build -t trtorch_test_image .
docker run --gpus all -it --name trtorch_test_container trtorch_test_image:latest bash
./Test
```

I sometimes succeed in it, so try it a few times if you want to reproduce it.

 - PyTorch Version (e.g., 1.0): 1.7
 - CPU Architecture: x86_64
 - OS (e.g., Linux): Ubuntu 20.04 (on Docker)
 - Build command you used (if compiling from source): bazel build //:libtrtorch --compilation_mode opt
 - CUDA version: 11.0
 - GPU models and configuration: RTX 2080ti
 - Nvidia driver version : 460

narendasan:
# Description

The memcpy for weights tensors over copies when the Tensor is not a FP32 tensor. This seems to be the root cause of segfaults observed in #326 as well as failing DLA tests on aarch64 when users try to compile a model that is already in FP16. 

Fixes #326 

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/Weights.cpp
+++ b/core/conversion/converters/Weights.cpp
@@ -81,7 +81,7 @@ Weights::Weights(ConversionCtx* ctx, at::Tensor t) {
   // Store the data in the conversion context so it remains until building is
   // complete
 
-  void* buf;
+  void* buf = nullptr;
 
   if (dtype_optional.value() == nvinfer1::DataType::kFLOAT) {
     buf = malloc(t_cpu.numel() * sizeof(float));
--- a/core/conversion/converters/Weights.cpp
+++ b/core/conversion/converters/Weights.cpp
@@ -80,9 +80,29 @@ Weights::Weights(ConversionCtx* ctx, at::Tensor t) {
 
   // Store the data in the conversion context so it remains until building is
   // complete
-  void* buf = malloc(t_cpu.numel() * sizeof(float));
+
+  void* buf;
+
+  if (dtype_optional.value() == nvinfer1::DataType::kFLOAT) {
+    buf = malloc(t_cpu.numel() * sizeof(float));
+    memcpy(buf, t_cpu.data_ptr(), t_cpu.numel() * sizeof(float));
+  } else if (dtype_optional.value() == nvinfer1::DataType::kHALF) {
+    buf = malloc(t_cpu.numel() * (sizeof(float) / 2));
+    memcpy(buf, t_cpu.data_ptr(), t_cpu.numel() * (sizeof(float) / 2));
+  } else if (dtype_optional.value() == nvinfer1::DataType::kINT8) {
+    buf = malloc(t_cpu.numel() * sizeof(char));
+    memcpy(buf, t_cpu.data_ptr(), t_cpu.numel() * sizeof(char));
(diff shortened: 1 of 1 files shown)


narendasan:
@andi4191 can you verify this change fixes the observed segfaults?

narendasan:
I would still say that the compiler should still not be considered thread-safe yet. There's at least a few parts that may still have issues (Logger, Runtime).

andi4191:
Tested it for a few iterations. It passes everytime.

core/conversion/converters/Weights.cpp
```diff
@@ -80,9 +80,29 @@ Weights::Weights(ConversionCtx* ctx, at::Tensor t) {
 
   // Store the data in the conversion context so it remains until building is
   // complete
-  void* buf = malloc(t_cpu.numel() * sizeof(float));
+
+  void* buf;
```
andi4191, line 84: Should be initialized to nullptr.

andi4191 (approved):
Minor comment. LGTM

PR #398, opened 2021-03-16
Adds support for PTQ through the PyTorch to_backend api.

narendasan:
# Description
Adds support for PTQ through the PyTorch to_backend api.

## Type of change

# Checklist:

### DIFF
--- a/py/trtorch/csrc/tensorrt_classes.h
+++ b/py/trtorch/csrc/tensorrt_classes.h
@@ -94,8 +94,16 @@ struct CompileSpec : torch::CustomClassHolder {
     input_ranges.push_back(*ir);
   }
 
+  int64_t getPTQCalibratorHandle() {
+    return (int64_t)ptq_calibrator;
+  }
+
   void setDeviceIntrusive(const c10::intrusive_ptr<Device>& d) {
     device = *d;
+  }
+
+  void setPTQCalibratorViaHandle(int64_t handle) {
+    ptq_calibrator = (nvinfer1::IInt8Calibrator*)handle;
   }
 
   ADD_ENUM_GET_SET(op_precision, DataType, static_cast<int64_t>(DataType::kChar));
--- a/tests/py/test_ptq_to_backend.py
+++ b/tests/py/test_ptq_to_backend.py
@@ -1,0 +1,97 @@
+import unittest
+import trtorch
+from trtorch.logging import *
+import torch
+import torch.nn as nn
+from torch.nn import functional as F
+import torchvision
+import torchvision.transforms as transforms
+from model_test_case import ModelTestCase
+
+
+class TestAccuracy(ModelTestCase):
+
+    def setUp(self):
+        self.input = torch.randn((1, 3, 32, 32)).to("cuda")
+        self.testing_dataset = torchvision.datasets.CIFAR10(root='./data',
+                                                            train=False,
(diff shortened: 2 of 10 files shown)


tests/py/test_ptq_to_backend.py
```diff
@@ -0,0 +1,97 @@
...
+        total = 0
+        correct = 0
+        loss = 0.0
+        class_probs = []
+        class_preds = []
+
+        with torch.no_grad():
+            idx = 0
+            for data, labels in testing_dataloader:
+                data, labels = data.cuda(), labels.cuda(non_blocking=True)
```
peri044, line 60: data.to(device) maybe to avoid warnings ?

tests/py/test_ptq_to_backend.py
```diff
@@ -0,0 +1,97 @@
...
+        trt_mod = torch._C._jit_to_backend("tensorrt", self.model, self.spec)
+        int8_test_acc = self.compute_accuracy(self.testing_dataloader, trt_mod)
+        log(Level.Info, "[TRT INT8 Backend] Test Acc: {:.2f}%".format(100 * int8_test_acc))
+        acc_diff = fp32_test_acc - int8_test_acc
+        self.assertTrue(abs(acc_diff) < 3)
+
+
+def test_suite():
+    suite = unittest.TestSuite()
+    suite.addTest(TestAccuracy.parametrize(TestAccuracy, model=torch.jit.load('./trained_vgg16.jit.pt')))
```
peri044, line 87: Can you please add this comment here and also to `test_ptq_dataloader_calibrator.py` as well ? 
```
# You need a pre-trained VGG cifar10 model to run this test. Please follow instructions at
# https://github.com/NVIDIA/TRTorch/tree/master/cpp/ptq/training/vgg16 to export this model.
```
I added it to `test_ptq_trt_calibrator.py` but forgot it at other place.

py/trtorch/csrc/tensorrt_classes.h
```diff
@@ -94,10 +94,18 @@ struct CompileSpec : torch::CustomClassHolder {
     input_ranges.push_back(*ir);
   }
 
+  int64_t getPTQCalibratorHandle() {
+    return (int64_t)ptq_calibrator;
```
peri044, line 98: Can you let me know why we do this cast to int64_t and type cast it back in the `setPTQCalibratorViaHandle` call ?

peri044 (changes requested):
LGTM

tests/py/test_ptq_to_backend.py
narendasan: this shouldnt throw warnings since its using the new api. the .cuda api wasnt deprecated it was just the async flag

py/trtorch/csrc/tensorrt_classes.h
narendasan: TorchBind cannot handle pointers as arguments so this was the cheapest way to get a pointer added to the struct. we get the int64_t casted pointer from the original struct and there is `void setPTQCalibratorViaHandle(int64_t handle)` to set the pointer from an int64_t in a struct owned by torchbind

narendasan: Both these functions dont get exposed to the user, they are purely used internally

Issue #388, opened 2021-03-04
✨[Feature] Setting the input data type of models, such as INT32, is not supported


guoruoqian:
**Is your feature request related to a problem? Please describe.**

When i tried to use TRTorch to convert some PyTorch models which input's data type is INT32 such as  BERT , it failed in `model.forward(inputs) `process. The error message is below.  It seems that TRT Model's input data type is related to [op_precision.](https://github.com/NVIDIA/TRTorch/blob/62b077e1873fab5c6095347482b2abe45f92b302/core/conversion/conversionctx/ConversionCtx.cpp#L61)

**Describe the solution you'd like**

I would like to add a corresponding interface to specify the type of input data.

**Describe alternatives you've considered**

**Additional context**

```
torch::Tensor in1 = torch::randint(0, 4, {128, 203}, torch::kCUDA).to(torch::kInt32);  
torch::Tensor in2 = torch::randint(0, 4, {128, 203}, torch::kCUDA).to(torch::kInt32);
std::vector<at::Tensor> inputs_trt;
inputs_trt.push_back(in1);
inputs_trt.push_back(in2);

std::vector<torch::jit::IValue> inputs_trt_;
for (auto in : inputs_trt) {
  inputs_trt_.push_back(torch::jit::IValue(in.clone()));
}

torch::jit::Module mod;
try {
  // Deserialize the ScriptModule from a file using torch::jit::load().
  mod = torch::jit::load(path);
} catch (const c10::Error& e) {
  std::cerr << "error loading the model\n";
}
mod.eval();
mod.to(torch::kCUDA);

auto trt_mod = trtorch::CompileGraph(mod, std::vector<trtorch::CompileSpec::InputRange>{in1.sizes(), in2.sizes()});
auto trt_out = trt_mod.forward(inputs_trt_);
```
When executing trt_mod.forward(), a bug appears
![image](https://user-images.githubusercontent.com/22525902/109931905-728b3900-7d04-11eb-90f9-a105a0e226f7.png)

peri044:
@narendasan 
Discovered the same bug while writing testcase for [aten::cast](https://github.com/NVIDIA/TRTorch/pull/382). I just had a workaround for cast testcases.
TRTorch assumes the input to be float by using `ctx->input_type`

I agree we need to provide an interface to provide input data types of tensors. 

[...]

PR #403, opened 2021-03-17
fix bugs in embedding converter

guoruoqian:

# Description
In my practice, the `indicesTensor ` in aten::embedding sometimes is IValue. So i replaced the ITensor with ITensorOrFreeze.
According to the [document ](https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_tensor.html#a1004b03f9249097f6d146abc3772819a), After calling setType, the type of a tensor is unchanged if the tensor is not a network input tensor, or marked as an output tensor or shape output tensor. So i replaced the setType with addIdentity to avoid change the input data type and ensure the indicesTensor is set as INT32.
 
In addition, i use the linters to ensure that my code matches the style guidelines. The linters changed the entire file(core/conversion/converters/impl/select.cpp) . I only modified the aten::embedding part in select.cpp.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/select.cpp
+++ b/core/conversion/converters/impl/select.cpp
@@ -174,9 +174,11 @@ auto select_registrations TRTORCH_UNUSED =
             {"aten::embedding(Tensor weight, Tensor indices, int padding_idx=-1, bool scale_grad_by_freq=False, bool sparse=False) -> (Tensor)",
              [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
                auto embeddingTensor = args[0].ITensorOrFreeze(ctx);
-               auto indicesTensor = args[1].ITensor();
+               auto indicesTensor = args[1].ITensorOrFreeze(ctx);
                // Set datatype for indices tensor to INT32
-               indicesTensor->setType(nvinfer1::DataType::kINT32);
+               auto identity = ctx->net->addIdentity(*indicesTensor);
+               identity->setOutputType(0, nvinfer1::DataType::kINT32);
+               indicesTensor = identity->getOutput(0);
 
                // IGatherLayer takes in input tensor, the indices, and the axis of input tensor to take indices from
                auto gather_layer = ctx->net->addGather(*embeddingTensor, *indicesTensor, 0);
--- a/tests/core/conversion/converters/test_select.cpp
+++ b/tests/core/conversion/converters/test_select.cpp
@@ -106,7 +106,7 @@ TEST(Converters, ATenEmbeddingConvertsCorrectly) {
   auto jit_results = trtorch::tests::util::RunGraph(g, params, {jit_in});
 
   // Run TensorRT
-  auto options_trt = torch::TensorOptions().device(torch::kCUDA, 0).dtype(torch::kI32);
+  auto options_trt = torch::TensorOptions().device(torch::kCUDA, 0).dtype(torch::kFloat);
   auto trt_in = at::tensor({0, 1, 2}, options_trt);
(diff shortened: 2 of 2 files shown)


core/conversion/converters/impl/select.cpp
```diff
@@ -65,118 +65,123 @@ bool add_split(ConversionCtx* ctx, const torch::jit::Node* n, args& args, bool s
...
-               auto indicesTensor = args[1].ITensor();
+               auto indicesTensor = args[1].ITensorOrFreeze(ctx);
                // Set datatype for indices tensor to INT32
-               indicesTensor->setType(nvinfer1::DataType::kINT32);
+               auto identity = ctx->net->addIdentity(*indicesTensor);
+               identity->setOutputType(0, nvinfer1::DataType::kINT32);
+               indicesTensor = identity->getOutput(0);
 
                // IGatherLayer takes in input tensor, the indices, and the axis of input tensor to take indices from
                auto gather_layer = ctx->net->addGather(*embeddingTensor, *indicesTensor, 0);
```
guoruoqian, line 187: I only modified this part in select.cpp.

tests/core/conversion/converters/test_select.cpp
```diff
@@ -106,7 +106,7 @@ TEST(Converters, ATenEmbeddingConvertsCorrectly) {
   auto jit_results = trtorch::tests::util::RunGraph(g, params, {jit_in});
 
   // Run TensorRT
-  auto options_trt = torch::TensorOptions().device(torch::kCUDA, 0).dtype(torch::kI32);
+  auto options_trt = torch::TensorOptions().device(torch::kCUDA, 0).dtype(torch::kFloat);
```
narendasan, line 109: Do the indices only come in float? can they come in other types? We should maybe instead duplicate instead of modifying the test

narendasan:
When I run the linter on my machine on your branch it switches everything back. How are you running the linter? do you perhaps have a global clang-format or something?

tests/core/conversion/converters/test_select.cpp
guoruoqian: When using setType function in embbedding converter, the input data type will be changed from kFloat to kInt32.  But  i have replaced setType function with addIdentity function, the input data type will not change. If i still use kInt32, it will get an error like that "Expected inputs[pyt_idx].dtype() == expected_type to be true but got false. Expected input tensors to have type Float, found type int". 
In my opinion, a single operation at one layer of the network directly changes the input data type of the entire network, which maybe is not a cool thing, right?  And the issue #388  can be resolved through set all input data type as kFloat.

guoruoqian:
> When I run the linter on my machine on your branch it switches everything back. How are you running the linter? do you perhaps have a global clang-format or something?

Previously I used clang-format10.0, now I have changed it to clang-format 9.0.1-12. Now everything in select.cpp is back. 

Besides, after i run this command "bazel run //tools/linter:cpplint -- //..." with clang-format 9.0.1-12, there are other some files  are changed:
image
I am curious why other files are also modified. Is it because of my clang-format version?

tests/core/conversion/converters/test_select.cpp
narendasan: Yeah this is a good catch, I was just wondering about the test and why it wasnt duplicated. I think its not a massive deal, other than that this patch lgtm

narendasan (approved):

narendasan:
Hmm, that is odd, I did not realize that different clang-format versions will give you different lintings. I will look more into what the set up of the canonical linter (the one that gets run on github on PRs) is and make it standard

narendasan (approved):
lgtm

PR #406, opened 2021-03-19
Automatic Fallback

bowang95:
# Description

Implemented Automatic Fallback, this feature will allow the compiler to identify which operations are supported by TRTorch and correctly segment out these graphs, compile each engine and then link together TorchScript and TRTorch engines.

- New feature

# Checklist:


core/partitioning/partitioning.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+torch::jit::Node* cloneNode(
+    torch::jit::Node* node,
+    std::shared_ptr<torch::jit::Graph>& graph,
+    std::unordered_map<torch::jit::Value*, torch::jit::Value*>& old_to_new) {
+  auto* block = graph->block();
+  auto env = [&](torch::jit::Value* v) { return getOrAddInputForValue(v, graph, old_to_new); };
+
+  auto new_node = block->appendNode(graph->createClone(node, env));
```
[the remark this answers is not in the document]
peri044: env looks like it is the new `torch::jit::Value` which has the same metadata as the node we are trying to clone from ? This should be renamed better.

core/partitioning/partitioning.cpp
```diff
@@ -0,0 +1,250 @@
...
+  return random_inputs;
+}
+
+void registerSegmentsInputsOutputs(
+    std::vector<SegmentedBlock>& segmented_blocks,
+    std::shared_ptr<torch::jit::Graph> g) {
+  // find the corresponding raw values in original global graph for this segmented block's inputs/outputs
+  std::set<torch::jit::Value*> input_values;
+  for (auto& seg_block : segmented_blocks) {
+    seg_block.registerInputs();
```
peri044, line 156: Can you explain why we have this function `registerInputs` ?
It looks like when we call `SegmentedBlock(kTensorRT, tensorrt_nodes)` , it will create a mini_graph `g_` with all nodes appended. 
Do we actually need a separate registerInputs() to add the graph_inputs to a `inputs_` vector ? 
Is there a problem with doing this in the constructor itself ?

core/partitioning/partitioning.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+  for (auto& mini_graph_input : input_values) {
+    for (auto& seg_block : segmented_blocks) {
+      if (std::find(seg_block.raw_inputs().begin(), seg_block.raw_inputs().end(), mini_graph_input) ==
+              seg_block.raw_inputs().end() &&
+          seg_block.contain_raw_input(mini_graph_input)) {
+        seg_block.registerOutput(mini_graph_input);
+      }
+    }
+  }
```
peri044, line 174: Can you add a comment with an example on what is happening here ? Seems like the stiching code.

core/partitioning/partitioning.cpp
```diff
@@ -0,0 +1,250 @@
...
+    segmented_blocks.emplace_back(SegmentedBlock::kTorch, pytorch_nodes);
+  }
+
+  // register input/output torch::jit::Value for segmetned graphs
+  registerSegmentsInputsOutputs(segmented_blocks, g);
+
+  // store the mapping from lowering graph torch::jit::Value => torch::jit::IValue that we get by running segments
+  std::unordered_map<torch::jit::Value*, torch::jit::IValue> ivalues_maps;
+
+  std::vector<torch::jit::IValue> random_inputs = generateRandomInputs(input_ranges);
```
peri044, line 235: Should we move this `generateRandomInputs`  function outside of fallback code for more general usage? something like utils ?

core/partitioning/partitioning.h
```diff
@@ -0,0 +1,121 @@
...
+
+  c10::ArrayRef<torch::jit::Value*> outputs() {
+    return g_->outputs();
+  }
+
+  const std::vector<torch::jit::Value*>& raw_inputs() const {
+    return inputs_;
+  }
+
+  const std::vector<torch::jit::Value*>& raw_outputs() const {
```
peri044, line 77: What would be the difference between raw_inputs() and input() values in practice ? Do we need both variants ?

peri044 (commented):
Took an initial pass (80% complete). Still have some functions left to review. 
In general, implementation looks good. 
- As discussed before, some functions can be restructured for testcases to directly call them.
- Comments (with an example) at certain places (eg: stitching ) for easier understanding for future users.

core/partitioning/partitioning.cpp
```diff
@@ -0,0 +1,250 @@
...
+    }
+    tensorrt_nodes.clear();
+  }
+}
+
+std::vector<SegmentedBlock> segment_graph(
+    std::shared_ptr<torch::jit::Graph> g,
+    std::vector<conversion::InputRange>& input_ranges,
+    const conversion::TorchFallback& fallback_info) {
+  auto min_block_size = fallback_info.min_block_size;
```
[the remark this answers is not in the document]
narendasan: I think min block size is fine, I think we can consider the actual API more as we get closer to merge

narendasan:
I took a pass as well, my high level comments are that in general it seems like there is a lot of uncoupled state floating around and being modified in various different void functions. It also seems like a lot of functions exist solely to keep the number of lines of code per function low vs doing clear distinct tasks so its easy to understand each step of the process. Also some of the naming is not clear, like not sure what `registerSegmentInOutIValues` is trying to do, it seems like this is the part doing the shape analysis. I am taking a stab at decoupling and organizing the code rn, here is my structure so far: 

```
//.../partitioning 
-> partitioning.h // Main interface file 
-> partitioning.cpp // Source for core graph segmentation code
-> SegmentedBlock.h // Header for the SegmentedBlock class
-> SegmentedBlock.cpp // Source for SegmentedBlock class (Also collapsed a few helper functions into appendNode)
-> shape_analysis.h // Header for code that will be used to do shape analysis 
-> shape_analysis.cpp // Source for shape analysis which right now contains a function that handles running provided graphs 
```

I think that its important that we start to centralize state and make it clear how it gets modified through the partitioning process

core/partitioning/partitioning.h
bowang95: raw_inputs() would be the torch::jit::Value that in the lowering global graph that is used as an input in our mini graph, when we construct the mini-graphs these Values in mini-graph will change because we are constructing new graph. We need them both because when we stitch mini-graph together, we need the raw_inputs to find the mappings, and we use inputs() to run the segments inferences and do something like replaceAllUsesWith() for mini-graph inputs.

core/partitioning/partitioning.cpp
bowang95: Yes, I think it's ok. BTW, I found that my getFunctionSchema function in partitioning.cpp is same with the GenerateGraphSchema function in compiler.cpp, I am not sure if we should delete both of these two and move them to utils?

core/partitioning/partitioning.cpp
bowang95: Yes, I will add comments for this part. This piece of code is used to identify each mini-graph's outputs by going through all input values that is necessary for all mini-graphs and if we find that input value is included in a block, we will output that value.

core/partitioning/partitioning.cpp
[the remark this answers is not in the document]
bowang95: Yes, graph->createClone is used to create a node for current graph by copying all the metadata from another node in another graph.  Env here is a value_map to translate inputs of original node to inputs of the cloned node. I will comment this part for better understanding.

core/partitioning/partitioning.cpp
bowang95: That's a good point. I think I used this because I didn't have SegmentedBlock(kTensorRT, tensorrt_nodes) constructer at first. So, I had to construct segmentedBlock first and then append node one by one, after all of this is done we can identify all required inputs and then register inputs. 
Currently if we don't append nodes one by one for SegmentedBlock I think maybe it's better to move this part to our constructor. Thanks!

core/compiler.cpp
```diff
@@ -108,17 +95,19 @@ void AddEngineToGraph(
   g->block()->appendNode(unpack_node);
 
   // If there are multiple output tensors from TensorRT we wrap them in a tuple
-  // to return
-  if (unpack_node->outputs().size() > 1) {
+  // to return, convert to tuple only when we only have 1 segmented graph
+  if (!engine_id && unpack_node->outputs().size() > 1) {
```
narendasan, line 99: Does the case where we have multiple TRT engines never have engine_id 0? We should not this if so.

core/compiler.cpp
```diff
@@ -151,7 +141,111 @@ std::string ConvertGraphToTRTEngine(const torch::jit::script::Module& mod, std::
...
+      auto g = graph_and_parameters.first;
+      auto params = graph_and_parameters.second;
+      auto named_params = conversion::get_named_params(g->inputs(), params);
+      auto convert_cfg = std::move(cfg.convert_info);
+      LOG_INFO(*g << "(LoweringGraph)\n");
+
+      // segment the graph and convert segmented TensorRT block
+      auto segmented_blocks = partitioning::Partition(g, convert_cfg.input_ranges, cfg.partition_info);
+      if (segmented_blocks.size() == 1 && segmented_blocks[0].target() == partitioning::SegmentedBlock::kTorch) {
+        return mod;
```
narendasan, line 199: Lets throw a warning here that the compiler did nothing.

core/compiler.cpp
```diff
@@ -151,7 +141,111 @@ std::string ConvertGraphToTRTEngine(const torch::jit::script::Module& mod, std::
...
+      int trt_engine_id = 1;
+      std::unordered_map<torch::jit::Value*, torch::jit::Value*> old_to_new_g;
+      // add global graph's input to old_to_new_g mapping
+      for (auto input : g->inputs()) {
+        util::getOrAddInputForValue(input, new_g, old_to_new_g);
+      }
+      for (auto& seg_block : segmented_blocks) {
+        LOG_INFO(*g << "(MiniGraphInSegmentedBlock)\n");
+        if (seg_block.target() == partitioning::SegmentedBlock::kTensorRT) {
+          std::vector<ir::InputRange> input_ranges;
```
narendasan, line 211: Does Dynamic shape work?

core/partitioning/SegmentedBlock.h
```diff
@@ -0,0 +1,89 @@
...
+  }
+  const std::vector<torch::jit::Value*>& raw_inputs() const {
+    return inputs_;
+  }
+  const std::vector<torch::jit::Value*>& raw_outputs() const {
+    return outputs_;
+  }
+  void eraseInput(size_t i);
+  void eraseOutput(size_t i);
+  bool contain_raw_value(torch::jit::Value* input) {
```
narendasan, line 61: what is a raw value?

core/partitioning/SegmentedBlock.h
```diff
@@ -0,0 +1,89 @@
...
+  void update_target(SegmentedBlockTarget new_target) {
+    target_ = new_target;
+  }
+  enum SegmentedBlockTarget target() {
+    return target_;
+  }
+
+ private:
+  SegmentedBlockTarget target_;
+  std::vector<ir::InputRange> in_shape_; // REVIEW: This should just be ir::InputRange
```
narendasan, line 79: You can remove the comment here

core/partitioning/partitioning.cpp
```diff
@@ -0,0 +1,232 @@
...
+            if (isTensorOrTensorList(node_output))
+              seg_block.registerOutput(node_output);
+          }
+          if (!seg_block.raw_outputs().empty())
+            break;
+        }
+      }
+    }
+  }
+  // erase segments which still have no output
```
narendasan, line 145: You can just use the DeadCodeElimination Pass to do this

core/partitioning/partitioning.cpp
```diff
@@ -0,0 +1,232 @@
...
+  std::unordered_set<std::string> forced_fallback_operators(
+      partition_info.forced_fallback_operators.begin(), partition_info.forced_fallback_operators.end());
+
+  auto nodes = g->block()->nodes();
+  std::vector<SegmentedBlock> segmented_blocks;
+
+  // segment the nodes
+  std::vector<torch::jit::Node*> tensorrt_nodes, pytorch_nodes;
+  for (const auto n : nodes) {
+    if (n->kind() == torch::jit::prim::Constant)
```
narendasan, line 167: Add brackets

core/partitioning/partitioning.cpp
```diff
@@ -0,0 +1,232 @@
...
+  // segment lowering global graph into blocks
+  std::vector<SegmentedBlock> segmented_blocks = segment_graph(g, partition_info);
+
+  // resolve nonTensor inputs/outputs
+  resolveNonTensorInputs(segmented_blocks, g);
+
+  // register input/output torch::jit::Value for segmented graphs
+  registerSegmentsOutputs(segmented_blocks, g);
+
+  // store the mapping from lowering graph torch::jit::Value => torch::jit::IValue that we get by running segments
```
narendasan, line 214: Can this be isolated in the shape analysis component or is it necessary else where?

core/util/trt_util.cpp
```diff
@@ -350,6 +350,58 @@ c10::optional<nvinfer1::DataType> toTRTDataType(caffe2::TypeMeta dtype) {
   }
 }
 
+c10::FunctionSchema GenerateGraphSchema(std::string method_name, std::shared_ptr<torch::jit::Graph>& g) {
```
narendasan, line 353: Move this to jit_util.h

core/compiler.cpp
bowang95: I don't think so. Currently we haven't considered the case when we have dynamic shapes in shape analysis.

core/partitioning/SegmentedBlock.h
bowang95: Raw value is the torch::jit::Value* from original lowered graph, we use this function to identify if current block produces any torch::jit::Value* that later block will use as inputs.

core/compiler.cpp
bowang95: Yes. I refactored this function by adding a default argument indicating whether we have fallback or not.

core/partitioning/partitioning.cpp
bowang95: DeadCodeElimination seems to erase dead code in a graph, but here we erase the useless SegmentedBlock. Maybe we can add that pass for every SegmentedBlock to erase dead code?

core/partitioning/partitioning.cpp
bowang95: We can move it into shape analysis. But this function runs on all SegmentedBlock while shape analysis solely runs on each SegmentedBlock, we may need to refactor shape analysis as well.

core/partitioning/partitioning.cpp
narendasan: Yeah we should do that

core/compiler.cpp
```diff
@@ -21,37 +21,24 @@
...
-
 void AddEngineToGraph(
     torch::jit::script::Module mod,
     std::shared_ptr<torch::jit::Graph>& g,
-    const std::string& serialized_engine) {
-  auto engine_ptr = c10::make_intrusive<runtime::TRTEngine>(mod._ivalue()->name(), serialized_engine);
+    const std::string& serialized_engine,
+    int engine_id = 0) {
+  auto engine_ptr =
+      c10::make_intrusive<runtime::TRTEngine>(mod._ivalue()->name() + std::to_string(engine_id), serialized_engine);
```
narendasan, line 36: Does engine_id just need to be unique or do we use the ids else where?

If they are just unique we should use the pointer trick to get something that is likely to be unique, therefore we dont really need to worry about conflicts

core/compiler.cpp
narendasan: Thats probably higher priority than loops then, I think since we have unrolling that can be enabled. Also I think its pretty achievable in the time we have

core/compiler.cpp
bowang95: Yeah, I think it should be unique. We only use the ids to differentiate between engines.
BTW, what do you mean by pointer trick? I think I'm new to this... What I can think about is generating some unique values by a class or using some library, is this good way?

PR #419, opened 2021-04-01
fix: support c10::List type of output in MarkOutputs

inocsin:
# Description

Fixes # (issue)

## Type of change

Support c10::List<c10::IValue> type of output type in **void MarkOutputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> outputs)**

# Checklist:

### DIFF
--- a/core/conversion/conversion.cpp
+++ b/core/conversion/conversion.cpp
@@ -173,6 +173,20 @@ void AddInputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> inputs
   }
 #endif
 }
+void MarkOutputsOfIvalue(ConversionCtx* ctx, c10::IValue out_ivalue, const torch::jit::Value* out) {
+  if (out_ivalue.isCustomClass()) {
+    std::string name = std::string("output_") + std::to_string(ctx->num_outputs);
+    auto output_container = out_ivalue.toCustomClass<TensorContainer>();
+    nvinfer1::ITensor* out_tensor = output_container.get()->tensor();
+    out_tensor->setName(name.c_str());
+    ctx->net->markOutput(*out_tensor);
+    LOG_INFO(
+        ctx->logger, "Marking Output " << out->debugName() << " named " << name << " in engine (ctx.MarkOutput)");
+    ctx->num_outputs += 1;
+  } else {
+    TRTORCH_THROW_ERROR("Unknown output type. Only a single tensor or a TensorList type is supported.");
+  }
+}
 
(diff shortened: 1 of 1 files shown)


core/conversion/conversion.cpp
```diff
@@ -173,24 +173,34 @@ void AddInputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> inputs
...
+    std::string name = std::string("output_") + std::to_string(ctx->num_outputs);
+    auto output_container = out_ivalue.toCustomClass<TensorContainer>();
+    nvinfer1::ITensor* out_tensor = output_container.get()->tensor();
+    out_tensor->setName(name.c_str());
+    ctx->net->markOutput(*out_tensor);
+    LOG_INFO(
+        ctx->logger, "Marking Output " << out->debugName() << " named " << name << " in engine (ctx.MarkOutput)");
+    ctx->num_outputs += 1;
+  } else {
+    TRTORCH_THROW_ERROR("Unknown output type. Only a single tensor or a TensorList type is supported.");
```
narendasan, line 187: We should make this Error a bit more clear, something like 

```
Unsupported output type, only Tensors or unwrapped collections of Tensors can be marked as engine outputs but found type: [Print type here]
```

core/conversion/conversion.cpp
```diff
@@ -173,24 +173,34 @@ void AddInputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> inputs
   }
 #endif
 }
+void MarkOutputsOfIvalue(ConversionCtx* ctx, c10::IValue out_ivalue, const torch::jit::Value* out) {
```
narendasan, line 176: rename to `MarkIValueOutputs`

narendasan (commented):
I am a bit concerned because say we have a TensorList Value `%out1`. The output of the TensorRT engine could have like 10 outputs associated with that one Value. If there is an additional Tensor Value output `%out2` the output of the engine will have 11 tensors but we don't really have a way to determine which are supposed to be in a list and which should be by themselves. Right now the operator interface for all engines is `trt::execute_engine(TRTEngine, Tensor[]) -> Tensor[]`. @peri044 I think we discussed  this regarding tuples as well. We would need to store more information about output formatting and change the graph construction step in `compile.cpp` to take the output of `trt::execute_engine` and reconstruct the expected output format, but that is a much larger PR i believe.

peri044:
Currently this PR can be put on hold as it involves a broader change across other components like graph synthesis as described in https://github.com/NVIDIA/TRTorch/issues/428

PR #430, opened 2021-04-10
Added Dockerfile for 21.02, updated WORKSPACE to be buildable on most…

borisfom:
… systems

# Description

Added Dockerfile for 21.02. 
Updated WORKSPACE to be buildable on most systems:
- used local repos for cudnn and TRT by default: the previous default assumed manual changes to WORKSPACE and manual downloads. This way TRTorch can be built on most systems (that have CUDNN and TRT installed) as is.
- changed default CUDA location to /usr/local/cuda : I do not have full 11.1 on my 11.2 system, for example. This way TRTorch can be built on any system that is compatible.
- pip3_import was failing with obscure diagnostics.  Same happened when I tried building in in Docker based on 11.02 
With upgraded versions of bazel_tools and rules_pkg, it seems to work both on my setup and yours, and in Docker as well.
- removed dead cublas repo

Fixes # (issue)

## Type of change

# Checklist:


WORKSPACE
```diff
@@ -108,41 +101,32 @@ http_archive(
 #    build_file = "third_party/libtorch/BUILD"
 #)
 
-#new_local_repository(
-#    name = "cudnn",
-#    path = "/usr/",
-#    build_file = "@//third_party/cudnn/local:BUILD"
-#)
+new_local_repository(
```
narendasan, line 104: Same thing here, we have no idea what versions of the libraries are installed on peoples systems and TRTorch does crash if you use the wrong ones. Thats why we suggest people download the tars and use `dist-dir`. If they know it is correct or on Jetson were we can ensure the environment is uniform then this is fine

WORKSPACE
```diff
@@ -35,16 +34,10 @@ git_repository(
 # CUDA should be installed on the system locally
 new_local_repository(
     name = "cuda",
-    path = "/usr/local/cuda-11.1/",
+    path = "/usr/local/cuda",
```
narendasan, line 37: Does this work if people have CUDA 11.2 installed as their primary cuda? I think we'd accidentally end up linking the wrong library. 

I think in general the way we have approached WORKSPACE is as a declaration of dependencies, and its incumbent on the user to correctly modify the WORKSPACE for their particular environment, all we can do is ensure that what we provides guarantees correct compilation on our conception of the default system environment (so Ubuntu 18.04 with latest CUDA install via debians). Given that latest CUDA is 11.2 and presumably be linked to `/usr/local/cuda`, we should set the specific path to ensure people are using the right version and let them change it if its different.

WORKSPACE
narendasan: Basically I'd rather people get a error from bazel that the system is not what was expected and them either modifying the WORKSPACE to point to what they have or realizing they dont have the right dependencies than a bunch of bug reports saying that people are getting runtime crashes because they inadvertently linked the wrong cuDNN.

narendasan: (Also I can share with you a WORKSPACE file which pulls at compile time from the build server for cuDNN and TensorRT, which makes sure you always have the right versions)

WORKSPACE
```diff
@@ -35,16 +34,10 @@ git_repository(
...
     build_file = "@//third_party/cuda:BUILD",
 )
 
-new_local_repository(
```
narendasan, line 42: I though cuBLAS was a dependency for TensorRT, do we not need this?

WORKSPACE
narendasan: This was particularly a problem when PyTorch was 6 months behind a major version update to CUDA

borisfom: Yes that worked for me - 11.1 path may work for people who had 11.1 installed, but I installed 11.2 as my first CUDA 11 and don't want to go through installing 11.1 and then reinstalling 11.2. 
I understand your considerations about not wanting people to get "successful" build while potentially linking wrong library - but people do expect GitHub projects to build out of the box these days and early build breaks may be discouraging. 
   The way you prefer people to build right now automatically means custom changes to WORKSPACE - not just uncommenting some lines, they actually need to decide where to put downloaded archive and how to refer to it (would have been so much better if there were no login requirements at developer.nvidia.com). Besides first build frustration, it also means content issues with contribution - one has to exclude those manual changes to WORKSPACE before submitting  PR...
   So, perhaps, better solution here would be to remove WORKSPACE from the project and provide WORKSPACE.local and WORKSPACE.archive templates instead, documenting that user should use one of them for their WORKSPACE and implications of using each ?

WORKSPACE
borisfom: Nope - yes cuBLAS is a dependency, but we have cuda:cublas target for it. As you can see, this one was not referred to from any BUILD file.

WORKSPACE
narendasan: > 11.1 path may work for people who had 11.1 installed, but I installed 11.2 as my first CUDA 11 and don't want to go through installing 11.1 and then reinstalling 11.2.

I think the solution for your particular case to avoid uninstalling an reinstalling is to install the `cuda-libraries-11-1 cuda-libraries-dev-11-1` packages (I think these are the right ones) which can be installed after 11.2 and wont override default cuda but give you the libraries in `/usr/local/cuda-11.1`. IIRC this is what did since like you 11.2 is my original CUDA. Ideally we could make CUDA hermetic too but I have no idea how we could do that. 

> but people do expect GitHub projects to build out of the box these days and early build breaks may be discouraging.

Yeah I get that but I think its something that you cant really control considering how building C++ projects work. We try to be as hermetic as possible because I find the downstream errors of slight library version mismatches or system configuration incompatibilities much more frustrating and difficult to debug than file path issues. We default to archive because I don't think its is too hard to download two tarfiles and put them in a directory and use the `--dist-dir` flag like we put in the documentation and README and be guaranteed they are working with the same libraries we are targeting https://nvidia.github.io/TRTorch/tutorials/installation.html#building-using-cudnn-tensorrt-tarball-distributions
(yeah totally agree if NVIDIA's distribution of cuDNN and TensorRT was more liberal it would be way nicer). I feel if people want to do something non standard or something we can't fully account for, they should explicitly make that choice and so if they hit errors they can refer back to this step when debugging. 

> it also means repeated issues with contributions - one has to exclude those manual changes to WORKSPACE before submitting PR...

Yeah I guess it means you can't do `git add -u`. `git add core/  tests/` is what I have been using. We can maybe add a git hook to catch attempting to commit WORKSPACE changes for you.

> So, perhaps, better solution here would be to remove WORKSPACE from the project and provide WORKSPACE.local and WORKSPACE.archive templates instead, documenting that user should use one of them for their WORKSPACE and implications of using each ?

We could do this and gitignore WORKSPACE @peri044 thoughts? My concern is its another file that needs to be updated every version bump since I'm still not convinced that leaving the path as `/usr/local/cuda` is the right thing to do for most users.

borisfom: Thanks for the ideas! 
I understand your concern about updating stuff in WORKSPACE when we bump version of anything. 
Can that be solved by some kind of WORKSPACE that includes .bzl files some of them has to be user-defined (like WORKSPACE.trt-deps.bzl) ? So that the bulk of WORKSPACE is checked in except for the configurable part that users have to make decision on and maintain themselves (I see two of them: WORKSPACE.libtorch.bzl default/local and WORKSPACE.trt-deps) 
I quickly tried including .bzl in WORKSPACE which by itself worked but then ran into some issues. So just sharing this idea here so Bazel pros can ponder :)

borisfom: > I don't think its is too hard to download two tarfiles and put them in a directory and use the --dist-dir flag like we put in the documentation and README and be guaranteed they are working with the same libraries we are targeting https://nvidia.github.io/TRTorch/tutorials/installation.html#building-using-cudnn-tensorrt-tarball-distributions

I have tried using --dist-dir : not really convenient to type every single time. Is there a way to put it to per-project rc file? 
Another - probably the simplest - way would be to direct users to put the tarballs in dist_dir/x86_64-linux-gnu and have checked in WORKSPACE refer to it - then first time, the build would break with meaningful diagnostic. That's what I do for myself locally - did not find the way to refer to the dir relatively to the project - though I did not try too hard to make it portable.

narendasan: We could do macros for versions, that might work. I would need to look into how that works exactly with WORKSPACE. 

You can add command aliases to the `.bazelrc`file. We could maybe do something like `--archive-deps-x86_64` or something shorter which aliases to `--dist-dir thrid_party/distdir/x86_64-linux-gnu` or add it as something that is added by default but people who use local would need to disable that. 

We could also explore something like this https://stackoverflow.com/questions/58937356/bazel-relative-local-path-as-url-in-http-archive

WORKSPACE
borisfom: Reverted local/archive changes.

WORKSPACE
borisfom: That certainly calls for another PR. Reverted local/archive changes.

WORKSPACE
```diff
@@ -3,25 +3,24 @@ workspace(name = "TRTorch")
...
+        name = "rules_python",
+        url = "https://github.com/bazelbuild/rules_python/releases/download/0.2.0/rules_python-0.2.0.tar.gz",
+        sha256 = "778197e26c5fbeb07ac2a2c5ae405b30f6cb7ad1f5510ea6fdac03bded96cc6f",
+    )
 
-load("@rules_python//python:pip.bzl", "pip_repositories", "pip3_import")
-pip_repositories()
+load("@rules_python//python:pip.bzl", "pip_install")
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
narendasan, line 15: We dont need to load this twice right?

WORKSPACE
```diff
@@ -3,25 +3,24 @@ workspace(name = "TRTorch")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
narendasan, line 3: We load http_archive here

WORKSPACE
```diff
@@ -3,25 +3,24 @@ workspace(name = "TRTorch")
...
-git_repository(
-    name = "rules_python",
-    remote = "https://github.com/bazelbuild/rules_python.git",
-    commit = "4fcc24fd8a850bdab2ef2e078b1de337eea751a6",
-    shallow_since = "1589292086 -0400"
-)
-
-load("@rules_python//python:repositories.bzl", "py_repositories")
-py_repositories()
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
narendasan, line 6: I think this is redundant

docker/WORKSPACE.cu.docker
```diff
@@ -3,25 +3,24 @@ workspace(name = "TRTorch")
...
-git_repository(
-    name = "rules_python",
-    remote = "https://github.com/bazelbuild/rules_python.git",
-    commit = "4fcc24fd8a850bdab2ef2e078b1de337eea751a6",
-    shallow_since = "1589292086 -0400"
-)
-
-load("@rules_python//python:repositories.bzl", "py_repositories")
-py_repositories()
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
narendasan, line 6: Similar redundant http_archive

docker/WORKSPACE.cu.docker
```diff
@@ -3,25 +3,24 @@ workspace(name = "TRTorch")
...
+        name = "rules_python",
+        url = "https://github.com/bazelbuild/rules_python/releases/download/0.2.0/rules_python-0.2.0.tar.gz",
+        sha256 = "778197e26c5fbeb07ac2a2c5ae405b30f6cb7ad1f5510ea6fdac03bded96cc6f",
+    )
 
-load("@rules_python//python:pip.bzl", "pip_repositories", "pip3_import")
-pip_repositories()
+load("@rules_python//python:pip.bzl", "pip_install")
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
narendasan, line 15: Same here

narendasan (commented):
LGTM other than the multiple http_archives

WORKSPACE
borisfom: Removed.

narendasan (approved):
LGTM

PR #431, opened 2021-04-17
New API to register arbitrary TRT engines as TorchScript modules

### LINKED ISSUE #46: API to register arbitrary TRT engines as modules
It would be nice to be able to accept TRT engines built outside TRTorch and register them in the execution system.

narendasan:
# Description

This PR adds the ability to register TRT engines from outside TRTorch as TorchScript modules. Modules created through this method inherit all the abilities (and limitations) of TRTorch compiled TorchScript modules including serialization and portability through `torch.jit.load` and deployment with `libtrtorchrt.so`

Fixes #46 

## Type of change

# Checklist:

### DIFF
--- a/core/compiler.cpp
+++ b/core/compiler.cpp
@@ -173,6 +173,20 @@ torch::jit::script::Module CompileGraph(const torch::jit::script::Module& mod, C
   return new_mod;
 }
 
+torch::jit::script::Module EmbedEngineInNewModule(std::string& engine) {
+  std::ostringstream engine_id;
+  engine_id << reinterpret_cast<int*>(&engine);
+  torch::jit::script::Module new_mod("tensorrt_engine_mod_" + engine_id.str());
+  auto new_g = std::make_shared<torch::jit::Graph>();
+  AddEngineToGraph(new_mod, new_g, engine);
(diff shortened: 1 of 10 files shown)


core/compiler.cpp
```diff
@@ -173,6 +173,20 @@ torch::jit::script::Module CompileGraph(const torch::jit::script::Module& mod, C
   return new_mod;
 }
 
+torch::jit::script::Module EmbedEngineInNewModule(const std::string& engine) {
+  std::ostringstream engine_id;
+  engine_id << reinterpret_cast<const int*>(&engine);
+  torch::jit::script::Module new_mod("tensorrt_engine_mod_" + engine_id.str());
```
peri044, line 179: Can you let me know if there is any reason behind using `engine_id.str()` ?. It doesn't seem to give useful info but is it just to differentiate engines when multiple engines are being loaded ?   
```
tensorrt_engine_mod_0x7fffffffd900
```

peri044 (approved):
LGTM

core/compiler.cpp
narendasan: yeah its so engines dont clash by name. All it is is a "UID"

PR #434, opened 2021-04-18
Add aten::layer_norm support

chengyute:
# Description

Add aten::layer_norm support

Fixes # 278

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/BUILD
+++ b/core/conversion/converters/BUILD
@@ -40,6 +40,7 @@ cc_library(
         "impl/conv_deconv.cpp",
         "impl/element_wise.cpp",
         "impl/expand.cpp",
+        "impl/layer_norm.cpp",
         "impl/linear.cpp",
         "impl/matrix_multiply.cpp",
         "impl/pooling.cpp",
--- a/core/conversion/converters/impl/layer_norm.cpp
+++ b/core/conversion/converters/impl/layer_norm.cpp
@@ -81,23 +81,12 @@ auto layer_norm_registrations TRTORCH_UNUSED = RegisterNodeConversionPatterns().
 
       /* Layer_Norm normalizes over last N dimensions.
          normalizaed_shape could be (C,H,W), (H,W), or (W). */
-
       auto normalized_shape = args[1].unwrapToIntList();
       auto normalized_shape_vec = util::toVec(util::toDims(normalized_shape));
 
-      torch::Tensor gamma, beta;
-      gamma = args[2].unwrapToTensor();
-      beta = args[3].unwrapToTensor();
-
-      // Remove batch dimension from input shape for expand_size, which will
-      // be used to create weights for addScaleNd later.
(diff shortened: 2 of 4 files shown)


narendasan:
@chengyute do you know the feasibility of using a lowering pass to synthesize the converter for layer norm? I noticed we are repeating work for elementwise and some other sub components. In the future i think wed prefer layer decomposition vs adding custom converters for larger ops like layer_norm 

So since layer norm is `((x - E[x]) / (Var[x] + e) ^ (0.5)) * y + B` we could do something like 

```
graph(%x, %e, %y, %B):
    %half = prim::Constant[value=0.5]
    %mean = aten::mean(%x)
    %1 = aten::sub(%x, %mean)
    %var = aten::var(%x)
    %2 = aten::add(%var, %e)
    %3 = aten::pow(%2, %half)
    %4 = aten::div(%1, %3)
    %5 = aten::mul(%4, %y)
    %6 = aten::add(%5, %B)
    return %6
```
(Theres probably a lot more arguments for each op I missed)

chengyute:
@narendasan , 
I tried lowering pass, the issue of layer_norm is that the normalized_shape we feed in is dynamic. 
For example, in https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html,
the normalized_shape could be [5,10,10] [10,10], [10]. (note this is not index)
We need to hard code of use IF statement in instruction level. 
The code below could only fit [5,10,10] in the case.

graph(%input, %normalized_shape, %gamma, %beta, %eps, %cudnn):
            %gamma_trt = trt::const(%gamma)
            %beta_trt = trt::const(%beta)
            %0: int = prim::Constant[value=1]()  <- need to be hardcoded
            %1: int = prim::Constant[value=2]()  <- needs to be hardcoded
            %2: int = prim::Constant[value=3]()  <- needs to be hardcoded
            **%3 : int[] = prim::ListConstruct(%0, %1, %2)**  <- need to be hardcoded
            %4 : bool = prim::Constant[value=1]()
            %5 : None = prim::Constant()
            **%6 : Tensor = aten::mean(%input, %3, %4, %5)**
            %7: Scalar = prim::Constant[value=1]()
            %8 = aten::sub(%input, %6, %7)
            %9: Scalar = prim::Constant[value=2]()
            %10 = aten::pow(%8, %9)
            %11 = aten::mean(%10, %3, %4, %5)
            %12: Scalar = prim::Constant[value=1]()
            %13 = aten::add(%11, %eps, %12)
            %14 = aten::sqrt(%13)
            %15 = aten::div(%8, %14)
            %16 = aten::mul(%15, %gamma_trt)
            %17: Scalar = prim::Constant[value=1]()
            %18 = aten::add(%16, %beta_trt, %17)
            return(%18))IR";

narendasan:
what part do we need to condition on? the number of dimensions? or the shape itself? if its the number can we run size on the `normalized_shape`?

PR #435, opened 2021-04-20
Implemented 1D conv, refactored pooling, 1D and special case for adap…

borisfom:
…tive pooling

# Description

Added support for 1d convolutions.
Also:
Fixed 1d case for batch normalization
Included a fix for prim::ListConstruct when arg is a Tuple (apparently valid use case)
Refactored pooling
1D pooling and special case for adaptive pooling -> global pooling

Fixes # (issue)

## Type of change

# Checklist:


core/conversion/converters/impl/pooling.cpp
```diff
@@ -9,100 +10,163 @@ namespace converters {
...
+  } else if (pool_type == nvinfer1::PoolingType::kAVERAGE) {
+    ceil_mode = args[4].unwrapToBool();
+    bool count_inlcude_pad = args[5].unwrapToBool();
+
+    new_layer = ctx->net->addPoolingNd(*in, pool_type, kernel_size);
+    TRTORCH_CHECK(new_layer, "Unable to create Avg Pooling layer from node: " << *n);
+    new_layer->setAverageCountExcludesPadding(!count_inlcude_pad);
+    // if (!(args[6].IValue()->isNone())) {
+    //  LOG_WARNING("Divisor override is now handled by Avg Pooling Converter");
+    // }
```
peri044, line 166: Maybe remove these lines ?

core/conversion/converters/converter_util.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    LOG_DEBUG("Original shape: " << dims << ", reshaping to: " << newDims);
+    auto shuffle_layer = ctx->net->addShuffle(*tensor);
+    TRTORCH_CHECK(shuffle_layer, "Unable to create shuffle layer");
+    shuffle_layer->setReshapeDimensions(newDims);
+    shuffle_layer->setZeroIsPlaceholder(use_zeros);
+    shuffle_layer->setName((util::node_info(n) + " [Reshape to " + util::toStr(newDims) + ']').c_str());
+    return shuffle_layer;
+  } else
+    return nullptr;
```
narendasan, line 29: Not sure about returning `nullptr`. Could we maybe return a ITensor* instead of an ILayer* and pass through the original tensor if padding is not needed? Or we could maybe just assert `dims.nbDims < nDim` and fail?

core/conversion/converters/impl/batch_norm.cpp
```diff
@@ -40,17 +41,9 @@ auto batch_norm_registrations TRTORCH_UNUSED = RegisterNodeConversionPatterns().
...
-        // expand spatial dims from 1D to 2D
-        auto new_shape = util::toDimsTailPad(util::toVec(orig_shape), 4);
-        LOG_DEBUG(
-            "Input shape is less than 4D got: "
-            << orig_shape << ", inserting shuffle layer to reshape to 4D tensor shape: " << new_shape);
-        auto in_shuffle = ctx->net->addShuffle(*input);
-        in_shuffle->setReshapeDimensions(new_shape);
-        in_shuffle->setName(std::string("[Reshape input to " + util::toStr(new_shape) + ']').c_str());
-        input = in_shuffle->getOutput(0);
+      auto expandDims = addPaddingLayer(ctx, n, input, 4);
```
narendasan, line 44: I see why you did `ILayer*/nullptr` here, i feel like we should maybe have people explicitly decide to pad based on their own logic and so addPaddingLayer will always give you a layer/tensor back. Either way we need to have logic on if padding should/has happened but in this case, you might get a `nullptr` back which if not handled properly could cause problems.

core/conversion/converters/impl/conv_deconv.cpp
```diff
@@ -14,15 +14,46 @@ bool add_conv_deconv(ConversionCtx* ctx, const torch::jit::Node* n, args& args)
...
+  }
+  if (stride.nbDims==1)
+    stride = util::unsqueezeDims(stride, 1, 1);
+  if (dilation.nbDims==1)
+    dilation = util::unsqueezeDims(dilation, 1, 1);
+  if (padding.nbDims==1)  
+    padding = util::unsqueezeDims(padding, 1, 0);
+  if (out_padding.nbDims==1)
+    out_padding = util::unsqueezeDims(out_padding, 1, 0);
+
```
narendasan, line 50: I think its worth having the logging for what the shapes are when pulled right out of the args structure and then adding additional logging that says we are unsqueezing to 2D. That way its explicit what is happening

narendasan: I know we do that already for the input dims but I think it would be nice for the rest too

core/conversion/converters/impl/pooling.cpp
```diff
@@ -9,100 +10,163 @@ namespace converters {
...
     stride = util::toDims(args[1].unwrapToIntList());
   }
-  LOG_DEBUG("stride: " << stride);
 
-  bool ceil_mode = args[4].unwrapToBool();
-  bool count_inlcude_pad = args[5].unwrapToBool();
+  bool shuffle_back = false;
+  if (kernel_size.nbDims == 1) {
+    kernel_size = util::unsqueezeDims(kernel_size, 0, 1);
+    if (shuffle)
```
narendasan, line 127: same style comment

core/conversion/converters/impl/pooling.cpp
```diff
@@ -9,100 +10,163 @@ namespace converters {
...
+      shuffle_back = true;
+    LOG_DEBUG("kernel_size.nbDims < 2, padding:" << kernel_size);
+    LOG_DEBUG("kernel_size: " << kernel_size);
+  }
+  if (padding.nbDims == 1)
```
narendasan, line 132: same style comment

core/conversion/converters/impl/pooling.cpp
```diff
@@ -9,100 +10,163 @@ namespace converters {
...
+    padding = util::unsqueezeDims(padding, 0, 0); 
+  if (stride.nbDims == 1)
```
narendasan, line 134: same style comment

core/conversion/converters/impl/pooling.cpp
```diff
@@ -9,100 +10,163 @@ namespace converters {
...
 
-  auto new_layer = ctx->net->addPoolingNd(*in, nvinfer1::PoolingType::kAVERAGE, kernel_size);
-  TRTORCH_CHECK(new_layer, "Unable to create Avg Pooling layer from node: " << *n);
+  bool ceil_mode;
+  nvinfer1::IPoolingLayer* new_layer;
+
+  if (pool_type == nvinfer1::PoolingType::kMAX) {
+    auto dilation = util::toDims(args[4].unwrapToIntList());
+
+    TRTORCH_ASSERT(
```
narendasan, line 147: This should be a TRTORCH_CHECK

core/conversion/evaluators/prim.cpp
```diff
@@ -48,7 +48,13 @@ auto prim_registrations =
                           c10::List<int64_t> list;
                           list.reserve(num_inputs);
                           for (auto in : n->inputs()) {
-                            list.emplace_back(std::move(args.at(in).unwrapToInt()));
+                            if (args.at(in).IValue()->isInt()) {
+                              list.emplace_back(std::move(args.at(in).unwrapToInt()));
+                            } else if (args.at(in).IValue()->isTuple()) {
```
narendasan, line 53: Is this valid in TorchScript? like I guess this would be:

```
%a : (Int) = prim::Constant[value=(1,2,3)]
%b : (Int) = prim::Constant[value=(4,5,6)]
%x : Int[] = prim::ListConstruct(%a, %b)
```
Where `%x` is `[1,2,3,4,5,6]` or something? 

Wouldn't this also let you do: 
```
%a : (Int) = prim::Constant[value=(1,2,3)]
%b : Int = prim::Constant[value=4]
%x : Int[] = prim::ListConstruct(%a, %b)
```

Where `%x` would be `[1, 2, 3, 4]`

narendasan: Naively I would think that the first case should result in `[(1,2,3),(4,5,6)]` and the second case shouldnt be allowed

core/util/trt_util.cpp
```diff
@@ -208,56 +160,37 @@ nvinfer1::Dims unpadDims(const nvinfer1::Dims& d) {
...
-  while (i <= d.nbDims) {
-    if (j != pos) {
-      dims.d[j] = d.d[i];
-      i++;
-    } else {
-      // add new dimension at pos
-      dims.d[j] = 1;
+  for (int i = 0, j = 0; j <= d.nbDims; j++) {
+    // add new dimension at pos
+    if (j == pos)
```
narendasan, line 170: same syntax comment

core/util/trt_util.cpp
```diff
@@ -208,56 +160,37 @@ nvinfer1::Dims unpadDims(const nvinfer1::Dims& d) {
   return dims;
 }
 
-nvinfer1::Dims unsqueezeDims(const nvinfer1::Dims& d, int pos) {
+nvinfer1::Dims unsqueezeDims(const nvinfer1::Dims& d, int pos, int val, bool use_zeros) {
```
narendasan, line 163: Why do we need both val and use_zeros?

core/util/trt_util.cpp
```diff
@@ -208,56 +160,37 @@ nvinfer1::Dims unpadDims(const nvinfer1::Dims& d) {
...
-      // add new dimension at pos
-      i++;
-      if (i <= d.nbDims) {
-        dims.d[j] = d.d[i];
-      }
-    }
-    i++;
-    j++;
+  for (int i = 0; i < d.nbDims; i++) {
+    if (i != pos)
```
narendasan, line 190: Same syntax comment

core/conversion/converters/impl/pooling.cpp
```diff
@@ -9,100 +10,163 @@ namespace converters {
 namespace impl {
 namespace {
 
-bool MaxPoolingConverter(ConversionCtx* ctx, const torch::jit::Node* n, args& args) {
+bool GlobalPoolingConverter(ConversionCtx* ctx, const torch::jit::Node* n, args& args, nvinfer1::PoolingType pool_type)
```
narendasan, line 13: Is GlobalPooling for when window size is the size of the input Tensor or when the output is a 0d tensor or something ?

core/conversion/converters/converter_util.cpp
borisfom: What's wrong with returning nullptr ? The idea is that you call it on 3 or 4 dimensional input and it returns shuffle layer if shuffle is needed and nullptr otherwise. If we operate on Tensors and/or consider 4d input to be error, that would put conditions in the caller (which may be a good thing making it more explicit ), but would hide shuffle layer from the caller. Also, net->addXxx methods return ILayers, so I thought it would be more intuitive.

core/conversion/evaluators/prim.cpp
borisfom: That stemmed out from my early attempts to use rewriting rules where I followed some existing rewriting sample - so apparently it is valid. If there are any doubts - we do not have to include this fix, as I ended up not doing rewrites.

core/util/trt_util.cpp
borisfom: use _zeros controls whether we should be using 0 instead of -1 on the shape. I have found that for convolutions, we do need to use 0s when there are more than 1 dynamic dimensions. Note it also sets zeroIsPlaceholder attribute for the shuffle.  Some pooling cases failed with that setting though, therefore I made it optional.

core/conversion/converters/converter_util.cpp
narendasan: Really its about reducing sharp edges. In theory power users of TRTorch might interact with these tools in either their own converters or as contributors. Conditionally returning `nullptr` could cause components missing the logic to handle it or with poor testing coverage to fail unexpectedly (to the user) and with a segfault vs some descriptive error. 

"that would put conditions in the caller (which may be a good thing making it more explicit )". Yeah agree, id rather add caller conditions and be explicit. So its on the converter writer to decided if a shuffle layer is needed. Or we can set clear conditions for usage and if those conditions aren't met we error out with a meaningful message so its clear what happened. 

RE: ILayer v. ITensor, I think the argument for tensor is that I dont think people will do a whole lot extra with the ILayer itself (name is set for them and all the settings are set for them). So really we are just asking people to add the `layer->getOutput(0)` boiler plate. I think theres a couple other places (tensor_to_const) where we dont bother returning the layer we just return the relevant ITensor since thats the piece that people really care about.

borisfom: I actually deleted my initial comment - I guess, not fast enough :) Anyway, it does look better now that I followed your suggestion. I have removed 'Layer' from the name, too.

core/conversion/converters/impl/batch_norm.cpp
borisfom: Agreed about people better decide explicitly.

core/conversion/converters/impl/conv_deconv.cpp
borisfom: That I think would be too much logging, and not very useful.

core/conversion/converters/impl/pooling.cpp
borisfom: Currenly, i only route the corner case when out shape is all 1's here. There may be more, yes.

core/conversion/converters/converter_util.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    LOG_DEBUG("Original shape: " << dims << ", reshaping to: " << newDims);
+    auto shuffle_layer = ctx->net->addShuffle(*tensor);
+    TRTORCH_CHECK(shuffle_layer, "Unable to create shuffle layer");
+    shuffle_layer->setReshapeDimensions(newDims);
+    shuffle_layer->setZeroIsPlaceholder(use_zeros);
+    shuffle_layer->setName((util::node_info(n) + " [Reshape to " + util::toStr(newDims) + ']').c_str());
+    return shuffle_layer->getOutput(0);
+  } else {
+    return nullptr;
```
narendasan, line 27: I think that we should either pass through the original tensor here (`tensor`) or error out instead of `nullptr`. What do you think makes the most sense?

narendasan: Looks like we pass through below, we should make them consistent then.

core/conversion/converters/impl/pooling.cpp
```diff
@@ -9,361 +10,222 @@ namespace converters {
...
+    TRTORCH_CHECK(new_layer, "Unable to create Max Pooling layer from node: " << *n);
+  } else if (pool_type == nvinfer1::PoolingType::kAVERAGE) {
+    ceil_mode = args[4].unwrapToBool();
+    bool count_inlcude_pad = args[5].unwrapToBool();
 
+    new_layer = ctx->net->addPoolingNd(*in, pool_type, kernel_size);
+    TRTORCH_CHECK(new_layer, "Unable to create Avg Pooling layer from node: " << *n);
+    new_layer->setAverageCountExcludesPadding(!count_inlcude_pad);
+  } else {
+    TRTORCH_ASSERT(0, "Unsupported pool mode!");
```
narendasan, line 163: We can use TRTORCH_THROW_ERROR for this instead of an always failing assert

narendasan (commented):
Think everything looks almost ready to go. It seems like you are still returning nullptr for the padding helper though you intended to pass through the original tensor. Also changing some of the error checking macros to ones that would be more relevant for what is being done

core/conversion/converters/impl/pooling.cpp
narendasan: Bump

core/conversion/converters/converter_util.cpp
borisfom: Fixed.

core/conversion/converters/impl/pooling.cpp
borisfom: Fixed.

narendasan (approved):
LGTM!

PR #462, opened 2021-05-13
fix: fix the bug when TensorRT segment depends a nonTensor input which is produced in an unsupported op

bowang95:
# Description

There are some cases that TensorRT segments use some nonTensor input which is produced in Torch segments. However, the node in the dependencies for these nonTensor input might not be supported in TRTorch. So, we need to segment this TensorRT segment again and add the dependency nodes. 

Fixes # (issue)
Some bugs related to cannot find IValues for listConstruct operation.

## Type of change

# Checklist:

### DIFF
--- a/core/partitioning/partitioning.cpp
+++ b/core/partitioning/partitioning.cpp
@@ -10,9 +10,9 @@ namespace core {
 namespace partitioning {
 
 struct usage_info {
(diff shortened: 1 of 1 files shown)


narendasan (approved):
LGTM

PR #477, opened 2021-05-18
feat(//core/plugins): Add adaptive_max_pool2d plugin, enable the plugins to run on GPU

peri044:
# Description

Add adaptive max pool2d plugin to the new plugins interface. Add changes in the PR https://github.com/NVIDIA/TRTorch/pull/328 to the latest codebase. 

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/pooling.cpp
+++ b/core/conversion/converters/impl/pooling.cpp
@@ -60,77 +60,53 @@ bool AdaptivePoolingConverter(
   auto in_shape = util::toVec(in->getDimensions());
   nvinfer1::ILayer* new_layer = nullptr;
 
-  if (ctx->input_is_dynamic) {
-#if NV_TENSORRT_MAJOR < 7 || (NV_TENSORRT_MAJOR == 7 && NV_TENSORRT_MINOR < 1)
-    LOG_WARNING(
-        "Adaptive pooling layer will be run through ATen, via not TensorRT, performace will be lower than expected. Consider switching either to static input shape or moving to non adaptive pooling if this is an issue");
-#else
-    LOG_WARNING(
-        "Adaptive pooling layer will be run through ATen (on CPU), via not TensorRT, performace will suffer. Consider switching either to static input shape or moving to non adaptive pooling");
-#endif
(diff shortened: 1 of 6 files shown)


core/conversion/converters/impl/pooling.cpp
```diff
@@ -60,77 +60,53 @@ bool AdaptivePoolingConverter(
...
+  fc.nbFields = f.size();
+  fc.fields = f.data();
+  /*====== PLUGIN PARAMETERS CONFIGURATION COMPLETED ======*/
 
-    std::vector<int64_t> window(out_size.nbDims);
-    for (int64_t i = 0; i < out_size.nbDims; i++) {
-      window[window.size() - 1 - i] =
-          in_shape[in_shape.size() - 1 - i] - (out_size.d[out_size.nbDims - 1 - i] - 1) * stride[stride.size() - 1 - i];
-    }
+  LOG_WARNING(
```
narendasan, line 102: This warning is unclear, seems like some text is missing from the original

narendasan (commented):
Minor comment but seems fine in general

core/conversion/converters/impl/pooling.cpp
narendasan: Also static shape would not trigger adaptive pooling

peri044: Fixed it now.

peri044: Fixed it now

peri044: Removed the static shape handling via TensorRT. That was the reason for adaptive avg pool bugs as listed in #328 . Adaptive pool (max/avg) regardless of static/dynamic shapes run via aten library now.

narendasan: Is there a reason for this? We should be able to calculate the parameters if we know the input shape

peri044: No reason. I followed the PR and replaced it but I can look into calculating the parameters. In the PR, the tests were failing for some shapes with the static shape code.

PR #484, opened 2021-05-25
feat(//core)!: Added support for Device meta data serialization and d…

andi4191:
…eserialization implicitly

## Description

BREAKING CHANGE: Change in the signature of EmbedEngineInNewModule API will result in a breaking change.
Runtime has added support for Cuda Device metadata serialization and deserialization.  TRTEngine constructor is modified.

Design discussion can be found here: 
https://github.com/NVIDIA/TRTorch/discussions/311

## Type of change

# Checklist:


core/runtime/runtime.h
```diff
@@ -11,6 +11,73 @@ namespace runtime {
 
 using EngineID = int64_t;
 
+typedef enum { DeviceIdx = 0, EngineIdx } SerializedInfoIndex;
```
narendasan, line 14: Can we make these all caps?

core/runtime/runtime.h
```diff
@@ -11,6 +11,73 @@ namespace runtime {
...
+
+struct CudaDevice {
+  int64_t id; // CUDA device id
+  int64_t major; // CUDA compute major version
+  int64_t minor; // CUDA compute minor version
+  nvinfer1::DeviceType device_type;
+  size_t device_name_len;
```
narendasan, line 21: Why do you need this?

cpp/api/include/trtorch/trtorch.h
```diff
@@ -517,14 +517,15 @@ TRTORCH_API std::string ConvertGraphToTRTEngine(
...
+ * @param info: trtorch::CompileSpec - Compilation settings
  *
  * Takes a pre-built serialized TensorRT engine and embeds it in a TorchScript
  * module. Registers execution of the engine as the forward method of the module
  * Forward is defined as: forward(Tensor[]) -> Tensor[]
  *
  * @return: A new module trageting a TensorRT engine
  */
-TRTORCH_API torch::jit::Module EmbedEngineInNewModule(const std::string& engine);
+TRTORCH_API torch::jit::Module EmbedEngineInNewModule(const std::string& engine, CompileSpec info);
```
narendasan, line 528: We should just do CompileSpec::Device since the rest is unused, and allow this to work without the device parameter where we just select a device that fits the settings for the engine

core/compiler.cpp
```diff
@@ -220,7 +222,9 @@ torch::jit::script::Module CompileGraphWithFallback(const torch::jit::script::Mo
           convert_cfg.input_ranges = input_ranges;
           auto engine = conversion::ConvertBlockToEngine(seg_block.block(), convert_cfg, named_params);
           auto temp_g = std::make_shared<torch::jit::Graph>();
-          AddEngineToGraph(new_mod, temp_g, engine, trt_engine_id.str(), true);
+          auto device_spec = convert_cfg.engine_settings.device;
+          auto cuda_device = runtime::get_device_info(device_spec.gpu_id, device_spec.device_type);
```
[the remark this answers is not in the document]
andi4191: Yes, if the compile spec specifies device type DLA and gpu_id as Xavier.

core/plugins/impl/normalize_plugin.cpp
```diff
@@ -1,295 +1,295 @@
...
-  return nullptr;
-}
-
-REGISTER_TRTORCH_PLUGIN(NormalizePluginCreator);
-
-} // namespace impl
-} // namespace plugins
-} // namespace core
-} // namespace trtorch
+#include "core/plugins/impl/normalize_plugin.h"
```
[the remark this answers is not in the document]
andi4191: This file was changed by linter automatically.

core/compiler.h
```diff
@@ -22,7 +23,7 @@ std::string ConvertGraphToTRTEngine(const torch::jit::script::Module& mod, std::
 
 torch::jit::script::Module CompileGraph(const torch::jit::script::Module& module, CompileSpec cfg);
 
-torch::jit::script::Module EmbedEngineInNewModule(const std::string& engine);
+torch::jit::script::Module EmbedEngineInNewModule(const std::string& engine, CompileSpec cfg);
```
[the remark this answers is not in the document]
andi4191: CudaDevice is internal to core/runtime. If the end-user wants to use CudaDevice, we would need to add a mapper from the external device struct to the internal CudaDevice.

core/runtime/register_trt_op.cpp
```diff
@@ -10,8 +10,92 @@ namespace trtorch {
...
+}
+
+int select_cuda_device(const CudaDevice& conf_device) {
+  int device_id = 0;
+  int num_devices = 0;
+  // SM Compute capability <major,minor> pair
+  std::unordered_map<std::string, std::string> dla_supported_SM;
+
+  // Xavier SM Compute Capability
+  dla_supported_SM.insert(std::make_pair("7.2", "Xavier"));
```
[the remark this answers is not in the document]
andi4191: Since Volta architecture also shares the SM capability. If we don't add Xavier/Orin to it will cause errors for dGPUs based on compute capability 7.2  such as Titan V cards.

I cannot think of any other way around this. For Orin, we will have to add that to the list as well.

core/runtime/register_trt_op.cpp
```diff
@@ -10,8 +10,92 @@ namespace trtorch {
...
+  auto status = cudaGetDeviceCount(&num_devices);
+  TRTORCH_CHECK((status == cudaSuccess), "Unable to read CUDA capable devices. Return status: " << status);
+
+  cudaDeviceProp device_prop;
+
+  for (int i = 0; i < num_devices; i++) {
```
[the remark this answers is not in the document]
andi4191: Outside execute_engine or as soon as execute_engine starts?

core/runtime/register_trt_op.cpp
```diff
@@ -10,8 +10,92 @@ namespace trtorch {
...
+    }
+  }
+  return device_id;
+}
+
 std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intrusive_ptr<TRTEngine> compiled_engine) {
   LOG_DEBUG("Attempting to run engine (ID: " << compiled_engine->name << ")");
+
+  CudaDevice curr_device;
+  get_cuda_device(curr_device);
```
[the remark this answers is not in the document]
andi4191: Sure

core/runtime/runtime.h
andi4191: ok

core/runtime/runtime.h
andi4191: I need to know the byte size I am supposed to deserialize for the device_name field.

cpp/api/include/trtorch/trtorch.h
andi4191: ok, that can be done

core/runtime/runtime.h
narendasan: Can you just get the size of the string?

core/runtime/register_trt_op.cpp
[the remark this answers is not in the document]
narendasan: Like when the runtime is loaded on start up. We can query all available devices and store them so we are doing this during execution

core/runtime/register_trt_op.cpp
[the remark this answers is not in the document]
narendasan: Can you make a static DLA supported SM map then like this: https://github.com/NVIDIA/TRTorch/blob/master/core/conversion/conversion_ignorelist.cpp

core/compiler.h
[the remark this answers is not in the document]
narendasan: the C++ API can access the CudaDevice struct, we should just go straight from CompileSpec::Device to runtime::CudaDevice

core/plugins/impl/normalize_plugin.cpp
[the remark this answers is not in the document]
narendasan: can you remove it from the PR

core/runtime/runtime.h
andi4191: how will that work during deserialization? It needs to understand how many bytes form part of device_name.  For other primitive data types, it makes sense as it will be parsing based on sizeof(datatype) but for string it will vary, I don't think we have such an option.

core/plugins/impl/normalize_plugin.cpp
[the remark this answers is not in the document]
andi4191: It comes back after running linter. Resetting this file generates Bot requested changes

core/compiler.h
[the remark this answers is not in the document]
andi4191: Incorporated review comment

core/runtime/register_trt_op.cpp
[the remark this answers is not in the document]
andi4191: Done

core/runtime/register_trt_op.cpp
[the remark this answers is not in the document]
andi4191: Added a singleton DeviceList class to track available Cuda devices for runtime.

core/runtime/runtime.h
andi4191: Resolving this conversation as per discussion.

core/compiler.cpp
```diff
@@ -27,13 +27,35 @@
 namespace trtorch {
 namespace core {
 
+static std::unordered_map<int, runtime::CudaDevice> cuda_device_list;
```
narendasan, line 30: We need to move this into the runtime module, if people are using libtrtorchrt they will not have access to this code

core/compiler.cpp
```diff
@@ -27,13 +27,35 @@
...
 void AddEngineToGraph(
     torch::jit::script::Module mod,
     std::shared_ptr<torch::jit::Graph>& g,
     const std::string& serialized_engine,
+    runtime::CudaDevice& device_info,
     std::string engine_id = "",
     bool fallback = false) {
-  auto engine_ptr = c10::make_intrusive<runtime::TRTEngine>(mod._ivalue()->name() + engine_id, serialized_engine);
+  // Scan and Update the list of available cuda devices
+  update_cuda_device_list();
```
narendasan, line 56: We should do this once when the runtime library is loaded, i dont think theres a case where the devices will change during runtime

cpp/api/include/trtorch/trtorch.h
```diff
@@ -517,14 +517,15 @@ TRTORCH_API std::string ConvertGraphToTRTEngine(
  * in a TorchScript module
  *
  * @param engine: std::string - Pre-built serialized TensorRT engine
+ * @param info: CompileSepc::Device - Device information
```
narendasan, line 520: param is device

core/compiler.cpp
andi4191: Moved this inside runtime.

core/compiler.cpp
andi4191: Done

cpp/api/include/trtorch/trtorch.h
andi4191: changed

PR #501, opened 2021-06-17
Enable LogLevel.GRAPH in Python API

letian-jiang:
# Description

Add certain codes to enable `LogLevel.GRAPH` in python. 

## Type of change

# Checklist:

### DIFF
--- a/py/trtorch/logging.py
+++ b/py/trtorch/logging.py
@@ -13,6 +13,7 @@ class Level(Enum):
     Warning = LogLevel.WARNING
     Info = LogLevel.INFO
     Debug = LogLevel.DEBUG
+    Graph = LogLevel.GRAPH
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


narendasan:
We initially did not expose graph level logs without modifying the build since its only really useful for lowering debugging which is not user alterable. If you think this is helpful for users, I am totally okay with exposing it, but we should do the same for the C++ API as well.

py/trtorch/logging.py
```diff
@@ -26,6 +27,8 @@ def _to_internal_level(external) -> LogLevel:
             return LogLevel.INFO
         if external == Level.Debug:
             return LogLevel.DEBUG
+        if external == Level.Graph:
+            return LogLevel.GRAPH
```
narendasan, line 31: We should probably add an exception or something here if the program gets this far into the cases.

narendasan (approved):

PR #518, opened 2021-06-30
chore: Add efficientnet b0 and VIT to testsuite

peri044:
# Description

Add efficientnet and vit to testsuite

## Type of change

# Checklist:

### DIFF
--- a/tests/modules/hub.py
+++ b/tests/modules/hub.py
@@ -66,8 +66,12 @@ models = {
         "model": models.detection.fasterrcnn_resnet50_fpn(pretrained=True),
         "path": "script"
     },
-    "vit": {
+    "efficientnet_b0": {
         "model": timm.create_model('efficientnet_b0', pretrained=True),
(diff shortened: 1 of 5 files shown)


narendasan (commented):
Can we change these tests slightly so that the threshold is an argument? would be good not to lower the threshold for the existing networks

peri044:
@narendasan The threshold needs to be increased for VIT. I think it might be because of GeLU. The difference for a test image that I saw is around (6~7)e-3. Accuracy on imagenet seems to be almost similar to pytorch.

narendasan:
Yeah so we should parameterize the threshold like we do the model and the input size so that only VIT uses the lower threshold

peri044:
> Yeah so we should parameterize the threshold like we do the model and the input size so that only VIT uses the lower threshold

Done.

narendasan (approved):
lgtm

PR #520, opened 2021-07-01
feat(serde)!: Refactor CudaDevice struct, implement ABI versioning, serde cleanup

narendasan:
# Description

BREAKING CHANGE: This commit cleans up the WIP CudaDevice class,
simplifying implementation and formalizing the seralized format for CUDA
devices.

It also implements ABI Versioning. The first entry in the serialized
format of a TRTEngine now records the ABI that the engine was compiled
with, defining expected compatibility with the TRTorch runtime. If the
ABI version does not match, the runtime will error out asking to
recompile the program.

ABI version is a monotonically increasing integer and should be
incremented everytime the serialization format changes in some way.

This commit cleans up the CudaDevice class, implementing a number of
constructors to replace the various utility functions that populate the
struct. Descriptive utility functions remain but solely call the
relevant constructor.
Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/compiler.cpp
+++ b/core/compiler.cpp
@@ -223,7 +223,7 @@ torch::jit::script::Module CompileGraphWithFallback(const torch::jit::script::Mo
           auto engine = conversion::ConvertBlockToEngine(seg_block.block(), convert_cfg, named_params);
           auto temp_g = std::make_shared<torch::jit::Graph>();
           auto device_spec = convert_cfg.engine_settings.device;
-          auto cuda_device = runtime::get_device_info(device_spec.gpu_id, device_spec.device_type);
+          auto cuda_device = runtime::CudaDevice(device_spec.gpu_id, device_spec.device_type);
           AddEngineToGraph(new_mod, temp_g, engine, cuda_device, trt_engine_id.str(), true);
 
           seg_block.update_graph(temp_g);
@@ -265,7 +265,7 @@ torch::jit::script::Module CompileGraph(const torch::jit::script::Module& mod, C
       auto engine = ConvertGraphToTRTEngine(mod, method.name(), cfg);
       auto new_g = std::make_shared<torch::jit::Graph>();
       auto device_spec = cfg.convert_info.engine_settings.device;
-      auto cuda_device = runtime::get_device_info(device_spec.gpu_id, device_spec.device_type);
+      auto cuda_device = runtime::CudaDevice(device_spec.gpu_id, device_spec.device_type);
       AddEngineToGraph(new_mod, new_g, engine, cuda_device);
       auto new_method = new_mod._ivalue()->compilation_unit()->create_function(method.name(), new_g);
       auto schema = util::GenerateGraphSchema(new_method->name(), new_g);
@@ -277,9 +277,7 @@ torch::jit::script::Module CompileGraph(const torch::jit::script::Module& mod, C
   return new_mod;
 }
 
-torch::jit::script::Module EmbedEngineInNewModule(
-    const std::string& engine,
-    trtorch::core::runtime::CudaDevice cuda_device) {
+torch::jit::script::Module EmbedEngineInNewModule(const std::string& engine, runtime::CudaDevice cuda_device) {
   std::ostringstream engine_id;
   engine_id << reinterpret_cast<const int*>(&engine);
   torch::jit::script::Module new_mod("tensorrt_engine_mod_" + engine_id.str());
--- a/core/runtime/CudaDevice.cpp
+++ b/core/runtime/CudaDevice.cpp
@@ -34,9 +34,6 @@ CudaDevice::CudaDevice(int64_t gpu_id, nvinfer1::DeviceType device_type) {
   // Set Device name
   this->device_name = device_name;
 
-  // Set Device name len for serialization/deserialization
-  this->device_name_len = device_name.size();
-
   // Set Device Type
   this->device_type = device_type;
 }
@@ -70,14 +67,20 @@ CudaDevice::CudaDevice(std::string device_info) {
 }
 
 std::string CudaDevice::serialize() {
+  std::vector<std::string> content;
+  content.resize(DEVICE_NAME_IDX + 1);
+
+  content[ID_IDX] = std::to_string(id);
+  content[SM_MAJOR_IDX] = std::to_string(major);
+  content[SM_MINOR_IDX] = std::to_string(minor);
+  content[DEVICE_TYPE_IDX] = std::to_string((int64_t)device_type);
+  content[DEVICE_NAME_IDX] = device_name;
+
   std::stringstream ss;
-  // clang-format off
-  ss << id << DEVICE_INFO_DELIM \
-     << major << DEVICE_INFO_DELIM \
-     << minor << DEVICE_INFO_DELIM \
-     << (int64_t) device_type << DEVICE_INFO_DELIM
-     << device_name;
-  // clang-format on
+  for (size_t i = 0; i < content.size() - 1; i++) {
+    ss << content[i] << DEVICE_INFO_DELIM;
+  }
+  ss << content[DEVICE_NAME_IDX];
 
   std::string serialized_device_info = ss.str();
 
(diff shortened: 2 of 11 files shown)


core/runtime/register_trt_op.cpp
```diff
@@ -10,80 +10,88 @@ namespace trtorch {
...
     } else {
-      LOG_ERROR("Unkown device type detected from the compiled engine");
+      TRTORCH_THROW_ERROR("Unknown target device type detected from the compiled program (runtime.select_cuda_device)");
       break;
     }
   }
-  return device_id;
+
+  // REVIEW: THIS DOES NOT LIST DLA PROBABLY, WHICH WE SHOULD
+  TRTORCH_CHECK(
```
narendasan, line 78: We should figure out a way to add DLA to the index or have a DLA index or something, I think the debugability for GPUs is good now but for DLA it kind of sucks

core/runtime/CudaDevice.cpp
```diff
@@ -0,0 +1,103 @@
...
+  // Set Device name
+  this->device_name = device_name;
+
+  // Set Device name len for serialization/deserialization
+  this->device_name_len = device_name.size();
+
+  // Set Device Type
+  this->device_type = device_type;
+}
+
```
narendasan, line 43: Switched to a delimted string vs. a byte indexed string, should be easier to parse

core/runtime/CudaDevice.cpp
```diff
@@ -0,0 +1,103 @@
...
+  // Compute capability minor version
+  this->minor = device_prop.minor;
+
+  std::string device_name(device_prop.name);
+
+  // Set Device name
+  this->device_name = device_name;
+
+  // Set Device name len for serialization/deserialization
+  this->device_name_len = device_name.size();
```
narendasan, line 38: I think we can remove this now

core/runtime/register_trt_op.cpp
```diff
@@ -10,80 +10,88 @@ namespace trtorch {
...
-      auto conf_sm = std::to_string(conf_device.major) + "." + std::to_string(conf_device.minor);
+      auto conf_sm = conf_device.getSMCapability();
       if (compute_cap == conf_sm && device.second.device_name == conf_device.device_name) {
         device_id = device.second.id;
+        new_target_device = CudaDevice(device_id, nvinfer1::DeviceType::kGPU);
         break;
       }
     } else {
-      LOG_ERROR("Unkown device type detected from the compiled engine");
+      TRTORCH_THROW_ERROR("Unknown target device type detected from the compiled program (runtime.select_cuda_device)");
```
narendasan, line 72: the LOG_ERROR would not stop the program from executing, not sure if we should continue if we dont know the device type

core/compiler.cpp
```diff
@@ -277,9 +277,7 @@ torch::jit::script::Module CompileGraph(const torch::jit::script::Module& mod, C
   return new_mod;
 }
 
-torch::jit::script::Module EmbedEngineInNewModule(
-    const std::string& engine,
-    trtorch::core::runtime::CudaDevice cuda_device) {
+torch::jit::script::Module EmbedEngineInNewModule(const std::string& engine, runtime::CudaDevice cuda_device) {
```
andi4191, line 280: Have you ran a linter on this?

core/runtime/CudaDevice.cpp
andi4191: Agreed

core/runtime/CudaDevice.cpp
andi4191: It still requires parsing based on the positional encoding of the parameters during deserialization.

core/runtime/CudaDevice.cpp
```diff
@@ -0,0 +1,103 @@
...
+  id = std::stoi(tokens[ID_IDX]);
+  major = std::stoi(tokens[SM_MAJOR_IDX]);
+  minor = std::stoi(tokens[SM_MINOR_IDX]);
+  device_type = (nvinfer1::DeviceType)(std::stoi(tokens[DEVICE_TYPE_IDX]));
+  device_name = tokens[DEVICE_NAME_IDX];
+
+  LOG_DEBUG("Deserialized Device Info: " << *this);
+}
+
+std::string CudaDevice::serialize() {
```
andi4191, line 72: It seems inconsistent with the deserialization in terms of position-based parsing?

How about
`std::vector<std::string> serial_data_vec(<PARAM>_IDX);
serial_data_vec[ID_IDX] = id;
`

In the end we can append the strings with delimiter based on index position. 

This way mapping will be independent of the order of parameter parsing/embedding. Let me know your thought?

core/runtime/register_trt_op.cpp
andi4191: From a TRT perspective, DLA is configurable using the GPU id. Even if we add a mechanism for DLA deviceType. Internally we would still be validating against Xavier's GPU id.

It makes sense to keep track of DLA Core instead which is already taken care of. (Use Case: Hybrid device configuration for inference)

core/runtime/TRTEngine.cpp
```diff
@@ -30,6 +32,12 @@ TRTEngine::TRTEngine(std::vector<std::string> serialized_info)
           std::string("[] = "),
           util::logging::get_logger().get_reportable_severity(),
           util::logging::get_logger().get_is_colored_output_on()) {
+  TRTORCH_CHECK(
+      serialized_info.size() == ENGINE_IDX + 1, "Program to be deserialized targets an incompatible TRTorch ABI");
```
andi4191, line 36: Why serialized_info.size() is being compared against ENGINE_IDX + 1?
Shouldn't this be versioned along with TRTorch?

core/runtime/register_trt_op.cpp
```diff
@@ -10,80 +10,87 @@ namespace trtorch {
...
   }
 
   // GPU case
   if (conf_device.device_type == nvinfer1::DeviceType::kGPU) {
     if (curr_device.device_name != conf_device.device_name) {
       LOG_WARNING(
-          "TRTEngine compiled for " << conf_device.device_name << " but current CUDA device is "
-                                    << curr_device.device_name << ". Switching the device context");
+          "Program compiled for " << conf_device.device_name << " but current CUDA device is " << curr_device
+                                  << ". Attempting to switch device context for better compatibility");
```
andi4191, line 30: Here "better compatibility" shows that cards with different compute capabilities are compatible which is not the case.

How about "to a matching compute capablity"

core/runtime/register_trt_op.cpp
```diff
@@ -10,80 +10,87 @@ namespace trtorch {
...
+  auto dla_supported = get_dla_supported_SMs();
 
-  auto device_list = cuda_device_list.instance().get_devices();
+  auto device_list = get_available_device_list().get_devices();
+
+  CudaDevice new_target_device;
 
   for (auto device : device_list) {
-    auto compute_cap = std::to_string(device.second.major) + "." + std::to_string(device.second.minor);
+    auto compute_cap = device.second.getSMCapability();
```
andi4191, line 54: Is getSMCapability function consistent with other function naming guidelines such as select_cuda_device?

andi4191 (changes requested):

core/runtime/register_trt_op.cpp
narendasan: This is a device name switch, so like switching from RTX 3070 to RTX 3080, the compute capability should be the same

core/runtime/TRTEngine.cpp
narendasan: wdym versioned? ENGINE_IDX is the largest index, so it verifies the correct size for the array, if we prepend new data, this will stay true, otherwise having some constant would have the same update issue as updating the index

core/runtime/register_trt_op.cpp
narendasan: In general I use camel case for methods and snake for functions, but use whatever feels most natural in that specific case.

andi4191 (approved):

PR #574, opened 2021-08-12
fix logic error of exception elimination

seanxcwang:
# Description

If there is no exception node in both two blocks of If node，the elimination check function will always return true


narendasan:
Hi @seanxcwang, thanks for the contribution! Can you sign your commit? Also a test verifying the change would be good

PR #616, opened 2021-08-31
Example tensors

### LINKED ISSUE #514: ✨[Feature] Let users set input shape, dtype and format using example tensors

Now that we are going to be able to specify tensor properties for inputs such as data type and format, it should be easier to use a tensor I already have as an example of the shape dtype and format I want.

compile should be able to accept a list of Tensors to represent my graph inputs. From those tensors compile should be able to infer shape, data type and tensor format for the user. These should be logged in the debug log for easy debugging. 

Note: I dont think theres a way to support dynamic shape using this method, so users will need to use `trtorch.Input` for that 

**Additional context**

narendasan:
# Description

Adds a new convience constructor to both Python and C++ APIs to allow users to provide example tensors to implicitly set shape, dtype and layout. Does not allow for dynamic shape, there the existing Input constructors should be used

Fixes #514 

## Type of change

# Checklist:

### DIFF
--- a/cpp/include/trtorch/trtorch.h
+++ b/cpp/include/trtorch/trtorch.h
@@ -427,7 +427,7 @@ struct TRTORCH_API CompileSpec {
     Input(c10::ArrayRef<int64_t> shape, DataType dtype, TensorFormat format = TensorFormat::kContiguous);
 
     /**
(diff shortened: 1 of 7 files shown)


peri044 (approved):
LGTM. Tests pass

PR #625, opened 2021-09-16
Refactored Dockerfiles

borisfom:

# Description
Refactored Dockerfiles into a single one with build arg for base 
Added a script to build both .tar.gz and wheel distribution - used in Dockerfile.

Fixes # (issue)

## Type of change

# Checklist:


borisfom:
Also, setup.py wanted pytorch < 1.10, which would make it to try downgrading pytorch in the container - fixed that.

py/requirements.txt
```diff
@@ -1,3 +1,3 @@
 -f https://download.pytorch.org/whl/torch_stable.html
-torch==1.9.0+cu111
+torch>=1.9.0+cu111
```
peri044, line 2: Why is this change required ?

docker/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+COPY . /workspace/trtorch/src
+WORKDIR /workspace/trtorch/src
+RUN cp ./docker/WORKSPACE.cu.docker  WORKSPACE
+
+# This script builds both libtrtorch bin/lib/include tarball and the Pythin wheel, in dist/  
+RUN ./docker/dist-build.sh
+
+# This currently does not work with BUILDKIT=1
+# RUN ./docker/dist-test.sh
```
peri044, line 28: Is the intention here only to run python tests ? If so, why ? 
Besides, I'm not sure if this would work. I used to run python tests directly. @narendasan  can comment more on this.

peri044 (changes requested):
Please update the README.md with updated instructions on how to use this docker file and versions etc. Tested 21.07 container and it builds successfully.

docker/Dockerfile
narendasan: Yeah the bazel based python tests usually hit errors related to the bazel sandbox. we usually use pytest directly.

py/requirements.txt
borisfom: This is required for 21.07 container.

docker/Dockerfile
borisfom: It won't work, I will clean those up.

borisfom:
> Please update the README.md with updated instructions on how to use this docker file and versions etc. Tested 21.07 container. I face one intermediate error
> 
> ```
> Downloading: "https://github.com/NVIDIA/DeepLearningExamples/archive/torchhub.zip" to /root/.cache/torch/hub/torchhub.zip
> Downloading checkpoint from https://api.ngc.nvidia.com/v2/models/nvidia/ssd_pyt_ckpt_amp/versions/20.06.0/files/nvidia_ssdpyt_amp_200703.pt
>   File "./hub.py", line 60, in <module>
>     "model": torch.hub.load('NVIDIA/DeepLearningExamples:torchhub', 'nvidia_ssd', model_math="fp32"),
>   File "/opt/conda/lib/python3.8/site-packages/torch/hub.py", line 364, in load
>     model = _load_local(repo_or_dir, model, *args, **kwargs)
>   File "/opt/conda/lib/python3.8/site-packages/torch/hub.py", line 393, in _load_local
>     model = entry(*args, **kwargs)
>   File "/root/.cache/torch/hub/NVIDIA_DeepLearningExamples_torchhub/PyTorch/Detection/SSD/src/entrypoints.py", line 187, in nvidia_ssd
>     ckpt = torch.load(ckpt_file)
>   File "/opt/conda/lib/python3.8/site-packages/torch/serialization.py", line 608, in load
>     return _legacy_load(opened_file, map_location, pickle_module, **pickle_load_args)
>   File "/opt/conda/lib/python3.8/site-packages/torch/serialization.py", line 787, in _legacy_load
>     result = unpickler.load()
>   File "/opt/conda/lib/python3.8/site-packages/torch/serialization.py", line 743, in persistent_load
>     deserialized_objects[root_key] = restore_location(obj, location)
>   File "/opt/conda/lib/python3.8/site-packages/torch/serialization.py", line 175, in default_restore_location
>     result = fn(storage, location)
>   File "/opt/conda/lib/python3.8/site-packages/torch/serialization.py", line 151, in _cuda_deserialize
>     device = validate_cuda_device(location)
>   File "/opt/conda/lib/python3.8/site-packages/torch/serialization.py", line 135, in validate_cuda_device
>     raise RuntimeError('Attempting to deserialize object on a CUDA '
> RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU.
> ```
> 
> but rest is fine.

Moved hub.py call to the  test script - CUDA is not available during Docker build.

py/setup.py
```diff
@@ -234,7 +234,7 @@ def run(self):
       long_description=long_description,
       ext_modules=ext_modules,
       install_requires=[
-          'torch>=1.9.0+cu111,<1.10.0',
+          'torch>=1.9.0+cu111,<1.11.0',
```
narendasan, line 237: I dont think we want this change in master until we retarget the mainline codebase to PyTorch 1.10 (post PyTorch's 1.10 release). @andi4191 this might be a good time to start working through the DLFW release workflow. We can retarget this branch to be merged into a `DLFW-21.10` branch (lets all agree on a convention for these branch names @borisfom @andi4191 @ptrblck) and then cherry-pick any changes we want in the mainline at a later date (Like the single unified Dockerfile I think will be super useful)

docker/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+RUN apt-get update && apt-get install -y curl gnupg
+RUN curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor > /etc/apt/trusted.gpg.d/bazel.gpg
+RUN echo "deb [arch=amd64] https://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
+
+RUN apt-get update && apt-get install -y bazel-4.0.0
+RUN ln -s /usr/bin/bazel-4.0.0 /usr/bin/bazel
+
+# Workaround for bazel expecting both static and shared versions, we only use shared libraries inside container
+RUN cp /usr/lib/x86_64-linux-gnu/libnvinfer.so /usr/lib/x86_64-linux-gnu/libnvinfer_static.a
```
narendasan, line 14: For DLFW we should revert this once building is complete

docker/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# copy source repo
+COPY . /workspace/trtorch
+COPY --from=trtorch-builder  /workspace/trtorch/src/dist/ .
+
+RUN conda init bash
+
+RUN pip install ipywidgets --trusted-host pypi.org --trusted-host pypi.python.org --trusted-host=files.pythonhosted.org
+RUN jupyter nbextension enable --py widgetsnbextension
+
+RUN mkdir -p /opt/trtorch && tar xvf libtrtorch.tar.gz --strip-components 2 -C /opt/trtorch --exclude=LICENSE && pip3 install *.whl && rm -fr /workspace/trtorch/dist/*
```
narendasan, line 38: Do we want this library to be in /usr/ or /opt?

docker/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+ENV LD_LIBRARY_PATH /opt/conda/lib/python3.8/site-packages/torch/lib:/opt/trtorch/lib:${LD_LIBRARY_PATH}
+ENV PATH /opt/trtorch/bin:${PATH}
+
+WORKDIR /workspace/trtorch/
```
narendasan, line 43: Do we need to do any sort of clean up for size reduction? Do we conventionally leave the source in container?

narendasan: Like things like bazel are probably not needed post testing and install in the container itself

docker/dist-build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+mkdir -p dist
+
+bazel build //:libtrtorch //:bin --compilation_mode opt
```
narendasan, line 5: Were you not finding trtorchc in the tarball?

docker/Dockerfile
borisfom: This is for trtorch-builder auxiliary container only - neither bazel nor this will go to the Pytorch container

py/setup.py
borisfom: Why is this critical? It just allows TRTorch from master branch to be used in 21.07, 21.08, 21.09 containers, and it breaks nothing ?

docker/Dockerfile
borisfom: This is up for you to decide. Since various parts of TRTorch are going to be pulled to other containers (Triton, Riva SM), it's easier to handle if it's under its own root - but I am fine with either way you choose.

docker/Dockerfile
borisfom: Bazel is only installed in trtorch-builder image, not in target container. I do leave the source in TRTorch container - won't in Pytorch MR.

docker/dist-build.sh
borisfom: Oh you mean //:libtrtorch includes //bin ? I will clean this up though it's probably a no-op

docker/Dockerfile
narendasan: If there is a standing convention in DLFW containers to put libraries we add in /opt then I am fine with /opt. My preference is /usr otherwise since it seems more conventional and easier to link against

docker/Dockerfile
narendasan: ok

py/setup.py
narendasan: Because in the typical case (like not for the state of the repo as it is right now) we only guarantee compatibility with the latest released pytorch since we depend on internal APIs. There have been cases in the past where changes needed to support pytorch-next break support for the previous version. We arent checking against pytorch ToT and if some user has a newer build installed, it might not work even though setup.py says the dependency is compatible. I'd prefer if we only make changes specifically targeted for DLFW in their own branches so that we can keep the guarantees we give to users simple to understand.

borisfom: Well, the concern is understood. However, the effect of user trying to install current ToT TrTorch in 21.07+ container would be not be non-working TrTorch - it would be messing user's container (it will uninstall Pytorch 1.10a and will try to install 1.9). So I believe we'll get less users frustrated if we relax the requirement. Especially that I have checked both 21.07 and 21.08 and at least mostly it works.

docker/Dockerfile
borisfom: Yes we did put things in /opt, like triton and such. Let's use that for now.

py/setup.py
andi4191: @borisfom: I agree with @narendasan  on this. We don't have control over the end-user use cases here. It is safe to proceed with what we can guarantee that works.

Also, can you re-target this PR to release/ngc/21.10 instead of DLFW-21.10 ?

narendasan:
This got upstreamed during the rebrand

PR #674, opened 2021-10-19
feat: support setting input types of subgraph in fallback, handle Ten…

inocsin:
…sor type in evaluated_value_map branch in MarkOutputs

Signed-off-by: vcheungyi@163.com <vcheungyi@163.com>

# Description

When I dig into the "Unkown output type" issue, I found it wasn't caused by Collection, instead, it is caused by prim::NumToTensor.

prim::NumToTensor return an at::Tensor and stored in ctx->evaluated_value_map in MarkOutputs function. So we need to handle such case in MarkOutputs.

Besides, I found that when construct tensor graph, we won't set the input data type, and this data type will be kFloat in default. So I modified some code to support other type like kInt which is returned from prim::NumToTensor.

Fixes https://github.com/NVIDIA/TRTorch/issues/539

## Type of change

# Checklist:


inocsin:
@narendasan

narendasan:
Merged in staging branch

PR #704, opened 2021-11-11
feat: support aten::extend evaluator

guoruoqian:

# Description

Support aten::extend evaluator. The Function schema is `"aten::extend.t(t[](a!) self, t[] other) -> ()"`

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/evaluators/aten.cpp
+++ b/core/conversion/evaluators/aten.cpp
@@ -295,6 +295,7 @@ auto aten_registrations TORCHTRT_UNUSED =
                         for (int64_t i = 0; i < other_size; i++) {
                           self.push_back(other.get(i));
                         }
+                        return self;
                       } else {
                         TORCHTRT_THROW_ERROR(
                             "Unimplemented data type for aten::extend.t evaluator: "
--- a/core/conversion/evaluators/aten.cpp
+++ b/core/conversion/evaluators/aten.cpp
@@ -285,6 +285,26 @@ auto aten_registrations TORCHTRT_UNUSED =
                     EvalOptions().validSchemas({
                         "aten::append.t(t[](a!) self, t(c -> *) el) -> (t[](a!))",
                     })})
+        .evaluator({c10::Symbol::fromQualString("aten::extend"),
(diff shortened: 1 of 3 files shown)


core/conversion/evaluators/aten.cpp
```diff
@@ -285,6 +285,26 @@ auto aten_registrations TORCHTRT_UNUSED =
...
+
+                        for (int64_t i = 0; i < other_size; i++) {
+                          self.push_back(other.get(i));
+                        }
+                      } else {
+                        TORCHTRT_THROW_ERROR(
+                            "Unimplemented data type for aten::extend.t evaluator: "
+                            << args.at(n->input(0)).IValue()->type()->str() << ", "
+                            << args.at(n->input(1)).IValue()->type()->str());
+                      }
```
narendasan, line 303: The function doesnt return anything. While you might be modifying state correctly, we want to make transforms as functional as possible. Create a copy of self, modify it and return that

narendasan (changes requested):

core/conversion/evaluators/aten.cpp
guoruoqian: Hi @narendasan , i tried to return `self`, but the output of node `aten::extend.t(t[](a!) self, t[] other) -> ()` is None. It will occurs an error in [there](https://github.com/NVIDIA/Torch-TensorRT/blob/da15fa5aacf8f524457b132325502e4e6e40a31d/core/conversion/conversion.cpp#L426) like `C++ exception with description "vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)"`

guoruoqian: > The function doesnt return anything. While you might be modifying state correctly, we want to make transforms as functional as possible. Create a copy of self, modify it and return that

I return `self` in `aten::extend` evaluator and add a condition to estimate the number of ouputs in [ConvertBlockToNetDef](https://github.com/NVIDIA/Torch-TensorRT/blob/f0da646a604292335ddbb6c103a677f7b00aace6/core/conversion/conversion.cpp#L401).

PR #774, opened 2021-12-16
chore: Change unsupported operators log level to DEBUG

peri044:
Signed-off-by: Dheeraj Peri <peri.dheeraj@gmail.com>

# Description

Change unsupported operators log level to DEBUG

## Type of change

# Checklist:


narendasan:
I think we should just use the suppress_errors message in the fallback case instead of throwing this in debug. We just need to find the right code path here

narendasan (commented):
Do you have a script that would output the confusing error before this patch but not after?

PR #780, opened 2021-12-21
refactor: Restructured python perf benchmark tool

andi4191:

# Description
Re structured the python perf benchmark suite. Moved it to tools 

Fixes # (issue)

## Type of change

# Checklist:


narendasan:
Do you think we need the C++ benchmark as well?, maybe we have two subdirectories in //tools/perf one for the tool you wrote (we could call this benchmark or if you have an name you like) and the other one could be like cpp_bench or something

andi4191:
@narendasan : This benchmark tool simply expects the .jit.pt or TRT engine file in the configuration to run perf tests. CPP benchmark example has a reference sample to generate and export a TS module. Hence, this tool simply refers to it for example.

narendasan:
Yeah I mean we can change the role of the cpp_benchmark tool. The only reason we output an engine is to test with trtexec. It wasnt *really* intended to be an example other than to define a benchmarking procedure.

andi4191:
@narendasan : Can you take a look at this MR now? I have incorporated changes against the review comment.

narendasan (commented):
Why is the training recipe being moved?

andi4191:
> Why is the training recipe being moved?

Training recipe is still inside the examples/ and not being moved to tools/
It was under int8. It is a simple training recipe. Why does it need to be in examples/int8?

narendasan:
I mean its not really used for anything else from a user perspective

andi4191:
This PR is tracked here: https://github.com/NVIDIA/Torch-TensorRT/pull/879

Closing this PR

PR #808, opened 2022-01-14
Changed the recommended Docker environment to NGC PyTorch container

svenchilton:
# Description

Changed the README file for the notebooks directory to advise the user to run the NGC PyTorch container rather than build an image from the local Dockerfile and run a container from that. Also removed multiple calls to pip install in the notebooks, updated version numbers in calls to pip install scikit-image and timm, and removed a call to torch.hub in the EfficientNet notebook, since that notebook uses timm rather than torch.hub to obtain the pretrained model.

Fixes # (issue)

## Type of change

# Checklist:


narendasan:
Don't we already ship the notebooks in the container? Maybe it is worth keeping the build instructions just in case someone wants to use master branch but still recommending the NGC container as the default

svenchilton:
@narendasan, regarding your comment this morning, it turns out that we do, in fact, ship the notebooks with the PyTorch NGC container, specifically in the container's `/workspace/examples/torch_tensorrt/notebooks` directory. Of course, the notebooks that ship with the container don't contain my recent updates, though strangely enough, they don't encounter the package version issue which necessitated my updates in the first place. Maybe the best thing would be to modify the notebooks README file to outline 3 options: 
1. Run the NGC container without mounting the Torch-TensorRT repo
2. Run the NGC container and mount the Torch-TensorRT repo
3. Build a container from the provided Dockerfile, run it and mount the Torch-TensorRT repo.
How does that sound?

svenchilton:
I augmented the

svenchilton:
I augmented the notebook README file just now as follows: 
Added instructions to help users choose from the following 3 options:
    1. Run the notebooks as provided in the NGC PyTorch container
    2. Mount the repo to the NGC PyTorch container, then run the repo
    notebooks
    3. Build the container from source, mount the repo to that container,
    then run the repo notebooks

narendasan (commented):
I think content wise this is good. I would suggest adding some headings or something to break up the different methods so its easier to navigate at a glance

svenchilton:
@narendasan, I implemented your suggestions in the notebook README file and just pushed my changes.

narendasan (approved):
LGTM

Issue #469, opened 2021-05-14
🐛 [Bug] Unknown output type. Only a single tensor or a TensorList type is supported.


kiddyjinjin:
##  Bug Description

I meet some error msg when trying to trans a fastbert model using trtorch:

^[[1;35mDEBUG: ^[[0m[TRTorch Conversion Context] - Evaluating %4 : Tensor?[] = prim::ListConstruct(%rel_diff_idxs.13)
^[[1;35mDEBUG: ^[[0m[TRTorch Conversion Context] - Found the value to be: [<__torch__.torch.classes._trtorch_eval_ivalue_types.TensorContainer object at 0x876f0f10>]
terminate called after throwing an instance of 'trtorch::Error'
  what():  [enforce fail at core/conversion/conversion.cpp:196] Unknown output type. Only a single tensor or a TensorList type is supported.

## Expected behavior

support different kinds of graph output. not only tensor and tensorlist.

## Environment

> Build information about the TRTorch compiler can be found by turning on debug messages

 - TRTorch Version (e.g. 0.2.0):  latest master version
 - PyTorch Version (e.g. 1.0):  1.8.1
 - CPU Architecture: 
 - OS (e.g., Linux): linux  centos6u3
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source):  source
 - Build command you used (if compiling from source):  
 - Are you using local sources or building from archives:  local sources
 - Python version:  python 3.6.5
 - CUDA version: cuda 10.2
 - GPU models and configuration:  fastbert
 - Any other relevant information:

letian-jiang:
I met the same error when I tried to compile `torchvision.models,inception_v3()`.

PR #919, opened 2022-03-08
fix: skip None output instead of throw an error

void-main:
# Description
For some models, there may be `None` output for scripted model, for example, in `torchvision.inception_v3`, the second output is a None constant.

Current implementation throws error as reported in issue #469 .

I added a special case for `out_ivalue.isNone()` and skip the output to make the model compile.

## Call for discussion
Personally I don't know if it's a good idea to skip an output, it may cause error in client code (out of index error when try to get the output), but I don't know how to build a `None` tensor with TensorRT.

Fixes # (issue)
#469 

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/conversion.cpp
+++ b/core/conversion/conversion.cpp
@@ -229,6 +229,11 @@ void MarkOutputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> outp
           LOG_INFO(
               ctx->logger, "Marking Output " << out->debugName() << " named " << name << " in engine (ctx.MarkOutput)");
           ctx->num_outputs += 1;
+        } else if (out_ivalue.isNone()) { // skip a None output layer
+          std::string name = std::string("output_") + std::to_string(ctx->num_outputs);
+          LOG_INFO(
+              ctx->logger, "Skipping None Output " << out->debugName() << " named " << name << " in engine (ctx.MarkOutput)");
+          ctx->num_outputs += 1;
         } else {
           TORCHTRT_THROW_ERROR("Unknown output type. Only a single tensor or a TensorList type is supported.");
         }
--- a/core/conversion/conversion.cpp
+++ b/core/conversion/conversion.cpp
@@ -231,7 +231,7 @@ void MarkOutputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> outp
           ctx->num_outputs += 1;
         } else if (out_ivalue.isNone()) { // skip a None output layer
           std::string name = std::string("output_") + std::to_string(ctx->num_outputs);
-          LOG_INFO(
+          LOG_WARNING(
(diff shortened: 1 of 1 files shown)


core/conversion/conversion.cpp
```diff
@@ -229,6 +229,11 @@ void MarkOutputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> outp
           LOG_INFO(
               ctx->logger, "Marking Output " << out->debugName() << " named " << name << " in engine (ctx.MarkOutput)");
           ctx->num_outputs += 1;
+        } else if (out_ivalue.isNone()) { // skip a None output layer
+          std::string name = std::string("output_") + std::to_string(ctx->num_outputs);
+          LOG_INFO(
```
peri044, line 234: can you make it LOG_WARNING instead ?

peri044 (approved):

peri044:
@narendasan  the fix seems ok to me. 
The graph for inceptionv3 scripted has the following 
```
%self.Conv2d_1a_3x3.conv.bias.1 : NoneType = prim::Constant()
%1611 : str = prim::Constant[value="Scripted Inception3 always returns Inception3 Tuple"]()
....
....
return (%1955, %self.Conv2d_1a_3x3.conv.bias.1)
```

Maybe this is one weird case of scripting inception models ?. This fix seems reasonable to me.

core/conversion/conversion.cpp
void-main: Sure, I'll do a quick patch.

void-main:
> @narendasan the fix seems ok to me. The graph for inceptionv3 scripted has the following
> 
> ```
> %self.Conv2d_1a_3x3.conv.bias.1 : NoneType = prim::Constant()
> %1611 : str = prim::Constant[value="Scripted Inception3 always returns Inception3 Tuple"]()
> ....
> ....
> return (%1955, %self.Conv2d_1a_3x3.conv.bias.1)
> ```
> 
> Maybe this is one weird case of scripting inception models ?. This fix seems reasonable to me.

yep, the scripted model looks pretty weird, do you have any idea on how to create a `NoneType` tensor?

peri044:
@void-main  @narendasan Looks like this is a special case where they are using `aux` for training which is by default optional null tensor 
https://github.com/pytorch/vision/blob/main/torchvision/models/inception.py#L135-L155

narendasan:
I think what we really need to make sure is that the graph construction phase doesnt break because the input indexes dont meet expectations. For example what happens if the tensor in the middle is None instead of one on the end. Would producing a dummy output be a better option here vs skipping? Maybe we can explicitly produce a NoneTensor in the graph?

void-main:
@narendasan totally agree that we should produce a NoneTensor in the graph. Could you please point out how to create a NoneTensor? I'll change my code.

void-main:
Got it, I'll try to make these changes later this week.

PR #926, opened 2022-03-13
Add PartitioinCtx to correct BlockID errors

bowang95:
# Description
Currently, the SegmentedBlock's BlockID is not correct and could be misleading in debugging process, so fix it in a simple way right now. In the future we might want to add more features in the PartitionCtx class. 

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/partitioning/partitioning.cpp
+++ b/core/partitioning/partitioning.cpp
@@ -200,6 +200,10 @@ void resolveNonTensorInputs(PartitionedGraph& segmented_blocks) { // , std::shar
   }
   segmented_blocks.clear();
   segmented_blocks.insert(segmented_blocks.begin(), segmented_blocks_list.begin(), segmented_blocks_list.end());
+  PartitionCtx partition_context;
+  for (auto& seg_block : segmented_blocks) {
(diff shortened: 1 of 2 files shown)


core/partitioning/partitioning.cpp
```diff
@@ -200,6 +200,10 @@ void resolveNonTensorInputs(PartitionedGraph& segmented_blocks) { // , std::shar
   }
   segmented_blocks.clear();
   segmented_blocks.insert(segmented_blocks.begin(), segmented_blocks_list.begin(), segmented_blocks_list.end());
+  PartitionCtx partition_context;
```
narendasan, line 203: How urgent is this change? I think we might as well start moving the PartitionGraph object, the IValue maps etc. into the context now and pass this one object around in this patch vs just leaving it for later

bowang95: I don't think this one is so urgent, we could leave it for later.

narendasan: What I mean is we should put in the work now to move everything into the context as long as it isnt blocking you instead of making half a patch

PR #955, opened 2022-04-01
Fix bugs of ReduceToOperation

### LINKED ISSUE #954: 🐛 [Bug] Encountered bug when lowering `aten::to`
##  Bug Description

The lowering pass for `aten::to` could cause unexpected results. 
The error is as follows:
```bash
Successfully load torch model:
def forward(self,
    x: Tensor,
    y: Tensor) -> Tensor:
  basenet = self.basenet
  _0 = torch.to((basenet).forward(x, ), dtype=6, layout=0, device=torch.device("cuda:0"))
  return _0

TRTorch {self.trt_dtype} compile begin >>>
WARNING: [Torch-TensorRT] - Dilation not used in Max pooling converter
  File "test.py", line 55, in <module>
    trt_ts_module = torch_tensorrt.compile(torch_script_module, **compile_settings)
  File "/opt/conda/lib/python3.7/site-packages/torch_tensorrt/_compile.py", line 115, in compile
  File "/opt/conda/lib/python3.7/site-packages/torch_tensorrt/ts/_compiler.py", line 116, in compile
RuntimeError: [Error thrown at core/util/trt_util.cpp:283] Expected type to be true but got false
Unsupported ATen data type Byte
```

## To Reproduce

```cpp
import torch
import math
import numpy as np
from torchvision import models

class Net(torch.nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        """ Base network """
        self.basenet = models.resnet50(pretrained=True)
        self.basenet.fc = torch.nn.Identity()

    def forward(self, x, y, test=True):
        """ Base network """
        x = self.basenet(x)
        return x.to(y.device)

model = Net()
model = model.eval()
model = model.cuda()

input = np.zeros((1, 3, 224, 224)).astype(np.float32)
label = torch.ones(1).int().cuda()
input = torch.from_numpy(input).cuda()
output = model(input, label)

torch_script_module = torch.jit.trace(model, (input, label))
torch.jit.save(torch_script_module, "torch_script_module.ts")
import torch_tensorrt 
torch_tensorrt.logging.set_reportable_log_level(torch_tensorrt.logging.Level.Graph)

import torch
torch_script_module = torch.jit.load("torch_script_module.ts", map_location='cuda').eval()
        
print('Successfully load torch model:')
trt_dtype = torch.float16
            #self.trt_dtype = torch.float32

compile_settings = {
        shape=[1, 3, 224, 224],  # TODO: depends on the model size
        dtype=torch.float32,  # Datatype of input tensor. Allowed options torch.(float|half|int8|int32|bool)
    ),
    torch_tensorrt.Input(
        shape=[1],
        dtype=torch.int32
    )],
    "enabled_precisions": {torch.float32},  # Run with FP16
}

print(torch_script_module.code)
print("TRTorch {self.trt_dtype} compile begin >>>")
trt_ts_module = torch_tensorrt.compile(torch_script_module, **compile_settings)

torch_script_output = torch_script_module(input, label)
trt_ts_output = trt_ts_module(input, label)
diff = abs(torch_script_output - trt_ts_output).mean()
print('Diff between torchscript output and trt_ts_module output is {:.6f}'.format(diff))

```

## Expected behavior

Compilation should go smoothly without error, and numerical difference shall be negligible.

## Environment

 - Torch-TensorRT Version (e.g. 1.0.0): 1.1.0a0+ba9f730f
 - PyTorch Version (e.g. 1.0): 1.10.1
 - CPU Architecture: x86_64
 - OS (e.g., Linux): Linux
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source): pip
 - Python version: 3.7
 - CUDA version: 11.3
 - GPU models and configuration: A100
 - TensorRT version: 8.2.2.1

Njuapp:
# Description
`ReduceToOperation()` contain bugs which will cause compilation failure. After inspecting the lowered graph of `ReduceToOperation`, we found it has mistakenly swapped the place of `device` and `dtype`. Which `device=0` and `dtype=6`, it will be reduced to `dtype=0` instead of `dtype=6`.

Fixes #954

## Type of change

# Checklist:

### DIFF
--- a/core/lowering/passes/reduce_to.cpp
+++ b/core/lowering/passes/reduce_to.cpp
@@ -9,20 +9,20 @@ namespace passes {
 
 void ReduceToOperation(std::shared_ptr<torch::jit::Graph>& graph) {
   std::string to_device_pattern = R"IR(
-        graph(%x, %device, %dtype, %nb, %copy, %format):
-            %out : Tensor = aten::to(%x, %device, %dtype, %nb, %copy, %format)
+        graph(%x, %dtype, %device, %nb, %copy, %format):
+            %out : Tensor = aten::to(%x, %dtype, %device, %nb, %copy, %format)
             return (%out))IR";
   std::string to_dtype_pattern = R"IR(
-        graph(%x, %device, %dtype, %nb, %copy, %format):
+        graph(%x, %dtype, %device, %nb, %copy, %format):
             %out : Tensor = aten::to(%x, %dtype, %nb, %copy, %format)
             return (%out))IR";
   std::string to_dtype_layout_pattern = R"IR(
-        graph(%x, %device, %dtype, %layout, %pm, %nb, %copy, %format):
-            %out : Tensor = aten::to(%x, %device, %dtype, %layout, %pm, %nb, %copy, %format)
+        graph(%x, %dtype, %device, %layout, %pm, %nb, %copy, %format):
+            %out : Tensor = aten::to(%x, %dtype, %device, %layout, %pm, %nb, %copy, %format)
             return (%out))IR";
 
   std::string to_dtype_multi_input_pattern = R"IR(
-        graph(%x, %device, %dtype, %layout, %pm, %nb, %copy, %format):
+        graph(%x, %dtype, %device, %layout, %pm, %nb, %copy, %format):
             %out : Tensor = aten::to(%x, %dtype, %nb, %copy, %format)
             return (%out))IR";
 
--- a/tests/core/lowering/test_reduce_to_pass.cpp
+++ b/tests/core/lowering/test_reduce_to_pass.cpp
@@ -8,11 +8,11 @@
 
 TEST(LoweringPasses, ReduceToCorrectly) {
   std::string source_graph = R"IR(
-    graph(%x, %device, %dtype, %nb, %copy, %format):
-        %out : Tensor = aten::to(%x, %device, %dtype, %nb, %copy, %format)
+    graph(%x, %dtype, %device, %nb, %copy, %format):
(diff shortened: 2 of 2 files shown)


bowang95:
Hi @Njuapp , I don't think simply switching `dtype` and `device` will fix this issue. 
As you can refer here: https://github.com/pytorch/pytorch/blob/72f7193f4d977c8dd4fc0b99bd06a26b81e7d6cd/aten/src/ATen/native/native_functions.yaml#L5694,  there is no signature as `graph(%x, %dtype, %device, %nb, %copy, %format):`. 
Also, for `graph(%x, %dtype, %device, %layout, %pm, %nb, %copy, %format):`, it will induce the device missing issue if we lower it to `Tensor = aten::to(%x, %dtype, %nb, %copy, %format)` as what is explained here. 
I will raise a PR for this soon.

Njuapp:
Thanks @bowang95 for detailed explanations. 
I have checked the references you provided, yes, the first signature `graph(%x, %dtype, %device, %nb, %copy, %format)` I write is non-existent. As for second one, device should not be erased after lowering. This is also what confuses me when looking at this lowering pass. 
I will close this PR, and again thanks for the reference you provided.

Njuapp:
But I also doubt whether lowering the 2nd signature `graph(%x, %dtype, %layout, %device, %pm, %nb, %copy, %format)` to `graph(%x, %device, %dtype, %nb, %copy, %format):` is safe? I erases `%layout` and `%pm`, which may cause error if it is does not come with default values.

bowang95:
Hi @Njuapp , did you get an error after erasing the %layout or %pm? I don't think we will get an error in conversion stage since we are not using it in converters even if we keep it as what it was.

PR #959, opened 2022-04-02
fix: fix the bug in prim::NumToTensor which introduces kLong Tensor

### LINKED ISSUE #956: 🐛 [Bug] LongType introduced in Conversion process
##  Bug Description

 We have a graph like this: 

`
  %14 : int = aten::size(%t.1, %15) 
  %16 : Tensor = prim::NumToTensor(%14) 
  %17 : Tensor[] = prim::ListConstruct(%13, %16)
  %image_size.1 : Tensor = aten::stack(%17, %3) 

`
This fails with an error: 
`
terminate called after throwing an instance of 'c10::bad_optional_access'
  what():  bad optional access
`

This happens because we get an int from aten::size, then in prim::NumtoTensor we are using this function https://github.com/pytorch/pytorch/blob/8a7c9a5e01a24d465126210234aa9d3775b25032/aten/src/ATen/ScalarOps.h#L29, which will introduce a long tensor that incur the error. 

## Expected behavior

Truncate the long type here, should be fixed soon.

bowang007:

# Description
In our evaluation pass the prim::NumToTensor will use the function here https://github.com/pytorch/pytorch/blob/8a7c9a5e01a24d465126210234aa9d3775b25032/aten/src/ATen/ScalarOps.h#L29 to convert a number to a Tensor. However, this will convert int to Tensor kLong, this will incur the conversion phase break down since we don't support kLong types in TRT. 
We can either add a truncate_long_and_double in conversion phase right after evaluation or change the native function from pytorch to fix this issue. 

Fixes #956 

## Type of change

# Checklist:


core/conversion/evaluators/eval_util.cpp
```diff
@@ -118,6 +118,34 @@ void checkSequenceSize(int64_t n, int64_t dim, int64_t seq_size) {
   }
 }
 
+at::Tensor scalar_to_tensor_util(const at::Scalar& s, const at::Device device = at::kCPU) {
```
narendasan, line 121: This needs to be done in a safe manner since we are messing with types. We should at least throw an warning saying we are truncating the type. Ideally we could just optionally turn it on, but we don't have ConversionCtx available

bowang007: maybe we can also use the truncate_long_and_double in compile_settings to turn it on and throw an warning when this is enable, however, this means that we must have access to compile_settings from this function in the prim.cpp.

narendasan: I was considering adding ConversionCtx for evaluators. Maybe we just do that

PR #962, opened 2022-04-05
fix: Use user provided dtype when we can't infer it from the graph

peri044:
Signed-off-by: Dheeraj Peri <peri.dheeraj@gmail.com>

# Description

Use user provided dtype when we can't infer it from the graph

Fixes https://github.com/NVIDIA/Torch-TensorRT/issues/911#issuecomment-1059609092

## Type of change

# Checklist:

### DIFF
--- a/core/compiler.cpp
+++ b/core/compiler.cpp
@@ -328,7 +328,8 @@ void MapInputsAndDetermineDTypes(
         spec.dtype = nvinfer1::DataType::kFLOAT;
       } else if (spec.dtype_is_user_defined && cfg.partition_info.enabled) {
         if (!est_type_opt) {
-          LOG_INFO("Cannot infer input tensor dtype in graph, unable to verify user input dtype settings");
+          LOG_INFO("Cannot infer input tensor dtype in graph. Using user provided input dtype settings");
+          first_use_type_map[in] = {util::TRTDataTypeToScalarType(cfg.convert_info.inputs.find(in)->second.dtype)};
         } else {
           if (util::TRTDataTypeToScalarType(cfg.convert_info.inputs.find(in)->second.dtype) != est_type_opt.value()) {
             std::stringstream ss;
--- a/core/ir/Input.cpp
+++ b/core/ir/Input.cpp
@@ -40,6 +40,13 @@ bool valid_dtype_format_combo(nvinfer1::DataType dtype, nvinfer1::TensorFormat f
         default:
           return false;
       }
+    case nvinfer1::DataType::kBOOL: // Supports Linear (NCHW)
+      switch (format) {
+        case nvinfer1::TensorFormat::kLINEAR:
+          return true;
+        default:
+          return false;
+      }
     default:
       return false;
   }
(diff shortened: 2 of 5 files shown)


py/setup.py
```diff
@@ -239,8 +239,6 @@ def run(self):
         libraries=["torchtrt"],
         include_dirs=[
             dir_path + "torch_tensorrt/csrc", dir_path + "torch_tensorrt/include",
-            dir_path + "/../bazel-TRTorch/external/tensorrt/include",
```
narendasan, line 242: Might as well keep this since if people dont redownload the repo after the namechange its going to break the build

core/compiler.cpp
```diff
@@ -328,7 +328,8 @@ void MapInputsAndDetermineDTypes(
         spec.dtype = nvinfer1::DataType::kFLOAT;
       } else if (spec.dtype_is_user_defined && cfg.partition_info.enabled) {
         if (!est_type_opt) {
-          LOG_INFO("Cannot infer input tensor dtype in graph, unable to verify user input dtype settings");
+          LOG_INFO("Cannot infer input tensor dtype in graph. Using user provided input dtype settings");
+          first_use_type_map[in] = {util::TRTDataTypeToScalarType(cfg.convert_info.inputs.find(in)->second.dtype)};
```
narendasan, line 332: how does this work with this patch? https://github.com/NVIDIA/Torch-TensorRT/pull/902

core/ir/Input.cpp
```diff
@@ -40,6 +40,13 @@ bool valid_dtype_format_combo(nvinfer1::DataType dtype, nvinfer1::TensorFormat f
         default:
           return false;
       }
+    case nvinfer1::DataType::kBOOL: // Supports Linear (NCHW)
+      switch (format) {
```
narendasan, line 44: We probably should be more clear to the user why a BOOL / NHWC combo doesnt work

core/compiler.cpp
peri044: Your patch works in the case where the estimated type is not same as the user provided type. In this case, you override with user provided dtype. 
In this PR. if the user dtype is defined (dtype_is_user_defined=True) and if the est_type_opt is null (which happens for the bitwise OR graph I linked in the PR where inferred type is empty, I use user provided types.

core/ir/Input.cpp
peri044: NHWC is only bound to FP32 according to https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/namespacenvinfer1.html#ac3e115b1a2b1e578e8221ef99d27cd45

py/setup.py
peri044: added it back. But it would be nice to clean this up at some point in time.

core/ir/Input.cpp
narendasan: Yeah, I understand, I mean we need to throw a descriptive error or something

narendasan: Probably put this at the API level

narendasan (approved):
LGTM

PR #1004, opened 2022-04-27
fix(//core/runtime): Support more delimiter variants

### LINKED ISSUE #930: 🐛 [Bug] Compilation causes error: `ValueError: stoi`
##  Bug Description
Compiling the below graph causes the following error:
```
ValueError: stoi
```
It's not clear to me what the root cause is yet; I've seen this appear in several of our models internally, but this is the first time I've been able to isolate the error outside of our repo.

## To Reproduce
```
  import torch
  import torch_tensorrt.logging as logging
  
  logging.set_reportable_log_level(logging.Level.Warning)
  
  torch.manual_seed(0)
  
  DEVICE = torch.device("cuda:0")
  SHAPE = (1, 2)
  
  
  class Model(torch.nn.Module):
      def __init__(self):
          super().__init__()
  
      def forward(self, a):
          y = torch.squeeze(a, -1)
          return y - y
  
  
  if __name__ == "__main__":
      tensor = torch.randn(SHAPE, dtype=torch.float32, device=DEVICE)
  
      model = Model().eval().to(DEVICE)
      out = model(tensor)
      print(f"Model: {out}")
  
      model_trt = torchtrt.compile(
          model,
          inputs=[
              torchtrt.Input(shape=SHAPE),
          ],
          enabled_precisions={torch.float},
      )
      out_trt = model(tensor)
      print(f"Model TRT: {out_trt}")
  
      assert torch.max(torch.abs(out - out_trt)) < 1e-6
```
Outputs the following:
```
root@1ccd9cb0f739:/workspace# python /scripts/squeeze.py
Model: tensor([[0., 0.]], device='cuda:0')
WARNING: [Torch-TensorRT] - Cannot infer input type from calcuations in graph for input a.1. Assuming it is Float32. If not, specify input type explicity
  File "/scripts/squeeze.py", line 31, in <module>
    model_trt = torchtrt.compile(
  File "/usr/local/lib/python3.8/dist-packages/torch_tensorrt/_compile.py", line 115, in compile
  File "/usr/local/lib/python3.8/dist-packages/torch_tensorrt/ts/_compiler.py", line 116, in compile
ValueError: stoi
```

## Expected behavior
Compilation should not error out, and should output the same values as in the non-TRT version.

## Environment
Ubuntu 18.04 x86-64
Built from source:
- Master (`11bcb98d3cd680c3c34e6cc4c4efdc7512c144cc`) using TRT NGC `22.02-py3`
- v1.0 using TRT NGC  `21.11-py3`

Error appears in both.

## Additional context
In this particular case, it looks like we can fix the error by adding a third dimension to the tensor so that it is squeezable, ie.
```
SHAPE = (1, 2, 1)
```

### LINKED ISSUE #982: 🐛 [Bug] Failed to use stoi to get the engine ID in TRTEngine function
##  Bug Description

Got this  error: 
```
INFO: [Torch-TensorRT - Debug Build] - [MemUsageChange] TensorRT-managed allocation in IExecutionContext creation: CPU +0, GPU +132, now: CPU 0, GPU 540 (MiB)
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi
Aborted (core dumped)
```

Debugged and found that it fails at this line: https://github.com/NVIDIA/Torch-TensorRT/blob/c95229144432d96f4bdaa71fb1d242242d42bc29/core/runtime/TRTEngine.cpp#L63 

## To Reproduce

 MaskRCNN model from detectron2. 

## Expected behavior

should figure out why we cannot find a number from this engine.

narendasan:
# Description
Fixes an issue where different classes of delimiters would not properly be handled when deserializing TRT engines. 

Fixes #930
Fixes #982 

## Type of change

# Checklist:

### DIFF
--- a/core/ir/ir.cpp
+++ b/core/ir/ir.cpp
@@ -21,7 +21,7 @@ InputSpecMap pair_input_vals_with_specs(std::vector<const torch::jit::Value*> va
 
   std::unordered_map<const torch::jit::Value*, core::ir::Input> a;
   for (size_t i = 0; i < vals.size(); i++) {
-    LOG_DEBUG("Paring " << i << ": " << vals[i]->debugName() << " : " << specs[i]);
+    LOG_DEBUG("Pairing " << i << ": " << vals[i]->debugName() << " : " << specs[i]);
     a.insert({vals[i], specs[i]});
   }
   return a;
--- a/core/runtime/TRTEngine.cpp
+++ b/core/runtime/TRTEngine.cpp
@@ -63,7 +63,10 @@ TRTEngine::TRTEngine(std::string mod_name, std::string serialized_engine, CudaDe
     auto delim = bind_name.find(".");
     if (delim == std::string::npos) {
       delim = bind_name.find("_");
-      TORCHTRT_CHECK(delim != std::string::npos, "Unable to determine binding index for input " << bind_name << "\nEnsure module was compile with Torch-TensorRT.ts");
+      TORCHTRT_CHECK(
+          delim != std::string::npos,
+          "Unable to determine binding index for input " << bind_name
+                                                         << "\nEnsure module was compile with Torch-TensorRT.ts");
     }
 
     std::string idx_s = bind_name.substr(delim + 1);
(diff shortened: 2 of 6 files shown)


bowang007:
@narendasan looks good, however, I rebased locally it still cannot resolve the issue that I have now. There is no delimiter such as '.' or '_' in the bind_name that I get in the model, maybe I still need to look into what happens in this model.
At least the error message is clearer for me now.

narendasan:
We need inputs to have an index because tensorrt does not guarantee binding order. So you will need to modify whatever is coming in to have indexes

bowang007:
I checked the log where we have this in conversion process:
```
DEBUG: [Torch-TensorRT TorchScript Conversion Context] - Input Dimension Specs: {
    argument_1.1 : Input(shape: [3, 1024, 1024], dtype: Float32, format: NCHW\Contiguous\Linear),}
INFO: [Torch-TensorRT TorchScript Conversion Context] - Adding Input argument_1.1 (named: input_0): Input(shape: [3, 1024, 1024], dtype: Float32, format: NCHW\Contiguous\Linear) in engine (conversion.AddInputs)
```

it seems that we have input_0, then there comes serialization and deserialization, after deserialization the binding name became `0` for input_0. 
How can we modify this? since this whole serialization and deserialization are done through TensorRT. @narendasan

narendasan:
The binding is just 0?

bowang007:
> The binding is just 0?

yes.

narendasan:
hmm there is something odd there i feel. I would expect that the name would be input_0 or something

PR #1022, opened 2022-05-04
feat: Optimize hub.py download

andi4191:
Signed-off-by: Anurag Dixit <a.dixit91@gmail.com>

# Description
Optimizing hub.py to download model repository only if required. The models are downloaded and deserialized only when the model_snapshot file is missing (first time) OR the version recorded is different.

This PR will reduce the CI pipeline jobs turnaround time.

Fixes # (issue)

## Type of change

# Checklist:


.gitignore
```diff
@@ -57,4 +57,4 @@ examples/int8/qat/qat
 examples/int8/training/vgg16/data/*
 examples/int8/datasets/data/*
 env/**/*
-bazel-Torch-TensorRT-Preview
```
andi4191, line 60: @narendasan : I believe we can get rid of this now.

.gitignore
```diff
@@ -57,4 +57,4 @@ examples/int8/qat/qat
...
\ No newline at end of file
+model_snapshot.txt
```
narendasan, line 60: can we call this model_manifest.txt or something?

narendasan: or torch_version.txt

tests/modules/hub.py
```diff
@@ -4,9 +4,30 @@
...
+import os
+import sys
 
 torch.hub._validate_not_a_forked_repo = lambda a, b, c: True
 
+torch_version = torch.__version__
+snapshot_file = 'model_snapshot.txt'
+skip_download = False
+
+# If model repository already setup
```
narendasan, line 16: I feel like what this should be is a list of models which have been downloaded and we check if the model we are downloading is in the list, because we can always add more. And this would require someone to delete the file to redownload

tests/modules/hub.py
```diff
@@ -4,9 +4,30 @@
...
+
+# In case of existing model repository, skip the download
+if skip_download:
+    print('Skipping re-download of model repository')
+    sys.exit()
+else:
+    with open(snapshot_file, 'w') as f:
+        f.write(torch_version)
+
 models = {
```
narendasan, line 31: We should add our hand written modules to this dictionary instead of just having something separate below

.gitignore
andi4191: Changed it to model_manifest.json

tests/modules/hub.py
andi4191: Added the hand written model to the manifest. It has prefix torchtrt_
The model file gets updated only if the version mismatches with torch version installed or if the file is missing.

andi4191:
@narendasan: I refactored the Model download script and added tracking of downloaded files.

tests/modules/hub.py
```diff
@@ -186,34 +156,152 @@ def forward(self, x):
...
-    torchscript=True,
-)
-
-model = BertModel(config)
-model.eval()
-model = BertModel.from_pretrained("bert-base-uncased", torchscript=True)
-
-traced_model = torch.jit.trace(model, [tokens_tensor, segments_tensors])
-torch.jit.save(traced_model, "bert_base_uncased_traced.jit.pt")
+class ModelManifest:
```
narendasan, line 159: Why not just add these to the dictionary?

andi4191: Do we want to trace and script the local custom models on every run regardless of the model file present?
I am referring to model['custom_model']['model'] field

narendasan: You can just pick one or the other

tests/modules/hub.py
andi4191: Updated the code.

tests/modules/hub.py
andi4191: I don't follow. What will be in model['custom_model_x']['model'] field in this case?

narendasan: The path field in the dictionary controls if you get script, trace or both

tests/modules/hub.py
```diff
@@ -4,9 +4,16 @@
...
 
 torch.hub._validate_not_a_forked_repo = lambda a, b, c: True
 
+torch_version = torch.__version__
+
+# Downloads all model files again if manifest file is not present
+MANIFEST_FILE = 'model_manifest.json'
+
 models = {
     "alexnet": {
```
narendasan, line 18: So for example you can add 
```
    "loop_fallback_eval": {
        "model": LoopFallbackEval(),
        "path": "script",
     },
} 
```

andi4191: @narendasan : I added the same. I have segregated the custom model file from hub.py.
I am not sure about where to put the Bert-base-uncased case

Issue #1030, opened 2022-05-07
✨[Feature] Move `torchtrtc` to using new atol + rtol tolerance measure


narendasan:
**Is your feature request related to a problem? Please describe.**

Enhancement

**Describe the solution you'd like**

`torchtrtc` should use the new metric for tolerance

**Describe alternatives you've considered**

N/A

**Additional context**

andi4191:
Closing this feature request. Change is already merged in to the master branch.

PR #1052, opened 2022-05-10
feat (//cpp): Using atol and rtol based tolerance threshold for torchtrtc

andi4191:
# Description

Added rtol + atol tolerance check for torchtrtc. Similar to what we have for test benchmarks.

This PR is a replica of https://github.com/NVIDIA/Torch-TensorRT/pull/1038

Fixes # (issue)
#1030 

## Type of change

# Checklist:

### DIFF
--- a/cpp/bin/torchtrtc/accuracy.cpp
+++ b/cpp/bin/torchtrtc/accuracy.cpp
@@ -19,8 +19,23 @@ bool check_rtol(const at::Tensor& diff, const std::vector<at::Tensor> inputs, fl
   return diff.abs().max().item<float>() <= threshold * maxValue;
 }
 
-bool almost_equal(const at::Tensor& a, const at::Tensor& b, float threshold) {
-  return check_rtol(a - b, {a, b}, threshold);
+bool almost_equal(
+    const at::Tensor& computed_tensor,
+    const at::Tensor& gt_tensor, // gt_tensor : Ground Truth Tensor
+    float atol,
+    float rtol) {
+  auto computed_tensor_float = computed_tensor.toType(at::kFloat);
+  auto gt_tensor_float = gt_tensor.toType(at::kFloat);
+
+  auto diff = computed_tensor_float - gt_tensor_float;
+  auto result = diff.abs().max().item<float>();
+  auto threshold = atol + (rtol * gt_tensor.abs().max().item<float>());
+
+  torchtrt::logging::log(torchtrt::logging::Level::kDEBUG, std::string("Max Difference: ") + std::to_string(result));
+  torchtrt::logging::log(
+      torchtrt::logging::Level::kDEBUG, std::string("Acceptable Threshold: ") + std::to_string(threshold));
+
+  return result <= threshold;
 }
 
 } // namespace accuracy
--- a/cpp/bin/torchtrtc/main.cpp
+++ b/cpp/bin/torchtrtc/main.cpp
@@ -119,11 +119,16 @@ int main(int argc, char** argv) {
       parser, "num_iters", "Number of averaging timing iterations used to select kernels", {"num-avg-timing-iters"});
   args::ValueFlag<uint64_t> workspace_size(
       parser, "workspace_size", "Maximum size of workspace given to TensorRT", {"workspace-size"});
-  args::ValueFlag<double> threshold(
+  args::ValueFlag<double> atol(
       parser,
-      "threshold",
-      "Maximum acceptable numerical deviation from standard torchscript output (default 2e-5)",
-      {'t', "threshold"});
+      "atol",
+      "Absolute tolerance threshold for acceptable numerical deviation from standard torchscript output (default 1e-8)",
+      {"atol"});
+  args::ValueFlag<double> rtol(
+      parser,
+      "rtol",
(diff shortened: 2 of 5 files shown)


cpp/bin/torchtrtc/accuracy.cpp
```diff
@@ -19,9 +19,19 @@ bool check_rtol(const at::Tensor& diff, const std::vector<at::Tensor> inputs, fl
...
 }
 
-bool almost_equal(const at::Tensor& a, const at::Tensor& b, float threshold) {
-  return check_rtol(a - b, {a, b}, threshold);
+bool almost_equal(const at::Tensor& a, const at::Tensor& b, float atol, float rtol) {
+  auto a_float = a.toType(at::kFloat);
+  auto b_float = b.toType(at::kFloat);
+
+  auto diff = a_float - b_float;
+  auto result = diff.abs().max().item<float>();
```
narendasan, line 27: Is this really the calculation?

narendasan: This looks like $$max(a - b) \leq a_{tol} + r_{tol} * \max|b|$$

But I am not sure how this check both rtol and atol

cpp/bin/torchtrtc/main.cpp
```diff
@@ -119,11 +119,16 @@ int main(int argc, char** argv) {
       parser, "num_iters", "Number of averaging timing iterations used to select kernels", {"num-avg-timing-iters"});
   args::ValueFlag<uint64_t> workspace_size(
       parser, "workspace_size", "Maximum size of workspace given to TensorRT", {"workspace-size"});
-  args::ValueFlag<double> threshold(
+  args::ValueFlag<double> atol(
```
narendasan, line 122: This needs to be noted as a breaking change in the commit

cpp/bin/torchtrtc/accuracy.cpp
```diff
@@ -19,9 +19,19 @@ bool check_rtol(const at::Tensor& diff, const std::vector<at::Tensor> inputs, fl
...
+  auto threshold = atol + (rtol * b.abs().max().item<float>());
+
+  std::cout << "Max Difference: " << result << std::endl;
```
narendasan, line 30: Print this out in LOG_DEBUG

cpp/bin/torchtrtc/main.cpp
```diff
@@ -432,12 +441,12 @@ int main(int argc, char** argv) {
...
+                jit_results[i], trt_results[i].reshape_as(jit_results[i]), atol_val, rtol_val)) {
           std::ostringstream threshold_ss;
-          threshold_ss << threshold_val;
+          threshold_ss << "atol: " << atol_val << " rtol: " << rtol_val;
           torchtrt::logging::log(
               torchtrt::logging::Level::kWARNING,
-              std::string("Maximum numerical deviation for output exceeds set threshold (") + threshold_ss.str() +
+              std::string("Maximum numerical deviation for output exceeds tolerance thresholds (") + threshold_ss.str() +
                   std::string(")"));
         }
```
narendasan, line 451: Can we add a case where it meets the threshold and prints out what the difference was and the tolerance to LOG_INFO?

cpp/bin/torchtrtc/accuracy.cpp
andi4191: The check is as per:
https://pytorch.org/docs/stable/generated/torch.allclose.html

It uses the threshold limit based on absolute tolerance limit which is directly added to calibrated reference value ( rtol * other)

Ideally, it should be true for all the values in a tensor but in this case I am checking for the maximum difference value if it is less than (adjusted + calibrated threshold) or not.

andi4191: Using this metric, please note that

almost_equal(a, b) is not the same as almost_equal(b, a)

cpp/bin/torchtrtc/main.cpp
andi4191: Shouldn't we dump the threshold and difference only in case of failure?
It will flood the logs with so many test cases.

cpp/bin/torchtrtc/main.cpp
andi4191: Added as a Breaking change in the commit

cpp/bin/torchtrtc/accuracy.cpp
andi4191: Changed to LOG_DEBUG

cpp/bin/torchtrtc/accuracy.cpp
narendasan: Can we change a and b to computed_tensor and gt_tensor?

cpp/bin/torchtrtc/main.cpp
narendasan: We should log it but users won't see it unless they turn on verbose mode

andi4191: Updated the logging.

cpp/bin/torchtrtc/accuracy.cpp
andi4191: Changed the names for the tensors.

PR #1067, opened 2022-05-14
feat: refactoring segmentation in partitioning

### LINKED ISSUE #1031: ✨[Feature] Refactoring the Graph Segmentation in Partitioning
What we are doing now for graph segmentation in partitioning is going through each node one by one then divide them into 2 parts: TensorRT supported/ TensorRT not supported. This kind of naive segmentation induces many small issues like what we had in these 2 issues: 
#1018 
#1024

Although these bugs could be fixed by patching NonTensorInputs function in partitioning, we have many other side effects:
```
    Target: TensorRT

    Graph: graph(%z.1 : Tensor,
      %y.1 : Tensor):
  %0 : Tensor = aten::lt(%z.1, %y.1) # test_resolve.py:12:10
  %3 : Tensor?[] = prim::ListConstruct(%0)
  %4 : int = prim::dtype(%z.1)
  return ()

```
For example, the graph above is a TensorRT segment, however, 2 nodes out of 3 is useless in this segment because they are producing Non-Tensor outputs which could not be used by later segments since TensorRT engines can't have non-Tensor outputs. As a result, even though we do the calculations for these 2 nodes, they are useless. 

We would want to apply some more advanced graph algorithm to identify the segmentation more ideally. In other words, we should find a better interface to segment the graph. This could lead to less resolve Non-Tensor Inputs related issues and could optimize each divided segments to ensure there are no useless node in these segments. 

**Additional context**
Looking into them now, this is not complex, while we might need to do some testing after the refactoring.

bowang007:

# Description

Fixes #1031 
We are doing a mess in resolveNonTensorInput and resolveTensorListInput, there are many different issues related to these 2 functions. Moreover, keep fixing these issues are introducing more and more complexities in our code base. We hope to solve this related issues in a gentle way and could simplify our code base as well. 

## Type of change

# Checklist:


core/partitioning/partitioning.cpp
```diff
@@ -75,7 +75,8 @@ void find_all_fallback_nodes(std::unordered_map<torch::jit::Node*, int>& fallbac
     q.pop();
     // for every node that produces this fallback node's NonTensor input, they should fallback too
     for (auto input : cur_node->inputs()) {
-      if (!isTensor(input) && fallback_nodes.insert({input->node(), 4}).second) {
+      if (!isTensor(input) && input->node()->kind() != torch::jit::prim::Constant &&
```
mfeliz-cruise, line 78: I think this input->node() traversal misses cases where an op modifies an input.

Ex.
%0 : ListConstruct()
%1 : aten::append(%0, %val)
%2 : aten::append(%1, %val)
%3 : aten::cat(%0)

Looking at just the input->node() of the cat skips over the appends which could result in a change in model behavior.

core/partitioning/partitioning.cpp
```diff
@@ -145,254 +94,26 @@ void getDirtyNodes(
...
-    }
-
-    // For each non-tensor value in the usage_counts map, keep updating the produce_id to the earliest segmented block
-    // that has/produces it.
-    for (auto& use : usage_counts) {
-      // Set the produce_id to the segmented block index that contains/produces this non-tensor torch::jit::Value
-      if (segmented_blocks[i].contain_raw_value(use.first)) {
-        use.second.produce_id = i;
+      if (!inputs_to_resolve.empty()) {
+        std::vector<torch::jit::Node*> dependency_nodes = getDependencyNodes(inputs_to_resolve);
```
mfeliz-cruise, line 110: getDependencyNodes will miss any non-modifying dependency ops as it is currently written.

bowang007: thanks @mfeliz-cruise ! 
Let me check if there is any APIs that we can use in Torchscript to find all the modifying ops as well.

core/partitioning/partitioning.cpp
bowang007: Hi @mfeliz-cruise did you hit this kind of issue in your model?
I built a model locally with this graph:
```
graph(%x.1 : Tensor,
      %y.1 : Tensor):
  %2 : int = prim::Constant[value=0]() 
  %mod_list.1 : Tensor[] = prim::ListConstruct(%x.1)
  %z.1 : Tensor = aten::__getitem__(%mod_list.1, %2) 
  %5 : Tensor[] = aten::append(%mod_list.1, %x.1) 
  %6 : Tensor[] = aten::append(%mod_list.1, %y.1) 
  %7 : Tensor[] = aten::append(%mod_list.1, %z.1) 
  %8 : Tensor = aten::cat(%mod_list.1, %2) 
  return (%8)
```
with forced fallback on aten::cat, the model runs fine. It's true that the input->node() of aten::cat could skip over the aten::append, however, since aten::cat dependency node is prim::ListConstruct, and prim::ListConstruct's output is used later by aten::append, as a result, aten::append will also fallback.

I think there are several cases and these cases could all be covered: 

%0: nodeA
%1: nodeB_modifyinput(%0)
%2: nodeC(%1)

1: nodeC fallbacks, then there are 2 senarios: 
1.1: %1 is Tensor, in this case, nodeA should not fallback, nodeB which modifies input should also not fallback. We cannot get them fallback by traversing. 
1.2: %1 is not Tensor, which could be a TensorList like what happens here. So, in this case, nodeA will fallback. Since nodeA fallback, nodeB which modifies %0 will also fallback finally because nodeB is in nodeA's output list. 

2: nodeB fallbacks, 2 cases too:
2.1 if the input that nodeB is modifying is a Tensor, then nodeA and nodeC will not fallback.
2.2 if the input that nodeB is modifying is not a Tensor, then nodeA will fallback since nodeA is nodeB's input, since nodeA fallbacks, nodeC will also fallback since it's in nodeA's output. I also did a local test with forced fallback on aten::append, which also works fine. 

Did I miss any case here in the analysis?

mfeliz-cruise: I think you're right that these cases should be handled by the forward fallback propagation from LIstConstruct. The cases where I've seen issues are caused by GetDependencyNodes where the ListConstruct node is identified as a dependency and copied into the segment of the cat without the appends, changing the behavior of the model.

mfeliz-cruise: Thanks Bo, could we use the AliasInfo embedded in the node->schema to find these ops on the fly rather than looking it up from a static list? That seems like it could be more robust.

I'm looking at what's done in AliasDB here to get AliasInfo for an input: https://github.com/pytorch/pytorch/blob/d26c575ff581e4df0e9c72b339a25999c6cae59e/torch/csrc/jit/ir/alias_analysis.cpp#L816

And then check if it's a write:
https://github.com/pytorch/pytorch/blob/d26c575ff581e4df0e9c72b339a25999c6cae59e/torch/csrc/jit/ir/alias_analysis.cpp#L847

https://caffe2.ai/doxygen-c/html/classc10_1_1_alias_info.html

mfeliz-cruise: This would also let you check which inputs are modified by an op rather than assuming if a value is used by a modifying op it will be modified.

bowang007: Thanks @mfeliz-cruise ! This is really helpful! 
The method your proposed is much more gentle than the current solution. 
Thanks for sharing this AliasInfo APIs, I'm going to take a look and try to use them to find the modifying nodes.

PR #1092, opened 2022-05-27
added deployment on triton example

tanayvarshney:
# Description

Added a simple example of optimizing with Torch TRT and deploying using Triton. Would recommend adding this to docs

Fixes # (issue)

## Type of change

--examples

# Checklist:


narendasan:
Why not put together an rst file to add to the tutorial section? seems like the code sections are short enough where you can have everything inline.

This seems like a good reference https://pytorch.org/tutorials/intermediate/flask_rest_api_tutorial.html

PR #1109, opened 2022-06-13
Fix parsing error caused by dot-separated variable debugNames

### LINKED ISSUE #1112: 🐛 [Bug] Parse failure in fallback cases
##  Bug Description

When fallback is encountered, the torchscript may contain `.` in the name of parameters, e.g., `x.1` and `input.1`. Such torchscript models, after saved to disk, cannot be loaded successfully because paramter names containing `.` is invalid.

```bash
  File "main.py", line 56, in <module>
    main()
  File "main.py", line 39, in main
    model = torch.jit.load('resnet_withfallback.trt.ts')
  File "/opt/conda/lib/python3.8/site-packages/torch/jit/_serialization.py", line 162, in load
    cpp_module = torch._C.import_ir_module(cu, str(f), map_location, _extra_files)
RuntimeError: expected ) but found 'number' here:
Serialized   File "code/__torch__/torchvision/models/resnet.py", line 7
  __torch___torchvision_models_resnet_ResNet_trt_engine_0x5595a2dc5070 : __torch__.torch.classes.tensorrt.Engine
  def forward(self_1: __torch__.torchvision.models.resnet.ResNet_trt,
    x.1: Tensor) -> Tensor:
     ~~ <--- HERE
    __torch___torchvision_models_resnet_ResNet_trt_engine_0x5595a2dc4ed0 = self_1.__torch___torchvision_models_resnet_ResNet_trt_engine_0x5595a2dc4ed0
    _0 = ops.tensorrt.execute_engine([x.1], __torch___torchvision_models_resnet_ResNet_trt_engine_0x5595a2dc4ed0)
```

## To Reproduce

Run the following python file:

```python
import torch
import torch_tensorrt
import torchvision
import time

torch_tensorrt.logging.set_reportable_log_level(torch_tensorrt.logging.Level.Debug)

warmup_time = 10
test_time = 100
def main():
    torch.manual_seed(2022)
    im_rand = torch.rand((16, 3, 224, 224)).cuda()
    resnet_model = torchvision.models.resnet50()
    resnet_model = resnet_model.eval().cuda()
    torch_script_module = torch.jit.trace(resnet_model, (im_rand))
    # torch.jit.save(torch_script_module.half(), 'resnet18.pt')

    # torch_script_module = torch.jit.load('resnet.pt')
    input = torch.rand((32, 3, 224, 224)).cuda()
    ts_output = torch_script_module(input)
    print('result:', ts_output)

    compile_settings = {
        # "inputs": [input],
            # min_shape=[1, 3, 224, 224],  # TODO: depends on the model size
            # opt_shape=[32, 3, 224, 224],
            # max_shape=[32, 3, 224, 224],
            shape=[1, 3, 224, 224],
            dtype=torch.half,  # Datatype of input tensor. Allowed options torch.(float|half|int8|int32|bool)
        )],
        'torch_executed_ops': ['aten::adaptive_avg_pool2d'],
        "enabled_precisions": {torch.half}
    }

    print("Compile: FP16")
    model = torch_tensorrt.compile(torch_script_module.half(), **compile_settings)
    torch.jit.save(model, 'resnet_withfallback.trt.ts')
    model = torch.jit.load('resnet_withfallback.trt.ts')

    for i in range(warmup_time):
        trt_output = model(input.half())

    print("Compile: FP16")
    t1 = time.time()
    for i in range(test_time):
        trt_output = model(input.half())
    t2 = time.time()
    print('trt_output: ', trt_output)

    diff = abs(ts_output - trt_output)
    print('diff: {}'.format(diff.mean()))


if __name__ == '__main__':
    main()

```

## Expected behavior

The compilation should end without failure.

## Environment

 - Torch-TensorRT Version (e.g. 1.1.0):
 - PyTorch Version (e.g. 1.11):
 - CPU Architecture: x86_64
 - OS (e.g., Linux): Linux
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source): pip
 - Build command you used (if compiling from source):  bazel
 - Python version: 3.8
 - CUDA version: 11.3

Njuapp:
# Description

Sometimes when fallback is triggered, the debugName of parameter names could contain `.`, e.g., `input.1` and `input_ids.1`. Such torchscript files, after conversion and saving to disk, cannot be parsed successfully. We propose to replace such `.` with `_`, which can be a solution.

Fixes #1112 

## Type of change

# Checklist:

### DIFF
--- a/core/compiler.cpp
+++ b/core/compiler.cpp
@@ -447,6 +447,15 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg) 
         auto engine = conversion::ConvertBlockToEngine(g->block(), cfg.convert_info, static_params);
         AddEngineToGraph(new_mod, new_g, engine, cuda_device);
       }
+      for (auto input : new_g->block()->inputs()) {
+        LOG_DEBUG("input name:" << input->debugName());
+        if (input->debugName().find(".") != std::string::npos) {
+          auto pos = input->debugName().find(".");
+          auto newName = input->debugName().replace(pos, 1, "_");
+          input->setDebugName(newName);
+          LOG_DEBUG("changed name:" << input->debugName());
+        }
+      }
       auto new_method = new_mod._ivalue()->compilation_unit()->create_function(method.name(), new_g);
       auto schema = util::GenerateGraphSchema(new_method->name(), new_g);
       new_mod.type()->addMethod(new_method);


core/compiler.cpp
```diff
@@ -447,6 +447,15 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg)
         auto engine = conversion::ConvertBlockToEngine(g->block(), cfg.convert_info, static_params);
         AddEngineToGraph(new_mod, new_g, engine, cuda_device);
       }
+      for (auto input : new_g->block()->inputs()) {
+        LOG_DEBUG("input name:" << input->debugName());
+        if (input->debugName().find(".") != std::string::npos) {
+          auto pos = input->debugName().find(".");
+          auto newName = input->debugName().replace(pos, 1, "_");
```
narendasan, line 454: This seems hacky, @bowang007 thoughts? We might just be doing something wrong upstream causing these issues.

bowang007: what happens is that when there is any fallback we will copy the Metadata of the inputs of the lowered graph (include debugname()) to the inputs of the new graph through this [line](https://github.com/pytorch/TensorRT/blob/a3432e250e7fe35618607ea95d632a1833d225c4/core/compiler.cpp#L237) As a result, the inputs' names' are same with the Lowered Graph inputs, which has some values as x.1, these names cannot be deserialized by torch.jit.load(). This explains why we are not getting this error when there is no fallback, since all the inputs are substituted by input_0, input_1 when there is no fallback. 
Another solution to this error would be simply deleting the mapping [here](https://github.com/pytorch/TensorRT/blob/a3432e250e7fe35618607ea95d632a1833d225c4/core/compiler.cpp#L237). I did some tests locally which could help resolve this issue. However, not sure if this deletion would induce other issues, will run the test cases to ensure it works.

bowang007:
raised another PR to fix it in upstream code. #1148

PR #1122, opened 2022-06-16
Add trt_executed_modules & default_torch_execution interface

Njuapp:
Signed-off-by: Cheng Hang <calvinhance@gmail.com>

# Description

According to the current design, every module is by default compiled to TensorRT execpt those modules included in torch_executed_modules.

However, in some cases, for example, a multi-modal model where ResNet and BERT are present, BERT is optimized by custom op implementation ([FasterTransformer](https://github.com/NVIDIA/FasterTransformer)). Developers may only want to compile ResNet, and leave everything in BERT (embedding, custom op, FC) run in torch. Although it is fine to explicitly assign values for torch_executed_modules, it will be much more convenient to have an interface like trt_executed_modules, and a switch-mode interface like default_torch_execution.

When the switch is turned on, every module would by default run in Torch, and only modules explicitly included in trt_executed_modules would be compiled.

Fixes # ()

## Type of change

# Checklist:

### DIFF
--- a/core/lowering/lowering.cpp
+++ b/core/lowering/lowering.cpp
@@ -39,7 +39,7 @@ void LowerGraph(std::shared_ptr<torch::jit::Graph>& g, LowerInfo lower_info) {
   }
   torch::jit::EliminateDeadCode(g);
   if (lower_info.forced_fallback_modules.size() > 0) {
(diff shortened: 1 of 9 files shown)


narendasan:
Notes left on the design in the RFC #1119. This PR should be WIP until the design is finalized

PR #1141, opened 2022-06-22
[FX] refactor the fx path in compile function

frank-wei:
# Description

Separate the FX lowering process to a single function `compile` to align with existing ts compile.
 
## Type of change

# Checklist:


py/torch_tensorrt/fx/compile.py
```diff
@@ -0,0 +1,98 @@
...
+    Keyword Arguments:
+        inputs (List[torch.Tensor]): for fixed shape scenario, inputs shapes can not change
+        enabled_precision (torch.dtype): The datatype that TensorRT can use when selecting kernels. If torch.float is chosen, the kernel is running with fp32; If torch.float16 is chosen, the kernel is running with fp16 or fp32 which selected by TensorRT
+        ir (str): The requested strategy to compile. (default is ts - TorchScript with scripting path, fx is FX based path)
+        **kwargs: Additional settings for the specific requested strategy (See submodules for more info)
+
+    Returns:
+        torch.nn.Module: Compiled Module, when run it will execute via TensorRT
+    """
+    acc_model = acc_tracer.trace(module, inputs)
```
yinghai, line 35: Where do we pass the lowering settings?

frank-wei: En, it is not used here.  I am thinking we probably do not need the compile function but use `lower_to_trt()`? 
But I'd like add an extra parameter minimum_segment which controls whether to use TRT if the number of segments(after splitting) are larger than a threshold? For ex, if number of segments is larger than 8, performance is relatively bad and the whole graph will fall to eager model. The default value will be None which means does not check. And the user could set it if they want. I'd like to hear your opinions.

yinghai: In general there is a lot of goodie in lowering settings which can make the lowering quite flexible. But yeah, this is advanced usage. Maybe leave a placeholder for lowering settings so that we can use later.

frank-wei: Will revise the PR.

py/torch_tensorrt/fx/lower.py
```diff
@@ -100,12 +102,12 @@ def __call__(self, mod, input, split_name) -> TRTInterpreterResult:
                         self.lower_setting.max_batch_size,
                         self.lower_setting.max_batch_size,
                     ),
+                    self.lower_setting.opt_profile_replica,
                 )
-                if self.lower_setting.explicit_batch_dimension
+                if self.lower_setting.explicit_batch_dimension and self.lower_setting.dynamic_batch
```
frank-wei, line 107: dynamic_batch is added to differentiate two cases: with or w/o dynamic shape on batch dim (dim=0). cc @wushirong. I keep the dynamic_batch=True as default value so it will not change the previous behavior in production. Please have a review.

py/torch_tensorrt/fx/example/lower_example.py
```diff
@@ -198,6 +198,6 @@ def run_configuration_benchmark(
 
 
 if __name__ == "__main__":
-    test_model = torchvision.models.resnet101()
-    input = [torch.cuda.FloatTensor(1024, 3, 224, 224)]  # type: ignore[attr-defined]
-    benchmark(test_model, input, 100, 1024)
+    test_model = torchvision.models.resnet18(pretrained=True)
```
frank-wei, line 201: Reduce the workload to get better user experience. No long wait time.

wushirong (commented):
Add a description for the new field 'dynamic_batch', and I am a bit unsure whether we should set it default to true.

py/torch_tensorrt/fx/lower.py
yinghai: What's the use case of w/o dyanmic shape on batch dim with explicit_batch_dimension=True? What's the different in terms of behavior in TensorRT? Basically, if I have explicit_batch_dimension=True while all my input dims are positive, how does TRT interprets it? 

Maybe a question to @narendasan too.

frank-wei: Here is my understanding, firstly, explicit_batch_dimension=True will become the default in next year and there is no explicit_batch_dimension=False(implicit) mode.
Secondly, 
> if I have explicit_batch_dimension=True while all my input dims are positive, how does TRT interprets it?

TRT will treat it as fixed shape. And that is what I tested for all the torchdynamo [benchmarks](https://github.com/pytorch/torchdynamo/blob/5fb502660e52a2e1f93ab0f148fd8776e1b56297/torchdynamo/optimizations/backends.py#L380)

frank-wei:
> Add a description for the new field 'dynamic_batch', and I am a bit unsure whether we should set it default to true.

addressed.

yinghai (approved):

PR #1153, opened 2022-07-01
Add .circleci/config.yml

narendasan:
# Description

Adding the rest of Torch-TensorRT to CI

Fixes # (issue)

## Type of change

# Checklist:


narendasan:
@frank-wei Two thoughts for future work: 
1. Do you want to have a separate workflow for `--fx-only` at least temporarily? We cant switch to nightly until we finish the 1.2.0 release so this might allow you to continue without tests failing
2. Can you take a look at rigging your fx tests in `nox` similar to our current TS API tests? I would like to run all "L0" tests together so if there is an `l0_fx_tests` session that would be a lot easier. It would also let devs test faster (`nox -s l0_fx_tests`)

Issue #1121, opened 2022-06-15
✨[Feature] Do not remove nodes which are in-place operations (classified in  TRT segments) in fallback


peri044:
**Is your feature request related to a problem? Please describe.**
Currently for eg: if aten::append is in prim::If block, it gets removed because it is classified as TRT segment. No output is registered for it because aten::append is an inplace op. Make modifications to registerSegmentsOutputs
```
Segmented Graph: graph(%self_1 : __torch__.ModifyingOpError_trt,
      %x.1 : Tensor,
      %y.1 : Tensor):
  %mod_list.1 : Tensor[] = prim::ListConstruct(%x.1)
  %__torch___ModifyingOpError_trt_engine_0x557361771710 : __torch__.torch.classes.tensorrt.Engine = prim::GetAttr[name="__torch___ModifyingOpError_trt_engine_0x557361771710"](%self_1)
  %5 : Tensor[] = prim::ListConstruct(%x.1, %y.1)
  %6 : Tensor[] = tensorrt::execute_engine(%5, %__torch___ModifyingOpError_trt_engine_0x557361771710)
  %7 : Tensor = prim::ListUnpack(%6)
  %8 : bool = aten::Bool(%7) # mod.py:12:11
   = prim::If(%8)
    block0():
      -> ()
    block1():
      -> ()
  %9 : int = prim::Constant[value=0]()
  %10 : Tensor = aten::cat(%mod_list.1, %9) # mod.py:14:12
  return (%10)
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

peri044:
Also add this as a test case for this PR https://github.com/pytorch/TensorRT/pull/1067
```
class ModifyingOpError(nn.Module):
    def __init__(self):
        super(ModifyingOpError, self).__init__()
    def forward(self, x, y):
        mod_list = [x]
        if x.sum() > y.sum():
            mod_list.append(y)
        z = torch.cat(mod_list)
        return z
```

ncomly-nvidia:
Resolved by #1140. Closing

bowang007:
@ncomly-nvidia I don't this this one should be closed. 
It refers to whether we should delete these several lines https://github.com/pytorch/TensorRT/blob/0738caacf8d0911521ff73b0f67b071b8c2b7b42/core/partitioning/partitioning.cpp#L202

narendasan:
@gs-olive / @bowang007 How does this relate to index_put_?

[...]

PR #1174, opened 2022-07-12
refactor: remove the code that covers the no output segmented block

bowang007:

# Description

Before this feature https://github.com/pytorch/TensorRT/issues/1031 is supported, there might be some cases that TensorRT segmented block has NonTensor input, so we go through the nodes' output one by one to determine the output for this block if there is no output for this block. 
However, this this PR merged https://github.com/pytorch/TensorRT/pull/1140,  we can ensure that every TensorRT segmented block has Tensor input, which means there will never be no output for segmented block. 

Therefore, we can erase these lines. 

support #1121 

# Checklist:


narendasan (changes requested):
Seems like `test_resolve_nontensor_inputs` is failing, perhaps needs a rebase?

Issue #1173, opened 2022-07-11
✨[Feature] Erase the NonTensor input/output because of min_block_size != 1


bowang007:
When we implement this feature https://github.com/pytorch/TensorRT/issues/1031, we raised a PR here: https://github.com/pytorch/TensorRT/pull/1140

In this PR, we used DFS for re-segmentation. However, as we can see from here: 
 
```
bool check_node_fallback(torch::jit::Node* n, const std::unordered_map<torch::jit::Node*, int>& fallback_nodes) {
  if (fallback_nodes.count(n)) {
    if (fallback_nodes.at(n) == 0) {
      LOG_GRAPH("Node not supported by conversion: " << util::node_info(n));
    } else if (fallback_nodes.at(n) == 1) {
      LOG_GRAPH("Node explicitly set to run in torch: " << util::node_info(n));
    } else if (fallback_nodes.at(n) == 2) {
      LOG_GRAPH("Node is within a module set to run in torch: " << util::node_info(n));
    } else {
      LOG_GRAPH(
          "Node fallback to Torch because the NonTensor dependencies with other fallback nodes: "
          << util::node_info(n));
    }
    return false;
  }
  LOG_GRAPH("Node is going to run in TensorRT: " << util::node_info(n));
  return true;
}
```

In this function, it doesn't cover cases that nodes fallback because of min_block_size. Ff we do DFS first then segment graph according to min_block_size, then there would be NonTensor inputs and outputs again. 

So, what we are going to do here is peudo-segment graph first with min_block_size then use DFS to determine the fallback nodes. 
There are at lease 2 passes for segmentation. Many factors need to be considered here for implementation.

bowang007:
closing per #1195

PR #1195, opened 2022-07-21
feat: support min_block_size != 1 caused fallback nodes re-segmentation

bowang007:

Support the feature request: #1173 

This helps fix some of the error in #922

## Type of change

# Checklist:


core/partitioning/partitioning.cpp
```diff
@@ -111,7 +115,7 @@ void find_all_fallback_nodes(std::unordered_map<torch::jit::Node*, int>& fallbac
     // for every node that produces this fallback node's NonTensor input, they should fallback too
     for (auto input : cur_node->inputs()) {
       if (!isTensor(input) && input->node()->kind() != torch::jit::prim::Constant &&
-          fallback_nodes.insert({input->node(), 4}).second) {
+          global_fallback_nodes.insert({input->node(), 4}).second) {
```
mfeliz-cruise, line 118: nit: Consider an enum rather than using raw values for fallback reason.

bowang007: good idea! you mean we should use enum rather than raw values like 4 here right? thanks!

peri044 (approved):
LGTM

PR #1228, opened 2022-08-04
feat: keep model's original input signature

bowang007:

# Description

There are some requests that users doesn't want to change the models' inputs signature for input, which could make it difficult to use in a deployment environment. We can keep original inputs signature. 

## Type of change

# Checklist:

### DIFF
--- a/core/compiler.cpp
+++ b/core/compiler.cpp
@@ -28,6 +28,16 @@
 namespace torch_tensorrt {
 namespace core {
 
+void AssociateInputSignature(torch::jit::Graph* g, torch::jit::Graph* new_g) {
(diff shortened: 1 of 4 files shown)


narendasan:
Does this solution not change the input names anyway?

bowang007:
> Does this solution not change the input names anyway?

yes, that's what we expect.

narendasan:
@bowang007 how does this effect runtime? does this change our convention for providing inputs? I don't think it should but just to make sure

bowang007:
> @bowang007 how does this effect runtime? does this change our convention for providing inputs? I don't think it should but just to make sure

I don't think so.

PR #1247, opened 2022-08-10
refactor: Changed the hardcoded values to macros for DLA memory sizes

andi4191:
Signed-off-by: Anurag Dixit <a.dixit91@gmail.com>

# Description
Refactored the hard coded values to macro for DLA specific memory sizes

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/conversionctx/ConversionCtx.h
+++ b/core/conversion/conversionctx/ConversionCtx.h
@@ -11,6 +11,10 @@
 #include <cuda_runtime.h>
 #include "core/util/prelude.h"
 
+#define DLA_SRAM_SIZE 1048576
+#define DLA_LOCAL_DRAM_SIZE 1073741824
+#define DLA_GLOBAL_DRAM_SIZE 536870912
+
 namespace torch_tensorrt {
 namespace core {
 namespace conversion {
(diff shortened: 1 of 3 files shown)


core/conversion/conversionctx/ConversionCtx.h
```diff
@@ -11,6 +11,10 @@
 #include <cuda_runtime.h>
 #include "core/util/prelude.h"
 
+#define DLA_SRAM_SIZE 1048576
```
narendasan, line 14: Maybe we should put these constants somewhere in utils if they will get used in more than one place

andi4191: I think these values would be used only in the conversion context for building the TRT engine. Hence, I kept it encapsulated in ConversonCtx. Do you think it will be used in converters, evaluators or runtime too?

narendasan: These are user definable values right? Presumably compile spec would use this

andi4191: Yeah, moved them to core/utils

narendasan (approved):
LGTM

Issue #949, opened 2022-03-31
✨[Feature] Print out Python file name, line number, and/or op for warnings, compilation failures


chaoz-dev:
**Is your feature request related to a problem? Please describe.**
By default, when running `Torch-TensorRT` from Python, most errors encountered during compilation will print a stack trace originating from either the Python compilation command (eg. printing something like `torch_tensorrt_pypi/torch_tensorrt/ts/_compiler.py", line 119, in compile  compiled_cpp_mod = _C.compile_graph(module._c, _parse_compile_spec(spec)). RuntimeError: ....`) or the CPP file backing the implementation. 

Warnings during runtime may be printed without _any_ identifying information.

The current stack traces and output by themselves do not provide enough information to diagnose or pinpoint which PyTorch op is causing the offending behavior. This is especially egregious when attempting to apply Torch-TensorRT to large and complex models, where failures with the conversion process are currently difficult to identify without unwinding and stepping through the model op by op to locate the root cause of conversion failures. 

**Describe the solution you'd like**
Ideally, when an error or warning occurs, we should print output identifying the offending op (at minimum), as well as the file and line number at which the op occurs, in addition to other stack traces, errors, and warnings. 

**Describe alternatives you've considered**
Currently, I've tried log printing at the graph level to better identify errors during the conversion process. However, there are issues with this approach:
1. We may encounter the error before printing the next node or subgraph where the error occurs;
2. The output appears correct, and does not identify the offending op, anyway

**Additional context**
I _think_ this ask is potentially complicated but somewhat possible... On the one hand, Torch-TensorRT converts using the JIT output, which means obtaining metadata could be more complex than just reading the model directly. On the other hand, looking at the [Interpreting Graphs](https://pytorch.org/docs/stable/jit.html#interpreting-graphs) section of the documentation, it appears file and line number information _should_ be accessible somewhere...

narendasan:
[...]

PR #1263, opened 2022-08-14
Centralizing Partitioning State

narendasan:
# Description

It is incredibly hard to track the state of the partitioning process because the pipeline relies on a number of shared lookup tables that are modified in place during the course of partitioning. 

`PartitioningCtx`, similar to `ConversionCtx` centralizes key information like decisions about node executors, user settings, the in progress graph etc. so that this information is uniformly managed and queryable at any point during the pipelines execution. This means it is hard to client code to alter state in unpredictable ways and it also reduces the amount of arguments that need to be passed to each phase. 

Addresses #949 

## Type of change

# Checklist:


narendasan:
@bowang007 This should be roughly equivalent to our current system. There are a couple improvements on this that I'd like to do that I want your input on. 

1. I renamed a bunch of stuff to hopefully help make it clearer to new devs, let me know if you disagree about anything. 
2. I want every node in the `global_fallback_map` (now called `node_executor_decision_map`) and make an explicit decision on if it will run in torch or tensorrt (this can be changed like if a block would run in trt but is not large enough). That way in debugging or otherwise I can just ask about a particular node and get back what it will run on. Right now I am having trouble doing this because there is an assumption in the system that if the node is not in the map it must run in tensorrt. 
3. There is still work to be done on being able to easily dump state at any given time. 
4. Are there other data sources that we might want to have managed by the context?

core/partitioning/partitioningctx/PartitioningCtx.cpp
```diff
@@ -0,0 +1,123 @@
...
+    return false;
+  } else {
+    return true;
+  }
+}
+
+std::ostream& operator<<(std::ostream& os, const NodeExecutorDecision& format) {
+  switch (format) {
+    case NodeExecutorDecision::kUNSUPPORTED:
+      return os << "to run torch due to lack of converter support";
```
narendasan, line 95: "run in torch"

core/compiler.cpp
```diff
@@ -272,8 +269,8 @@ GraphAndMapping ConstructFallbackGraph(
         // convert the 2 blocks in prim::if and get the converted graph with mappings
         std::vector<GraphAndMapping> graph_and_mappings;
         for (auto cur_block : if_node->blocks()) {
-          graph_and_mappings.push_back(
-              ConstructFallbackGraph(new_mod, cur_block, example_tensor_map, cfg, static_params, fallback_nodes));
+          graph_and_mappings.push_back(ConstructFallbackGraph_(
```
narendasan, line 272: Not wildly pressing but is there a way to do all the partitioning beforehand then go through and compile specific blocks? Having them mixed is not as easy to debug

core/compiler.cpp
```diff
@@ -219,19 +219,16 @@ void AddIfBlockToGraph(
   return;
 }
 
-GraphAndMapping ConstructFallbackGraph(
+GraphAndMapping ConstructFallbackGraph_(
```
narendasan, line 222: Have a more distinguishing name for this maybe?

core/compiler.cpp
narendasan: @bowang007 thoughts here?

narendasan: Perhaps what we do is recursively partition, then recursively compile the final graph. Not sure if graph stitching can handle this right now.

bowang007:
> @bowang007 This should be roughly equivalent to our current system. There are a couple improvements on this that I'd like to do that I want your input on.
> 
> 1. I renamed a bunch of stuff to hopefully help make it clearer to new devs, let me know if you disagree about anything.
> 2. I want every node in the `global_fallback_map` (now called `node_executor_decision_map`) and make an explicit decision on if it will run in torch or tensorrt (this can be changed like if a block would run in trt but is not large enough). That way in debugging or otherwise I can just ask about a particular node and get back what it will run on. Right now I am having trouble doing this because there is an assumption in the system that if the node is not in the map it must run in tensorrt.
> 3. There is still work to be done on being able to easily dump state at any given time.
> 4. Are there other data sources that we might want to have managed by the context?

Went through just now, I think the general architecture is very clear and comprehensive, here are some thoughts:
2. We can have a map for every nodes in the graph, logic might be a little bit different but shouldn't be to complicated. 
3. more details?

narendasan:
> > @bowang007 This should be roughly equivalent to our current system. There are a couple improvements on this that I'd like to do that I want your input on.
> > 
> > 1. I renamed a bunch of stuff to hopefully help make it clearer to new devs, let me know if you disagree about anything.
> > 2. I want every node in the `global_fallback_map` (now called `node_executor_decision_map`) and make an explicit decision on if it will run in torch or tensorrt (this can be changed like if a block would run in trt but is not large enough). That way in debugging or otherwise I can just ask about a particular node and get back what it will run on. Right now I am having trouble doing this because there is an assumption in the system that if the node is not in the map it must run in tensorrt.
> > 3. There is still work to be done on being able to easily dump state at any given time.
> > 4. Are there other data sources that we might want to have managed by the context?
> 
> Went through just now, I think the general architecture is very clear and comprehensive, here are some thoughts: 2. We can have a map for every nodes in the graph, logic might be a little bit different but shouldn't be to complicated. 3. more details on this?

I want to be able to enable features like https://github.com/pytorch/TensorRT/discussions/1257
which would be able to show a user info like inputs to a block, node assignments, shape calculation results etc.

bowang007:
@narendasan If we store SegmentedBlocks https://github.com/pytorch/TensorRT/blob/1a4084b39a4725cf4818c2ebcab455cd1793031e/core/partitioning/partitioningctx/PartitioningCtx.h#L51  in PartitioningCtx, what about the sub-blocks' segmentedBlocks? Looks like it could get messed up if there is some node like If::node which contains sub-blocks.

narendasan:
> @narendasan If we store SegmentedBlocks
> 
> https://github.com/pytorch/TensorRT/blob/1a4084b39a4725cf4818c2ebcab455cd1793031e/core/partitioning/partitioningctx/PartitioningCtx.h#L51
> in PartitioningCtx, what about the sub-blocks' segmentedBlocks? Looks like it could get messed up if there is some node like If::node which contains sub-blocks.

Where do we put sub blocks now?

bowang007:
> > @narendasan If we store SegmentedBlocks
> > https://github.com/pytorch/TensorRT/blob/1a4084b39a4725cf4818c2ebcab45 5cd1793031e/core/partitioning/partitioningctx/PartitioningCtx.h#L51
> > 
> > in PartitioningCtx, what about the sub-blocks' segmentedBlocks? Looks like it could get messed up if there is some node like If::node which contains sub-blocks.
> 
> Where do we put sub blocks now?

We are doing recursion for sub-blocks, it that stack ends then the sub-block will be stitched and put back to the if nodes.

bowang007:
> @bowang007 This should be roughly equivalent to our current system. There are a couple improvements on this that I'd like to do that I want your input on.
> 
> 1. I renamed a bunch of stuff to hopefully help make it clearer to new devs, let me know if you disagree about anything.
> 2. I want every node in the `global_fallback_map` (now called `node_executor_decision_map`) and make an explicit decision on if it will run in torch or tensorrt (this can be changed like if a block would run in trt but is not large enough). That way in debugging or otherwise I can just ask about a particular node and get back what it will run on. Right now I am having trouble doing this because there is an assumption in the system that if the node is not in the map it must run in tensorrt.
> 3. There is still work to be done on being able to easily dump state at any given time.
> 4. Are there other data sources that we might want to have managed by the context?

For 4: I'm now trying to peel off the stitching part out of compiler.cpp so the LUT for mapping between old value and new value might be stored in the context as well.

PR #1298, opened 2022-08-22
feat: rewriting param to a Constant if it's a introduced input

### LINKED ISSUE #1190: ✨[Feature] Support cases where LowerGraph() introduces input whose values are stored in Param
#1024
There are some models that LowerGraph() might introduce new input to the entire graph after lowering, which may cause the missing value error. As what is happening here https://github.com/pytorch/TensorRT/issues/922#issuecomment-1184833720 and here: https://github.com/pytorch/TensorRT/issues/922#issuecomment-1157063427. 

For now, in most cases it fails at the shape analysis phase since Torch-TensorRT cannot find the value for the missing input. We could introduce a pass to modify the graph and substitute the input by some constant or we can just add some mechanism to use the values from Params directly.

In the future Torch-TensorRT could handle these cases.

bowang007:

# Description

In some cases, some parameters are graphs' input. We can rewrite them into constants to make sure it works. 

Fixes #1190 

## Type of change

# Checklist:


narendasan:
@bowang007 seems like linting is failing, can you set up the pre-commit system?

narendasan (commented):
Is there a test case we can write for this?

core/compiler.cpp
```diff
@@ -434,6 +451,9 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg)
           (!(cfg.lower_info.forced_fallback_modules.size() == 0 &&
              cfg.partition_info.forced_fallback_operators.size() == 0 && isBlockConvertible) ||
            outputIsCollection)) {
+        if (!static_params.empty()) {
+          RewriteInputsWithParams(g, params);
```
narendasan, line 455: Should this be a lowering pass?

bowang007: It could be a lowering pass. However, I'm not so sure about what happens with conversion stage if the input are stored in params. Any thought on this? @narendasan

bowang007: cool let me change it to a lowering pass.

bowang007:
This PR should be good to merge. @narendasan @peri044

PR #1308, opened 2022-08-25
[feat] add support for aten::reciprocal(int)

mfeliz-cruise:
# Description

The unary layer does not support integer inputs to RECIP. Pytorch implicitly casts integer inputs to float for aten::reciprocal so we can add the same cast here to add support.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/unary.cpp
+++ b/core/conversion/converters/impl/unary.cpp
@@ -49,6 +49,21 @@ auto abs_registration TORCHTRT_UNUSED = RegisterNodeConversionPatterns().pattern
        }
      }});
 
+auto reciprocal_registration TORCHTRT_UNUSED = RegisterNodeConversionPatterns().pattern(
+    {"aten::reciprocal(Tensor self) -> Tensor", [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
+       auto in = args[0].ITensorOrFreeze(ctx);
+       if (in->getType() == nvinfer1::DataType::kINT32) {
+         // pytorch implicitly casts to float for aten::reciprocal(int)
+         in = castITensor(ctx, in, nvinfer1::DataType::kFLOAT);
+       }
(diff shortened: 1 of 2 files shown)


core/conversion/converters/impl/unary.cpp
```diff
@@ -49,6 +49,21 @@ auto abs_registration TORCHTRT_UNUSED = RegisterNodeConversionPatterns().pattern
        }
      }});
 
+auto reciprocal_registration TORCHTRT_UNUSED = RegisterNodeConversionPatterns().pattern(
+    {"aten::reciprocal(Tensor self) -> Tensor", [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
+       auto in = args[0].ITensorOrFreeze(ctx);
+       if (in->getType() == nvinfer1::DataType::kINT32) {
```
narendasan, line 55: This might be a large change but would it make sense to just add this to the macro for other unary ops? @peri044 thoughts on what the repercussions would be?

mfeliz-cruise: In this case the behavior matches pytorch. For other ops (ex. abs implemented element-wise above) this behavior would be incorrect.

peri044: There are two ops which have some restrictions according to the doc but other unary ops must have floating point inputs. https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_network_definition.html#a77831224c9a72ad02587a56ded93c672
```
Generally the input must have a floating-point type (or kINT8 as a quantized float), except for the following operations:
kSIGN accepts a floating-point or Int32 tensor.
kNOT requires a Bool tensor.
```

peri044: We could add the above restrictions in the code in addition to what Michael added, to cover the cases in the doc completely.

narendasan: I was mostly talking from the pytorch perspective since seems like theres at least a few ops where int inputs are valid

PR #1358, opened 2022-09-14
Remove Bazel related file from Python wheel packaging

### LINKED ISSUE #1262: 🐛 [Bug] Can't install torch_tensorrt python wheel with Bazel and rules_python
##  Bug Description

When installed with `rules_python` the python wheel`torch_tensorrt` is not made available in the runfiles folder and thefore an import error is raised: 
```
ModuleNotFoundError: No module named 'torch_tensorrt'
```

## To Reproduce

Simple repo https://github.com/gdippolito/reproduce-rules-python-bug/tree/main

## Expected behavior

The package is made available to Bazel targets.

## Environment

 - Torch-TensorRT Version (e.g. 1.0.0): v1.1.0
 - PyTorch Version (e.g. 1.0): 1.12.1
 - CPU Architecture: x86_64
 - OS (e.g., Linux): Linux
 - How you installed PyTorch: pip (via `rules_python`)
 - Are you using local sources or building from archives: Using python wheel releases
 - Python version: 3.8.12
 - CUDA version: 1.11.0
 - GPU models and configuration: T4

## Additional context

Not sure if the issue is with `rules_python` or TorchRT I have created an equivalent ticket in the rules_python repository https://github.com/bazelbuild/rules_python/issues/788

gdippolito:
# Description

I'm not sure if the issue is with `rules_python` or Bazel but it is not possible to install a Python wheels with Bazel and rules_python if the wheels contains Bazel related files (e.g WORKSPACE, BUILD) on the root folder. 

This patch will remove the two files from the Python wheel packaging (reducing it size by a very marginal amount 😄 ) 

I have tested that the python wheel can be built successfully and installed in a Bazel environment. 

Fixes #1262 

## Type of change

# Checklist:

### DIFF
--- a/py/setup.py
+++ b/py/setup.py
@@ -428,8 +428,6 @@ setup(
             "include/torch_tensorrt/core/util/*.h",
             "include/torch_tensorrt/core/util/logging/*.h",
             "bin/*",
-            "BUILD",
-            "WORKSPACE",
         ],
     },
     exclude_package_data={


narendasan:
We ship these files so that users who install the wheel file can pull in the headers and .so files as a `local_repository` for C++ projects. For example this is in our WORKSPACE. https://github.com/pytorch/TensorRT/blob/8498287e70dc1e83efe68a6c5ce02953d8f65a45/WORKSPACE#L34
Which allows us to do testing against pre-compiled versions of torch_tensorrt. For us I guess we can just make this a `new_local_repository` and just reference the BUILD file in tree but seems like something that may be useful to others (ala the cmake dir in pytorch's package).

narendasan:
Was thinking about why this might be a problem and it could be rules_python is generating a BUILD file and placing it in the unpacked package dir and this causes a conflict. iirc there's a build filename hierarchy so perhaps we can use a different lower priority filename for the build file and this would solve the issue

janickm:
> it could be rules_python is generating a BUILD file and placing it in the unpacked package dir and this causes a conflict

yes, this is actually what's happening at wheel extraction / bazel repo generation time via `rules_python` - I agree with the OP that these files should not live out of the box inside the wheel's root like this, but these target rules should rather be hooked up after extraction via the usual `additive_build_content` ([docs](https://rules-python.readthedocs.io/en/stable/api/rules_python/python/pip.html#package_annotation.additive_build_content))

as an ugly workaround one can patch these files away at wheel extraction time using `pip.override()` ([link](https://rules-python.readthedocs.io/en/stable/api/rules_python/python/extensions/pip.html#pip.override))

narendasan:
These files were removed a while back, @ph03 are you still hitting issues here?

janickm:
Oh, since which version? We're hitting this on at least 2.7.0

narendasan:
https://github.com/pytorch/TensorRT/pull/3641

Might have just missed 2.8 Cant remember exactly

PR #1359, opened 2022-09-14
Add converter support for aten::maximum/minimum

mfeliz-cruise:
Add converter support for aten::maximum/minimum. Refactor element_wise ops to reduce repeated code.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/element_wise.cpp
+++ b/core/conversion/converters/impl/element_wise.cpp
@@ -25,6 +25,22 @@ nvinfer1::ITensor* clamp_util(
   return clamp_layer_out;
 }
 
+bool two_tensor_input_single_elment_wise_layer(
+    ConversionCtx* ctx,
+    const torch::jit::Node* n,
+    args& args,
+    nvinfer1::ElementWiseOperation op) {
+  auto self = args[0].ITensorOrFreeze(ctx);
+  auto other = args[1].ITensorOrFreeze(ctx);
+  auto element_wise = add_elementwise(ctx, op, self, other, util::node_info(n));
+  TORCHTRT_CHECK(element_wise, "Unable to create element_wise layer from node: " << *n);
+
+  element_wise->setName(util::node_info(n).c_str());
+  auto out = ctx->AssociateValueAndTensor(n->outputs()[0], element_wise->getOutput(0));
+  LOG_DEBUG("Output tensor shape: " << out->getDimensions());
+  return true;
+}
+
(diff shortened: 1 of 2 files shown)


core/conversion/converters/impl/element_wise.cpp
```diff
@@ -25,6 +25,22 @@ nvinfer1::ITensor* clamp_util(
   return clamp_layer_out;
 }
 
+bool two_tensor_input_single_elment_wise_layer(
```
narendasan, line 28: Feel like I'd rather this be a macro similar to unary that just takes the torchscript schema and the enum.

core/conversion/converters/impl/element_wise.cpp
```diff
@@ -25,6 +25,22 @@ nvinfer1::ITensor* clamp_util(
...
+    ConversionCtx* ctx,
+    const torch::jit::Node* n,
+    args& args,
+    nvinfer1::ElementWiseOperation op) {
+  auto self = args[0].ITensorOrFreeze(ctx);
+  auto other = args[1].ITensorOrFreeze(ctx);
+  auto element_wise = add_elementwise(ctx, op, self, other, util::node_info(n));
+  TORCHTRT_CHECK(element_wise, "Unable to create element_wise layer from node: " << *n);
```
narendasan, line 36: This reduces the descriptiveness of our error messages. I would like to know which converter is failing, not some converter that does elementwisen ops.

mfeliz-cruise: Doesn't the node in the message give you more information than the layer name if you're looking for the converter that failed? There are still 5 "mul" layer error messages in this file in different converters.

This is easy enough to add back if we switch to a macro, but I want to understand the requirement.

narendasan: Well we still dump the node info in either TORCHTRT_CHECK, but node info does not dump the schema iirc. The issue is that there can be many say `aten::div` variants and the node info would not tell you per say which one. I've found is that when debugging the high order bit at least for devs is which converter has an issue so explicitly stating the failing converter helps us determine a root cause faster.

peri044: It does dump the complete info.. Example node_info output: %2 : Tensor = aten::minimum(%0, %1) 
We could use `util::node_info(n)` instead of `*n` just to be consistent although both are same.

peri044 (approved):

PR #1360, opened 2022-09-16
Add a "no_conversion" flow to torch-tensorrt

mfeliz-cruise:
Adds a "no_conversion" option to torch-tensorrt which when enabled will replace the standard conversion and engine insertion with an embedded function call for each convertible segment. This allows inspection of the partition without running conversion and the possibility to convert each engine individually in subsequent runs.

Future work:
Allow this flow to run without a GPU to enable TRT convertibility/partitioning linting flows on host machines
Partition without running shape propagation when in the no-convert flow

Fixes # (#1361)

### DIFF
--- a/core/compiler.cpp
+++ b/core/compiler.cpp
@@ -168,6 +168,37 @@ void AddSegmentedBlockToGraph(
   return;
 }
 
+void AddSegmentedBlockToGraphAsFunction(
+    torch::jit::script::Module& new_mod,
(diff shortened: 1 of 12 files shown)


narendasan:
@bowang007 Could something like this solve our graph stitching issues? Say if the way partitioning works is it creates a bunch of methods but a method is either 100% PyTorch or 100% TRT. May make things like collections way easier too.

Issue #1233, opened 2022-08-05
❓ [Question] How to install "tensorrt" package? 


domef:
## ❓ Question

I'm trying to install `torch-tensorrt` on a Jetson AGX Xavier. I first installed `pytorch` 1.12.0 and `torchvision` 0.13.0 following this [guide](https://forums.developer.nvidia.com/t/pytorch-for-jetson-version-1-11-now-available/72048). Then I installed `torch-tensorrt` following this [guide](https://pytorch.org/TensorRT/tutorials/installation.html#installation), and the compilation completed succesfully.

When I try to import `torch_tensorrt` it throws an error, saying it can't find a module named `tensorrt`. Where I can find this package?

## Environment

I'm using a Jetson AGX Xavier with Jetpack 5.0.1.

narendasan:
Are you able to import the tensorrt package explicitly on its own?

domef:
I don't have `tensorrt` in my environment, in fact my question is: how can I install `tensorrt`?

narendasan:
The tensorrt python package should come prepacked with Jetpack. Are you using the system python install or something like conda?

domef:
I created a new environment with python3.8 (I installed `torch-tensorrt` in the virtual env).

narendasan:
Can you try importing tensorrt with the system python?

domef:
I will try, but how can I install `tensorrt` on another virtual environment?

narendasan:
I am not sure, TensorRT gets install via Jetpack (a debian package internally). You could try to extract the wheel to install. Alternatively you can check if it is distributed for aarch64-jetson on `nvidia-pyindex`

domef:
I tried to import `tensorrt` with the system python and it is imported succesfully.
To install `tensorrt` in my environment i tried this:

```
pip install nvidia-pyindex
pip install nvidia-tensorrt
```

but I get this error:

```
Looking in indexes: https://pypi.org/simple, https://pypi.ngc.nvidia.com
Collecting nvidia-tensorrt
  Downloading nvidia-tensorrt-0.0.1.dev5.tar.gz (7.9 kB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [17 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
[...]

Issue #1214, opened 2022-07-29
🐛 [Bug] Unable to import Torch-TensorRT


valeriosofi:
##  Bug Description
Hello, I was trying to install and import Torch-TensorRT in a colab notebook, but after following all the steps when I import torch_tensorrt I get the following error: 
`ImportError: libnvinfer_plugin.so.8: cannot open shared object file: No such file or directory`
In some other issues was suggested to import tensorrt before importing torch_tensorrt, but even in this case I get an error:
`ImportError: /usr/local/lib/python3.7/dist-packages/torch_tensorrt/_C.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZNSt18basic_stringstreamIcSt11char_traitsIcESaIcEEC1Ev`
How can I solve this? I also tried on an aws machine and got the same issue.
I can share the notebook where the issue can be easily reproduced: https://colab.research.google.com/drive/1tHRq8Sbe-kE-UoQ4mNlI_IEyDAOoW2sK?usp=sharing

## To Reproduce

1. Start a colab environment with gpu
2. Install pytorch 1.11: pip install torch==1.11.0+cu113 torchvision==0.12.0+cu113 torchaudio==0.11.0 --extra-index-url https://download.pytorch.org/whl/cu113
3. Install tensorrt: pip install --upgrade nvidia-tensorrt
4. Install pytorch-tensorrt: pip install torch-tensorrt -f https://github.com/pytorch/TensorRT/releases

## Environment

 - Torch-TensorRT Version: 1.1.0
 - PyTorch Version: 1.11
 - How you installed PyTorch: pip
 - Python version: 3.7
 - CUDA version: 11.2
 - GPU models and configuration: Tesla P100

RolandGao:
I have the same problem

narendasan:
This issue is solved in Torch-TensorRT 1.2. 

Adding the following to install Torch-TRT from our CI should give you a build to try out.

I did notice that CoLab is still on CUDA 11.2 so not sure if you will run into issues re: CUDA incompatibility

valeriosofi:
Thank you @narendasan, I tried with that wheel and it works even on colab! Do you have wheels like that also for python 3.8 and 3.9?

narendasan:
Yes you can pull from here: https://app.circleci.com/pipelines/github/pytorch/TensorRT/464/workflows/a07892ac-38f6-4ac9-9d48-69bbc0ce3acb/jobs/1396/artifacts

[...]

PR #1400, opened 2022-10-13
Add correct pip install instructions

msaroufim:
# Description

I've seen a few people (including myself get bit by not having clear instructions to pip install the library)

https://github.com/pytorch/TensorRT/issues/1233
https://github.com/pytorch/TensorRT/issues/1214

I can't update the release notes to fix the actual issue but regardless I believe installation instructions should be on the front page

If you like I can also update the dependencies here https://github.com/pytorch/TensorRT/blob/master/pyproject.toml so installation instructions will be a 1 liner

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,12 @@ These are the following dependencies used to verify the testcases. Torch-TensorR
 
 Releases: https://github.com/pytorch/TensorRT/releases
 
+```
+pip install nvidia-pyindex
+pip install nvidia-tensorrt==8.4.3.1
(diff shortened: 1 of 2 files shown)


narendasan:
> If you like I can also update the dependencies here https://github.com/pytorch/TensorRT/blob/master/pyproject.toml so installation instructions will be a 1 liner

Sure that would be appreciated.

msaroufim:
@narendasan does this look good to you?

narendasan (approved):

Issue #1346, opened 2022-09-12
✨[Feature] Support auto data type transformation between int64 <-> int32


bowang007:
**Is your feature request related to a problem? Please describe.**
In this graph:

```
INFO: [Torch-TensorRT - Debug Build] - Partitioned Graph: [Segment Block @0:
    Target: TensorRT

    Graph: graph(%index.1 : Tensor,
      %data.1 : Tensor):
  %2 : int = prim::Constant[value=4]() # test_int64.py:28:0
  %3 : bool = prim::Constant[value=0]() # test_int64.py:28:0
  %4 : NoneType = prim::Constant()
  %index : Tensor = aten::to(%index.1, %2, %3, %3, %4) # test_int64.py:28:0
  %data.3 : Tensor = aten::mul(%data.1, %data.1) # test_int64.py:29:0
  return (%index, %data.3)

Segment Block @1:
    Target: Torch

    Graph: graph(%data.3 : Tensor,
      %index : Tensor):
  %2 : int = prim::Constant[value=1]() # test_int64.py:30:0
  %0 : Tensor = aten::scatter(%data.3, %2, %index, %2) # test_int64.py:30:0
  return (%0)
```

`%index` is converted to int32, but in block 1, `scatter` function needs data type int64 but got int32. 
This is because TensorRT doesn't support int64, so Torch-TensorRT will cast all int64=>int32 to run them in TensorRT. However, when partitioning is enabled, some functions in Torch they still need type int64 to run. 

**Describe the solution you'd like**
This could be supported if every aten::to operation is recorded and then cast the types between torch and tensorrt.

ncomly-nvidia:
@inocsin  for viz.

inocsin:
We should also record which value has been truncated in converson process

peri044:
@bowang007  Is this support already in the master ? https://github.com/pytorch/TensorRT/blob/master/core/partitioning/shape_analysis.cpp#L231-L259

Christina-Young-NVIDIA:
This one is indeed already supported in the master.  @bowang007 to confirm and close.

bowang007:
supported in #1407 , closing

PR #1407, opened 2022-10-16
feat: support int64 <=> int32 auto conversion

### LINKED ISSUE #1382: ✨[Feature] Automatic conversion for int32<->int64 in fallback
As we know, if there are some operators that torch-tensorrt doesn't support, the model will be partitioned into tensorrt and torch subgraphs. TensorRT doesn't support int64 value and will truncate int64 to int32.

In some cases, the operators in the torch subgraph consume int64 value(like aten::index), and this value is produced from tensorrt subgraph(truncated into int32), this will cause an error. We need to track the data type conversion and automatic convert the data type back to the origianl type between torch and tensorrt.

Here is a typical case
```python
import torch
import torch.nn as nn
import torch_tensorrt


class Net(nn.Module):
  def __init__(self):
    super(Net, self).__init__()

  def forward(self, data, index):
    src = 1
    index = index.to(torch.int64)
    data = data * data
    data = data.scatter_(1,index,src)
    data = data + 1
    return data

data = torch.randn([5,5])
index = torch.randint(0,4,[2,2], dtype = torch.int32)

compile_spec = {
    "inputs": None,
    "device": {
        "device_type": torch_tensorrt.DeviceType.GPU,
        "gpu_id": 0,
        "allow_gpu_fallback": False,
        "disable_tf32": False
    },
    "torch_executed_ops": ["aten::scatter_", "aten::scatter"],
    "min_block_size": 1
}

net = Net()
model = torch.jit.trace(net, (data, index))

torch_type = torch.float32
min_shape = [5,5]
data2 = torch_tensorrt.Input(shape=min_shape, dtype=torch_type)
torch_type = torch.int32
index2 = torch_tensorrt.Input(shape=min_shape, dtype=torch_type)

inputs = [data2, index2]

compile_spec["inputs"] = inputs

    trt_mod = torch_tensorrt.ts.compile(model, **compile_spec)
inputs = [data.cuda(), index.cuda()]
output = trt_mod(*inputs)
print(output)
```
subgraph log
```
INFO: [Torch-TensorRT - Debug Build] - Partitioned Graph: [Segment Block @0:
    Target: TensorRT

    Graph: graph(%index.1 : Tensor,
      %data.1 : Tensor):
  %2 : int = prim::Constant[value=4]() # test_int64.py:28:0
  %3 : bool = prim::Constant[value=0]() # test_int64.py:28:0
  %4 : NoneType = prim::Constant()
  %index : Tensor = aten::to(%index.1, %2, %3, %3, %4) # test_int64.py:28:0
  %data.3 : Tensor = aten::mul(%data.1, %data.1) # test_int64.py:29:0
  return (%index, %data.3)

Segment Block @1:
    Target: Torch

    Graph: graph(%data.3 : Tensor,
      %index : Tensor):
  %2 : int = prim::Constant[value=1]() # test_int64.py:30:0
  %0 : Tensor = aten::scatter(%data.3, %2, %index, %2) # test_int64.py:30:0
  return (%0)

Segment Block @2:
    Target: TensorRT

    Graph: graph(%1 : Tensor):
  %2 : Tensor = prim::Constant[value={1}]() # test_int64.py:31:0
  %3 : int = prim::Constant[value=1]() # test_int64.py:30:0
  %0 : Tensor = aten::add(%1, %2, %3) # test_int64.py:31:0
  return (%0)

]
```

bowang007:

# Description
Support int64 <=> int32 type conversion.

Fixes #1382 

## Type of change

# Checklist:

### DIFF
--- a/core/partitioning/shape_analysis.cpp
+++ b/core/partitioning/shape_analysis.cpp
@@ -76,6 +76,42 @@ torch::jit::Node* getUpstreamCastNode(torch::jit::Value* val) {
       }
     }
   }
+  return nullptr;
+}
+
+torch::jit::Node* createCastNode(SegmentedBlock& seg_block, size_t index, bool is_input) {
+  torch::jit::Node* cast_node = getUpstreamCastNode(seg_block.raw_inputs()[index]);
+  auto g = seg_block.g();
+  // if we can find upstream aten::to node, we use it's parameters for creating new cast node
+  if (cast_node) {
+    std::unordered_map<torch::jit::Value*, torch::jit::Value*> value_map;
+    value_map.insert({cast_node->inputs()[0], g->inputs()[index]});
+    if (!is_input) {
+      // if this value is output, we need to cast it to int32
+      auto const_val = g->insertConstant(3);
+      value_map.insert({cast_node->inputs()[1], const_val});
+    }
+    auto env = [&](torch::jit::Value* v) { return util::getOrAddInputForValue(v, g, value_map); };
+    cast_node = g->createClone(cast_node, env);
+    //    auto cast_node = g->prependNode(g->createClone(cast_node, env));
+  } else {
+    // if there is no explicit cast aten::to operation, we need to create a node
+    auto const_type = is_input ? g->insertConstant(4) : g->insertConstant(3);
+    auto const_zero = g->insertConstant(0);
+    const_zero->setType(torch::jit::BoolType::get());
+    auto none_val = g->insertNode(g->createNone())->output();
+    cast_node = g->create(torch::jit::aten::to, {g->inputs()[index], const_type, const_zero, const_zero, none_val});
+
+    //    auto cast_node = g->prependNode(g->create(torch::jit::aten::to, {g->inputs()[i], const_type, const_zero,
+    //    const_zero, none_val})); seg_block.inputs()[i]->replaceAllUsesAfterNodeWith(cast_node,
+    //    cast_node->outputs()[0]); LOG_DEBUG(seg_block << " in shape analysis");
+  }
+  if (is_input) {
+    g->prependNode(cast_node);
+  } else {
+    g->appendNode(cast_node);
+  }
+  return cast_node;
 }
 
 void getSegmentsOutputByRunning(
@@ -164,42 +200,29 @@ void getSegmentsOutputByRunning(
   }
 
   // auto int64 <=> int32 conversion
-  if (seg_block.target() == SegmentedBlock::kTorch) {
-    // Firstly, check if there is Int64 input
+  if (seg_block.target() == SegmentedBlock::kTorch && partitioning_info.truncate_long_and_double) {
+    // First, check if there is Int64 input
     for (size_t i = 0; i < seg_block.inputs().size(); ++i) {
       if (ivalues_maps[seg_block.raw_inputs()[i]].isTensor()) {
         auto cur_ivalue = ivalues_maps[seg_block.raw_inputs()[i]];
         at::ScalarType t = cur_ivalue.toTensor().scalar_type();
         if (t == at::kLong) {
           // we add a cast operation to cast the type to Int64
-          auto inner_g = seg_block.g();
-          torch::jit::Node* cast_node = getUpstreamCastNode(seg_block.raw_inputs()[i]);
-          std::unordered_map<torch::jit::Value*, torch::jit::Value*> value_map;
-          value_map.insert({cast_node->inputs()[0], seg_block.inputs()[i]});
-          auto env = [&](torch::jit::Value* v) { return util::getOrAddInputForValue(v, inner_g, value_map); };
-          auto new_cast_node = inner_g->prependNode(inner_g->createClone(cast_node, env));
-          seg_block.inputs()[i]->replaceAllUsesAfterNodeWith(new_cast_node, new_cast_node->outputs()[0]);
+          auto cast_node = createCastNode(seg_block, i, true);
+          seg_block.inputs()[i]->replaceAllUsesAfterNodeWith(cast_node, cast_node->outputs()[0]);
         }
       }
     }
-    // TODO: This part might be necessary for some model, now checkint to verify
-    //    for (size_t i = 0; i < seg_block.outputs().size(); ++i) {
-    //      if (ivalues_maps[seg_block.raw_outputs()[i]].isTensor()) {
-    //        auto cur_ivalue = ivalues_maps[seg_block.raw_outputs()[i]];
-    //        at::ScalarType t = cur_ivalue.toTensor().scalar_type();
-    //        if (t == at::kLong) {
-    //          auto inner_g = seg_block.g();
-    //          torch::jit::Node* cast_node = getUpstreamCastNode(seg_block.raw_outputs()[i]);
-    //          std::unordered_map<torch::jit::Value*, torch::jit::Value*> value_map;
-    //          value_map.insert({cast_node->inputs()[0], seg_block.outputs()[i]});
-    //          auto const_val = inner_g->insertConstant(3);
-    //          value_map.insert({cast_node->inputs()[1], const_val});
-    //          auto env = [&](torch::jit::Value* v) { return util::getOrAddInputForValue(v, inner_g, value_map); };
-    //          auto new_cast_node = inner_g->appendNode(inner_g->createClone(cast_node, env));
-    //
-    //        }
-    //      }
-    //    }
+    for (size_t i = 0; i < seg_block.outputs().size(); ++i) {
+      if (ivalues_maps[seg_block.raw_outputs()[i]].isTensor()) {
+        auto cur_ivalue = ivalues_maps[seg_block.raw_outputs()[i]];
+        at::ScalarType t = cur_ivalue.toTensor().scalar_type();
+        if (t == at::kLong) {
+          auto cast_node = createCastNode(seg_block, i, false);
+          seg_block.g()->block()->replaceOutput(i, cast_node->outputs()[0]);
+        }
+      }
+    }
   }
 
   // set input shape for each segmented block so we wil use it in conversion process
--- a/core/partitioning/shape_analysis.cpp
+++ b/core/partitioning/shape_analysis.cpp
@@ -66,7 +66,9 @@ torch::jit::Node* getUpstreamCastNode(torch::jit::Value* val) {
     auto cur_val = q.front();
     q.pop();
     auto node = cur_val->node();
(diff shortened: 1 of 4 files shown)


core/partitioning/shape_analysis.cpp
```diff
@@ -142,16 +163,56 @@ void getSegmentsOutputByRunning(
...
+          torch::jit::Node* cast_node = getUpstreamCastNode(seg_block.raw_inputs()[i]);
+          std::unordered_map<torch::jit::Value*, torch::jit::Value*> value_map;
+          value_map.insert({cast_node->inputs()[0], seg_block.inputs()[i]});
+          auto env = [&](torch::jit::Value* v) { return util::getOrAddInputForValue(v, inner_g, value_map); };
+          auto new_cast_node = inner_g->prependNode(inner_g->createClone(cast_node, env));
+          seg_block.inputs()[i]->replaceAllUsesAfterNodeWith(new_cast_node, new_cast_node->outputs()[0]);
+        }
+      }
+    }
+    // TODO: This part might be necessary for some model, now checkint to verify
```
narendasan, line 185: @bowang007 should this be uncommented?

narendasan:
@inocsin Can you verify these changes on key models?

narendasan:
Seems like tests are failing for partitioning?

inocsin:
> @inocsin Can you verify these changes on key models?

Sure, we have asked users to test this pr with their models

core/partitioning/shape_analysis.cpp
bowang007: yes, optimized and refactored, this part is now included.

gs-olive:
When compiling BART (https://huggingface.co/facebook/bart-base) using Torch TensorRT, this PR currently segfaults on my machine. Will add an additional comment with the line which is causing the issue

core/partitioning/shape_analysis.cpp
```diff
@@ -57,6 +58,62 @@ std::unordered_map<const torch::jit::Value*, torch::jit::IValue> generateRandomI
...
 }
 
+torch::jit::Node* getUpstreamCastNode(torch::jit::Value* val) {
+  std::queue<torch::jit::Value*> q;
+  q.push(val);
+  std::unordered_set<torch::jit::Node*> visited;
+  while (!q.empty()) {
+    auto cur_val = q.front();
+    q.pop();
+    auto node = cur_val->node();
```
gs-olive, line 68: For BART (https://huggingface.co/facebook/bart-base), the segfault occurs here, when the `cur_val` pointer is accessed. When `getSegmentsOutputByRunning` calls `createCastNode` on line 221, the index passed in is a `seg_blocks.output()` index, but on line 83, `createCastNode` uses that index for `seg_block.raw_inputs()`. For the block in question, the number of inputs and outputs differs, so this may contribute to the segfault.

core/partitioning/shape_analysis.cpp
```diff
@@ -57,6 +58,62 @@ std::unordered_map<const torch::jit::Value*, torch::jit::IValue> generateRandomI
...
+      for (auto input : node->inputs()) {
+        q.push(input);
+      }
+    }
+  }
+  return nullptr;
+}
+
+torch::jit::Node* createCastNode(SegmentedBlock& seg_block, size_t index, bool is_input) {
+  torch::jit::Node* cast_node = getUpstreamCastNode(seg_block.raw_inputs()[index]);
```
gs-olive, line 83: Add an if/else here to use `raw_inputs()` if `is_input = true` otherwise use `raw_outputs()`

core/partitioning/shape_analysis.cpp
```diff
@@ -57,6 +58,62 @@ std::unordered_map<const torch::jit::Value*, torch::jit::IValue> generateRandomI
...
+  auto g = seg_block.g();
+  // if we can find upstream aten::to node, we use it's parameters for creating new cast node
+  if (cast_node) {
+    std::unordered_map<torch::jit::Value*, torch::jit::Value*> value_map;
+    value_map.insert({cast_node->inputs()[0], g->inputs()[index]});
+    if (!is_input) {
+      // if this value is output, we need to cast it to int32
+      auto const_val = g->insertConstant(3);
+      value_map.insert({cast_node->inputs()[1], const_val});
```
gs-olive, line 92: Throws an error when the upstream `aten::to` node does not have `dtype` as its second argument. For example, the schema `aten::to.prim_Device(Tensor(a) self, Device? device, int? dtype=None, bool non_blocking=False, bool copy=False) -> Tensor(b|a)` has `Device` as its second value, and this insertion causes it to be transformed to an invalid schema. We need to differentiate between schemas to ensure the `dtype` is placed in the right position. It seems that valid schemas for `aten::to` have `dtype` as either the second or third argument, or not at all. I believe there should be a check should be in `getUpstreamCastNode` to see if dtype is any of the arguments, and then a second check here to see if it is second or third argument in the schema.

The check here could be something like an if/else checking the debugName at the second index, as in:
```cpp
if (cast_node->inputs()[1]->node()->output()->type()->kind() == torch::jit::TypeKind::DeviceObjType) {
  value_map.insert({cast_node->inputs()[2], const_val});
} else {
  value_map.insert({cast_node->inputs()[1], const_val});
}
```

core/partitioning/shape_analysis.cpp
```diff
@@ -57,6 +58,62 @@ std::unordered_map<const torch::jit::Value*, torch::jit::IValue> generateRandomI
...
 
+torch::jit::Node* getUpstreamCastNode(torch::jit::Value* val) {
+  std::queue<torch::jit::Value*> q;
+  q.push(val);
+  std::unordered_set<torch::jit::Node*> visited;
+  while (!q.empty()) {
+    auto cur_val = q.front();
+    q.pop();
+    auto node = cur_val->node();
+    if (node->kind().toQualString() == std::string("aten::to")) {
```
gs-olive, line 69: May need an additional check to ensure that the `aten::to` schema is valid for `dtype` insertion, as some of these schemas do not take an integer `dtype` at all, for example:

- `aten::to(Tensor(a) self, bool non_blocking=False, bool copy=False) -> Tensor(b|a)` 
- `aten::to(Tensor(a) self, Device device, ScalarType dtype, bool non_blocking=False, bool copy=False, MemoryFormat? memory_format=None) -> Tensor(a)`
- `aten::to(Tensor(a) self, Tensor other, bool non_blocking=False, bool copy=False, MemoryFormat? memory_format=None) -> Tensor(a)`

A check could be something like an additional `&&` with

```cpp
(node->inputs()[1]->node()->output()->type()->kind() == torch::jit::TypeKind::IntType) ||
(node->inputs()[2]->node()->output()->type()->kind() == torch::jit::TypeKind::IntType)
```

core/partitioning/shape_analysis.cpp
gs-olive, line 68: For BART (https://huggingface.co/facebook/bart-base), the segfault occurs here, on line 68, when the `cur_val` pointer is accessed. When `getSegmentsOutputByRunning` calls `createCastNode` on line 221, the index passed in is a `seg_blocks.output()` index, but on line 83, `createCastNode` uses that index for `seg_block.raw_inputs()`. For the block in question, the number of inputs and outputs differs, so this may contribute to the segfault.

gs-olive (changes requested):
With the suggested edits, PR is functioning for BART model and successfully casts Long tensors to Int tensors. Suggestions are related to bugs arising from `input` vs `output` checking and usage of different `aten::to` schemas

ncomly-nvidia:
Fixes #1346

core/partitioning/shape_analysis.cpp
bowang007: Hey @gs-olive Any reproducer for this? 
What I'm not sure about is that for getUpstreamNode() function when we pass in a int32 value will the first `cast` node be the cast node that casts this value to int64? If that's the case, then we don't need this check. 
In other words, is it possible that the first `cast` node involving the passed value is to cast some other value? If the first `cast` node is not the cast node that casts to int64, will the second `cast` node be what we want?

gs-olive: Hi @bowang007 - yes, this did throw an error on my end, when compiling the [BART](https://huggingface.co/docs/transformers/model_doc/bart) model with the changes made in [this PR](https://github.com/pytorch/TensorRT/pull/1416), to fix an unrelated issue. The linked PR uses the second `aten::to` listed above, which does not take an integer dtype, so inserting the integer dtype there throws an error. Specifically, these lines in the PR cause the issue: https://github.com/pytorch/TensorRT/blob/9e7822aa842ee50316f1cac5d9d841ed04353c15/core/lowering/passes/device_casting.cpp#L28-L29

My thought was that we do need this check you have, but maybe it should be more stringent - something like:

```cpp
    if ((node->kind().toQualString() == std::string("aten::to")) &&
        ((node->inputs()[1]->node()->output()->type()->kind() == torch::jit::TypeKind::IntType) ||
        (node->inputs()[2]->node()->output()->type()->kind() == torch::jit::TypeKind::IntType))) {
```

core/partitioning/shape_analysis.cpp
```diff
@@ -57,6 +58,59 @@ std::unordered_map<const torch::jit::Value*, torch::jit::IValue> generateRandomI
...
+        q.push(input);
+      }
+    }
+  }
+  return nullptr;
+}
+
+torch::jit::Node* createCastNode(SegmentedBlock& seg_block, size_t index, bool is_input) {
+  auto cast_raw_value = is_input ? seg_block.raw_inputs()[index] : seg_block.raw_outputs()[index];
+  auto cast_subgraph_value = is_input ? seg_block.outputs()[index] : seg_block.outputs()[index];
```
gs-olive, line 84: Please update this line to:

```cpp
auto cast_subgraph_value = is_input ? seg_block.inputs()[index] : seg_block.outputs()[index];
```

Currently, it is using the outputs regardless of the truth value of `is_input`. With this change, the PR (used along with PR #1416 is working for compilation + inference with the BART model)

core/partitioning/shape_analysis.cpp
gs-olive: @bowang007 Please let me know what you think about the above

gs-olive (changes requested):
One additional bugfix requested and one minor optional comment, and then the PR is successful when used for compilation + inference on BART.

core/partitioning/shape_analysis.cpp
bowang007: @gs-olive I got your point now, let me update this part.

gs-olive (approved):

gs-olive:
Looks good! Fully functional now on BART!

PR #1416, opened 2022-10-25
fix: Device casting issues with certain `aten` operators

gs-olive:
# Description

- Investigated issue arising with BART-base model (https://huggingface.co/facebook/bart-base) where certain tensor inputs to TensorRT were on the cpu, despite users explicitly casting all inputs properly
- Traced issue to internally-generated 0D tensors, mask tensors, and operations returning CPU tensors passed between Torch and Torch-TensorRT engines
- Added lowering passes to ensure function edge cases are appropriately dealt with, and added validation check in runtime to avoid models crashing at runtime due to device mismatches

## Type of change

  - Fixed device location errors relating to specific `aten` functions and casting
  - Added runtime check to ensure all tensor inputs to TorchTRT are on GPU

# Checklist:

### DIFF
--- a/core/lowering/passes/device_casting.cpp
+++ b/core/lowering/passes/device_casting.cpp
@@ -8,23 +8,29 @@ namespace core {
 namespace lowering {
 namespace passes {
 
-void UnpackAndCastMaskedFill(std::shared_ptr<torch::jit::Graph>& graph) {
+void UnpackAndCastMaskedFill(std::shared_ptr<torch::jit::Graph>& graph, std::string target_device_name) {
   std::string masked_fill_pattern = R"IR(
     graph(%self, %mask, %value):
       %out: Tensor = aten::masked_fill_(%self, %mask, %value)
       return (%out))IR";
 
   // Calls to masked_fill_ often utilize CPU tensors, and as such
-  // should be casted to CUDA to avoid device mismatch errors
-  std::string unpacked_pattern = R"IR(
+  // should be moved to gpu to avoid device mismatch errors
+
+  // Separate string into portions to insert device name
+  std::string clean_pattern_part_1 = R"IR(
     graph(%self, %mask, %value):
-      %device: Device = prim::Constant[value="cuda"]()
+      %device: Device = prim::Constant[value=")IR";
+
+  std::string clean_pattern_part_2 = R"IR("]()
       %dtype: NoneType = prim::Constant()
       %false: bool = prim::Constant[value=0]()
       %mask_cuda: Tensor = aten::to(%mask, %device, %dtype, %false, %false)
       %self_cuda: Tensor = aten::to(%self, %device, %dtype, %false, %false)
-      %out: Tensor = aten::masked_fill_(%self_cuda, %mask_cuda, %value)
+      %out: Tensor = aten::masked_fill(%self_cuda, %mask_cuda, %value)
       return (%out))IR";
+
+  auto unpacked_pattern = clean_pattern_part_1 + target_device_name + clean_pattern_part_2;
 
   torch::jit::SubgraphRewriter masked_fill_rewriter;
   masked_fill_rewriter.RegisterRewritePattern(masked_fill_pattern, unpacked_pattern);
@@ -32,22 +38,28 @@ void UnpackAndCastMaskedFill(std::shared_ptr<torch::jit::Graph>& graph) {
   LOG_GRAPH("After unpack and cast masked_fill_: " << *graph);
 }
 
-void UnpackAndCastNumToTensor(std::shared_ptr<torch::jit::Graph>& graph) {
+void UnpackAndCastNumToTensor(std::shared_ptr<torch::jit::Graph>& graph, std::string target_device_name) {
   std::string num_to_tensor_cast_pattern = R"IR(
     graph(%1: Scalar):
       %2: Tensor = prim::NumToTensor(%1)
       return (%2))IR";
 
-  // 0D Tensors are initialized on cpu, and need to be casted to CUDA
+  // 0D Tensors are initialized on cpu, and need to be moved to gpu
   // to avoid device mismatch issues
-  std::string num_to_tensor_clean_pattern = R"IR(
+
+  // Separate string into portions to insert device name
+  std::string clean_pattern_part_1 = R"IR(
     graph(%1: Scalar):
       %2: Tensor = prim::NumToTensor(%1)
-      %device: Device = prim::Constant[value="cuda"]()
+      %device: Device = prim::Constant[value=")IR";
+
+  std::string clean_pattern_part_2 = R"IR("]()
       %dtype: NoneType = prim::Constant()
       %false: bool = prim::Constant[value=0]()
       %3: Tensor = aten::to(%2, %device, %dtype, %false, %false)
       return (%3))IR";
+
+  auto num_to_tensor_clean_pattern = clean_pattern_part_1 + target_device_name + clean_pattern_part_2;
 
   torch::jit::SubgraphRewriter num_to_tensor_cast_rewriter;
   num_to_tensor_cast_rewriter.RegisterRewritePattern(num_to_tensor_cast_pattern, num_to_tensor_clean_pattern);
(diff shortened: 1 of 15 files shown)


core/lowering/passes/device_casting.cpp
```diff
@@ -0,0 +1,103 @@
...
+  // Calls to masked_fill_ often utilize CPU tensors, and as such
+  // should be casted to CUDA to avoid device mismatch errors
+  std::string unpacked_pattern = R"IR(
+    graph(%self, %mask, %value):
+      %device: Device = prim::Constant[value="cuda"]()
+      %dtype: NoneType = prim::Constant()
+      %false: bool = prim::Constant[value=0]()
+      %mask_cuda: Tensor = aten::to(%mask, %device, %dtype, %false, %false)
+      %self_cuda: Tensor = aten::to(%self, %device, %dtype, %false, %false)
+      %out: Tensor = aten::masked_fill_(%self_cuda, %mask_cuda, %value)
```
gs-olive, line 26: This could be replaced by `aten::masked_fill`, for which a converter exists, however there are a few bugs which arise in edge cases when `%value` is a float but `%self_cuda` is an int, and similar such scenarios which are not directly handled by the converter and can cause errors in TRT. As a result, I opted for the unimplemented `aten::masked_fill_` version with casted tensors, in the meantime.

core/lowering/passes/device_casting.cpp
```diff
@@ -0,0 +1,103 @@
...
+    graph(%1: Tensor):
+      %2: Scalar = aten::item(%1)
+      return (%2))IR";
+
+  torch::jit::SubgraphRewriter scalar_implicit_cast_rewriter;
+  scalar_implicit_cast_rewriter.RegisterRewritePattern(scalar_implicit_cast_pattern, scalar_implicit_clean_pattern);
+  scalar_implicit_cast_rewriter.runOnGraph(graph);
+
+  LOG_GRAPH("After unpack and cast full: " << *graph);
+}
```
gs-olive, line 98: Testing for this function has not been added despite many attempts to do so, as the Torchscript interpreter will not recognize a 0D tensor as a Tensor (it takes the Tensor as an `int`). This causes the `aten::ScalarImplicit` schema to throw an error. It seems that in practice, with arguments being passed between engines, these 0D tensors are interpreted as Tensors, but this has been difficult to replicate. An example of IR which fails to compile in the test case is:

[Intended to be extremely similar to what is done in real model compilations which use `ScalarImplicit`]
```
      graph(%x.1: Tensor):
        %5 : int = prim::Constant[value=0]()
        %false : bool = prim::Constant[value=0]()
        %none : NoneType = prim::Constant()
        %cuda : Device = prim::Constant[value="cuda"]()
        %3 : int = aten::size(%x.1, %5)
        %y.2 : Tensor = prim::NumToTensor(%3)
        %y.1 : Tensor = aten::to(%y.2, %cuda, %none, %false, %false)
        %19 : Tensor[] = prim::ListConstruct(%x.1, %y.1)
        %21 : Tensor, %22 : Tensor = prim::ListUnpack(%19)

        %2 : Scalar = aten::ScalarImplicit(%22)
        return (%2))IR";
```

core/lowering/passes/device_casting.cpp
```diff
@@ -0,0 +1,103 @@
...
+  std::string masked_fill_pattern = R"IR(
+    graph(%self, %mask, %value):
+      %out: Tensor = aten::masked_fill_(%self, %mask, %value)
+      return (%out))IR";
+
+  // Calls to masked_fill_ often utilize CPU tensors, and as such
+  // should be casted to CUDA to avoid device mismatch errors
+  std::string unpacked_pattern = R"IR(
+    graph(%self, %mask, %value):
+      %device: Device = prim::Constant[value="cuda"]()
```
narendasan, line 21: What happens in the case of multi-gpu systems?

narendasan: Potentially could add an argument to take the target device.

narendasan: Take a look at snprintf and modifying `core::CompileSpec::lower_info` to add a device field which replicates the device info from the external API, then you should able to determine the target device at lower time.

core/lowering/passes/device_casting.cpp
narendasan, line 26: Right now this would handle the in place case. Can handle the functional case here too? https://github.com/pytorch/TensorRT/blob/5a7f00e8db1ca1d974e4583657a1f03e810f15a9/core/conversion/converters/converter_util.cpp#L62

core/lowering/passes/device_casting.cpp
```diff
@@ -0,0 +1,103 @@
...
+    graph(%1: Scalar):
+      %2: Tensor = prim::NumToTensor(%1)
+      return (%2))IR";
+
+  // 0D Tensors are initialized on cpu, and need to be casted to CUDA
+  // to avoid device mismatch issues
+  std::string num_to_tensor_clean_pattern = R"IR(
+    graph(%1: Scalar):
+      %2: Tensor = prim::NumToTensor(%1)
+      %device: Device = prim::Constant[value="cuda"]()
```
narendasan, line 46: See above about correct device

core/lowering/passes/device_casting.cpp
```diff
@@ -0,0 +1,103 @@
...
+  std::string full_cast_pattern = R"IR(
+    graph(%1, %2, %3, %4, %5, %6):
+      %out: Tensor = aten::full(%1, %2, %3, %4, %5, %6)
+      return (%out))IR";
+
+  // Tensors created via aten::full are initialized on cpu, and need to be casted to CUDA
+  // to avoid device mismatch issues
+  std::string full_clean_pattern = R"IR(
+    graph(%1, %2, %3, %4, %5, %6):
+      %cuda: Device = prim::Constant[value="cuda"]()
```
narendasan, line 69: ""

core/runtime/execute_engine.cpp
```diff
@@ -63,16 +63,41 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
...
+    for (size_t i = 0; i < inputs.size(); i++) {
+      at::Tensor* in = &inputs[i];
+      std::string current_tensor_device = in->device().str();
+
+      // If current device string does not match target device, display warning and move tensor accordingly
+      if (current_tensor_device != target_device) {
+        LOG_WARNING(
+            "Input " << i << " of engine " << compiled_engine->name << " was found to be on " << current_tensor_device
+                     << " but should be on " << target_device
+                     << ". This tensor is being moved manually by the runtime but "
```
narendasan, line 94: ~~manually~~

core/lowering/passes/device_casting.cpp
gs-olive: Thanks for the suggestion - this actually turns out to be a bug in the `aten::masked_fill` converter, as it behaves differently than Torch when the types of the input and value are different. Specifically, the converter throws an error whereas Torch just inherits the type of the first argument. I will make a new PR + Test Cases for this, as it is an unrelated bug.

gs-olive: Fix in PR #1430

cpp/src/compile_spec.cpp
```diff
@@ -126,6 +126,7 @@ torchtrt::core::CompileSpec to_internal_compile_spec(CompileSpec external) {
   internal.partitioning_info.forced_fallback_operators = std::move(external.torch_executed_ops);
   internal.partitioning_info.truncate_long_and_double = external.truncate_long_and_double;
   internal.lower_info.forced_fallback_modules = std::move(external.torch_executed_modules);
+  internal.lower_info.target_device_name = ":" + std::to_string(external.device.gpu_id);
```
gs-olive, line 129: Doesn't seem to be running before lowering phase

core/lowering/passes/device_casting.cpp
```diff
@@ -0,0 +1,121 @@
...
+  std::string clean_pattern_part_1 = R"IR(
+    graph(%self, %mask, %value):
+      %device: Device = prim::Constant[value=")IR";
+
+  std::string clean_pattern_part_2 = R"IR("]()
+      %dtype: NoneType = prim::Constant()
+      %false: bool = prim::Constant[value=0]()
+      %mask_cuda: Tensor = aten::to(%mask, %device, %dtype, %false, %false)
+      %self_cuda: Tensor = aten::to(%self, %device, %dtype, %false, %false)
+      %out: Tensor = aten::masked_fill(%self_cuda, %mask_cuda, %value)
```
gs-olive, line 30: Dependent on PR #1430 - will not work otherwise

core/lowering/passes/device_casting.cpp
```diff
@@ -0,0 +1,121 @@
...
+      %2: Tensor = prim::NumToTensor(%1)
+      %device: Device = prim::Constant[value=")IR";
+
+  std::string clean_pattern_part_2 = R"IR("]()
+      %dtype: NoneType = prim::Constant()
+      %false: bool = prim::Constant[value=0]()
+      %3: Tensor = aten::to(%2, %device, %dtype, %false, %false)
+      return (%3))IR";
+
+  auto num_to_tensor_clean_pattern = clean_pattern_part_1 + target_device_name + clean_pattern_part_2;
```
gs-olive, line 62: Had to use this paradigm instead of `snprintf` because the `%` symbols in the IR are registered as formatting for `snprintf`, which made it difficult to insert the device string

core/runtime/execute_engine.cpp
```diff
@@ -63,16 +63,40 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
...
     set_cuda_device(device);
 
-    std::string target_device = "cuda:" + std::to_string(device.id);
+    // Target device is new device
+    target_device += std::to_string(device.id);
 
     for (auto& in : inputs) {
       in = in.to(torch::Device(target_device));
     }
+  } else {
```
narendasan, line 80: Doesn't need to be an else, could just be a second check.

gs-olive: Updated the else block to just assign the `cuda` target device name, and now the runtime device check is applied as a second check

narendasan:
fixes: #1446

PR #1442, opened 2022-11-04
fix: Error with `aten::div` when using truncation with Int32 tensor inputs

### LINKED ISSUE #1441: 🐛 [Bug] Bug encountered when compiling `Jasper10x5dr` network
##  Bug Description

When compiling the [Jasper10x5dr](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/stt_en_jasper10x5dr) speech recognition model with Torch-TRT, the following error is displayed:

```python
ERROR: [Torch-TensorRT TorchScript Conversion Context] - 3: (Unnamed Layer* 22) [Unary]: operation FLOOR not allowed on type Int32
...
RuntimeError: [Error thrown at core/conversion/converters/impl/unsqueeze.cpp:25] Expected dim <= nbDims && dim >= -(nbDims + 1) to be true but got false
Dimension out of range (expected to be in range of [-1, 0], but got 1)
```

## To Reproduce

1. Run torch_tensorrt.compile with `Jasper10x5dr` model as input, using fp32 precision.
2. Choose fixed input sizes of [1, 64, 256] and [1,] and enable truncate_long_and_double with 8 GB workspace.

## Expected behavior

Model should successfully compile to Torch-TRT. Specifically, the floor operation should not throw an error on the input

## Environment

- Torch-TensorRT Version: 1.3.0a0+b5bcccf9
- PyTorch Version: 1.14.0.dev20221018+cu116
- OS: Ubuntu 20.04
- Build command you used: python setup.py develop
- Python version: 3.8.13
- CUDA version: 11.6

gs-olive:
# Description
- `aten::div` with truncation on integer tensor inputs currently throws an error if both inputs are integer type, as the TRT unary operations for absolute value and floor do not apply to Int32 or Bool types
- For absolute value, this is a legitimate bug as `aten::abs` is functional for integer types
- For the floor operation, `aten::floor` does not explicitly support integer inputs, and `torch.floor()` does not work with Int32 inputs by default (on `1.13.0.dev20220921+cu116`). However, `torch.div(..., rounding_mode="trunc")` with integer tensors does return an integer value, and so the corollary Torch-TRT converter should behave similarly
- Modified `aten:abs` converter logic to be a utility (moved file location), as the operator is used in multiple locations
- Added regression test to ensure truncation divide with two integer tensors is functional

Note: The behavior of `torch.floor()` on Int32 types differs between `1.13.0.dev20220921+cu116` and `1.14.0.dev20221018+cu116`: the former does not by default support this operation, while the latter does. This PR does not fix the general `aten::floor` operator for Int32 inputs, but instead fixes the `aten::div` truncation operator only.

Fixes #1441 
Note: The issue was traced to a problem with `aten::div` with truncation enabled, and not `aten::floor`

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/converter_util.cpp
+++ b/core/conversion/converters/converter_util.cpp
@@ -156,6 +156,38 @@ nvinfer1::ILayer* add_elementwise(
   return ele;
 }
 
+nvinfer1::ITensor* add_abs(
+    ConversionCtx* ctx,
+    const torch::jit::Node* n,
+    nvinfer1::ITensor* self,
+    const std::string& name) {
+  nvinfer1::ILayer* absolute_value_layer;
+
+  // Check if TRT Unary ops support the input type
+  bool unary_supported_input = (self->getType() == nvinfer1::DataType::kFLOAT) ||
+      (self->getType() == nvinfer1::DataType::kHALF) || (self->getType() == nvinfer1::DataType::kINT8);
+  if (unary_supported_input) {
+    absolute_value_layer = ctx->net->addUnary(*self, nvinfer1::UnaryOperation::kABS);
+    TORCHTRT_CHECK(absolute_value_layer, "Unable to create abs layer from node: " << *n);
+    absolute_value_layer->setName(name.c_str());
+  } else {
+    LOG_GRAPH(
+        "Tensor is of unsupported type "
+        << self->getType() << " for IUnaryLayer::kABS. Using backup implementation via IElementWise (max(x, -x)");
+    // For types not supported by kABS, use an elementwise implementation abs(x) = max(x, -1 * x)
+    at::Tensor neg_one = torch::full({1}, -1).to(util::TRTDataTypeToScalarType(self->getType()));
+    auto neg_one_const = tensor_to_const(ctx, neg_one);
+    auto neg_layer = add_elementwise(
+        ctx, nvinfer1::ElementWiseOperation::kPROD, self, neg_one_const, util::node_info(n) + std::string("_Negation"));
+    TORCHTRT_CHECK(neg_layer, "Unable to create prod layer from node: " << *n);
+    absolute_value_layer =
+        add_elementwise(ctx, nvinfer1::ElementWiseOperation::kMAX, self, neg_layer->getOutput(0), name);
+    TORCHTRT_CHECK(absolute_value_layer, "Unable to create max layer from node: " << *n);
+  }
+
+  return absolute_value_layer->getOutput(0);
+}
+
 nvinfer1::ITensor* applyIdentityOp(ConversionCtx* ctx, nvinfer1::ITensor* tensor, const std::string& tensor_name) {
   auto id_layer = ctx->net->addIdentity(*tensor);
   auto id_out_tensor = id_layer->getOutput(0);
--- a/core/conversion/converters/impl/element_wise.cpp
+++ b/core/conversion/converters/impl/element_wise.cpp
@@ -326,15 +326,27 @@ auto element_wise_registrations TORCHTRT_UNUSED =
                } else if (rounding_mode == "trunc") {
                  // trunc = floor(abs(div)) * sign(div)
                  auto tmp_div = add_elementwise(ctx, nvinfer1::ElementWiseOperation::kDIV, self, other, "tmp_div");
-                 auto abs = ctx->net->addUnary(*tmp_div->getOutput(0), nvinfer1::UnaryOperation::kABS);
-                 auto floor = ctx->net->addUnary(*abs->getOutput(0), nvinfer1::UnaryOperation::kFLOOR);
+                 auto abs = add_abs(ctx, n, tmp_div->getOutput(0), util::node_info(n) + "_absolute_val");
+
+                 // In this case, we allow the floor unary on non-TRT Unary types, as it is needed for this
+                 // specific function. Floor applied to non-float types equates to identity
+                 nvinfer1::ITensor* floor;
+
+                 if ((abs->getType() == nvinfer1::DataType::kINT32) || (abs->getType() == nvinfer1::DataType::kBOOL)) {
+                   LOG_DEBUG(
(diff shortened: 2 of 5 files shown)


core/conversion/converters/impl/element_wise.cpp
```diff
@@ -326,8 +326,24 @@ auto element_wise_registrations TORCHTRT_UNUSED =
...
+                   LOG_GRAPH(
+                       "Tensor is of unsupported type " << abs->getOutput(0)->getType()
+                                                        << " for IUnaryLayer::kFLOOR. Using identity instead.");
+                   floor = ctx->net->addIdentity(*abs->getOutput(0));
+                   TORCHTRT_CHECK(floor, "Unable to create identity layer from node: " << *n);
+                 } else {
+                   floor = ctx->net->addUnary(*abs->getOutput(0), nvinfer1::UnaryOperation::kFLOOR);
+                   TORCHTRT_CHECK(floor, "Unable to create floor layer from node: " << *n);
+                 }
+                 floor->setName((util::node_info(n) + "_floor").c_str());
```
gs-olive, line 345: In this code block, both the `abs` and the `floor` operators were encountering errors when both inputs are integer types. The solution for `abs` was a converter utility, whereas for `floor`, the solution only appears here. The reasoning for this choice was that Torch support for `torch.floor()` applied to an Int32 type differs between `1.13.0.dev20220921+cu116` and `1.14.0.dev20221018+cu116`, so it is unclear if the `aten::floor` converter should generally support Int32 inputs or not, currently.

core/conversion/converters/impl/element_wise.cpp
```diff
@@ -326,8 +326,24 @@ auto element_wise_registrations TORCHTRT_UNUSED =
...
-                 auto abs = ctx->net->addUnary(*tmp_div->getOutput(0), nvinfer1::UnaryOperation::kABS);
-                 auto floor = ctx->net->addUnary(*abs->getOutput(0), nvinfer1::UnaryOperation::kFLOOR);
+                 auto abs = add_absolute_value(ctx, n, tmp_div->getOutput(0), util::node_info(n) + "_absolute_val");
+
+                 // In this case, we allow the floor unary on non-TRT Unary types, as it is needed for this
+                 // specific function. Floor applied to non-float types equates to identity
+                 nvinfer1::ILayer* floor;
+                 if ((abs->getOutput(0)->getType() == nvinfer1::DataType::kINT32) ||
+                     (abs->getOutput(0)->getType() == nvinfer1::DataType::kBOOL)) {
+                   LOG_GRAPH(
```
narendasan, line 336: LOG_DEBUG instead of LOG_GRAPH

core/conversion/converters/impl/element_wise.cpp
```diff
@@ -326,8 +326,24 @@ auto element_wise_registrations TORCHTRT_UNUSED =
                } else if (rounding_mode == "trunc") {
                  // trunc = floor(abs(div)) * sign(div)
                  auto tmp_div = add_elementwise(ctx, nvinfer1::ElementWiseOperation::kDIV, self, other, "tmp_div");
-                 auto abs = ctx->net->addUnary(*tmp_div->getOutput(0), nvinfer1::UnaryOperation::kABS);
-                 auto floor = ctx->net->addUnary(*abs->getOutput(0), nvinfer1::UnaryOperation::kFLOOR);
+                 auto abs = add_absolute_value(ctx, n, tmp_div->getOutput(0), util::node_info(n) + "_absolute_val");
+
+                 // In this case, we allow the floor unary on non-TRT Unary types, as it is needed for this
+                 // specific function. Floor applied to non-float types equates to identity
+                 nvinfer1::ILayer* floor;
```
narendasan, line 333: Work with ITensor* instead ideally

core/conversion/converters/impl/element_wise.cpp
narendasan, line 345: This needs to only be applied to the unary layer on line 342 so as not to overwrite the info from the abs on 329

core/conversion/converters/converter_util.cpp
```diff
@@ -156,6 +156,38 @@ nvinfer1::ILayer* add_elementwise(
...
+    auto neg_one_const = tensor_to_const(ctx, neg_one);
+    auto neg_layer = add_elementwise(
+        ctx, nvinfer1::ElementWiseOperation::kPROD, self, neg_one_const, util::node_info(n) + std::string("_Negation"));
+    TORCHTRT_CHECK(neg_layer, "Unable to create prod layer from node: " << *n);
+    absolute_value_layer =
+        add_elementwise(ctx, nvinfer1::ElementWiseOperation::kMAX, self, neg_layer->getOutput(0), name);
+    TORCHTRT_CHECK(absolute_value_layer, "Unable to create max layer from node: " << *n);
+  }
+
+  return absolute_value_layer->getOutput(0);
```
gs-olive, line 188: Switched function schema to return the output of the absolute value layer and not the layer itself

core/conversion/converters/impl/element_wise.cpp
```diff
@@ -326,15 +326,27 @@ auto element_wise_registrations TORCHTRT_UNUSED =
...
+
+                 // In this case, we allow the floor unary on non-TRT Unary types, as it is needed for this
+                 // specific function. Floor applied to non-float types equates to identity
+                 nvinfer1::ITensor* floor;
+
+                 if ((abs->getType() == nvinfer1::DataType::kINT32) || (abs->getType() == nvinfer1::DataType::kBOOL)) {
+                   LOG_DEBUG(
+                       "Tensor is of unsupported type " << abs->getType()
+                                                        << " for IUnaryLayer::kFLOOR. Using identity instead.");
+                   floor = abs;
```
gs-olive, line 339: Instead of using the identity function, the floor output tensor is just set to the absolute value result, to avoid unnecessary computation.

narendasan (approved):
LGTM

PR #1450, opened 2022-11-15
Improve batch_norm fp16 accuracy

mfeliz-cruise:
# Description

Use float types for compile-time calculations around batch_norm. Improves fp16 accuracy relative to pytorch. Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/batch_norm.cpp
+++ b/core/conversion/converters/impl/batch_norm.cpp
@@ -20,15 +20,28 @@ void _batch_norm(
     const torch::Tensor& mean,
     const torch::Tensor& var,
     const float eps) {
-  auto scale = gamma / torch::sqrt(var + eps);
-  auto bias = beta - mean * scale;
+  auto orig_dtype = var.dtype();
+  // perform compile-time weight calculations in float to improve accuracy
+  // resulting weights will be embedded as the original dtype
+  auto calculation_gamma = gamma;
+  auto calculation_beta = beta;
+  auto calculation_mean = mean;
+  auto calculation_var = var;
+  if (orig_dtype == torch::kHalf) {
+    calculation_gamma = calculation_gamma.to(torch::kFloat);
+    calculation_beta = calculation_beta.to(torch::kFloat);
+    calculation_mean = calculation_mean.to(torch::kFloat);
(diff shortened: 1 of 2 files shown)


core/conversion/converters/impl/batch_norm.cpp
```diff
@@ -20,15 +20,28 @@ void _batch_norm(
...
-  auto scale = gamma / torch::sqrt(var + eps);
-  auto bias = beta - mean * scale;
+  auto orig_dtype = var.dtype();
+  // perform compile-time weight calculations in float to improve accuracy
+  // resulting weights will be embedded as the original dtype
+  auto calculation_gamma = gamma;
+  auto calculation_beta = beta;
+  auto calculation_mean = mean;
+  auto calculation_var = var;
+  if (orig_dtype == torch::kHalf) {
```
narendasan, line 30: Quick question is this different than the normal pytorch behavior? If so can we add a debug message here saying that we are doing this to improve accuracy?

mfeliz-cruise: It looks like the cudnn implementation at least asserts that the weight is fp32 which would force similar calculations to fp32:
https://github.com/pytorch/pytorch/blob/4bfe2a24505049fa4fe43d24c2e3a5f5d99d9f00/aten/src/ATen/native/cudnn/BatchNorm.cpp#L110

narendasan: Ok

narendasan (approved):
Changes LGTM

narendasan:
Seems like the BN FP16 test is failing, maybe we need to change the metric for this test since I guess by definition the numerics will be different?

mfeliz-cruise:
> Seems like the BN FP16 test is failing, maybe we need to change the metric for this test since I guess by definition the numerics will be different?

I lowered the bar to 2e-4 which may be more reasonable. Both pass locally for me so it may depend on GPU and the numerics of the tactic chosen by TensorRT.

mfeliz-cruise:
I rebased this today, but it looks like there's a CI failure on pytorch 1.14 currently:
ERROR: Could not find a version that satisfies the requirement torch==1.14.0.dev20221114+cu116

PR #1458, opened 2022-11-17
make padding layer converter more efficient

FelixXidddd:
# Description

In the current padding layer converter for version > 8.2, there are 3 padding layers: pre_pad + mid_pad + post_pad to do the converter. But please consider this case, we want to pad tensor from (2048, 628, 20) to (2048, 628, 32), the pre_pad and mid_pad can be erased because they are doing the opposite operation and waste time.
And consider that from version 8.2, the start of the slice layer can support negative. So let's use one padding layer to do this.

Fixes # (issue)
As described above, we can improve perf significantly.
## Type of change

-Perf improves, so no additional functional unit test is needed.

# Checklist:

### DIFF
--- a/py/torch_tensorrt/fx/converters/acc_ops_converters.py
+++ b/py/torch_tensorrt/fx/converters/acc_ops_converters.py
@@ -397,46 +397,29 @@ def acc_ops_pad_with_slice_layer(
     )
 
     input_shape = input_val.shape
-    pre_start = tuple(i - 1 for i in input_shape)
     prefix_len = len(input_shape) - len(pad) // 2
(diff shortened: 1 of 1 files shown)


FelixXidddd:
@narendasan Could you please help to point out what should I do to trigger the CI? I don't think the CI is launched.
I remembered that CI should be launched automatically.

narendasan:
Not sure why none of the actions are running

FelixXidddd:
Aha, let's wait one day. If no actions still, I will close this PR and file a new one to see.
Anyway, thanks a lot.

PR #1479, opened 2022-11-24
Fix build

yinghai:
# How to debug
1. Follow https://circleci.com/docs/ssh-access-jobs/ to get into the host where test failed (e.g. https://app.circleci.com/pipelines/github/pytorch/TensorRT/1461/workflows/877d58af-c765-48df-b6c0-e6328723a7fd/jobs/6968) 
2. Repeat the failed command in the circleci host: `python3 -c "import torch_tensorrt; torch_tensorrt.dump_build_info()"` and see that issue can be reproduced. 
3. Repeat with `LD_DEBUG=libs python3 -c "import torch_tensorrt; torch_tensorrt.dump_build_info()"` and locate the torch libraries at `/opt/circleci/.pyenv/versions/3.9.4/lib/python3.9/site-packages/torch/lib/libc10_cuda.so`. 
4. Do `nm -C /opt/circleci/.pyenv/versions/3.9.4/lib/python3.9/site-packages/torch/lib/libc10_cuda.so | grep CUDACachingAllocator |grep T` and noticed that it seems weird as it doesn't contain the symbol ` c10::cuda::CUDACachingAllocator::allocator`. It seems that the libs are not from torch 1.14 nightly. 
5. Check the torch version `python3 -c "import torch; print(torch.__version__)"` and it showed `1.13.0+cu117`. This is very weird as we are supposed to install `1.14.0+cu116`. 
6. Check the CI pipeline at noticed that during the `Install torch-tensorrt` stage, we are actually uninstalling the previously installed torch 1.14 

> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> torchaudio 0.14.0.dev20221114+cu116 requires torch==1.14.0.dev20221114, but you have torch 1.13.0 which is incompatible.

7. Check `py/setup.py` and find that we are forcing torch version to be "torch>=1.13.0.dev0,<1.14.0". Fix that. 
 

## Type of change

# Checklist:

### DIFF
--- a/core/BUILD
+++ b/core/BUILD
@@ -33,8 +33,8 @@ cc_library(
         "//core/util/logging",
         "@tensorrt//:nvinfer",
     ] + select({
-        ":use_pre_cxx11_abi": ["@libtorch_pre_cxx11_abi//:libtorch"],
-        "//conditions:default": ["@libtorch//:libtorch"],
+        ":use_pre_cxx11_abi": ["@libtorch_pre_cxx11_abi//:libtorch", "@libtorch_pre_cxx11_abi//:c10_cuda"],
+        "//conditions:default": ["@libtorch//:libtorch", "@libtorch//:c10_cuda"],
     }),
     alwayslink = True,
 )
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -367,7 +367,7 @@ commands:
       - run:
           name: Test torch
           command: |
-            python3 -c "import torch; print(torch.cuda.is_available()); print(torch.cuda.device_count())"
+            python3 -c "import torch; print(torch.cuda.is_available()); print(torch.cuda.device_count()); print(torch.__version__)"
 
       - run:
           name: Get torch-tensorrt version information
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -946,7 +946,7 @@ parameters:
   # Nightly platform config
   torch-build:
     type: string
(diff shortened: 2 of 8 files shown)


yinghai:
@narendasan `//tests/core/conversion/converters:test_einsum` failed. Please help check. Thanks.

core/BUILD
```diff
@@ -33,8 +33,8 @@ cc_library(
         "//core/util/logging",
         "@tensorrt//:nvinfer",
     ] + select({
-        ":use_pre_cxx11_abi": ["@libtorch_pre_cxx11_abi//:libtorch"],
-        "//conditions:default": ["@libtorch//:libtorch"],
+        ":use_pre_cxx11_abi": ["@libtorch_pre_cxx11_abi//:libtorch", "@libtorch_pre_cxx11_abi//:c10_cuda"],
+        "//conditions:default": ["@libtorch//:libtorch", "@libtorch//:c10_cuda"],
```
frank-wei, line 37: Did you forget to mention in the diagnosis that we also need to link with c10_cuda?  :-)

frank-wei (approved):
Thanks @yinghai  for your time on deep diving in this issue. It helps us a lot on unblocking the 1.3 release!

core/BUILD
yinghai: We probably don't need that. But it doesn't hurt.

Issue #1455, opened 2022-11-17
🐛 [Bug] Compilation Error on GPT-2


gs-olive:
##  Bug Description

When converting the GPT-2 network (https://huggingface.co/gpt2) from TorchScript to Torch-TRT, the following error is encountered:

```python
compiled_cpp_mod = _C.compile_graph(module._c, _parse_compile_spec(spec))
RuntimeError: [Error thrown at core/partitioning/shape_analysis.cpp:167] Unsupported input data type unsigned char
```

## To Reproduce

1. Run torch_tensorrt.compile with GPT-2 model as input, using fp32 precision.
2. Choose fixed input size of [1, 128] and enable truncate_long_and_double with 12 GB workspace.
3. Pass in model keyword args to disable attention and hidden state outputs

## Expected behavior

Model should successfully compile to Torch-TRT. Specifically, internal (non-user-provided) type-casting issues should not cause errors.

## Environment

- Torch-TensorRT Version: 1.3.0a0+e3b99294
- PyTorch Version: 1.13.0.dev20220921+cu116
- CPU Architecture: Intel Xeon CPU
- OS: Ubuntu 20.04
- How you installed PyTorch: pip
- Build command you used: `python setup.py develop`
- Are you using local sources or building from archives: local
- Python version: 3.8.13
- CUDA version: 11.6

## Additional context

The problematic data in GPT-2 seems to be this [bias term](https://github.com/huggingface/transformers/blob/5e012f8e3c3a8d8bb5d6d95c58ff59a4a3156519/src/transformers/models/gpt2/modeling_gpt2.py#L128-L133), instantiated in the attention module, which has type `uint8`. In both the TorchScript IR and the model code ([example 1](https://github.com/huggingface/transformers/blob/5e012f8e3c3a8d8bb5d6d95c58ff59a4a3156519/src/transformers/models/gpt2/modeling_gpt2.py#L196), [example 2](https://github.com/huggingface/transformers/blob/5e012f8e3c3a8d8bb5d6d95c58ff59a4a3156519/src/transformers/models/gpt2/modeling_gpt2.py#L246)), it seems that this bias term is generally cast to a bool. The error is thrown in this code segment:
https://github.com/pytorch/TensorRT/blob/e3b992941b3ae5f1863de271fc9032829834ec6a/core/partitioning/shape_analysis.cpp#L258-L260
[...]

PR #1533, opened 2022-12-07
fix: `aten::where` with differing-shape inputs bugfix

gs-olive:
# Description
- Behavior of Torch-TRT differs from that of Torch in the case where the input tensors to `aten::where` have different rank
- Torch automatically broadcasts tensors to the highest-rank variant whereas the TRT Select layer requires tensors of the same rank and throws an error
- Add dimension checking and unsqueeze operator to ensure broadcasting is enabled
- Add test case to catch error

Resolves Part 1 of solution to Issue #1455 

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/select.cpp
+++ b/core/conversion/converters/impl/select.cpp
@@ -736,8 +736,41 @@ auto select_registrations TORCHTRT_UNUSED =
             {"aten::where.self(Tensor condition, Tensor self, Tensor other) -> (Tensor)",
              [](ConversionCtx* ctx, const torch::jit::Node* n, args& args) -> bool {
                auto condition = args[0].ITensorOrFreeze(ctx);
+               auto c_nbDims = condition->getDimensions().nbDims;
                auto x = args[1].ITensorOrFreeze(ctx);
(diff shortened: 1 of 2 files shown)


core/conversion/converters/impl/select.cpp
```diff
@@ -736,8 +736,41 @@ auto select_registrations TORCHTRT_UNUSED =
...
+                   // unsqueeze dimensions by 1
+                   for (auto i = dimensions.nbDims; i < max_nbDims; i++) {
+                     dimensions = util::unsqueezeDims(dimensions, 0, 1, false);
+                   }
+
+                   // Reshape to the unsqueezed dimensions
+                   shuffle_layer->setReshapeDimensions(dimensions);
+                   tensor = shuffle_layer->getOutput(0);
+                 }
+               };
```
gs-olive, line 768: Could be too similar to `util::addPadding`

gs-olive: Refactored implementation to use `addPadding`

narendasan (approved):
LGTM

Issue #1543, opened 2022-12-12
✨[Feature] Improve Integer Range Specification for `Input` class


gs-olive:
Adding FX support for specifying integer range + adding lowering pass to support int64 inputs with automatic casting.

1/2 MVP + Phase 1 of RFC #1425 

See RFC for details.

gs-olive:
Addressed by PR #1551, Phase 1 already exists as discussed in https://github.com/pytorch/TensorRT/discussions/1425#discussioncomment-4661390

Christina-Young-NVIDIA:
Also a duplicate of 1531.

PR #1551, opened 2022-12-15
feat: Automatically cast user inputs to inferred data type

### LINKED ISSUE #1546: ✨[Feature] Support INT64 inputs at the graph input level
Currently if the graph is given int64 inputs, the compilation fails with unsupported type. 

Solution we are expecting : Cast INT64 inputs to INT32 and issue a warning to users about this instead of breaking compilation.

**Additional context**

gs-olive:
# Description

- Add post-lowering pass to insert `aten::to` operators for Tensor inputs determined to require float or int inputs
- Specifically, if the user provides an non-float input to a float-dtype input field and has `truncate_long_and_double=True`, a Torch-executed graph block will be inserted which casts that input to a float in-place.
- This operation modifies user-provided tensors and provides a warning as such
- Currently, the feature is only functional for Tensor inputs (not input signatures) and only casts to int and float types - if the input is specified as any other type, a cast will not be inserted
- Modify compiler to extract inferred data types for each input
- Add testing to ensure casts are inserted correctly and run in Torch

Fixes #1546 
Addresses most of #1543

## Type of change

# Checklist:


core/lowering/lowering.cpp
```diff
@@ -26,6 +26,56 @@ void LowerBlock(torch::jit::Block* b) {
...
+      LOG_DEBUG("Inserting aten::to casting " << input->debugName() << " to dtype " << dtype);
+
+      // Generate cast node sending input tensors to the inferred or specified datatype (float or int)
+      auto const_type = g->insertConstant(dtype);
+      auto const_false = g->insertConstant(0);
+      const_false->setType(torch::jit::BoolType::get());
+      auto cuda = g->insertConstant(target_device_name);
+      cuda->setType(torch::jit::DeviceObjType::get());
+      auto none_val = g->insertNode(g->createNone())->output();
+      auto cast_node = g->create(torch::jit::aten::to, {input, cuda, const_type, const_false, const_false, none_val});
```
gs-olive, line 59: This inserted `aten::to` could be altered to return a copy of the original tensor, so as not to modify user-provided inputs, however this could have performance impacts since a copy would be made of _every_ input tensor, every time inference is run, and regardless of whether the user-provided type was correct for the input. Instead, we currently display a warning that inputs which are of different type than their corresponding engine are casted in-place, and this casting is disabled by default since it requires `truncate_long_and_double=True`

core/lowering/lowering.cpp
```diff
@@ -26,6 +26,56 @@ void LowerBlock(torch::jit::Block* b) {
...
+      // Mark the cast node to run in PyTorch for ease of casting
+      LOG_GRAPH("Marking autocast node " << util::node_info(cast_node) << " to run in PyTorch");
+      cast_node->i_(c10::Symbol::attr("to_compile"), (int64_t) false);
+    }
+  }
+
+  LOG_WARNING(
+      "Input tensors to this Torch-TRT engine may have their data types in-place modified "
+      << "if the type does not match the determined required type for the model. To disable this "
+      << "automatic casting, specify truncate_long_and_double=False");
```
gs-olive, line 74: Warning referenced in the above comment, highlighting that the autocast feature can modify the data type of input tensors.

core/compiler.cpp
```diff
@@ -307,7 +310,14 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg)
       // Infer the type of an input from the weights of the calculation
       auto first_use_types = ir::get_block_first_calc_dtypes_opt_collection(g->block());
 
-      MapInputsAndDetermineDTypes(cfg, g, static_params, first_use_types);
+      // Extract map of IValue to DType
+      auto type_map = MapInputsAndDetermineDTypes(cfg, g, static_params, first_use_types);
+
+      // Use dtype map to autocast inputs to the correct type
+      if (cfg.partitioning_info.enabled && cfg.partitioning_info.truncate_long_and_double) {
```
gs-olive, line 317: An alternative to having the `truncate_long_and_double` flag control this feature is to add a new flag solely for `autocast` in case users want to have internal tensors truncated, but don't want the overhead of an additional engine to truncate inputs.

narendasan:
@peri044 is the reviewer for this feature

core/lowering/lowering.cpp
```diff
@@ -26,6 +26,63 @@ void LowerBlock(torch::jit::Block* b) {
...
+      // Ensure the data type to be casted to exists in the type map
+      if (dtype_input == input_type_map.end() || !dtype_input->second) {
+        LOG_DEBUG("No inferred input dtype for tensor " << input->debugName() << ", skipping autocast");
+        continue;
+      }
+
+      auto dtype = dtype_input->second.value();
+      // Currently, we do not autocast inputs for which the determined type is not long
+      if (dtype != at::kLong) {
+        continue;
```
peri044, line 51: Can you also issue a LOG_DEBUG statement here saying we are skipping autocasting because dtype is not at::kLong ?

core/lowering/lowering.cpp
```diff
@@ -26,6 +26,63 @@ void LowerBlock(torch::jit::Block* b) {
...
+
+      // Mark the cast node to run in PyTorch for ease of casting
+      LOG_GRAPH("Marking autocast node " << util::node_info(cast_node) << " to run in PyTorch");
+      cast_node->i_(c10::Symbol::attr("to_compile"), (int64_t) false);
+      num_autocasts++;
+    }
+  }
+
+  LOG_WARNING(
+      "Input tensors to this Torch-TRT engine may have their data types in-place modified "
```
peri044, line 77: Does this data-types in-place modified statement sound correct ? I am wondering if this might be confusing because we are explicitly inserting aten::to casts here ? Let me know what you think or open to any other suggestions.

cpp/src/types.cpp
```diff
@@ -87,6 +87,25 @@ nvinfer1::DataType toTRTDataType(DataType value) {
   }
 }
 
+at::ScalarType toAtDataType(DataType value) {
```
peri044, line 90: Suggestion - rename it to toAtenDataType ?

py/torch_tensorrt/csrc/tensorrt_classes.cpp
```diff
@@ -40,6 +44,27 @@ nvinfer1::DataType toTRTDataType(DataType value) {
   }
 }
 
+at::ScalarType toAtDataType(DataType value) {
+  switch (value) {
```
peri044, line 48: suggestion - rename it to toAtenDataType ?

py/torch_tensorrt/_Input.py
```diff
@@ -242,6 +244,9 @@ def _parse_dtype(dtype: Any) -> _enums.dtype:
                 + str(type(dtype))
             )
 
+    def is_TRT_dtype(self) -> bool:
```
peri044, line 247: rename it to is_trt_dtype  ?

peri044 (changes requested):

peri044:
@mfeliz-cruise Please check the description for more details on the extent of int64 dtype support. This PR should resolve the usecases that we discussed.

core/lowering/lowering.cpp
gs-olive: I originally included this statement since users will almost certainly have INT64 input tensors returned as INT32 tensors, in-place on account of the inserted `aten::to` operators (I have verified this with the Python API as well). I suppose this warning may be redundant, however, since Autocast currently requires `truncate_long_and_double`, so it might already be implicit that casting can occur

peri044: How about this ? We issue a warning when `num_autocasts>0`
LOG_WARNING("Data types for input tensors have been modified by inserting aten::to operations which cast INT64 inputs to INT32. To disable this, please recompile using INT32 inputs")

peri044 (changes requested):

core/lowering/lowering.cpp
```diff
@@ -48,18 +48,26 @@ int AutocastLongInputs(
...
       // Generate cast node sending input tensors to the inferred or specified datatype (long)
+      torch::jit::Value *const_false, *cuda, *none_val;
+      if (num_autocasts == 0) {
+        // Only generate constants once and reuse for all autocasts
+        const_false = g->insertConstant(0);
+        const_false->setType(torch::jit::BoolType::get());
+        cuda = g->insertConstant(target_device_name);
+        cuda->setType(torch::jit::DeviceObjType::get());
+        none_val = g->insertNode(g->createNone())->output();
+      }
```
peri044, line 68: Shouldn't this block of code be at line 33 before the for loop begins in order to be reused ?

core/lowering/lowering.cpp
```diff
@@ -48,18 +48,26 @@ int AutocastLongInputs(
...
+
       auto const_type = g->insertConstant(dtype);
```
peri044, line 70: Should this be dtype instead of INT32 ? coz dtype here would be at::kLong correct ?

core/lowering/lowering.cpp
gs-olive: I don't think this should be outside of the `for` loop, because we only want to insert it if the checks on lines 39, 43, and 50 pass, in which case an autocast will be inserted. The code already increments the `num_autocasts` variable when a cast is inserted (line 80), so this code should only run maximally once. What are your thoughts on this?

core/lowering/lowering.cpp
gs-olive: It is correct that `dtype = at::kLong` in this case, but the intent was to cast to `at::kLong` here, because this autocast was written to insert a Torch block at the beginning of the graph.

The reason is for the following case: if partitioning determines that the first block executed in the graph should be TensorRT, but the user specifies an INT64 input type, we currently can't insert a cast to INT32 without adding a Torch-executed block first. By casting to INT64 and marking the cast node to run in PyTorch, we ensure that a Torch block is inserted first, and partitioning's INT64 <-> INT32 conversion takes care of the rest.

core/lowering/lowering.cpp
peri044: Oh got it. I was confused a bit. Makes sense

peri044 (approved):
LGTM.

gs-olive:
Fixes #1543

PR #1599, opened 2023-01-19
fix: Allow full model compilation with collection outputs

### LINKED ISSUE #1368: 🐛 [Bug]  Support for modules with multiple outputs seems broken in v1.2.0
##  Bug Description

It appears that modules with multiple outputs no longer compile when using dynamic input shapes in v1.2.0. 

The following example works in **v1.1.1** but fails in **v1.2.0**
```
import torch
import torch.nn as nn
import torch_tensorrt as trt

from torch import Tensor
from typing import List, Tuple

trt.logging.set_reportable_log_level(trt.logging.Level.Debug)

class Net(nn.Module):
    def __init__(self):
        nn.Module.__init__(self)

        self.h = nn.Conv2d(1, 4, 3, padding=1)
        self.g = nn.Conv2d(1, 4, 3, padding=1)

    def forward(self, x) -> Tuple[Tensor, Tensor]:
        return self.h(x), self.g(x)

model = Net().eval()
model = torch.jit.trace(model, torch.randn(1, 1, 128, 128))
model = trt.compile(
    model.cuda(),
    inputs=[
        trt.Input(min_shape=(1, 1, 128, 128),
                  opt_shape=(4, 1, 256, 256),
                  max_shape=(8, 1, 512, 512))
    ],
    min_block_size=1
    require_full_compilation=True
)

```
Fails with error:
> RuntimeError: [Error thrown at core/conversion/conversion.cpp:230] Tuple type. Only a single tensor or a TensorList type is supported.

In *v1.1.1*, the graph returns two output tensors - while in **v1.2.0** it creates an intermediate node to (%13) and returns a single TupleConstruct output. Unfortunately MarkOutputs in core/conversion/converter.cpp now only gets a single tuple output and throws an error.

Graphs are given below:

**v1.1.1**
```
  %11 : Tensor = aten::_convolution(%x, %self.h.weight, %self.h.bias, %3, %3, %3, %5, %2, %4, %5, %5, %6, %6), scope: __module.h # /home/cjenkins/dev/mat/venv/lib/python3.8/site-packages/torch/nn/modules/conv.py:453:0
  %12 : Tensor = aten::_convolution(%x, %self.g.weight, %self.g.bias, %3, %3, %3, %5, %2, %4, %5, %5, %6, %6), scope: __module.g # /home/cjenkins/dev/mat/venv/lib/python3.8/site-packages/torch/nn/modules/conv.py:453:0
  return (%11, %12)
```

**v1.2.0**
```
  %11 : Tensor = aten::_convolution(%x, %self.h.weight, %self.h.bias, %3, %3, %3, %5, %2, %4, %5, %5, %6, %6), scope: __module.h # /home/cjenkins/dev/mat/venv/lib/python3.8/site-packages/torch/nn/modules/conv.py:453:0
  %12 : Tensor = aten::_convolution(%x, %self.g.weight, %self.g.bias, %3, %3, %3, %5, %2, %4, %5, %5, %6, %6), scope: __module.g # /home/cjenkins/dev/mat/venv/lib/python3.8/site-packages/torch/nn/modules/conv.py:453:0
  %13 : (Float(1, 4, 128, 128, strides=[65536, 16384, 128, 1], requires_grad=0, device=cpu), Float(1, 4, 128, 128, strides=[65536, 16384, 128, 1], requires_grad=0, device=cpu)) = prim::TupleConstruct(%11, %12)
  return (%13)
```

## Expected behavior

A return type of Tuple[Tensor, Tensor] should be treated as two separate outputs - not one.

## Environment

 - Torch-TensorRT Version: **v1.1.1** and **v1.2.0**
 - PyTorch Version (e.g. 1.0): **1.12.1**
 - CPU Architecture: AMD x86_64
 - OS (e.g., Linux): Linux / Ubuntu 22.0.4
 - How you installed PyTorch: source (w/ c++11 abi)
 - Build command you used (if compiling from source):  n/a
 - Are you using local sources or building from archives: local source
 - Python version: **3.8**
 - CUDA version: **11.7**
 - GPU models and configuration: RTX A6000

### LINKED ISSUE #1598: 🐛 [Bug] Differing behavior for fully-supported models with collection outputs, when toggling `require_full_compilation`
##  Bug Description

For model types which are fully-supported in Torch-TRT, but which use complex collections of outputs, for example `Tuple[Tensor, Tensor]`, or `Tuple[Tensor, Tuple[Tensor, Tensor]]`, the compilation behavior of Torch-TRT is different when the flag `require_full_compilation` is enabled versus not.

When `require_full_compilation=False`, the model compiles successfully, and the only operations executed in Torch are collections-processing operations. When `require_full_compilation=True`, however, the model fails with:

```python
RuntimeError: [Error thrown at core/conversion/conversion.cpp:230] Tuple type. Only a single tensor or a TensorList type is supported.
```

## To Reproduce

1. Define a fully convertible Torch model with forward function having the following form:
```python
def forward(self, input : Tensor) -> Tuple[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
```
2. Define the Torch-TRT Inputs and compilation settings using `require_full_compilation=True`, then compile the scripted model:
```python
compile_settings = {
    "inputs": [torch_tensorrt.Input((5, 5), dtype=torch.float)],
    "require_full_compilation": True,
}
trt_ts_module = torch_tensorrt.ts.compile(scripted_model, **compile_settings)
```

## Expected behavior

Model should compile with `require_full_compilation=True`, when containing nested Tuple collection.

## Environment

- Torch-TensorRT Version: 1.4.0.dev0+f43be5b6
- PyTorch Version: 1.14.0.dev20221114+cu116
- OS: Ubuntu 20.04
- Build command you used: `python setup.py develop`
- Python version: 3.8.13
- CUDA version: 11.6

## Additional context

This bug is related to, but not the same as #1595, as this bug relates to nested collection outputs and not inputs. The resolution to both of these bugs should resolve the overall issue of having fully-compiled models with complex collections of inputs and outputs.

This particular bug also appears when compiling the [HuggingFace BERT uncased model](https://huggingface.co/bert-base-uncased) with `require_full_compilation=True`.

gs-olive:
# Description
- Update graph-building in compiler to account for case where all operations are supported by Torch-TRT, but the output is a collection
- Enable 'psuedo-partitioning' for nearly-fully-compiled models for which the only non-supported aspect of the model is the format of the output (TRT cannot output complex collections)
- Define a small subset of operation schemas which are allowed despite the flag `require_full_compilation`. These operations are packing and unpacking of Tuples/Lists, and some are already used in cases of `require_full_compilation`
- Display warnings to users if any portion of the `pseudo-partitioning` is unexpected, for example the model being partitioned ends up in more than 3 segments (maximally - a Torch segment to preprocess collection inputs, a TRT segment to perform model logic, a Torch segment to post-process collection outputs) or if schemas falling outside of the collection subset are encountered in a Torch segment
- Add end-to-end test case with minimal reproducing example of a failing model, repaired with the changes to the compiler
- Add minor fix to lowering to remediate c++ compiler warning

This fix was designed to minimally alter the existing phases of model conversion and does not manually flatten/reconstruct complex collection outputs, but instead uses the existing partitioning infrastructure and engine-stitching paradigm to accomplish this.

Fixes #1598 
Fixes #1368 

## Type of change

- MVP for New feature

# Checklist:

### DIFF
--- a/core/compiler.cpp
+++ b/core/compiler.cpp
@@ -143,11 +143,6 @@ partitioning::GraphAndMapping BuildHybridGraph(
   auto convert_info = cfg.convert_info;
   auto partitioning_info = cfg.partitioning_info;
 
-  // Any nonzero block size is valid if full compilation to TRT is desired
-  if (expect_full_compilation) {
-    partitioning_info.min_block_size = 1;
-  }
-
   auto partitioning_ctx = partitioning::PartitioningCtx(block, partitioning_info);
   partitioning_ctx.input_types_map = first_use_types;
 
@@ -155,7 +150,7 @@ partitioning::GraphAndMapping BuildHybridGraph(
   // TODO: Combine this within partition call
   partitioning::populateInputIValues(&partitioning_ctx);
 
-  partitioning::partition(&partitioning_ctx);
+  partitioning::partition(&partitioning_ctx, expect_full_compilation);
 
   for (auto& partitioned_block : partitioning_ctx.partitioned_blocks) {
     partitioning::PartitionedGraph& segmented_blocks = partitioned_block.second;
@@ -197,9 +192,11 @@ partitioning::GraphAndMapping BuildHybridGraph(
         if (expect_full_compilation) {
           for (auto torch_node : seg_block.block()->nodes()) {
             if (partitioning::CollectionNodeKinds.find(torch_node->kind()) == partitioning::CollectionNodeKinds.end()) {
-              LOG_ERROR(
-                  "Full compilation specified but node " << torch_node->kind().toQualString()
-                                                         << " was executed in Torch.");
+              TORCHTRT_THROW_ERROR(
+                  "Full compilation specified but node "
+                  << *torch_node
+                  << " is set to run in PyTorch due to either lack of support in TensorRT or graph partitioning rules."
+                  << " Try recompiling with require_full_compilation=False.");
             }
           }
         }
@@ -209,10 +206,9 @@ partitioning::GraphAndMapping BuildHybridGraph(
     // If full compilation is expected, cannot have more than 2 Torch segments
     // (one for preprocessing inputs, one for post-processing outputs) and 1 TRT segment
     if (expect_full_compilation && !(num_torch_segments <= 2 && num_trt_segments == 1)) {
-      LOG_ERROR(
-          "Full compilation specified but number of torch segments was "
-          << num_torch_segments << " and number of trt segments was " << num_trt_segments
-          << ". Was expecting at most 2 Torch segments and 1 TRT segment.");
+      TORCHTRT_THROW_ERROR(
+          "Full compilation was requested but unable to convert all operations to TensorRT."
+          << " Try recompiling with require_full_compilation=False.");
     }
   }
 
@@ -224,7 +220,7 @@ ir::TypeMap MapInputsAndDetermineDTypes(
     std::shared_ptr<torch::jit::Graph>& g,
     ir::StaticParams& static_params,
     ir::CollectionTypeMap& first_use_type_map,
-    bool expect_full_compilation = false) {
+    bool requires_collection_handling = false) {
   cfg.convert_info.collection_input_spec_map =
       std::move(ir::associate_specs_with_collection_inputs(g, cfg.graph_inputs, static_params));
   cfg.partitioning_info.collection_input_spec_map =
@@ -259,7 +255,7 @@ ir::TypeMap MapInputsAndDetermineDTypes(
             "Cannot infer input type from calcuations in graph for input "
             << in->debugName() << ". Assuming it is Float32. If not, specify input type explicity");
         spec[i].dtype = at::kFloat;
-      } else if (spec[i].dtype_is_user_defined && (cfg.partitioning_info.enabled || expect_full_compilation)) {
+      } else if (spec[i].dtype_is_user_defined && (cfg.partitioning_info.enabled || requires_collection_handling)) {
         if (!est_type_opt[i]) {
           LOG_INFO("Cannot infer input tensor dtype in graph, compiler is going to use the user setting");
           std::stringstream ss;
@@ -330,6 +326,11 @@ std::string ConvertGraphToTRTEngine(const torch::jit::script::Module& mod, std::
   return engine;
 }
 
+bool userRequestedFallback(CompileSpec& cfg) {
+  return cfg.lower_info.forced_fallback_modules.size() != 0 ||
+      cfg.partitioning_info.forced_fallback_operators.size() != 0;
+}
+
 torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg) {
   torch::jit::Module new_mod(mod._ivalue()->name() + "_trt");
 
@@ -352,10 +353,13 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg) 
       // whether full compilation can be expected
       auto isBlockConvertible = conversion::VerifyConverterSupportForBlock(g->block(), true);
       auto outputIsCollection = conversion::OutputIsCollection(g->block());
-      auto nearly_full_compilation = (isBlockConvertible && outputIsCollection);
+      auto requires_collection_handling = (isBlockConvertible && outputIsCollection);
+
+      // Determine whether user specifications necessitate partitioning
+      auto isFallbackRequested = userRequestedFallback(cfg);
 
       // Extract map of IValue to DType
-      auto type_map = MapInputsAndDetermineDTypes(cfg, g, static_params, first_use_types, nearly_full_compilation);
+      auto type_map = MapInputsAndDetermineDTypes(cfg, g, static_params, first_use_types, requires_collection_handling);
 
       // Check whether any of the input types are Long
       bool user_requested_long = false;
@@ -369,21 +373,26 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg) 
         user_requested_long &= (casts_inserted > 0);
       }
 
-      if (cfg.partitioning_info.enabled && !user_requested_long &&
-          (cfg.lower_info.forced_fallback_modules.size() == 0 &&
-           cfg.partitioning_info.forced_fallback_operators.size() == 0 && isBlockConvertible) &&
-          !outputIsCollection) {
+      // Partitioning is required if:
+      // 1. User requested some modules/operators fallback
+      // 2. The block (graph) cannot be converted due to operator coverage
+      // 3. The output of the graph is a collection
+      // 4. The user requested a non-TRT data type input
+      auto isPartitioningRequired =
+          (isFallbackRequested || !isBlockConvertible || outputIsCollection || user_requested_long);
+
+      // The user did not require full compilation, but the model can be fully compiled
+      if (cfg.partitioning_info.enabled && !isPartitioningRequired) {
         LOG_INFO("Skipping partitioning since model is fully supported");
       }
 
-      if ((cfg.partitioning_info.enabled &&
-           (!(cfg.lower_info.forced_fallback_modules.size() == 0 &&
-              cfg.partitioning_info.forced_fallback_operators.size() == 0 && isBlockConvertible) ||
-            outputIsCollection || user_requested_long)) ||
-          nearly_full_compilation) {
+      // The user did not require full compilation, and the model can be fully compiled
+      // or, the user required full compilation but the I/O of the graph use collections
+      if ((cfg.partitioning_info.enabled && isPartitioningRequired) || requires_collection_handling) {
         // If the model is fully-compilable and the user has specified full compilation, run partitioning
         // to generate collection-processing code in Torch
-        auto expect_full_compilation = (nearly_full_compilation && !cfg.partitioning_info.enabled);
+        auto expect_full_compilation = (requires_collection_handling && !cfg.partitioning_info.enabled);
+
         auto graph_and_mapping =
             BuildHybridGraph(new_mod, g->block(), cfg, static_params, first_use_types, expect_full_compilation);
         new_g = graph_and_mapping.first;
--- a/core/compiler.cpp
+++ b/core/compiler.cpp
@@ -138,9 +138,15 @@ partitioning::GraphAndMapping BuildHybridGraph(
     torch::jit::Block* block,
     CompileSpec cfg,
     ir::StaticParams static_params,
-    ir::CollectionTypeMap first_use_types) {
+    ir::CollectionTypeMap first_use_types,
+    bool expect_full_compilation = false) {
   auto convert_info = cfg.convert_info;
   auto partitioning_info = cfg.partitioning_info;
+
+  // Any nonzero block size is valid if full compilation to TRT is desired
+  if (expect_full_compilation) {
+    partitioning_info.min_block_size = 1;
+  }
 
   auto partitioning_ctx = partitioning::PartitioningCtx(block, partitioning_info);
   partitioning_ctx.input_types_map = first_use_types;
@@ -153,6 +159,8 @@ partitioning::GraphAndMapping BuildHybridGraph(
 
   for (auto& partitioned_block : partitioning_ctx.partitioned_blocks) {
     partitioning::PartitionedGraph& segmented_blocks = partitioned_block.second;
+    int num_torch_segments = 0;
+    int num_trt_segments = 0;
 
     for (auto& seg_block : segmented_blocks) {
       LOG_INFO("Block segment:" << seg_block);
@@ -160,6 +168,7 @@ partitioning::GraphAndMapping BuildHybridGraph(
       trt_engine_id << reinterpret_cast<const int*>(&seg_block);
 
       if (seg_block.target() == partitioning::SegmentedBlock::kTensorRT) {
+        num_trt_segments++;
         auto inputs = seg_block.construct_inputs_spec();
         // update the input ranges for each segments
         convert_info.inputs = ir::associate_specs_with_inputs(seg_block.g(), inputs, static_params);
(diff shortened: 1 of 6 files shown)


core/compiler.cpp
```diff
@@ -180,8 +189,32 @@ partitioning::GraphAndMapping BuildHybridGraph(
...
+
+        // If full compilation is expected, ensure that all operators in Torch blocks are
+        // for collections processing
+        if (expect_full_compilation) {
+          for (auto torch_node : seg_block.block()->nodes()) {
+            if (partitioning::CollectionSchemas.find(torch_node->kind().toQualString()) ==
+                partitioning::CollectionSchemas.end()) {
+              LOG_WARNING(
+                  "Full compilation specified but node " << torch_node->kind().toQualString()
+                                                         << " was executed in Torch.");
```
gs-olive, line 203: Could potentially be a `LOG_ERROR` to increase severity or `TORCHTRT_CHECK` to completely halt compilation, since entering this `if` block indicates that the user specified `require_full_compilation=True`, but an operation was set to run in Torch which is not for the purpose of creating the collection.

core/compiler.cpp
```diff
@@ -180,8 +189,32 @@ partitioning::GraphAndMapping BuildHybridGraph(
...
       }
     }
+
+    // If full compilation is expected, cannot have more than 2 Torch segments
+    // (one for preprocessing inputs, one for post-processing outputs) and 1 TRT segment
+    if (expect_full_compilation && !(num_torch_segments <= 2 && num_trt_segments == 1)) {
+      LOG_WARNING(
+          "Full compilation specified but number of torch segments was "
+          << num_torch_segments << " and number of trt segments was " << num_trt_segments
+          << ". Was expecting at most 2 Torch segments and 1 TRT segment.");
```
gs-olive, line 216: Similarly to the above comment, this could alternatively be made to halt compilation, since it indicates the model is no longer being "fully-compiled" as there are multiple TRT engine blocks.

core/compiler.cpp
```diff
@@ -138,10 +138,16 @@ partitioning::GraphAndMapping BuildHybridGraph(
...
     ir::StaticParams static_params,
-    ir::CollectionTypeMap first_use_types) {
+    ir::CollectionTypeMap first_use_types,
+    bool expect_full_compilation = false) {
   auto convert_info = cfg.convert_info;
   auto partitioning_info = cfg.partitioning_info;
 
+  // Any nonzero block size is valid if full compilation to TRT is desired
+  if (expect_full_compilation) {
+    partitioning_info.min_block_size = 1;
```
narendasan, line 148: I don't quite understand this, if you expect full compilation, shouldnt the min block size be the full graph? min_block_size 1 would mean any singular layer can be by itself in trt

core/compiler.cpp
narendasan: I dont think we need this warning, it could probably be debug if all that is happening is collection packing/unpacking. But I think our criteria is more restrictive. It there should not really be any intermediate packing unpacking imo. Maybe there are some counter examples?

core/compiler.cpp
narendasan: This should be an error if it does not meet our criteria for full compilation

core/partitioning/partitioning.h
```diff
@@ -18,6 +18,19 @@ typedef std::unordered_map<const torch::jit::Value*, torch::jit::IValue> Example
 typedef std::pair<std::shared_ptr<torch::jit::Graph>, std::unordered_map<torch::jit::Value*, torch::jit::Value*>>
     GraphAndMapping;
 
+// Set of schemas allowed to be executed in Torch, even with require_full_compilation=true,
+// as necessary for returning collections of Tensors or other complex constructs, and for
+// processing inputs to TRT engines
+const std::unordered_set<c10::Symbol> CollectionSchemas = {
```
narendasan, line 24: Call this `CollectionNodeKinds` or there might be a different name for the op kind in JIT. Schema implies the full `aten::foo(Tensor x) -> (Tensor)`

narendasan (commented):
Looks mostly fine, just some ux and dev stuff

core/compiler.cpp
gs-olive: The purpose of this was to override the default of `min_block_size=3` in partitioning. I agree that the min block size should be the full graph, but if the number of non-collection operations in the graph is less than 3, this partitioning phase would not respect full compilation (it would cause those operations to be executed in Torch). This condition is a way around that issue - by allowing any size of TRT block, we allow any size of TRT graph to be valid in partitioning.

core/compiler.cpp
gs-olive: Since this message should only display if a non-collection-based operation was determined to run in Torch (which should not occur - this violates the `require_full_compilation=True` condition), I have made this log command a `LOG_ERROR`, similarly to the case below.

narendasan: If its an error then it should use THROW_ERROR and quit instead of continuing

core/compiler.cpp
narendasan: min_block_size only looks at the number of contiguous TRT operations. If this is an issue then it would probably effect non full compilation cases as well. There should be explicit handling of graphs that are smaller than `min_block_size` to just lower that threshold to 1, but that should probably be in partitioning and not specific to this use case.

gs-olive: I see - I agree that this should probably be in partitioning. I am wondering what your thoughts are on adding `expect_full_compilation` as an argument to `partitioning::partition`?

I added a commit that has a test case to elicit the specific behavior that I want to catch with this condition. Additionally, I have moved the condition into a narrower scope, closer to where it is used so that the code is more understandable (though still in `compiler.cpp`, for now).

core/compiler.cpp
```diff
@@ -315,8 +345,14 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg)
       // Infer the type of an input from the weights of the calculation
       auto first_use_types = ir::get_block_first_calc_dtypes_opt_collection(g->block());
 
+      // Determine if the block is convertible/has collection output, and based on the result,
+      // whether full compilation can be expected
+      auto isBlockConvertible = conversion::VerifyConverterSupportForBlock(g->block(), true);
+      auto outputIsCollection = conversion::OutputIsCollection(g->block());
+      auto nearly_full_compilation = (isBlockConvertible && outputIsCollection);
```
narendasan, line 352: rename this to `requires_collection_handling`

core/partitioning/partitioning.cpp
```diff
@@ -564,7 +564,21 @@ void populateInputIValues(PartitioningCtx* ctx) {
...
 
-void partition(PartitioningCtx* ctx) {
+void partition(PartitioningCtx* ctx, bool expect_full_compilation) {
+  // If full compilation is expected, overwrite minimum block size
+  // Any nonzero block size is valid if full compilation to TRT is desired
+  // Override the default min_block_size to ensure all TRT-supported operations are
+  // executed in TRT, regardless of the size of the graph
+  if (expect_full_compilation) {
+    // If minimum block size is different from the default, the user must have specified it
+    if (ctx->settings.min_block_size != 3) {
```
narendasan, line 574: Create an issue to centralize defaults somewhere in the core

core/compiler.cpp
```diff
@@ -180,8 +184,33 @@ partitioning::GraphAndMapping BuildHybridGraph(
...
+
+    // If full compilation is expected, cannot have more than 2 Torch segments
+    // (one for preprocessing inputs, one for post-processing outputs) and 1 TRT segment
+    if (expect_full_compilation && !(num_torch_segments <= 2 && num_trt_segments == 1)) {
+      TORCHTRT_THROW_ERROR(
+          "Full compilation specified but number of torch segments was "
+          << num_torch_segments << " and number of trt segments was " << num_trt_segments
+          << ". Was expecting at most 2 Torch segments and 1 TRT segment."
+          << " Try recompiling with require_full_compilation=False.");
+    }
```
narendasan, line 213: This error should be something like: `Full compilation was requested but unable to convert all operations to TensorRT. Try recompile with require_full_compilation=False`

narendasan (changes requested):

core/compiler.cpp
```diff
@@ -180,8 +184,33 @@ partitioning::GraphAndMapping BuildHybridGraph(
...
+        num_torch_segments++;
+
+        // If full compilation is expected, ensure that all operators in Torch blocks are
+        // for collections processing
+        if (expect_full_compilation) {
+          for (auto torch_node : seg_block.block()->nodes()) {
+            if (partitioning::CollectionNodeKinds.find(torch_node->kind()) == partitioning::CollectionNodeKinds.end()) {
+              TORCHTRT_THROW_ERROR(
+                  "Full compilation specified but node " << torch_node->kind().toQualString()
+                                                         << " was executed in Torch."
```
narendasan, line 197: change ` was executed in Torch` to ` is set to run in PyTorch due to either lack of support in TensorRT or graph partitioning rules `

narendasan (changes requested):

narendasan (approved):
LGTM, @bowang007 can you take a look?

bowang007 (approved):
LGTM

core/compiler.cpp
```diff
@@ -180,8 +184,32 @@ partitioning::GraphAndMapping BuildHybridGraph(
...
+                  << " Try recompiling with require_full_compilation=False.");
+            }
+          }
+        }
       }
     }
+
+    // If full compilation is expected, cannot have more than 2 Torch segments
+    // (one for preprocessing inputs, one for post-processing outputs) and 1 TRT segment
+    if (expect_full_compilation && !(num_torch_segments <= 2 && num_trt_segments == 1)) {
```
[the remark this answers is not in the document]
gs-olive: There should not be a case where multiple Torch segments appear for inputs/outputs, since `merge_adjacent_segments_of_same_type` addresses this case, as you had mentioned. Since the tensors in question are inputs, it should not arise that `segment.do_not_merge()` is True, since the only approved operators falling into these segments are for collection construction, and only the `prim::If` or `prim::Loop` operators can induce a non-merge situation.

core/partitioning/partitioning.cpp
gs-olive: No, the user would not get a warning message in that case. We currently don't have a way of knowing whether the user inputs a value or not, since the defaults are not centralized. There is an issue #1644 to address this, but as of now, your statement is correct. Additionally, it is worth noting that prior to this PR, if a user specified `min_block_size` and `require_compilation=True`, we would still ignore the `min_block_size`, but without warning.

core/compiler.cpp
```diff
@@ -330,20 +365,24 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg)
...
-      if (cfg.partitioning_info.enabled &&
-          (!(cfg.lower_info.forced_fallback_modules.size() == 0 &&
-             cfg.partitioning_info.forced_fallback_operators.size() == 0 && isBlockConvertible) ||
-           outputIsCollection || user_requested_long)) {
-        auto graph_and_mapping = BuildHybridGraph(new_mod, g->block(), cfg, static_params, first_use_types);
+      if ((cfg.partitioning_info.enabled &&
+           (!(cfg.lower_info.forced_fallback_modules.size() == 0 &&
+              cfg.partitioning_info.forced_fallback_operators.size() == 0 && isBlockConvertible) ||
+            outputIsCollection || user_requested_long)) ||
+          requires_collection_handling) {
```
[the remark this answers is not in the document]
gs-olive: I've updated this statement to make the conditions clearer, by removing the `!` and distributing it over the conditionals inside. Other than this, the statement cannot be reduced any further since the `requires_collection_handling` boolean is independent of `cfg.partitioning_info.enabled`.

core/compiler.cpp
```diff
@@ -330,20 +365,24 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg)
...
-      if (cfg.partitioning_info.enabled &&
-          (!(cfg.lower_info.forced_fallback_modules.size() == 0 &&
-             cfg.partitioning_info.forced_fallback_operators.size() == 0 && isBlockConvertible) ||
-           outputIsCollection || user_requested_long)) {
-        auto graph_and_mapping = BuildHybridGraph(new_mod, g->block(), cfg, static_params, first_use_types);
+      if ((cfg.partitioning_info.enabled &&
+           (cfg.lower_info.forced_fallback_modules.size() != 0 ||
+            cfg.partitioning_info.forced_fallback_operators.size() != 0 || !isBlockConvertible || outputIsCollection ||
+            user_requested_long)) ||
+          requires_collection_handling) {
```
gs-olive, line 379: Note the updates to the conditional logic to make it $$\(cfg.partitioning\\_info.enabled \wedge (x \vee y \vee z ... )\) \vee requires\\_collection\\_handling$$

core/compiler.cpp
```diff
@@ -330,20 +373,28 @@ torch::jit::Module CompileGraph(const torch::jit::Module& mod, CompileSpec cfg)
...
-          (cfg.lower_info.forced_fallback_modules.size() == 0 &&
-           cfg.partitioning_info.forced_fallback_operators.size() == 0 && isBlockConvertible) &&
-          !outputIsCollection) {
+      // Partitioning is required if:
+      // 1. User requested some modules/operators fallback
+      // 2. The block (graph) cannot be converted due to operator coverage
+      // 3. The output of the graph is a collection
+      // 4. The user requested a non-TRT data type input
+      auto isPartitioningRequired =
+          (isFallbackRequested || !isBlockConvertible || outputIsCollection || user_requested_long);
```
gs-olive, line 382: Coalesced partitioning logic for readability

core/compiler.cpp
```diff
@@ -297,6 +326,11 @@ std::string ConvertGraphToTRTEngine(const torch::jit::script::Module& mod, std::
   return engine;
 }
 
+bool userRequestedFallback(CompileSpec& cfg) {
+  return cfg.lower_info.forced_fallback_modules.size() != 0 ||
+      cfg.partitioning_info.forced_fallback_operators.size() != 0;
+}
```
gs-olive, line 332: Added helper function to determine if the user's input specifications require fallback

narendasan (approved):
LGTM

bowang007 (approved):
Looks great, thanks!

Issue #1613, opened 2023-01-24
🐛 [Bug] Could not Find IValue / Node Inputs Cannot be Evaluated


gs-olive:
**Summary:** Node inputs attempting to be evaluated at conversion time / IValue not found for a particular input during partitioning + conversion phases.

See NVBugs for more details.

PR #1617, opened 2023-01-25
fix: Issue in non-Tensor Input Resolution

### LINKED ISSUE #1612: 🐛 [Bug] Node Inputs Cannot be Evaluated at Conversion Time
See NVBugs

gs-olive:
# Description

- <s>In certain TensorRT-executed blocks, multiple non-Tensor inputs originate from the same parent Tensor in another Torch block, so a single pass of resolveTRTNonTensorInputs does not resolve all non-Tensor inputs, causing failures at compile-time</s>
- Issue was determined to trace to the use of `prim::Loop` in the TensorRT block and is fixed in PR #1691 
- Add function to `SegmentedBlock` to get the ID of a block, so re-inserted blocks retain the same ID as the previous block being replaced
- Add regression test case to elicit behavior

### Before Non-Tensor Input Resolution
```python
DEBUG: [Torch-TensorRT - Debug Build] - Finalizing in progress TensorRT block
DEBUG: [Torch-TensorRT - Debug Build] - Segment Block @2:
    Target: TensorRT

    Graph: graph(%1 : int,
      %9 : int[],
      %12 : Tensor):
```
### After 1 Round of Non-Tensor Input Resolution
```python
GRAPH: [Torch-TensorRT - Debug Build] - Running shape analysis on block Segment Block @2:
    Target: TensorRT

    Graph: graph(%1 : int[],
      %3 : Tensor):
```

### After 2 Rounds of Non-Tensor Input Resolution
```python
INFO: [Torch-TensorRT - Debug Build] - Segment Block @2:
    Target: TensorRT

    Graph: graph(%1 : Tensor):
```

Fixes #1612 
Related to: #1613 

## Type of change

# Checklist:

### DIFF
--- a/core/partitioning/partitioning.cpp
+++ b/core/partitioning/partitioning.cpp
@@ -273,21 +273,34 @@ void resolveTRTNonTensorInputs(PartitioningCtx* ctx, torch::jit::Block* block) {
   PartitionedGraph& cur_partitioned_block = ctx->partitioned_blocks[block];
   for (size_t i = 0; i < cur_partitioned_block.size(); ++i) {
     if (cur_partitioned_block[i].target() == SegmentedBlock::kTensorRT) {
-      std::vector<torch::jit::Value*> inputs_to_resolve;
-      for (auto input : cur_partitioned_block[i].raw_inputs()) {
-        if (!isTensor(input)) {
-          inputs_to_resolve.push_back(input);
+      // Store current number of inputs to block, and number of inputs after one round
+      // of reductions
+      auto current_input_size = cur_partitioned_block[i].inputs().size();
+      auto reduced_input_size = current_input_size;
+
+      // While the number of inputs to the block is strictly decreasing, continue to resolve inputs
+      do {
+        current_input_size = reduced_input_size;
+        std::vector<torch::jit::Value*> inputs_to_resolve;
+        for (auto input : cur_partitioned_block[i].raw_inputs()) {
+          if (!isTensor(input)) {
+            inputs_to_resolve.push_back(input);
+          }
         }
-      }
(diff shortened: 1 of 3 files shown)


peri044:
Expected this PR to make it into 23.03 release

bowang007:
This does not make sense in fact. 
Firstly, even if 2 non-Tensor inputs come from a same node, that node will be added into the segmented block. There is no need to do it for multiple passes. 
Secondly, I guess why it fails for a single pass is because that there is a Loop block in the TensorRT segment. This could introduce undefined behaviors. The real issue I think here is the Loop block, it shouldn't be in TensorRT segment at all. In our last code refactoring for partitioning, we set all nodes aggressively to be run in TensorRT, this introduces some issue like LOOP nodes default to be run in TensorRT, this could result in some issues like this.

core/partitioning/partitioning.cpp
```diff
@@ -286,7 +286,8 @@ void resolveTRTNonTensorInputs(PartitioningCtx* ctx, torch::jit::Block* block) {
             dependency_nodes.end(),
             cur_partitioned_block[i].raw_nodes().begin(),
             cur_partitioned_block[i].raw_nodes().end());
-        cur_partitioned_block[i] = SegmentedBlock(SegmentedBlock::kTensorRT, dependency_nodes);
+        cur_partitioned_block[i] =
+            SegmentedBlock(cur_partitioned_block[i].get_id(), SegmentedBlock::kTensorRT, dependency_nodes);
```
gs-olive, line 290: Added to ensure the block IDs of segmented blocks stay in order despite resolution of non-Tensor inputs.

bowang007 (approved):
LGTM, please squash your commits then merge.

Issue #1616, opened 2023-01-24
🐛 [Bug] Model Decoder Failure


gs-olive:
**Summary:** Model failing on `trt::const` with NoneType input, and other errors.

See NVBugs for more details.

gs-olive:
Applying #1619 and #1623 together resolve the compilation error on my machine

Christina-Young-NVIDIA:
2 open PRs need review.  George to find someone to review and close during sprint starting 1/30.

PR #1619, opened 2023-01-26
fix: Bugfix in Linear-to-AddMM Fusion Lowering Pass

gs-olive:
# Description

- Fix 2 bugs in linear-to-addmm lowering pass:
  - Lowering pass did not explore nested sub-blocks of a node, of the sort contained in `prim::If` when `bias=None`. As an example, these would be ignored:
```python
      %4 : Tensor = prim::If(%true)
        block0():
          %res = aten::linear(%input, %weight, %biasNone)
          -> (%res)
        block1():
          %res = aten::linear(%input, %weight, %biasNone)
          -> (%res)
```
  - Lowering pass did not insert fused linear code inside sub-blocks of `prim::If` even when the original function call occurred within such a block. As an example, the following translates to the subsequent graph, which contains an invalid computation:
```python
      %4 : Tensor = prim::If(%true)
        block0():
          %res = aten::linear(%input, %weight, %bias)
          -> (%res)
        block1():
          %res = aten::linear(%input, %invalid_weight, %bias)
          -> (%res)

=============== TRANSLATES TO ===============

  %13 : int = prim::Constant[value=1]()
  %14 : Tensor = aten::t(%weight)
  %15 : Tensor = aten::matmul(%input, %14)
  %16 : Tensor = trt::const(%bias)
  %17 : Tensor = aten::add(%16, %15, %13)
  %3 : bool = prim::Constant[value=1]()
  %4 : Tensor = aten::t(%weight)
  %8 : int = prim::Constant[value=1]()
  %9 : Tensor = aten::t(%4)
  %10 : Tensor = aten::matmul(%input, %9)
  %11 : Tensor = trt::const(%bias)
  %12 : Tensor = aten::add(%11, %10, %8)
  %5 : Tensor = prim::If(%3)
    block0():
      -> (%17)
    block1():
      -> (%12)

=============== LEADING TO ===============

%10 : Tensor = aten::matmul(%input, %9): last dimension of input0 = 7 and second to last dimension of input1 = 8 but must match.
```
  - The latter causes issues when the control-flow switches between two versions of `aten::linear`, only one of which is a valid operation. Thus, evaluating both branches can cause compilation to crash, as invalid Tensor shapes can be encountered
- Update implementation to run recursively through all nested blocks within all nodes
- Update implementation to remove the use of `RegisterRewritePattern` paradigm for Tensor biases, as the rewrite does not always place the subgraph in the desired location
- Add regression test cases to isolate both bugs

Addresses 1st Bug in #1616 

## Type of change

# Checklist:

### DIFF
--- a/core/lowering/passes/linear_to_addmm.cpp
+++ b/core/lowering/passes/linear_to_addmm.cpp
@@ -3,6 +3,7 @@
 #include "core/util/prelude.h"
 #include "torch/csrc/jit/api/function_impl.h"
 #include "torch/csrc/jit/ir/alias_analysis.h"
+#include "torch/csrc/jit/ir/irparser.h"
 #include "torch/csrc/jit/jit_log.h"
 #include "torch/csrc/jit/passes/constant_propagation.h"
 #include "torch/csrc/jit/passes/dead_code_elimination.h"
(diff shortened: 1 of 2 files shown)


narendasan (approved):
LGTM

PR #1647, opened 2023-02-03
feat(//core/conversion): Add support for aten::size with dynamic shaped models for Torchscript backend.

peri044:
# Description

https://github.com/pytorch/TensorRT/discussions/1577

Fixes issues listed in the above discussion 

## Type of change

# Checklist:


peri044:
cc: @mfeliz-cruise

core/conversion/evaluators/aten.cpp
```diff
@@ -314,22 +354,30 @@ auto aten_registrations TORCHTRT_UNUSED =
                  {"aten::size(Tensor self) -> (int[])", "aten::size.int(Tensor self, int dim) -> (int)"})})
         .evaluator(
             {c10::Symbol::fromQualString("aten::__getitem__"),
-             [](const torch::jit::Node* n, kwargs& args) -> c10::optional<torch::jit::IValue> {
-               auto list = args.at(n->input(0)).IValue()->to<c10::List<c10::IValue>>();
+             [](ConversionCtx* ctx, const torch::jit::Node* n, kwargs& args) -> c10::optional<torch::jit::IValue> {
+               auto list_input = args.at(n->input(0));
                auto idx = args.at(n->input(1)).unwrapToInt();
+               if (list_input.isIValue()){
```
mfeliz-cruise, line 360: It looks like this implements one layer of look back, are there cases where this would fail in a new way or do those fallback to list_input.isIValue() to preserve the existing functionality?

core/conversion/converters/impl/shuffle.cpp
```diff
@@ -67,33 +67,33 @@ static auto shuffle_registrations TORCHTRT_UNUSED =
...
-               auto out_tensor = ctx->AssociateValueAndTensor(n->outputs()[0], shuffle->getOutput(0));
-               LOG_DEBUG("Output tensor shape: " << out_tensor->getDimensions());
-
-               return true;
+                auto in = args[0].ITensorOrFreeze(ctx);
+                auto in_shape = util::toVec(in->getDimensions());
+                std::vector<int64_t> new_shape;
+                nvinfer1::ITensor* shape_tensor;
+                if (ctx->input_is_dynamic) {
+                  auto new_shape = args[1].unwrapToITensorList();
```
mfeliz-cruise, line 75: Should this check whether args[1] is an ITensorList? If we have some intermediate math on the shape input that is handled by evaluators could we end up in a case where input_is_dynamic but args[1] is an IntList input?

mfeliz-cruise:
This looks great @peri044. I left some questions about edge cases, but generally I fully support this change to add the ConversionCtx to the evaluator inputs. This opens up the possibility for us to implement alternative conversion paths for dynamic shapes for any evaluator. Thank you!

core/conversion/converters/BUILD
```diff
@@ -62,6 +62,7 @@ cc_library(
         "impl/constant_pad.cpp",
         "impl/conv_deconv.cpp",
         "impl/cumsum.cpp",
+        "impl/dual_ops.cpp",
```
mfeliz-cruise, line 65: Was this file removed? I don't see it in the PR.

core/conversion/evaluators/aten.cpp
andi4191, line 360: @peri044: Can we have test cases with such edge case scenarios to verify this?

core/conversion/converters/impl/shuffle.cpp
```diff
@@ -67,33 +67,33 @@ static auto shuffle_registrations TORCHTRT_UNUSED =
...
-
-               auto out_tensor = ctx->AssociateValueAndTensor(n->outputs()[0], shuffle->getOutput(0));
-               LOG_DEBUG("Output tensor shape: " << out_tensor->getDimensions());
-
-               return true;
+                auto in = args[0].ITensorOrFreeze(ctx);
+                auto in_shape = util::toVec(in->getDimensions());
+                std::vector<int64_t> new_shape;
+                nvinfer1::ITensor* shape_tensor;
+                if (ctx->input_is_dynamic) {
```
andi4191, line 74: Does this include the case where the non-batch dimension is -1 wildcard? For instance, [ batch_size, -1, spatial_x, spatial_y] in an NCHW format.

core/conversion/converters/BUILD
peri044: Yeah resolved this as a part of clean up

core/conversion/converters/impl/shuffle.cpp
peri044: Yes, it is handled. Eg: The following works
```
class ViewTest(torch.nn.Module):
    def __init__(self):
        super(ViewTest, self).__init__()

    def forward(self, x):
        return x.view(x.size(0), -1, 4)

model = ViewTest().eval().cuda()

    trt_mod = torch_tensorrt.compile(model, inputs=[torch_tensorrt.Input(min_shape=[1, 16, 16],
                                                                         opt_shape=[16, 16, 16],
                                                                         max_shape=[32, 16, 16], dtype=torch.float32)], truncate_long_and_double=True)
```

core/conversion/evaluators/aten.cpp
peri044: I'm not sure what other edge case scenarios might occur. the input can be a IValue for normal cases, or an ITensor (when there is a preceeding aten::size which emits ITensor). Do you have anything other than these in mind ?

peri044: By look back, I'm imagining a case where we have something like list [idx1] [idx2]. I don't have an answer here (how both ITensor and IValue interplay) as it depends on how it is represented in the torchscript graph

mfeliz-cruise: Worth adding a test?

core/conversion/converters/impl/shuffle.cpp
peri044: Yeah WIP. Currently, at multiple places, we unwrap without detecting the types. Investigating.

peri044: Added type checking now.

core/conversion/evaluators/aten.cpp
peri044: Added few utils which better detect IValue and new test cases which use them.

core/conversion/var/Var.cpp
```diff
@@ -21,6 +21,28 @@ Var::Var(nvinfer1::ITensor* p) : type_(Type::kITensor) {
   ptr_.tensor = p;
 }
 
+Var::IValueType Var::determineIValueType(torch::jit::IValue* p) {
```
narendasan, line 24: @peri044 you can implement a macro to do this similar to what we have in var_inl.h

narendasan: Vs duplicating PyTorch's type system

peri044: Made changes now to rely on pytorch's type system

PR #1649, opened 2023-02-04
fix: fix the inappropriate lowering pass of aten::to

### LINKED ISSUE #1530: aten::to type mismatch.
See NVBugs

bowang007:
# Description
When there is a aten::to operation in the model as: 
```
sizes = sizes.to(device=boxes.device)
```
like what happens here: https://github.com/facebookresearch/detectron2/blob/58e472e076a5d861fdcf773d9254a3664e045bf8/detectron2/modeling/poolers.py#L65, What this operation does is setting the device for sizes variable, since all the other parameters for aten::to op is not explicitly set, this results in an aten::to op in graph as:
```
 %sizes0.2 : Tensor = aten::to(%1954, %80, %80, %1955, %80, %6, %6, %80)
```
where:
```
%80 : NoneType = prim::Constant()
```
The NoneType is valid for aten::to candidate form:
```
aten::to.dtype_layout(Tensor(a) self, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None, bool non_blocking=False, bool copy=False, MemoryFormat? memory_format=None) -> Tensor(a)
```
but not valid for the other forms, and in our lowering pass we convert the form above into other forms in this line: https://github.com/pytorch/TensorRT/blob/a3436504714dc40885e80f70c93478d263eb1df1/core/lowering/passes/reduce_to.cpp#L35
That's why this error message was printed : 
```
RuntimeError: 0 INTERNAL ASSERT FAILED at "../torch/csrc/jit/ir/alias_analysis.cpp":608, please report a bug to PyTorch. We don't have an op for aten::to but it isn't a special case.  Argument types: Tensor, Device, NoneType, bool, bool, NoneType, 
```

Fixes #1530 

## Type of change

# Checklist:

### DIFF
--- a/core/lowering/passes/reduce_to.cpp
+++ b/core/lowering/passes/reduce_to.cpp
@@ -8,16 +8,6 @@ namespace lowering {
 namespace passes {
 
 void ReduceToOperation(std::shared_ptr<torch::jit::Graph>& graph) {
-  std::string to_dtype_layout_pattern = R"IR(
-        graph(%x, %dtype, %layout, %device, %pm, %nb, %copy, %format):
-            %out : Tensor = aten::to(%x, %dtype, %layout, %device, %pm, %nb, %copy, %format)
-            return (%out))IR";
-
-  std::string to_dtype_multi_input_pattern = R"IR(
-        graph(%x, %dtype, %layout, %device, %pm, %nb, %copy, %format):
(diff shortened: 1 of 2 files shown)


narendasan (approved):
LGTM

Issue #1629, opened 2023-01-30
✨broadcast utility in converter utilities to broadcast two tensors in masked_fill and make the dimensions equal to torch output dimensions


apbose:
**Feature context**

In the implementation of _aten::masked_fill(Tensor self, Tensor mask, Scalar value)_ in _core/conversion/converters/impl/select.cpp_ the converter implementation is such that the masked rank is padded to match the rank of self if the rank of mask is less than self Tensor. This creates an issue when the broadcastable is checked in _util::broadcastable_, since they might be broadcastable according to the _masked_fill_ torch rules, but the TensorRT select layer will restrict since they would not be broadcastable.

Reference to  [Issue 1573](https://github.com/pytorch/TensorRT/issues/1573) after [PR 1609 ](https://github.com/pytorch/TensorRT/pull/1609)

**Workaround**

At present padding both the self tensor, when mask tensor has dimension greater than the self tensor causes issue. eg: mask tensor is of shape [1,1,1,1488] and the self tensor is [1,512, 1488] then padding self to mask with additional 1s, causes error down the line in aten::matmul when last dimension of input0 is 1 (should be 512 )and second last dimension of input 1 is 512. 

```
if(mask->getDimensions().nbDims < self->getDimensions().nbDims) {
        mask = addPadding(ctx, n, mask, self->getDimensions().nbDims, false, true);
 } else {
         self = addPadding(ctx, n, self, mask->getDimensions().nbDims, false, true);
 }

```
Hence we would want a broadcasting utility in _converter_utils.cpp_ which would broadcast the self and mask tensor according to torch rules. It would remove the unnecessary 1s from the dimensions.

Issue #1573, opened 2023-01-04
Problem when using Citrinet example notebook


hamjam:
I have similar problem with https://github.com/pytorch/TensorRT/issues/763  when trying to adopt [Citrinet-Example.ipynb](https://github.com/pytorch/TensorRT/blob/master/notebooks/CitriNet-example.ipynb). 
My model is a fine-tuned [Citrinet512](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/stt_en_citrinet_512) from [NeMo](https://github.com/NVIDIA/NeMo).
By the way, my **pytorch_tensorrt version is 1.3.0**.
Here is my code:
```python3
import torch
import torch.nn as nn
import torch_tensorrt as torchtrt
import argparse

precisions = [torch.float, torch.half]
batch_sizes = [1,8,32,128]
model = torch.jit.load(model_path+f"{variant}.ts")
model.eval()
for precision in precisions:
    for batch_size in batch_sizes:
        compile_settings = {
            "inputs": [torchtrt.Input(shape=[batch_size, 80, 1488],  dtype=torch.int), torchtrt.Input(shape=[1, batch_size],  dtype=torch.int)],
            "enabled_precisions": {precision},
            "workspace_size": 2000000000,
            "truncate_long_and_double": True,
        }
        print(f"Generating Torchscript-TensorRT module for batchsize {batch_size} precision {precision}")
        trt_ts_module = torchtrt.compile(model, **compile_settings)
        torch.jit.save(trt_ts_module, f"{variant}_bs{batch_size}_{precision}.torch-tensorrt")
 ```
 Output:
 ```
 RuntimeError                              Traceback (most recent call last)
Input In [6], in <module>
     14 compile_settings = {
     15     "inputs": [torchtrt.Input(shape=[batch_size, 80, 1488],  dtype=torch.int), torchtrt.Input(shape=[1, batch_size],  dtype=torch.int)],
     16     "enabled_precisions": {precision},
     17     "workspace_size": 2000000000,
     18     "truncate_long_and_double": True,
     19 }
     20 print(f"Generating Torchscript-TensorRT module for batchsize {batch_size} precision {precision}")
---> 21 trt_ts_module = torchtrt.compile(model, **compile_settings)
     22 torch.jit.save(trt_ts_module, f"{variant}_bs{batch_size}_{precision}.torch-tensorrt")

[...]

PR #1654, opened 2023-02-08
Broadcast operations in aten::masked_fill and element_wise operations

apbose:
# Description

This PR addresses the broadcasting issue encountered in the NeMo Citrinet model. ([Bug 1573](https://github.com/pytorch/TensorRT/issues/1573)). 

Fixes # (issue)

This PR makes two main changes-
1. The TensorRT broadcast rules are more limited as compared to Torch broadcast rules - [#1629](https://github.com/pytorch/TensorRT/issues/1629). Hence the _aten::masked_fill_ is modified to make mask and self Tensors broadcastable according to TensorRT rules.At present only the mask tensor is padded if its no of dimensions is lesser than the self Tensor, but  not vice versa. 
**Modification** : The mask tensor is unpadded if its dimensions is greater than the self Tensor and it has leading 1 dimensions

2. The _element_wise_ operations broadcasts the two tensors to get to the same dimensions. 
**Modification** : The tensor having extra dimensions is checked if it requires the leading 1 dimensions and accordingly unpadded. Else the other tensor is padded.

The above model fails with the runtime error-RuntimeError: expand(CUDABoolType{[1, 1, 1, 1488]}, size=[1, 256, 1488]): the number of sizes provided (3) must be greater or equal to the number of dimensions in the tensor (4)
 
out dimension = 1, 256, 1488
mask dimension = 1,1,1,1488
Cases with output = out.masked_fill_(mask,2 ) fail while output = torch.masked_fill(mask,2) passes

### DIFF
--- a/core/conversion/converters/converter_util.cpp
+++ b/core/conversion/converters/converter_util.cpp
@@ -118,6 +118,7 @@ nvinfer1::ILayer* add_elementwise(
 
   auto selfDim = util::toVec(self->getDimensions());
   auto otherDim = util::toVec(other->getDimensions());
+
   if (selfDim.size() != otherDim.size()) {
     // other is with dynamic shape, need to expand its dimension now and get its
     // shape at runtime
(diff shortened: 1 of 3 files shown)


narendasan:
@apbose we need tests to verify these changes. Also the TS tests are failing

apbose:
Yes @narendasan , I am looking at debugging the failing ts test

core/conversion/converters/converter_util.cpp
```diff
@@ -151,9 +152,30 @@ nvinfer1::ILayer* add_elementwise(
...
-      other = otherShuffle->getOutput(0);
+
+      // first check if the selfDim can be reduced to the other
+      int diff = selfDim.size() - otherDim.size();
+      bool canSqueeze = true;
+      for (int i = 0; i < diff; i++) {
+        if (selfDim[i] != 1) {
+          canSqueeze = false;
+        }
+      }
```
gs-olive, line 163: Does this work if `diff < 0`?

apbose: Yes it probably should since the above condition ensures that the self has greater dimension than the other. But this is leading to dimensions getting mishandled in bert model, so I need to removing this logic

apbose:
The above model fails in torchscript. `aten::masked_fill` rules are different from `aten::_masked_fill`

PR #1683, opened 2023-02-20
Prototype reorg of the FX converters to clean dependency chain

narendasan:
# Description

This PR proposes a reorg of the converter library for FX so that the dependency chain is much cleaner considering there's 3 versions of ops being supported in FX (nn, acc, aten). 

Instead of having implementations call each other, the idea is to have an IR agnostic set of converters and then the `[aten/acc/nn]_ops_converter` files will just pack arguments properly.

Fixes # (issue)

## Type of change

# Checklist:


narendasan:
@apbose This is sort of what I have been talking about. We abstract out the various potential IRs in files like `activation.py` where as `[aten/acc/nn]_ops_converters.py` are just for input packing. The idea is torch2trt style converters will fill out files like `activation.py`

apbose (approved):

apbose (commented):
Why is the dyn_range_fn required? What would be the typical use case?

apbose:
Raised PR #1745

PR #1698, opened 2023-02-25
fix: Improve input handling for `input_signature`

### LINKED ISSUE #1595: 🐛 [Bug] Encountered bug when compiling model with nested Input type using `input_signature` argument
##  Bug Description

When compiling a model with input type `Tuple[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]`, compilation fails with the following error:
```python
RuntimeError: [Error thrown at core/ir/ir.cpp:46] Expected vals.size() == specs.size() to be true but got false
Expected dimension specifications for all input tensors, but found 1 input tensors and 2 dimension specs
```

## To Reproduce

1. Define a Torch model with forward function having the following form:
```python
def forward(self, input : Tuple[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]])
```
2. Define the Torch-TRT Inputs and compilation settings using `input_signature`, then compile the scripted model:
```python
compile_settings = {
    "input_signature": (torch_tensorrt.Input((5, 5), dtype=torch.float), (torch_tensorrt.Input((5, 5), dtype=torch.float), torch_tensorrt.Input((5, 5), dtype=torch.float)),),
}
trt_ts_module = torch_tensorrt.ts.compile(scripted_model, **compile_settings)
```

## Expected behavior

Model should compile with input signature containing nested Tuple collection.

## Environment

- Torch-TensorRT Version: 1.4.0.dev0+f43be5b6
- PyTorch Version: 1.14.0.dev20221114+cu116
- OS: Ubuntu 20.04
- Build command you used: `python setup.py develop`
- Python version: 3.8.13
- CUDA version: 11.6

## Additional context

The case of a nested Tuple containing a singleton Tensor followed by a Tuple of Tensors is not parsed as expected by the input signature interpreter, whereas regular tuple Inputs of Tensors are supported, as in:
https://github.com/pytorch/TensorRT/blob/b2a5da66cd22db22560bdb7b54f107fb4bcffb2e/tests/py/api/test_collections.py#L210-L213

Wrapping the input signature in another Tuple does not resolve the issue, and raises a different error message:
```python
RuntimeError: forward() Expected a value of type 'Tuple[Tensor, Tuple[Tensor, Tensor]]' for argument 'input.1' but instead found type 'Tuple[Tensor]'.
Position: 1
Declaration: forward.forward(__torch__.module self_1, (Tensor, (Tensor, Tensor)) input.1) -> ((Tensor, Tensor, Tensor) 7)
```

gs-olive:
# Description

- Improve error messaging for input signature arguments by verifying the nesting depth of the object is within the bounds specified in the TorchScript compiler
- Throw user-error more quickly to avoid less-interpretable error messages later on

Relevant GraphInputs code:
https://github.com/pytorch/TensorRT/blob/5fa63743fc4448e38b1e6b01e7403c0d74717158/core/ir/GraphInputs.cpp#L42-L54

Fixes #1595 

## Type of change

- Improved input handling + error messaging

# Checklist:

### DIFF
--- a/cpp/src/compile_spec.cpp
+++ b/cpp/src/compile_spec.cpp
@@ -36,13 +36,15 @@ CompileSpec::CompileSpec(torch::jit::IValue input_signature) {
   graph_inputs.input_signature = input_signature;
 }
 
-void to_internal_input_signature(torch::jit::IValue input_ivalue, torch::jit::IValue& converted_ivalue) {
+void to_internal_input_signature(torch::jit::IValue input_ivalue, torch::jit::IValue& converted_ivalue, int depth = 0) {
+  TORCHTRT_CHECK(
+      depth <= 2, "Input nesting depth exceeds max supported depth, use 1 level: [A, B], or 2 level: [A, (B, C)]")
   if (input_ivalue.isTuple()) {
     auto input_tuple = input_ivalue.toTuple();
     std::vector<torch::jit::IValue> converted_elements;
     for (auto item : input_tuple->elements()) {
       torch::jit::IValue converted_item;
-      to_internal_input_signature(item, converted_item);
+      to_internal_input_signature(item, converted_item, depth++);
       converted_elements.push_back(converted_item);
       auto tuple_ptr = c10::ivalue::Tuple::create(converted_elements);
       converted_ivalue = torch::jit::IValue(tuple_ptr);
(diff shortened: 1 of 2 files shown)


narendasan (approved):
Is 2 arbitrary or is there some sort of technical depth limit? If possible i think it would be nice to support deeper nesting like 3-4 but not a massive deal. LGTM

gs-olive:
> Is 2 arbitrary or is there some sort of technical depth limit?

The choice of maximum depth of 2 is based on the data structure used to store the inputs: `std::vector<std::vector<Input>>`. While this structure could be expanded via adding more nested vector structures, a more robust solution would be desirable (supporting arbitrarily complex collections).

Issue #1563, opened 2022-12-21
🐛 [Bug] Windows build broken after 1.2.0 - experimental:filesystem in TRTEngine


bharrisau:
##  Bug Description

At some point after v1.2.0 the `TensorRT\core\runtime\TRTEngine.h` file was added and it brought the `#include <experimental/filesystem>` with it. I had some issues trying to build due to this. A `#define` is needed to build, the `path` cannot implicitly be converted to a `std::string`, an extra `#include` is needed to use `std::stringbuilder` and there is no `stdc++fs` to link against.

## Environment

 - Torch-TensorRT Version: master f43be5b64dd97f09c05dd8756f01427a497ecda4
 - PyTorch Version: libtorch 1.13.1+cu117
 - CPU Architecture: amd64
 - OS: Windows 11
 - Python version:  Python 3.10.9
 - CUDA version: 11.8
 - cmake version: cmake version 3.21.21080301-MSVC_2
 - cl version: Microsoft (R) C/C++ Optimizing Compiler Version 19.30.30706 for x64
 - MSVC toolchain version: 14.30.30705

## Rough fix
The below changes were needed to get a compile. They probably would break the build on other systems - so maybe need to be based on some platform detection.

Additional CMAKE flag
```json
{
  "name": "CMAKE_CXX_FLAGS",
  "value": "-D_SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING",
  "type": "STRING"
}
```

Code changes
```diff
diff --git a/core/runtime/CMakeLists.txt b/core/runtime/CMakeLists.txt
index d21c661a..b3632e8f 100644
--- a/core/runtime/CMakeLists.txt
+++ b/core/runtime/CMakeLists.txt
@@ -33,7 +33,6 @@ target_link_libraries(${lib_name}
         TensorRT::nvinfer
         torch
         core_util
-        stdc++fs
 )

 # Install
diff --git a/core/runtime/TRTEngine.h b/core/runtime/TRTEngine.h
index 5615a824..1a2df1f2 100644
--- a/core/runtime/TRTEngine.h
+++ b/core/runtime/TRTEngine.h
@@ -25,7 +25,7 @@ struct TRTEngine : torch::CustomClassHolder {
   std::string name;
   RTDevice device_info;

-  std::string profile_path_prefix = std::experimental::filesystem::temp_directory_path();
+  std::string profile_path_prefix = std::experimental::filesystem::temp_directory_path().string();

[...]

PR #1711, opened 2023-03-03
Get windows build working

bharrisau:
# Description

Initial PR for #1563. Fixes Windows CMAKE build.

Finally had time to come back to this - build on a fresh Windows11 machine, all steps to setup/reproduce included.

The change to `core/plugins/CMakeLists.txt` seems like it should be unrelated to this. I was unable to build because 'cudnn.h' was required, but the include path was not being passed to the compiler as the `cuDNN` module wasn't specified. This might not be a problem in other targets because commonly the cudnn files are just installed in to the cuda include/lib/bin dirs.

Also note - windows is generally case insensitive. The normal output dir is `build/` but that doesn't work on Windows as there is a `BUILD` file. So I'm using `out/` as the output. But that isn't in the `.gitignore`, and I haven't included a change for that in this PR.

## Type of change

- Bug fix (incomplete and probably breaks other build targets)

# Notes to reproduce with Visual Studio Code (I didn't want to pull down the whole Visual Studio)
- [ ] Install Visual Studio Code
- [ ] Install Build Tools for Visual Studio 2022
    - Select "Desktop Development with C++"
        - Currently, this installs MSVC v143 - 2022. There are also options to install previous 2019/2017/2015 editions of MSVC
    - License term "1b Build Tools additional use right" allows using Build Tools to compile Open Source Dependencies
    - Also allows using Build Tools to develop and test Open Source Dependencies, to the minor extend of ensuring compatibility with Build Tools
- [ ] Install CUDA 11.7.1
- [ ] Install CuDNN 8.5.0.96
    - Set `cuDNN_ROOT_DIR`
- [ ] Install TensorRT 8.5.1.7
    - Set `TensorRT_ROOT`
    - Add `TensorRT_ROOT\lib` to PATH
- [ ] Install "libtorch-win-shared-with-deps-latest.zip" 
    - Select the CUDA 11.7 version
    - Set `Torch_DIR`
    - Add `Torch_DIR\lib` to PATH
- [ ] Clone TensorRT repo
- [ ] Install C++ and CMake Tools extensions from MS
    - Change build to RelWithDebInfo
- [ ] Update `.vscode\settings.json`
- [ ] Clean, configure, build

`/.vscode/settings.json`
```json
{
    "cmake.generator": "Ninja",
    "cmake.configureSettings": {
        "CMAKE_MODULE_PATH": {
            "type": "FILEPATH",
            "value": "$PWD\\cmake\\Modules"
        },
        "CMAKE_CXX_FLAGS": {
            "type": "STRING",
            "value": "-D_SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING"
        },
        "Torch_DIR": {
            "type": "FILEPATH",
            "value": "X:\\libtorch\\share\\cmake\\Torch"
        },
        "TensorRT_ROOT": {
            "type": "FILEPATH",
            "value": "X:\\path\\to\\tensorrt"
        },
        "cuDNN_ROOT_DIR": {
            "type": "FILEPATH",
            "value": "X:\\path\\to\\cudnn"
        },
        "CMAKE_CUDA_FLAGS": "-allow-unsupported-compiler"
    },
    "cmake.buildDirectory": "${workspaceFolder}/out"
}
```

### DIFF
--- a/core/plugins/CMakeLists.txt
+++ b/core/plugins/CMakeLists.txt
@@ -23,6 +23,7 @@ target_link_libraries(${lib_name}
         TensorRT::nvinfer_plugin
         torch
         core_util
+        cuDNN::cuDNN
     PRIVATE
         Threads::Threads
 )
--- a/core/runtime/CMakeLists.txt
+++ b/core/runtime/CMakeLists.txt
@@ -33,8 +33,14 @@ target_link_libraries(${lib_name}
         TensorRT::nvinfer
         torch
         core_util
-        #stdc++fs
 )
+
+if(NOT WIN32)
+    target_link_libraries(${lib_name}
+        PUBLIC
+            stdc++fs
+    )
+endif(NOT WIN32)
 
 # Install
 install(FILES ${HEADER_FILES} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/torch_tensorrt/core/runtime")
--- a/core/runtime/CMakeLists.txt
+++ b/core/runtime/CMakeLists.txt
@@ -33,7 +33,7 @@ target_link_libraries(${lib_name}
         TensorRT::nvinfer
         torch
         core_util
-        stdc++fs
+        #stdc++fs
 )
 
 # Install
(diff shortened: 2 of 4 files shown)


narendasan:
Awesome @bharrisau, thanks for opening this. Will take a look.

narendasan (approved):
Got a chance to try this out. LGTM! Thanks for taking the time to fix this!

PR #1738, opened 2023-03-15
feat: Add decorator utility to improve error messaging for legacy support

### LINKED ISSUE #1733: ✨[Feature] Create Decorator for Python functions Incompatible with Legacy Torch Versions
## Context
Some functions in the code base are incompatible with certain versions of Torch such as 1.13.1. Adding a general-purpose utility decorator which detects whether the Torch version is incompatible with certain key functions could be helpful in notifying the user when such mismatches occur.

## Desired Solution
A decorator of the form:
```python
@req_torch_version(2)
def f(...):
```
This decorator would throw a clear error when the detected Torch version differs from that required to use the decorated function.

## Temporary Alternative
A temporary alternative, which is already in use in PR #1731, is to add logic to each function with compatibility issues, raising an error if any versioning issues are detected. While this is a functional solution, it is difficult to scale and can lead to repetitive code.

## Additional Context
Certain imports are also not compatible across Torch versions (as in `import torchdynamo` in Torch 1.13 versus `import torch._dynamo` in 2.0). To resolve this issue, the imports can be moved inside the functions using them so as to encapsulate all version-specific code within one area.

gs-olive:
# Description
- Add general-purpose utility for intercepting invalid Torch version imports and displaying an error message to the user, halting compilation
- Decorator can accept variable input versions and checks whether the current version in use in the environment is at least as large as the minimum specified version
- Display clear error message upon calling the aten tracer when using a legacy Torch version, for which the Dynamo import is invalid

Fixes #1733 

## Type of change

# Checklist:

  - Manually tested - remainder is covered via CI


py/torch_tensorrt/fx/passes/pass_utils.py
```diff
@@ -179,6 +179,27 @@ def bounded_method(*args, **kwargs):
     return dec_for_method
 
 
+def req_torch_version(min_torch_version: str = "2.0") -> Decorator:
```
narendasan, line 182: Lets try to use some sort of semver style versioning here. String might not be the best type

py/torch_tensorrt/fx/passes/pass_utils.py
```diff
@@ -179,6 +179,27 @@ def bounded_method(*args, **kwargs):
...
+    def nested_decorator(f: Callable):
+        def function_wrapper(*args, **kwargs):
+            # Parse minimum and current Torch versions as tuples of strings
+            # Split out GPU version (denoted via +_____)
+            reqd_torch_version = tuple(min_torch_version.split("+")[0].split("."))
+            current_torch_version = tuple(torch.__version__.split("+")[0].split("."))
+
+            if current_torch_version < reqd_torch_version:
```
narendasan, line 190: Lets print the name of the function so users are informed that its just this part of the library that is not supported.

py/torch_tensorrt/fx/passes/pass_utils.py
narendasan, line 1: Lets put this in some higher level file

gs-olive: Moved function to `fx.utils`

py/torch_tensorrt/fx/passes/pass_utils.py
gs-olive: In the raised error, the function itself is printed directly. This includes the name of the function as well.

py/torch_tensorrt/fx/utils.py
```diff
@@ -104,3 +105,36 @@ def f(*inp):
...
+        def function_wrapper(*args, **kwargs):
+            # Parse minimum and current Torch versions
+            min_version = version.parse(min_torch_version)
+            current_version = version.parse(torch.__version__)
+
+            if current_version < min_version:
+                raise AssertionError(
+                    f"Expected Torch version {min_torch_version} or greater, "
+                    + f"when calling {f}. Detected version {torch.__version__}"
+                )
```
gs-olive, line 134: The function `f` is printed in the error message, which includes its name.

narendasan (approved):
Lgtm

PR #1763, opened 2023-03-23
feat: Implement Input class support for FX backend.

peri044:
# Description

1) This PR adds support for providing `torch_tensorrt.Input(shape=x, dtype=x)` to  FX backend.

Example workflow would be similar to TS backend. 
```
inputs = [torch_tensorrt.Input(shape=(1, 3, 3, 4), dtype=torch.float32)]
trt_mod = torch_tensorrt.compile(
            mod,
            ir="fx",
            inputs=inputs,
            min_acc_module_size=1,
        )
```
2) Internally FX performs a) standard lowering passes b) TRT lowering. The standard lowering passes traces through the model which require pytorch tensors for execution. 

For standard lowering passes 
a) User input of type `torch_tensorrt.Input` -> Example pytorch tensors (using `example_tensors()`) 
b) User input of type `torch.Tensors`  - this type will directly pass through tracing and hence we don't need to do anything.

For TRT lowering 
a) `self._trt_input` is used to convert `torch_tensorrt.Input` into an `InputTensorSpec`
b) This can now support dynamic shapes using the same front end interface as Torchscript. 

3) `TSInput` class is derived from `torch_tensorrt.Input` because the latter had C++ calls (`to_internal`). For FX backend, we don't need them for `--fx-only` installation.

4) `max_batch_size` - This has been removed and already deprecated in TRT (not available in TS backend) as well.

5) FX tracing and lowering previously assumes input tensors reside on the same device as the model parameters. In the case of `torch_tensorrt.Input`, when we extract example tensors, the default device used is `cuda:0`. This assumption might not be valid for some non standard use cases. More refactoring is required to ensure we specify the right device and it is being passed accordingly.

## Type of change

# Checklist:

### DIFF
--- a/examples/fx/lower_example.py
+++ b/examples/fx/lower_example.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass, field, replace
 
 import torch
 import torchvision
+import torch_tensorrt
 from torch_tensorrt.fx import compile
 from torch_tensorrt.fx.utils import LowerPrecision
 
@@ -98,13 +99,17 @@ def benchmark(
 
     model = model.cuda().eval()
     inputs = [x.cuda() for x in inputs]
-
+    # inputs = [torch_tensorrt.Input(shape=(128, 3, 224, 224), dtype=torch.float32)]
+    # inputs = [torch_tensorrt.Input(min_shape=[1, 3, 224, 224],
+    #                                opt_shape=[8, 3, 224, 224],
+    #                                max_shape=[32, 3, 224, 224],
+    #                                dtype=torch.float32)]
     # benchmark base configuration
     conf = Configuration(batch_iter=batch_iter, batch_size=batch_size)
 
     configurations = [
         # Baseline
-        replace(conf, name="CUDA Eager", trt=False),
+        # replace(conf, name="CUDA Eager", trt=False),
         # FP32
         replace(
             conf,
@@ -115,14 +120,14 @@ def benchmark(
             accuracy_rtol=1e-3,
         ),
         # FP16
-        replace(
-            conf,
-            name="TRT FP16 Eager",
-            trt=True,
-            jit=False,
-            fp16=True,
-            accuracy_rtol=1e-2,
-        ),
+        # replace(
+        #     conf,
+        #     name="TRT FP16 Eager",
+        #     trt=True,
+        #     jit=False,
+        #     fp16=True,
+        #     accuracy_rtol=1e-2,
+        # ),
     ]
 
     results = [
@@ -189,8 +194,12 @@ def run_configuration_benchmark(
             max_batch_size=conf.batch_size,
             lower_precision=LowerPrecision.FP16 if conf.fp16 else LowerPrecision.FP32,
             explicit_batch_dimension=True,
+            dynamic_batch=False,
         )
-        time = benchmark_torch_function(conf.batch_iter, lambda: lowered_module(*input))
+        random_inputs = [torch.randn((128, 3, 224, 224), dtype=torch.float32).cuda()]
+        time = benchmark_torch_function(
+            conf.batch_iter, lambda: lowered_module(*random_inputs)
+        )
     else:
         print("Lowering with JIT is not available!", "red")
 
--- a/examples/fx/lower_example.py
+++ b/examples/fx/lower_example.py
@@ -4,7 +4,6 @@ from dataclasses import dataclass, field, replace
 
 import torch
 import torchvision
-import torch_tensorrt
 from torch_tensorrt.fx import compile
 from torch_tensorrt.fx.utils import LowerPrecision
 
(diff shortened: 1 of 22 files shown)


examples/fx/lower_example.py
```diff
@@ -98,13 +99,17 @@ def benchmark(
 
     model = model.cuda().eval()
     inputs = [x.cuda() for x in inputs]
-
+    # inputs = [torch_tensorrt.Input(shape=(128, 3, 224, 224), dtype=torch.float32)]
```
narendasan, line 102: What are these comments for?

py/torch_tensorrt/_Input.py
```diff
@@ -41,6 +40,7 @@ class _ShapeMode(Enum):
     DOMAIN_OFFSET = 2.0
     low_tensor_domain_incl = 0.0
     high_tensor_domain_excl = low_tensor_domain_incl + DOMAIN_OFFSET
+    torch_dtype = None
```
narendasan, line 43: Should we derive torch_dtype from self.dtype?

py/torch_tensorrt/_Input.py
```diff
@@ -173,59 +176,6 @@ def __str__(self) -> str:
         else:
             raise RuntimeError("Unknown input shape mode")
 
-    def _to_internal(self) -> _C.Input:
```
narendasan, line 176: Why was this taken out?

py/torch_tensorrt/fx/tools/trt_minimizer.py
```diff
@@ -5,21 +5,22 @@
...
 def lower_mod_default(
     mod: torch.fx.GraphModule,
     inputs: Tensors,
-    batch_size: Any = 2048,
+    # batch_size: Any = 2048,
     use_experimental_rt: bool = False,
 ) -> TRTModule:
     interp = TRTInterpreter(
-        mod, InputTensorSpec.from_tensors(inputs), explicit_batch_dimension=True
+        mod, _Input.Input.from_tensors(inputs), explicit_batch_dimension=True
```
narendasan, line 21: Can this just be Input.from_tensors

py/torch_tensorrt/ts/ts_input.py
```diff
@@ -0,0 +1,108 @@
...
+
+import torch
+
+from torch_tensorrt import _C
+from torch_tensorrt import _enums
+from torch_tensorrt import _Input
+from torch_tensorrt._Input import Input
+
+
+class TSInput(Input):
```
narendasan, line 12: This should be a hidden class, we dont want people using this, handle any conversion to TSInput internally

py/torch_tensorrt/fx/fx2trt.py
```diff
@@ -153,7 +153,6 @@ def validate_conversion(self):
 
     def run(
         self,
-        max_batch_size=64,
```
frank-wei, line 156: I am afraid we can not do this change. We have to maintain backward compatibility on the API, otherwise, it will break our internal product.

py/torch_tensorrt/fx/input_tensor_spec.py
```diff
@@ -4,58 +4,59 @@
 
 from .types import Shape, ShapeRange
 from .utils import get_dynamic_dims
-
-
-def generate_input_specs(inputs, lower_setting, additional_inputs=None):
```
frank-wei, line 9: Same here. These API are used in many internal products.

frank-wei (commented):
Overall, I can not approve this change for some reasons:
1) fx2trt is used internally and core APIs are called by different service. So far, we did not build the Import mechanism to protect the internal service. (ImportIt is used in other Pytorch projects which will run internal CI that covers not only unit tests but also other prod tests for any PR). So we can not take the risk to take any potential big change since that will break our internal services.
2) Due to the above reason, any change for FX path needs to consider backward compatibility. Also consider to reduce the PR size for easy review.
cc @yinghai  @wushirong

py/torch_tensorrt/ts/ts_input.py
peri044: Yeah that's the intended usage. In the current design, for torchscript, they still use the same interface 
`torch_tensorrt.Input()` and we internally convert into `TSInput` in the `ts/_compile_spec.py` file. Anything different that you have in mind ?

py/torch_tensorrt/fx/tools/trt_minimizer.py
peri044: yeah

py/torch_tensorrt/ts/ts_input.py
narendasan: No this would be fine

narendasan:
@frank-wei can we expose the old API somewhere else for backwards compatibility but move to the unified one for users? One of the big problems right now is it is difficult to move from torchscript to fx since many of the settings are named differently or used differently

peri044:
Closing this in favour of https://github.com/pytorch/TensorRT/pull/1807

frank-wei:
> @frank-wei can we expose the old API somewhere else for backwards compatibility but move to the unified one for users? One of the big problems right now is it is difficult to move from torchscript to fx since many of the settings are named differently or used differently

Specifically, I am thinking the unified version could happen on aten2trt?

py/torch_tensorrt/fx/input_tensor_spec.py
```diff
@@ -116,6 +117,43 @@ def from_tensors(cls, tensors: Sequence[torch.Tensor]) -> List["InputTensorSpec"
         assert isinstance(tensors, (list, tuple))
         return [cls.from_tensor(t) for t in tensors]
 
+    @classmethod
```
frank-wei, line 120: good to have

PR #1770, opened 2023-03-24
fx2trt converters aten::slice,aten::select and aten::matmul

### LINKED ISSUE #1709: ↔ [Converter] Add support for `aten.matmul` in the FX `aten` path
## aten.matmul

- **Function Schema**: `torch.ops.aten.matmul: ((torch.float32, torch.float32), {})`

- **Original PyTorch API**: `torch.matmul`

- **Relevant TensorRT Documentation**: [IMatrixMultiplyLayer](https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Graph/Layers.html#imatrixmultiplylayer)

Add support for `matmul` as an [aten converter](https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/fx/converters/aten_ops_converters.py).

### LINKED ISSUE #1714: ↔ [Converter] Add support for Tensor slicing and selection operators in the FX `aten` path
## aten.slice, aten.select, aten.expand

- **Function Schema**:
  - `torch.ops.aten.expand.default: ((torch.int64,), {})`
  - `torch.ops.aten.slice.Tensor: ((torch.float32,), {})`
  - `torch.ops.aten.slice.Tensor: ((torch.int32,), {})`
  - `torch.ops.aten.slice.Tensor: ((torch.float32,), {})`
  - `torch.ops.aten.slice.Tensor: ((torch.int32,), {})`
  - `torch.ops.aten.select.int: ((torch.float32,), {})`

- **Original PyTorch API**: `torch.expand`, `torch.slice`, `torch.select`

- **Relevant TensorRT Documentation**: [IGatherLayer](https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Graph/Layers.html?highlight=igather#igatherlayer), [IShuffleLayer](https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Graph/Layers.html?highlight=ishuffle#ishufflelayer)

Add support for `slice`, `select`, and `expand` as [aten converters](https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/fx/converters/aten_ops_converters.py).

apbose:
# Description

Fixes #1709 
Fixed #1714


py/torch_tensorrt/fx/converters/activation.py
```diff
@@ -1,76 +1,197 @@
...
+        layer.beta = beta
+    set_layer_name(layer, target, name)
+
+    if input_val.dynamic_range is not None:
+        dyn_range = dyn_range_fn(input_val.dynamic_range)
+        mark_as_int8_layer(layer, dyn_range)
+    return layer.get_output(0)
 
 
-def common_activation(
```
frank-wei, line 12: Is there reason we want to drop this `common_activation` func and split it into many single functions?

frank-wei (commented):
1. I am seeing the common

py/torch_tensorrt/fx/converters/activation.py
apbose: Hi Wei, the `common_activation` function is not split, instead it is included in the `add_activation` function in activation.py. The idea is that the activation and the operator functions are split in two (activation.py and operator.py respectively) and the `converter_utils.py` is for the common functions across the two and other operations.

py/torch_tensorrt/fx/converters/operator.py
```diff
@@ -0,0 +1,1256 @@
...
+    if not isinstance(input_val, TRTTensor):
+        raise RuntimeError(
+            f"{operation_type} received input {input_val} that is not part "
+            "of the TensorRT region!"
+        )
+    layer = network.add_unary(input_val, operation_type)
+    set_layer_name(layer, target, name)
+    output = layer.get_output(0)
+    output.name = output.name + "_" + target.__name__
+    return layer.get_output(0)
```
gs-olive, line 548: Does `layer.get_output(0)` make a copy of the underlying object? If so, line 547 may have no effect, since it only modifies the `output` object. Should this be `return output` instead?

py/torch_tensorrt/fx/converters/operator.py
```diff
@@ -0,0 +1,1256 @@
...
+        preset_diff += 1
+        other_matrix_op = trt.MatrixOperation.VECTOR
+
+    input_val, other_val = broadcast(
+        network, input_val, other_val, f"{name}_input", f"{name}_other", preset_diff
+    )
+    layer = network.add_matrix_multiply(
+        input_val, input_matrix_op, other_val, other_matrix_op
+    )
+    set_layer_name(layer, target, name)
```
gs-olive, line 700: Some layer name changes use this function and others directly modify `layer.get_output(0).name`. Consider standardizing the usage across functions.

py/torch_tensorrt/fx/test/converters/aten_op/test_matmul_aten.py
gs-olive, line 1: Consider adding additional tests for this function since it is used so often. For example, adding multi-dimensional tensors and single-dimensional vector matrix multiplications as well.

gs-olive (commented):
Looks good overall, with lots of critical operators and tests. Just made a few comments on in-place modification of layer objects and matmul test coverage.

py/torch_tensorrt/fx/converters/operator.py
apbose: That makes sense.

apbose: That makes sense. Just that in the case of `set_layer_name` the name is changed directly for the layer, while appending `acc_ops.target.__name__` to the layer name with the name passed in the above function. Whereas in case of the other we change the name of `layer.get_output(0)` with only the `target.__name__`. @frank-wei any thoughts on this to integrate the two?

py/torch_tensorrt/fx/converters/operator.py
apbose: So the `layer.get_output(0) `does not make an explicit copy and modifying the `output.name` causes inplace modification of the `layer.get_output(0).name` too. Yes it could have been `return output` too I assume, but any change would be inconsequential.

py/torch_tensorrt/fx/converters/operator.py
```diff
@@ -0,0 +1,1256 @@
...
+
+    start_int = cast(int, kwargs["start"])
+    stop_int = cast(int, kwargs["stop"])
+    step_int = cast(int, kwargs["step"])
+    start = [0] * len(input_val.shape)
+    start[dim] = start_int
+    stride = [1] * len(start)
+    stride[dim] = step_int
+    output_shape = list(input_val.shape)
+    output_shape[dim] = (stop_int - start_int) // step_int + 1
```
gs-olive, line 1210: I think this equation could potentially give an incorrect output shape. I think it could instead be `max((stop_int - start_int) // step_int, 1)`. Consider the following:

```python
a = ["a", "b", "c"]
start = 0
stop = 2
step = 1

a[start:stop]
>>> ["a", "b"]

(stop - start) // step + 1
>>> 3

len(a[start:stop])
>>> 2

max((stop - start) // step, 1)
>>> 2
```
The dimension assigned here to `output_shape[dim]` is one larger than the actual dimension returned. This issue could be contributing to the CI failures for `aten.slice` on this PR ([link here](https://app.circleci.com/pipelines/github/pytorch/TensorRT/2306/workflows/d1388765-8fc6-4d85-b79f-8c2bad34ee14/jobs/10615)).

It would also be helpful to add a regression test case to `test_slice_aten.py` to catch this case, as the current ones seem to be passing despite the issue.

gs-olive (changes requested):
Changes requested in comment to (potentially) resolve CI failures.

py/torch_tensorrt/fx/converters/operator.py
```diff
@@ -0,0 +1,1256 @@
...
+        index_value.shape, to_numpy(index_value)
+    ).get_output(0)
+    layer = network.add_gather(input_val, indices_tensor, dim)
+    out = layer.get_output(0)
+    if len(out.shape) != 1:
+        layer = network.add_shuffle(out)
+    return layer.get_output(0)
+
+
+def add_slice(network, target, kwargs, name):
```
gs-olive, line 1179: It seems that the `aten` path is now failing in the CI ([as here](https://app.circleci.com/pipelines/github/pytorch/TensorRT/2402/workflows/cd7562f3-3ccf-4ca9-beb4-240a5e93d86f/jobs/11073)). Could be that `aten` includes the `stop` as an endpoint, while `acc` does not? If that is the case, consider adding a keyword arg to the function, like `aten=False`, which determines whether the equation is `output_shape[dim] = (stop_int - start_int) // step_int` or `output_shape[dim] = (stop_int - start_int) // step_int + 1`. Though, I am not sure why the two implementations would differ in the meaning of the endpoint (`stop_int`).

gs-olive (changes requested):
Changes requested below related to latest CI results.

py/torch_tensorrt/fx/converters/operator.py
```diff
@@ -1207,7 +1208,7 @@ def add_slice(network, target, kwargs, name):
     stride = [1] * len(start)
     stride[dim] = step_int
     output_shape = list(input_val.shape)
-    output_shape[dim] = (stop_int - start_int) // step_int
+    output_shape[dim] = math.ceil((stop_int - start_int) // step_int)
```
gs-olive, line 1211: I think this needs to be `math.ceil((stop_int - start_int) / step_int)`, since the `ceil` is doing the int-casting

gs-olive (changes requested):

frank-wei:
cc @wushirong for review.

wushirong:
In general, this diff is too big and includes multiple fixing purpose. This would make code sync on Meta side very likely to fail. Let't split this into a series of small diffs and make sure you only do one thing within a diff, i.e adding one converter and correspondent unit test. 
Also I noticed you are trying to move existing code pieces around, the intention was good but we would likely to avoid this. New features or converters is welcomed.

apbose:
Hi @wushirong, thanks for the review.

PR #1851, opened 2023-04-21
feat: Wrap dynamic size handling in a compilation flag

peri044:
# Description

This PR does the following

1) The aten::size dynamic shape handling is now disabled by default. `allow_shape_tensors=True` will enable it.

2) Modifications to aten::size include preserving the static dimensions in the input, thereby not converting any known types eg: `Int` into `ITensors`

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/evaluators/aten.cpp
+++ b/core/conversion/evaluators/aten.cpp
@@ -273,7 +273,8 @@ auto aten_registrations TORCHTRT_UNUSED =
                      if (ctx->settings.allow_shape_tensors) {
                        return dynamic_size_layer(ctx, n, args);
                      } else {
-                       LOG_WARNING("There may be undefined behavior using dynamic shape and aten::size ");
+                       LOG_WARNING(
+                           "There may be undefined behavior using dynamic shape and aten::size without setting allow_shape_tensors");
                      }
                    }
                    return util::toVec(tensor->getDimensions());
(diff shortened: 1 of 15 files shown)


bowang007 (approved):
LGTM

core/conversion/evaluators/aten.cpp
```diff
@@ -270,7 +270,11 @@ auto aten_registrations TORCHTRT_UNUSED =
                  if (tensor_var.isITensor()) {
                    auto tensor = tensor_var.ITensor();
                    if (ctx->input_is_dynamic) {
-                     return dynamic_size_layer(ctx, n, args);
+                     if (ctx->settings.allow_shape_tensors) {
+                       return dynamic_size_layer(ctx, n, args);
+                     } else {
+                       LOG_WARNING("There may be undefined behavior using dynamic shape and aten::size ");
```
narendasan, line 276: Is this undefined behavior due to using aten::size with dynamic shape without shape tensors or the other way around. Might need rewording

core/conversion/evaluators/aten.cpp
```diff
@@ -270,7 +270,11 @@ auto aten_registrations TORCHTRT_UNUSED =
                  if (tensor_var.isITensor()) {
                    auto tensor = tensor_var.ITensor();
                    if (ctx->input_is_dynamic) {
-                     return dynamic_size_layer(ctx, n, args);
+                     if (ctx->settings.allow_shape_tensors) {
```
narendasan, line 273: We should be checking if input is dynamic, shape tensors are enabled **and** the input contains a placeholder dimension before using the shape tensor code path. Otherwise static aten size is sufficient and will cause less errors

core/conversion/evaluators/aten.cpp
```diff
@@ -286,7 +290,11 @@ auto aten_registrations TORCHTRT_UNUSED =
                  auto dim = args.at(n->input(1)).unwrapToInt();
                  if (tensor_var.isITensor()) {
                    if (ctx->input_is_dynamic) {
-                     return dynamic_size_layer(ctx, n, args);
+                     if (ctx->settings.allow_shape_tensors) {
```
narendasan, line 293: Same comments here from above about conditions and warnings

narendasan (changes requested):

PR #1853, opened 2023-04-22
Fix: fix float point 16 precision type issue

bowang007:
# Description
For some networks running with FP16, some bugs like 
```
Expected input tensors to have type Half, found type float
```
are observed. 
This PR fixes the bug that output type may not match to the input type
in converters. Some of the converters should make sure that the input
tensor and output tensor have the same date type.

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/activation.cpp
+++ b/core/conversion/converters/impl/activation.cpp
@@ -16,8 +16,9 @@ namespace {
     TORCHTRT_CHECK(new_layer, "Unable to create " #act " layer from node: " << *n);                                  \
                                                                                                                      \
     new_layer->setName(util::node_info(n).c_str());                                                                  \
+    new_layer->setOutputType(0, in->getType());                                                                      \
     ctx->AssociateValueAndTensor(n->outputs()[0], new_layer->getOutput(0));                                          \
-    LOG_DEBUG("Output tensor shape: " << new_layer->getOutput(0)->getDimensions());                                  \
+    LOG_DEBUG("Output tensor: " << new_layer->getOutput(0)->getDimensions() << "/" << in->getType());                \
                                                                                                                      \
     return true;                                                                                                     \
   }                                                                                                                  \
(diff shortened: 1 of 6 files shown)


core/conversion/conversion.cpp
```diff
@@ -212,6 +212,17 @@ void AddInputs(ConversionCtx* ctx, c10::ArrayRef<const torch::jit::Value*> input
 }
 
 void MarkOutputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> outputs) {
+  nvinfer1::DataType output_type = ctx->net->getInput(0)->getType();
+  for (size_t i = 0; i < ctx->net->getNbInputs(); ++i) {
+    auto cur_input_type = ctx->net->getInput(i)->getType();
+    if (ctx->enabled_precisions.find(cur_input_type) == ctx->enabled_precisions.end()) {
+      TORCHTRT_THROW_ERROR("Input type is not found in enabled_precisions.");
+    }
```
gs-olive, line 220: This error/check can cause models to erroneously fail, as many models (e.g. transformers) require `torch.int32` inputs, but are compiled with `torch.float` or `torch.int` precision.

bowang007:
@narendasan  What actually is happening is that in Converters the output type for each layer is not correctly inferred. 
The updated PR addresses such issues by inferring output types for some converters instead of setting whole TRT engine's output type. 
Please take a look, and I think some other converters may need to be updated as well.

core/conversion/converters/impl/activation.cpp
```diff
@@ -75,9 +77,10 @@ auto acthardtanh TORCHTRT_UNUSED =
                     new_layer->setBeta(max);
 
                     new_layer->setName(util::node_info(n).c_str());
+                    new_layer->setOutputType(0, in->getType());
```
narendasan, line 80: What would cause this to not be the case? would this interfere with TensorRT's mixed precision behavior?

bowang007:
closing per #2469

Issue #1842, opened 2023-04-20
✨[Feature] Add Lowering Pass to Eliminate `If/Else` Blocks with Exceptions in TorchScript


gs-olive:
## Problem Context
For certain TorchScript graph control flow blocks, the `prim::RaiseException` primitive is used in one block to enforce an invariant, while the other block performs computation. One example of this is in the case of `nn.Upsample`, which is shown below.

<details>

<summary>
<b>Upsample Graph Snippet </b>
</summary>

```python
  %out1.1 : Tensor = prim::If(%45)
    block0():
      %51 : Tensor = aten::upsample_bilinear2d(%X.1, %18, %119, %115)
      -> (%51)
    block1():
      %53 : bool = aten::eq(%36, %24)
       = prim::If(%53)
        block0():
           = prim::RaiseException(%12, %11)
          -> ()
        block1():
          -> ()
      %56 : bool = aten::eq(%36, %25)
       = prim::If(%56)
        block0():
           = prim::RaiseException(%10, %11)
          -> ()
        block1():
          -> ()
      %59 : str = aten::format(%9, %36, %27)
       = prim::RaiseException(%59, %11)
      -> (%30)
```

</details>

In the graph above, the outermost `block0` performs the `aten::upsample_bilinear2d` computation, while the outermost `block1` consists entirely of `prim::RaiseException` calls to inform the user of dimensionality issues and other such occurrences. While helpful, our converter implementation of `aten::upsample_bilinear2d` should handle dimension issues and report these to the user, instead of depending on the `nn.Module` code to do so. As such, we can remove the `prim::RaiseException` calls here.

Note further that there are many dangling `prim::If` statements in the above code, which are never assigned to any variable. These seem difficult to remove, as `node->destroy()` seems to segfault on these.
## Desired Solution
The desired solution in this case is a lowering pass which detects whether a control-flow block has a guaranteed exception along one of the paths, and if so, eliminate the control flow entirely and replace the `prim::If` with the nodes contained in the valid path. We already use [`torch::jit::EliminateExceptions`](https://github.com/pytorch/pytorch/blob/main/torch/csrc/jit/passes/remove_exceptions.h), however this pass only replaces the control flow boolean and not its logic, and seems to halt computation indefinitely in certain cases (see #1823). We also use the following lowering pass:
https://github.com/pytorch/TensorRT/blob/1d78f436a9a423a8486338ecefd972fd08777f63/core/lowering/passes/exception_elimination.cpp#L103
[...]

PR #1785, opened 2023-03-29
fix: include str ivalue type conversion

bowang007:
# Description

Fixes #1756 

## Type of change

Please delete options that are not relevant and/or add your own.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

# Checklist:

- [ ] My code follows the style guidelines of this project (You can use the linters)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

Issue #1282, opened 2022-08-18
❓ [Question] How do you solve the error: Expected Tensor but got Uninitialized?


Mark-M2L:
## ❓ Question

Currently, I am compiling a custom segmentation model using torch_tensorrt.compile(), using a model script obtained from jit. The code to compile is as follows:

```
scripted_model = torch.jit.freeze(torch.jit.script(model))

inputs = [torch_tensorrt.Input(
            min_shape=[2, 3, 600, 400],
            opt_shape=[2, 3, 600, 400],
            max_shape=[2, 3, 600, 400],
            dtype=torch.float,
        )]
enabled_precisions = {torch.float, torch.half}

with torch_tensorrt.logging.debug():
    trt_ts_module = torch_tensorrt.compile(scripted_model, inputs=inputs, enabled_precisions=enabled_precisions)
```

The code fails to compile at the following step:
```
        a = self.compression(torch.cat(x_list, 1))
        b = self.shortcut(x)

        c = a + b

        return c
```
, throwing the following error:
```
Traceback (most recent call last):
  File "test.py", line 118, in <module>
    trt_ts_module = torch_tensorrt.compile(scripted_model, inputs=inputs, enabled_precisions=enabled_precisions)
  File "/home/oem/.pyenv/versions/ddrnet/lib/python3.8/site-packages/torch_tensorrt/_compile.py", line 115, in compile
    return torch_tensorrt.ts.compile(ts_mod, inputs=inputs, enabled_precisions=enabled_precisions, **kwargs)
  File "/home/oem/.pyenv/versions/ddrnet/lib/python3.8/site-packages/torch_tensorrt/ts/_compiler.py", line 113, in compile
    compiled_cpp_mod = _C.compile_graph(module._c, _parse_compile_spec(spec))
RuntimeError: Expected Tensor but got Uninitialized
```

It seems that some variable is uninitialized. However, the strange thing is that replacing the previous code with the following code pieces both compile:
```
        a = self.compression(torch.cat(x_list, 1))

        return a
```
and
```
        b = self.shortcut(x)

        return b
```
So, somehow taking the sum of these two tensors results in a failure to compile. Do you have any suggestions I can try such that this step compiles as well?

## What you have already tried
Tried adding the following two parameters to the compilation step as well:
```    
[...]

PR #1859, opened 2023-04-25
fix: Replace EliminateExceptions lowering pass

### LINKED ISSUE #1823: 🐛 [Bug] Cannot convert simple torchscript containing two torch.nn.Upsample operations
##  Bug Description

Scripting a simple "network" containing two `torch.nn.Upsample` modules and trying to convert the resulting torchscript does not work.

## To Reproduce

1. Generate a torchscript of the Network below, with `torch.jit.script`.
2. Try to convert to TensorRT with torch_tensorrt (I tried both in C++, on linux and in python)

```python
class Network(torch.nn.Module):
    def __init__(self):
        super(Network, self).__init__()
        # NOTE:
        # * Specifying a single float as scale_factor or the 2d tuple doesn't change the behavior
        self.upsample1 = torch.nn.Upsample(
            scale_factor=(2.0, 2.0), mode="bilinear", align_corners=False
        )
        self.upsample2 = torch.nn.Upsample(
            scale_factor=(2.0, 2.0), mode="bilinear", align_corners=False
        )

    def forward(self, X: torch.Tensor) -> torch.Tensor:
        # NOTE:
        # * Using the same self.upsample() module doesn't work as well
        # * Computing and returning out doesn't work as well
        out1 = self.upsample1(X)
        out2 = self.upsample2(X)
        # out = out1 + out2
        return out1
```

## Expected behavior

The conversions succeeds and a new valid torchscript is obtained.

## Environment

I managed to reproduce the bug both when using pytorch 1.11 and torch-tensorRT 1.1.0 and using pytorch 1.13.1 and torch-tensorRT main.

### Torch-TensorRT 1.1.0

 - Torch-TensorRT Version (e.g. 1.0.0): 1.1.0
 - PyTorch Version (e.g. 1.0): 1.11
 - CPU Architecture: x86_64
 - OS (e.g., Linux): Linux, Ubuntu 20.04
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source): pip for the python package used to generate the torchscript, source for the C++ dependency linked to Torch-TensorRT
 - Build command you used (if compiling from source): Compiled with CMake
 - Python version: 3.8.16
 - CUDA version: 11.3
 - GPU models and configuration: NVIDIA GeForce RTX 2070 SUPER, SM Capability: 7.5

When using torch-tensorRT 1.1.0, I get the following error:
```
DEBUG: [Torch-TensorRT] - Registering input/output torch::jit::Value for segmented graphs
terminate called after throwing an instance of 'c10::Error'
  what():  Expected Tensor but got Uninitialized


Exception raised from reportToTensorTypeError at /home/ubuntu/buildAgent/temp/buildTmp/conan_home/.conan/data/libtorch/1.11.0-5/cognex/stable/build/b8fcba865a68ae644e8a8bfa868ef00d13dc8a17/source_subfolder/aten/src/ATen/core/ivalue.cpp:908 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x6b (0x7f66313e708b in /mnt/caches/conan/data/libtorch/1.11.0-5/cognex/stable/package/b8fcba865a68ae644e8a8bfa868ef00d13dc8a17/lib/libc10.so)
frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xce (0x7f66313e2a5e in /mnt/caches/conan/data/libtorch/1.11.0-5/cognex/stable/package/b8fcba865a68ae644e8a8bfa868ef00d13dc8a17/lib/libc10.so)
frame #2: c10::IValue::reportToTensorTypeError() const + 0x64 (0x7f6634480064 in /mnt/caches/conan/data/libtorch/1.11.0-5/cognex/stable/package/b8fcba865a68ae644e8a8bfa868ef00d13dc8a17/lib/libtorch_cpu.so)
frame #3: torch_tensorrt::core::partitioning::getSegmentsOutputByRunning(torch_tensorrt::core::partitioning::SegmentedBlock&, std::unordered_map<torch::jit::Value const*, c10::IValue, std::hash<torch::jit::Value const*>, std::equal_to<torch::jit::Value const*>, std::allocator<std::pair<torch::jit::Value const* const, c10::IValue> > >&, torch_tensorrt::core::partitioning::PartitionInfo const&) + 0x15a7 (0x7f6639fd8797 in /mnt/caches/conan/data/torch-tensorrt/1.1.0-6/cognex/stable/package/c2bc9df050867ddb44d73bc998d3c8ceef4f763a/lib/libtorchtrt.so)
frame #4: torch_tensorrt::core::partitioning::runShapeAnalysis(std::vector<torch_tensorrt::core::partitioning::SegmentedBlock, std::allocator<torch_tensorrt::core::partitioning::SegmentedBlock> >&, std::unordered_map<torch::jit::Value const*, c10::IValue, std::hash<torch::jit::Value const*>, std::equal_to<torch::jit::Value const*>, std::allocator<std::pair<torch::jit::Value const* const, c10::IValue> > >&, torch_tensorrt::core::partitioning::PartitionInfo const&) + 0x81 (0x7f6639fda581 in /mnt/caches/conan/data/torch-tensorrt/1.1.0-6/cognex/stable/package/c2bc9df050867ddb44d73bc998d3c8ceef4f763a/lib/libtorchtrt.so)
frame #5: torch_tensorrt::core::partitioning::Partition(torch::jit::Block*, std::unordered_map<torch::jit::Value const*, c10::IValue, std::hash<torch::jit::Value const*>, std::equal_to<torch::jit::Value const*>, std::allocator<std::pair<torch::jit::Value const* const, c10::IValue> > >&, torch_tensorrt::core::partitioning::PartitionInfo const&) + 0x19b (0x7f6639fe587b in /mnt/caches/conan/data/torch-tensorrt/1.1.0-6/cognex/stable/package/c2bc9df050867ddb44d73bc998d3c8ceef4f763a/lib/libtorchtrt.so)
frame #6: torch_tensorrt::core::ConstructFallbackGraph(torch::jit::Module&, torch::jit::Block*, std::unordered_map<torch::jit::Value const*, c10::IValue, std::hash<torch::jit::Value const*>, std::equal_to<torch::jit::Value const*>, std::allocator<std::pair<torch::jit::Value const* const, c10::IValue> > >, torch_tensorrt::core::CompileSpec, std::map<torch::jit::Value*, c10::IValue, std::less<torch::jit::Value*>, std::allocator<std::pair<torch::jit::Value* const, c10::IValue> > >) + 0xff (0x7f663a00006f in /mnt/caches/conan/data/torch-tensorrt/1.1.0-6/cognex/stable/package/c2bc9df050867ddb44d73bc998d3c8ceef4f763a/lib/libtorchtrt.so)
frame #7: torch_tensorrt::core::CompileGraph(torch::jit::Module const&, torch_tensorrt::core::CompileSpec) + 0x980 (0x7f663a002a00 in /mnt/caches/conan/data/torch-tensorrt/1.1.0-6/cognex/stable/package/c2bc9df050867ddb44d73bc998d3c8ceef4f763a/lib/libtorchtrt.so)
frame #8: torch_tensorrt::torchscript::compile(torch::jit::Module const&, torch_tensorrt::torchscript::CompileSpec) + 0x5b7 (0x7f6639e9e7f7 in /mnt/caches/conan/data/torch-tensorrt/1.1.0-6/cognex/stable/package/c2bc9df050867ddb44d73bc998d3c8ceef4f763a/lib/libtorchtrt.so)
frame #9: <unknown function> + 0x5ec8 (0x56183fd18ec8 in ./build/src/interpolate_tensorrt)
frame #10: __libc_start_main + 0xf3 (0x7f65fe02f083 in /lib/x86_64-linux-gnu/libc.so.6)
frame #11: <unknown function> + 0x58ee (0x56183fd188ee in ./build/src/interpolate_tensorrt)
```

That looked kind of similar to [this issue](https://github.com/pytorch/TensorRT/issues/1282) and patching Torch-TensorRT with [this PR](https://github.com/pytorch/TensorRT/pull/1345) makes the behavior exactly the same as in the second case (i.e. when using pytorch 1.13.1 and torch-tensorRT main).

### Torch-TensorRT main (commit 861edd03a510c600146575836b02c993ac386b00)

 - Torch-TensorRT Version (e.g. 1.0.0): main, commit 861edd03a510c600146575836b02c993ac386b00
 - PyTorch Version (e.g. 1.0): 1.13.1
 - CPU Architecture: x86_64
 - OS (e.g., Linux): Linux, Ubuntu 20.04
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source): pip for the python package used to generate the torchscript, source for the C++ dependency linked to Torch-TensorRT
 - Build command you used (if compiling from source): Compiled with CMake
 - Python version: 3.8.16
 - CUDA version: 11.8
 - GPU models and configuration: NVIDIA GeForce RTX 2070 SUPER, SM Capability: 7.5

When using torch-TensorRT main, the conversion just hangs for ever after 
```
GRAPH: [Torch-TensorRT] - Torch-TensorRT.TorchScript Graph Lowering
```

## Additional context

Interestingly, when using the tracing mechanism of pytorch to generate the torchscript, everything _seems_ fine (I didn't check the results, but the conversion finishes properly).
Also, when scripting with pytorch 1.9, everything works fine 🤯 

The thing I noticed is that pytorch changed slightly the `torch.nn.interpolate` API and I am wondering if that could explain (at least partially) the problem:
- Before pytorch 1.11: `torch.nn.functional.interpolate(input, size=None, scale_factor=None, mode='nearest', align_corners=None, recompute_scale_factor=None)`
- PyTorch 1.11 and more recent: `torch.nn.functional.interpolate(input, size=None, scale_factor=None, mode='nearest', align_corners=None, recompute_scale_factor=None, antialias=False)`

See the attached .zip file containing a python file to generate the torchscript.
[upsample.zip](https://github.com/pytorch/TensorRT/files/11234555/upsample.zip)

Let me know if you need more details to reproduce the problem. Thanks!

gs-olive:
# Description

- Torch-TensorRT has evaluator support for exception elimination and this particular lowering pass causes certain models to halt compilation indefinitely
- Modify this lowering pass to use a safer node-replacement system via direct node input replacement

- Add new `EliminateExceptionsSafe` lowering pass, which has functionally the same task as that of `EliminateExceptions`, but with a safer replacement scheme
- Update EliminateExceptions to use direct node input replacement instead of `replaceAllUsesWith` to avoid issue with invalid IR causing program halting
- Add testing for new lowering pass

The [`torch::jit::EliminateExceptions`](https://github.com/pytorch/pytorch/blob/main/torch/csrc/jit/passes/remove_exceptions.h) lowering pass is listed as "risky" in the documentation, and as per #1823, can also cause compilation to halt on certain graphs. We should remove this pass while awaiting a more robust solution through the implementation of #1842 (giving the user the option to eliminate exceptions more effectively in the graph via a compile-time flag). Torch-TRT already has an evaluator for `prim::RaiseException`, and can thus handle these operators:
https://github.com/pytorch/TensorRT/blob/09e15b25e3eee0110bfc51db287dfc7e167a6b17/core/conversion/evaluators/prim.cpp#L337-L343

**Note:** This PR does not remove the existing `EliminateExceptionOrPassPattern` lowering pass, as that pass is intended for a more specialized exception case.

Addresses #1823 
Relevant to #1842

## Type of change

# Checklist:

- [ x] I have performed a self-review of my own code

### DIFF
--- a/core/lowering/passes/exception_elimination.cpp
+++ b/core/lowering/passes/exception_elimination.cpp
@@ -1,5 +1,6 @@
 #include "torch/csrc/jit/ir/alias_analysis.h"
 #include "torch/csrc/jit/jit_log.h"
+#include "torch/csrc/jit/passes/constant_pooling.h"
 #include "torch/csrc/jit/passes/constant_propagation.h"
 #include "torch/csrc/jit/passes/dead_code_elimination.h"
 #include "torch/csrc/jit/passes/guard_elimination.h"
@@ -106,6 +107,71 @@ void EliminateExceptionOrPassPattern(std::shared_ptr<Graph> graph) {
   if (graph) {
     LOG_GRAPH("Post Eliminate Exception or Pass Patterns: " << *graph);
   }
+}
+
+/*
+  Below is a fork of the torch::jit::EliminateExceptions pass, with node replacement
+  using replaceAllUsesDominatedByNodeWith instead of replaceAllUsesWith,
+  so as to not invalidate the IR in challenging cases, such as nested Ifs
+
+  Original Source from which it was adapted:
+  https://github.com/pytorch/pytorch/blob/c29ab84115f40614d04e4557ea2e1ac40b7aa75c/torch/csrc/jit/passes/remove_exceptions.cpp
+*/
+
+bool certainlyThrows(Block* block) {
+  // A block certainly throws an exception if it contains
+  // the prim::RaiseException operation
+  for (Node* n : block->nodes()) {
+    if (n->kind() == prim::RaiseException) {
+      return true;
+    }
+  }
+  return false;
+}
+
+void EliminateExceptionsSafe(Block* block) {
+  auto graph = block->owningGraph();
+  // Generate false and true constant placeholders
+  Value* false_const = graph->insertConstant(IValue(false));
+  Value* true_const = graph->insertConstant(IValue(true));
+
+  // For each prim::If node, if either block certainly throws an exception,
+  // replace input conditional of the node input with the logical opposite
+  for (Node* n : block->nodes()) {
+    if (n->kind() == prim::If) {
+      Block* true_block = n->blocks()[0];
+      Block* false_block = n->blocks()[1];
+      bool removed_exception = false;
+      Value* input_value_replacement;
+
+      // If the block throws an exception, replace input with logical opposite
+      if (certainlyThrows(true_block)) {
+        removed_exception = true;
+        input_value_replacement = false_const;
+      } else if (certainlyThrows(false_block)) {
+        removed_exception = true;
+        input_value_replacement = true_const;
+      }
+
+      // Log node and perform input replacement
+      if (removed_exception) {
+        LOG_WARNING("Detected and removing exception in TorchScript IR for node: " << util::node_info(n));
+        n->insertInput(0, input_value_replacement);
+        n->removeInput(1);
+      }
+    }
+
+    // Inspect and replace all instances within subblocks of the current node
+    for (Block* subblock : n->blocks()) {
+      EliminateExceptionsSafe(subblock);
+    }
+  }
+}
+
+void EliminateExceptionsSafe(std::shared_ptr<Graph>& graph) {
+  EliminateExceptionsSafe(graph->block());
+  ConstantPropagation(graph);
+  ConstantPooling(graph);
 }
 
 } // namespace passes
--- a/tests/core/lowering/test_exception_elimination_pass.cpp
+++ b/tests/core/lowering/test_exception_elimination_pass.cpp
@@ -1,6 +1,10 @@
 #include "core/lowering/passes/passes.h"
 #include "gtest/gtest.h"
+#include "tests/util/util.h"
 #include "torch/csrc/jit/ir/irparser.h"
+#include "torch/csrc/jit/passes/canonicalize.h"
+#include "torch/csrc/jit/passes/constant_pooling.h"
+#include "torch/csrc/jit/passes/remove_exceptions.h"
 
 TEST(LoweringPasses, EliminateExceptionOrPassPattern_Block0) {
   // parseIR does not support " = prim::If(%51)" with no return value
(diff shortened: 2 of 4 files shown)


gcuendet:
That seems to help in certain cases in which the compilation was hanging, however, I am also now seeing some 
> "Expected Tensor but got Uninitialized"

error. Looking at #1282 , it seems that adding `torch::jit::EliminateExceptions` was the solution to such errors... 

My understanding of the interactions of these different pieces is limited, but I am wondering whether removing the call to `torch::jit::EliminateExceptions` is not just fixing one thing and breaking (again) another one.

gs-olive:
@gcuendet - Thank you for the follow-up. We are looking into the `Expected Tensor but got Uninitialized` error simultaneously with this change, as it seems the usage of `torch::jit::EliminateExceptions` is more of a temporary fix for that problem (see: https://github.com/pytorch/TensorRT/pull/1785#pullrequestreview-1365562514).

Do you have any examples of models you are willing to share which fail with the Uninitialized error, when the `torch::jit::EliminateExceptions` pass is removed?

gcuendet:
Thanks for the answer @gs-olive . I am trying to simplify our failing model to something minimal that I can share, but I am not sure I'll be able to 😞 . If I do I'll share, of course!

gs-olive:
Hi @gcuendet - I found the issue in `torch::jit::EliminateExceptions` originates from the calls to `replaceAllUsesWith`, and I've addressed this by replacing it with the safer alternative `replaceAllUsesDominatedByNodeWith`. This resolves the issue on the `nn.Upsample` case on my machine. Please let me know if it resolves that case while still retaining the benefits of the original `EliminateExceptions` on your models.

gs-olive:
Thank you for the detailed response! I agree with your statement and assessment that a proper solution for #1842 would address this issue. Another alternative is to support nodes such as `%60 : Tensor = prim::Uninitialized()`, since these can also show up in other contexts and should not fail. @bowang007 is working on something similar now, and that fix, in conjunction with this PR should help to resolve the issues relating to exceptions and uninitialized tensors, while the other PR is pending.

gcuendet:
Do you plan to open a PR in pytorch with that fix?

gs-olive:
I intend to open an issue with PyTorch first to assess whether the proposed fix is a viable solution to the issue and whether the intended behavior of that function is retained, then I will open a PR for it.

gs-olive:
Hi @gcuendet - I've made some updates to this lowering pass and added a few test cases. I have also verified that for the case in #1823, the model no longer halts mid-compilation (and compiled successfully on my machine, using this PR). Could you please verify on the model with the new updates as well?

core/lowering/passes/exception_elimination.cpp
```diff
@@ -108,6 +109,59 @@ void EliminateExceptionOrPassPattern(std::shared_ptr<Graph> graph) {
...
+  // Replace all uses of the node input with the logical opposite
+  for (Node* n : block->nodes()) {
+    if (n->kind() == prim::If) {
+      Block* true_block = n->blocks()[0];
+      Block* false_block = n->blocks()[1];
+
+      if (certainlyThrows(true_block)) {
+        n->input(0)->replaceAllUsesDominatedByNodeWith(n, false_const);
+      } else if (certainlyThrows(false_block)) {
+        n->input(0)->replaceAllUsesDominatedByNodeWith(n, true_const);
```
bowang007, line 148: This does not make sense. 
If you check what's actually going on for `prim::If` [here](https://github.com/pytorch/pytorch/blob/8e0837cf84d9d2cc4996ba696ee41b5dd2d008b9/torch/csrc/jit/ir/ir_views.h), you will find that: 
1. every if node has some blocks and attributes
2. for if node, `IfNode->input(0)` is actually the condition
So, in PyTorch's pass implementation [here](https://github.com/pytorch/pytorch/blob/c29ab84115f40614d04e4557ea2e1ac40b7aa75c/torch/csrc/jit/passes/remove_exceptions.cpp#L28), what's happening is that if will go through the 2 blocks in `prim::If`, if one of that block contains a `prim::RaiseException` node, it will RESET the `prim::If` input(condition) to enforce it to run another flow. 
For example, if the original condition is `True` it should go to the first block(true block) but it finds out the first block contains a `prim::RaiseException` node, so it rewrite the condition to be `False` to make sure that it won't go to the first block to avoid the exception. 

Moreover, if you go to the definition of the function `replaceAllUsesDominatedByNodeWith` [here](https://github.com/pytorch/pytorch/blob/31f311a816c026bbfca622d6121d6a7fab44260d/torch/csrc/jit/ir/ir.h#L297), you will find that this actually replace the value dominated by the node. In graph theory, [dominate](https://en.wikipedia.org/wiki/Dominator_(graph_theory)) means that the value is dominated by the node. For example, node A dominate value x means that in that graph it must flow from node A to value x. So, for this line `        n->input(0)->replaceAllUsesDominatedByNodeWith(n, true_const);` what you are doing is that you are going to replace the `prim::If` condition value which is dominated by that `prim::If` node. However, `prim::If` condition value dominate that `prim::If` node, in other words, you are replacing nothing here because `n->input(0)` is never dominated by that `n` node. 
So I think you are basically doing nothing here. 
Above is my understanding, please correct me if I'm wrong.

gs-olive: Hi @bowang007 - thank you for the detailed review and comment. I agree with your observation about the replacement of all uses of the condition dominated by `n` as effectively doing nothing here. I've run + added some additional tests to the PR and I have now updated the call to **`replaceFirstUseWith`**, which I believe is the correct replacement in this case.

Based on your comments, I now think the reason for which the original halting bug in #1823 was occurring is the following:

```python
%out1.1 : Tensor = prim::If(%71)
    block0():
      ...
      -> (%73)
    block1():
         ...
        block0():
          ...
          -> ()
        block1():
          -> ()
        ...
       = prim::RaiseException(%76, %59)
      -> (%63)
  %out2.1 : Tensor = prim::If(%71)
    block0():
      ...
      -> (%82)
    block1():
      ...
```

In the graph above, the condition `%71` appears as the predicate of two _distinct_ `prim::If` nodes. When we call **`replaceAllUsesWith`**, as `torch::jit::passes::EliminateExceptions` does, this replaces downstream `prim::If` users of the same condition, and seems to cause an [infinite loop here](https://github.com/pytorch/pytorch/blob/31f311a816c026bbfca622d6121d6a7fab44260d/torch/csrc/jit/ir/ir.cpp#LL931C25-L931C25). By replacing only the **first** use of the condition, however, this would ensure _only_ the condition of the `prim::If` being directly analyzed will be replaced, avoiding this loop. What do you think of the new changes?

narendasan: Not too much to add here, the discussion seems accurate. Here is the official documentation on block and conditional semantics in PyTorch https://github.com/pytorch/pytorch/blob/main/torch/csrc/jit/OVERVIEW.md#block

tests/core/lowering/test_exception_elimination_pass.cpp
```diff
@@ -169,3 +173,189 @@ TEST(LoweringPasses, EliminateExceptionOrPassPattern_Negative) {
   }
   EXPECT_EQ(1, if_count);
 }
+
+TEST(LoweringPasses, EliminateExceptionsSafeIfBlock) {
```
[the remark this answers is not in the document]
gs-olive: The difference is that the graph will be collapsed since the exception will be removed from the graph. This is the same as the graph effect of the `torch::jit::EliminateExceptions` lowering pass. For instance, the change would be:

**Original Graph:**
```
    graph(%x, %y):
      %dim : int = aten::dim(%x)
      %48 : int = prim::Constant[value=2]()
      %66 : bool = aten::eq(%48, %dim)
      %45 : str = prim::Constant[value="EXCEPTION"]()
      %4 : Tensor = prim::If(%66)
        block0():
          = prim::RaiseException(%45)
          -> (%x)
        block1():
          %res = aten::mul(%x, %y)
          -> (%res)
      return (%4)
```

**New Graph:**
```
    graph(%x : Tensor,
              %y : Tensor):
      %6 : Tensor = aten::mul(%x, %y)
      return (%6)
```

gs-olive:
Hi @bowang007 - thanks for the comment, this makes sense.

1. I agree that the pass can be risky in the sense that it removes exceptions. The main rationale for removing these exceptions would be that our converters already handle the edge cases which these exceptions are checking for, so we shouldn't need the extra logic and conditionals they bring with them. For instance, the conditional shown in the comment above could cause unnecessary fallback which can affect performance. `TRT` and our converters already apply a large number of input validation checks to address these cases.
2. Removing the `torch::jit::EliminateExceptions` pass entirely could also be a viable solution to resolve the compilation halting here, I agree.
3. This is a good point, thanks for bringing this up. I've updated the implementation to just directly modify the `prim::If` node's input, to address the case you mentioned.

core/lowering/passes/exception_elimination.cpp
```diff
@@ -108,6 +109,62 @@ void EliminateExceptionOrPassPattern(std::shared_ptr<Graph> graph) {
...
+
+      // If the block throws an exception, replace input with logical opposite
+      if (certainlyThrows(true_block)) {
+        n->insertInput(0, false_const);
+        n->removeInput(1);
+      } else if (certainlyThrows(false_block)) {
+        n->insertInput(0, true_const);
+        n->removeInput(1);
+      }
+    }
```
gs-olive, line 153: Directly modifies inputs to the `prim::If` node _only_, to avoid incorrect boolean modifications elsewhere.

core/lowering/passes/exception_elimination.cpp
```diff
@@ -108,6 +109,71 @@ void EliminateExceptionOrPassPattern(std::shared_ptr<Graph> graph) {
...
+      } else if (certainlyThrows(false_block)) {
+        removed_exception = true;
+        input_value_replacement = true_const;
+      }
+
+      // Log node and perform input replacement
+      if (removed_exception) {
+        LOG_WARNING("Detected and removing exception in TorchScript IR for node: " << util::node_info(n));
+        n->insertInput(0, input_value_replacement);
+        n->removeInput(1);
```
gs-olive, line 160: Directly modifies inputs to the `prim::If` node _only_, to avoid incorrect boolean modifications elsewhere.

Logs a warning informing the user that an exception was automatically removed from the TorchScript IR.

narendasan:
Have we done testing on what happens if we just remove the pass? iirc, we treat exceptions as compile time errors though the evaluator, the original purpose was to have less nops in the graph and remove unnecessary conditionals

gs-olive:
I haven't done specific testing (aside from CI) on removing the pass, but I would expect a performance degradation on models which have conditional exception logic, since the `prim::If` would cause fallback in the graph, increasing graph segmentation.

core/lowering/passes/exception_elimination.cpp
```diff
@@ -1,3 +1,4 @@
+#include <torch/csrc/jit/passes/constant_pooling.h>
```
vadimpdv, line 1: Hey! Sorry to barge in, and thank you for handling this issue! (which I also encountered and found my way here after analyzing gdb stack trace)
Just a nit here, these should be probably changed to quotes.

bowang007 (approved):
LGTM

PR #1878, opened 2023-05-01
DLFW changes

apbose:
The following files are modified-
1. _docker/WORKSPACE.ngc_ changing the python lib files to python 3.10. Hardcoded at present
2. _noxfile.py_ changing the PYT_PATH to python3.10


apbose:
Any thoughts on removing the hardcoded paths? Should be do something like sym links as we do in torch TRT in DLFW?

narendasan (approved):
LGTM

PR #1903, opened 2023-05-10
Reorg for converters elu and selu (FX Converter Refactor [7/N]) <Target: converter_reorg_proto>

apbose:


wushirong (approved):

py/torch_tensorrt/fx/converters/aten_ops_converters.py
```diff
@@ -170,6 +170,33 @@ def aten_ops_div(
...
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+
+    if len(args) > 2:
+        return activation.selu(
+            network,
+            target,
+            SourceIR.ATEN,
+            name,
+            args[0],
+        )
```
wushirong, line 189: what is the purpose of this if logic?

apbose: The aten trace for both selu and elu comes as torch.ops.aten.elu.default, but with different args. elu has args[0] and args[1] for alpha, while selu has only one arg args[0]. Just a way to differentiate between the two.

wushirong (approved):

PR #1906, opened 2023-05-10
Moving elementwise core to impl - rsub(FX Converter Refactor [4/N]) <Target: converter_reorg_elementwise>

apbose:


py/torch_tensorrt/fx/test/converters/aten_op/test_rsub_aten.py
```diff
@@ -0,0 +1,29 @@
...
+    @parameterized.expand(
+        [
+            ("2d_dim_alpha", (2, 1), 2),
+            ("3d_dim_alpha", (2, 1, 2), 2),
+        ]
+    )
+    def test_rsub(self, _, x, alpha):
+        class rsub(nn.Module):
+            def forward(self, input):
+                return torch.rsub(input, input, alpha=alpha)
```
wushirong, line 18: Provide 2 different input tensor to test module, otherwise you already get 0 output.

py/torch_tensorrt/fx/converters/impl/elementwise/ops.py
```diff
@@ -139,3 +139,37 @@ def rsqrt(
...
+        )
+
+    output = convert_binary_elementwise(
+        network,
+        target,
+        source_ir,
+        f"{name}_output",
+        trt.ElementWiseOperation.SUB,
+        input,
+        scaled_tensor_output,
```
wushirong, line 172: according to https://github.com/pytorch/pytorch/issues/64775
rsub for (x-y) should be rsub(y, x)

py/torch_tensorrt/fx/test/converters/aten_op/test_rsub_aten.py
apbose: Since it is has an alpha, the output will not be 0 in all cases. But makes sense to create a test case with different inputs.

py/torch_tensorrt/fx/converters/impl/elementwise/ops.py
apbose: Ok, I will change this!

py/torch_tensorrt/fx/test/converters/aten_op/test_rsub_aten.py
apbose: Addressed this. Although I have created the new test case from the class `torch.testing._internal.common_utils.test_case` instead of `DispatchTestCase`, since` DispatchTestCase.run_test` does not seem to have the capability to handle multiple inputs. You could correct me if I am wrong, but I could not find a way to work that out

py/torch_tensorrt/fx/converters/impl/elementwise/ops.py
apbose: Although the above issue states this, the tests are passing with same as well as multiple inputs which makes it seem like it is taking x-y as rsub(x,y). Looking into this further.

gs-olive (commented):
It seems that AOT autograd in newer nightly versions is reducing `torch.ops.aten.rsub.Tensor` to `torch.ops.aten.sub.Tensor` with the arguments swapped. This causes an accuracy error due to some issues in our `torch.ops.aten.sub.Tensor` implementation. For instance, consider the expression `rsub(x, y, alpha=2)`. This should evaluate to `y - 2*x`, but the current `torch.ops.aten.sub.Tensor` implementation does not use the alpha, and will return `y-x` instead:
https://github.com/pytorch/TensorRT/blob/6ceaed8985092c70548c3cdec13a0ff67b8fde0c/py/torch_tensorrt/fx/converters/aten_ops_converters.py#L393-L405
The argument swapping is something like:
```python
%rsub : [#users=1] = call_function[target=torch.rsub](args = (%1, %2), kwargs = {alpha: 2})
##### Translated to 
%sub : [#users=1] = call_function[target=torch.ops.aten.sub.Tensor](args = (%2, %1), kwargs = {alpha: 2})
```
The second line above throws an accuracy error currently in #2070, and may show the same error here if rebased to `main`.

PR #1916, opened 2023-05-12
Moving slice to impl - slice (FX Converter Refactor [16/N]) - <Target: converter_reorg_elementwise>

apbose:
]

### DIFF
--- a/py/torch_tensorrt/fx/converters/aten_ops_converters.py
+++ b/py/torch_tensorrt/fx/converters/aten_ops_converters.py
@@ -23,6 +23,7 @@ from ..utils import get_dynamic_dims, torch_dtype_from_trt, torch_dtype_to_trt
 from .converter_utils import *  # noqa: F403
 import torch_tensorrt.fx.tracer.acc_tracer.acc_utils as acc_utils
 from torch_tensorrt.fx.converters.impl import activation
+from torch_tensorrt.fx.converters.impl.elementwise import trunc_div
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
@@ -161,9 +162,7 @@ def aten_ops_div(
             network, target, None, kwargs_new, name
         )
     elif rounding_mode == "trunc":
-        return acc_ops_converters.acc_ops_trunc_div(
-            network, target, None, kwargs_new, name
-        )
+        return trunc_div(network, target, SourceIR.ATEN, name, args[0], args[1])
     else:
         raise RuntimeError(
             f"Target {target} does not support rounding mode {rounding_mode}"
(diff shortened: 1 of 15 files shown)


wushirong (approved):

py/torch_tensorrt/fx/converters/aten_ops_converters.py
```diff
@@ -465,6 +502,27 @@ def aten_ops_sym_numel(
...
+    return slice_op(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+        args[2],
+        args[3],
+        args[4],
```
gs-olive, line 522: This should check if the arguments list is long enough to enable the `step` option. Many models do not specify the `step` and it is omitted from the args list, causing an `IndexError`. Potentially the `or_none` function in this file could be modified slightly to help here:

```python
def arg_bounds_check(args, i, replacement=None):
    return args[i] if len(args) > i else replacement

...

# Replace step
arg_bounds_check(args, 4, replacement=1):
```

py/torch_tensorrt/fx/converters/aten_ops_converters.py
```diff
@@ -465,6 +502,27 @@ def aten_ops_sym_numel(
...
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return slice_op(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+        args[2],
+        args[3],
```
gs-olive, line 521: For `args[3]`, as in the TorchScript `slice` converter, the `aten` opset sometimes uses `9223372036854775807` to represent the entire dimension. We might use a check like `args[3] if args[3] != 9223372036854775807 else args[0].shape[args[1]]` to deal with this case.

py/torch_tensorrt/fx/converters/aten_ops_converters.py
apbose: Ok I will implement this change. Do you have a model that can be run to come across this case?

py/torch_tensorrt/fx/converters/aten_ops_converters.py
apbose: Looking at this.

py/torch_tensorrt/fx/converters/aten_ops_converters.py
gs-olive: Yes, MobileNetV2, which is tested in the Dynamo E2E test cases elicits this scenario:
https://github.com/pytorch/TensorRT/blob/bd9c29ab046df4605107b8fc530f353c054a5fc3/py/torch_tensorrt/dynamo/test/test_dynamo_backend.py#L50-L78

narendasan:
Closing in favor of https://github.com/pytorch/TensorRT/pull/2070

PR #1917, opened 2023-05-12
Moving elementwise core to impl - matmul(FX Converter Refactor [17/N]) <Target: converter_reorg_elementwise>

apbose:


py/torch_tensorrt/fx/converters/impl/elementwise/ops.py
```diff
@@ -139,3 +141,37 @@ def rsqrt(
     )
 
     return output
+
+
+def matrix_multiply(
```
narendasan, line 146: Put matmul in its own file

apbose: Ok  I will change that.

Issue #1834, opened 2023-04-18
🐛 [Bug] Expected ivalues_maps.count(input) to be true but got false


Charlyo:
##  Bug Description

RuntimeError: [Error thrown at core/partitioning/shape_analysis.cpp:183] Expected ivalues_maps.count(input) to be true but got false
Could not find torch::jit::Value* hidden.1 produced from %hidden.1 : (Tensor, Tensor) = prim::TupleConstruct(%440, %440) in lowering graph for mini graph input.

## To Reproduce

```
hidden: Tuple[torch.Tensor, torch.Tensor] = (
            torch.zeros(
                size=(batch_size, self.hidden_size),
                dtype=torch.float,
                device=self.device
            ),
            torch.zeros(
                size=(batch_size, self.hidden_size),
                dtype=torch.float,
                device=self.device
            )
        )
```

## Environment

> Build information about Torch-TensorRT can be found by turning on debug messages

torch==2.1.0.dev20230418+cu117
torch-tensorrt==1.4.0.dev0+a245b861

 - CPU Architecture: Intel 10750h
 - OS (e.g., Linux):   Linux
 - Are you using local sources or building from archives: nvidia tensorrt 8.5.3.1
 - Python version: 3.8
 - CUDA version: 11.7

gs-olive:
Hello - could you please share any information/sample of the model being compiled which led to the error, and/or the full debug logs associated with the failure?

Related: #1815

Charlyo:
With just a simple class that has this part of code in the forward. Being `batch_size` and `self.hidden_size` any positive integer

Charlyo:
@gs-olive the hidden variable was input for an attentioncell:


`class AttentionCell(nn.Module):
    """Attention cell class."""

    def __init__(self, input_size, hidden_size, num_embeddings):
        super(AttentionCell, self).__init__()
        self.i2h = nn.Linear(input_size, hidden_size, bias=False)
        self.h2h = nn.Linear(hidden_size, hidden_size)
        # either i2i or h2h should have bias
        self.score = nn.Linear(hidden_size, 1, bias=False)
        self.rnn = nn.LSTMCell(input_size + num_embeddings, hidden_size)
        self.hidden_size = hidden_size

    def forward(
        self,
        prev_hidden: Tuple[torch.Tensor, torch.Tensor],
        batch_h,
        char_onehots
    ):
        # [batch_size x num_encoder_step x num_channel] ->
[...]

Issue #1815, opened 2023-04-10
🐛 [Bug] Could not find `torch::jit::Value*` error encountered during compilation


gs-olive:
##  Bug Description

When compiling the [transformer-xl model](https://huggingface.co/docs/transformers/model_doc/transfo-xl) using Torch-TensorRT, the following error is encountered:

```python
RuntimeError: [Error thrown at core/partitioning/shape_analysis.cpp:183] Expected ivalues_maps.count(input) to be true but got false
    Could not find torch::jit::Value* 1822 produced from %1822 : Tensor = aten::mul(%attn_score.2, %self.layers.0.dec_attn.scale)
```

## To Reproduce

1. Run torch_tensorrt.compile with transformer-xl model as input, using fp32 precision.
2. Choose fixed input sizes and enable truncate_long_and_double with 12 GB workspace.

## Expected behavior

Model should successfully compile to Torch-TRT. Specifically, this `torch::jit::Value*` error should not arise during compilation.

## Environment

 - Torch-TensorRT Version (e.g. 1.0.0): 09cd47a0
 - PyTorch Version (e.g. 1.0): `2.1.0.dev20230314+cu117`

Issue #1756, opened 2023-03-21
🐛 [Bug] _Expected to find type str for value why_not_sparsity_fast_path.76 but get nothing._ when trying to partially compile TransformerEncoder


narendasan:
##  Bug Description

When trying to partially compile transformer encoder, you are met with the following error:

```
Expected to find type str for value why_not_sparsity_fast_path.76 but get nothing.
```

## To Reproduce

```py

import torch
import torch.nn as nn

import torch_tensorrt

class TransformerModel(nn.Module):
    def __init__(self, input_dim, hidden_dim, num_layers, num_heads):
        super(TransformerModel, self).__init__()

        # define embedding layer
        self.embedding = nn.Embedding(input_dim, hidden_dim)

        # define transformer encoder
        self.transformer_encoder = nn.TransformerEncoder(
            nn.TransformerEncoderLayer(d_model=hidden_dim, nhead=num_heads),
            num_layers=num_layers
        )

        # define output layer
        self.fc = nn.Linear(hidden_dim, input_dim)

    def forward(self, x):
        # apply embedding layer
        x = self.embedding(x)

        # apply transformer encoder
        x = self.transformer_encoder(x)

        # apply output layer
        x = self.fc(x)

        return x


model = TransformerModel(input_dim=100, hidden_dim=128, num_layers=2, num_heads=4)

input_data = torch.randint(low=0, high=100, size=(32,10)) # sequence length of 10, batch size of 32

input_data = input_data.to("cuda").to(torch.int)
model.to("cuda")
output = model(input_data)
model.eval()

inputs = [
    torch_tensorrt.Input(
        min_shape=[32,10],
        opt_shape=[32,10],
        max_shape=[32,10],
        dtype=torch.int,
    )]

enabled_precisions = {torch.float, torch.half}  # Run with fp16


with torch_tensorrt.logging.graphs():
    trt_ts_module = torch_tensorrt.compile(
        model, inputs=inputs, enabled_precisions=enabled_precisions, require_full_compilation=True
    )

result = trt_ts_module(input_data)

with open("../saved_models/trt_ts_module.ts", "wb") as f:
    torch.jit.save(trt_ts_module, f)
```

## Environment

> Build information about Torch-TensorRT can be found by turning on debug messages

 - Torch-TensorRT Version (e.g. 1.0.0): main
 - PyTorch Version (e.g. 1.0): 2.0
 - CPU Architecture: x64
 - OS (e.g., Linux): Linux
[...]

PR #1933, opened 2023-05-19
Fix: fix the bug that uninitialized tensor cannot be found

bowang007:
This PR fixes the bug that uninitialized tensor cannot be found. Specifically, uninitialized value is copied between subgraphs previously, this introduces unnecessary complexity and might incur bugs. In this PR, this uninitialized value is handled like constants: Create one uninitialized value for each subgraph in need.

Fixes # (issue)

## Type of change

### DIFF
--- a/core/partitioning/partitioning.cpp
+++ b/core/partitioning/partitioning.cpp
@@ -70,7 +70,7 @@ void setExplicitFallbackNodes(PartitioningCtx* ctx, torch::jit::Block* block) {
   const auto to_compile_sym = c10::Symbol::attr("to_compile");
 
   for (const auto n : nodes) {
-    if (n->kind() == torch::jit::prim::Constant) {
+    if (isConstantOrUninitialized(n)) {
       continue;
     }
 
@@ -107,7 +107,7 @@ void setNonTensorConnectedNodes(PartitioningCtx* ctx, std::vector<torch::jit::No
     q.pop();
     // for every node that produces this fallback node's NonTensor input, they should fallback too
     for (auto input : cur_node->inputs()) {
-      if (!isTensor(input) && input->node()->kind() != torch::jit::prim::Constant &&
+      if (!isTensor(input) && !isConstantOrUninitialized(input->node()) &&
           ctx->shouldNodeRunInTensorRT(input->node())) {
         ctx->setNodeExecutorDecision(input->node(), NodeExecutorDecision::kNON_TENSOR);
         q.push(input->node());
@@ -118,7 +118,7 @@ void setNonTensorConnectedNodes(PartitioningCtx* ctx, std::vector<torch::jit::No
       if (!isTensor(output)) {
         for (auto use : output->uses()) {
           auto node = use.user;
-          if (node->kind() != torch::jit::prim::Constant && ctx->shouldNodeRunInTensorRT(node)) {
+          if (!isConstantOrUninitialized(node) && ctx->shouldNodeRunInTensorRT(node)) {
             ctx->setNodeExecutorDecision(node, NodeExecutorDecision::kNON_TENSOR);
             q.push(node);
           }
@@ -128,11 +128,13 @@ void setNonTensorConnectedNodes(PartitioningCtx* ctx, std::vector<torch::jit::No
   }
 }
 
-std::set<torch::jit::Node*> getDependentNodes(torch::jit::Node* n) {
-  std::set<torch::jit::Node*> dependent_nodes;
+std::set<torch::jit::Node*> getUserNodes(torch::jit::Node* n) {
+  std::set<torch::jit::Node*> user_nodes;
   for (auto val : n->outputs()) {
     for (auto use : val->uses()) {
-      dependent_nodes.insert(use.user);
+      if (use.user->owningBlock()->owningNode())
+        user_nodes.insert(use.user->owningBlock()->owningNode());
+      user_nodes.insert(use.user);
     }
   }
   if (const auto* schema = n->maybeSchema()) {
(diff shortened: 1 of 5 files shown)


core/partitioning/partitioningctx/PartitioningCtx.cpp
```diff
@@ -15,7 +15,8 @@ PartitioningCtx::PartitioningCtx(torch::jit::Block* b, PartitioningInfo info)
 }
 
 void PartitioningCtx::_load_nodes_into_decision_map(torch::jit::Block* b) {
-  if (b->owningNode() && b->owningNode()->kind() == torch::jit::prim::Loop)
+  // won't load nodes if these nodes are in prim::loop or if these nodes are 2-level nested
+  if (b->owningNode() && (b->owningNode()->kind() == torch::jit::prim::Loop || b->owningNode()->owningBlock()->owningNode()))
```
gs-olive, line 19: Are there ways that `b->owningNode()->owningBlock()->owningNode()` could cause a segfault? For example, I am wondering about the case where the whole graph is a single block, then would `b->owningNode()->owningBlock()` be a `nullptr`?

gs-olive (commented):
Overall, the changes look good and this seems to resolve the main issue (pending #1823 verification). I just had some robustness questions about the nested `prim::If` check (mentioned below).

Also, I tested this PR out on #1756 and the first bug (`RuntimeError: Expected Tensor but got Uninitialized`) no longer appears, and it now shows:

```python
RuntimeError: [Error thrown at core/partitioning/shape_analysis.cpp:183] Expected ivalues_maps.count(input) to be true but got false
Could not find torch::jit::Value* x.14 produced from %x.14 : Tensor = aten::embedding(%self.embedding.weight, %x.1, %padding_idx.42, %self.transformer_encoder.layers.0.training.21, %self.transformer_encoder.layers.0.training.21) # /usr/local/lib/python3.8/dist-packages/torch/nn/functional.py:2225:11 in lowering graph for mini graph input.
```

The above error is similar to #1815, #1834 - both models which I believe have nested `prim::If` substructure.

core/partitioning/partitioningctx/PartitioningCtx.cpp
bowang007: It won't. 
In that case, `b->owningNode()` is `null` and it won't check the rest part of that if statement.

bowang007:
> Overall, the changes look good and this seems to resolve the main issue (pending #1823 verification). I just had some robustness questions about the nested `prim::If` check (mentioned below).
> 
> Also, I tested this PR out on #1756 and the first bug (`RuntimeError: Expected Tensor but got Uninitialized`) no longer appears, and it now shows:
> 
> ```python
> RuntimeError: [Error thrown at core/partitioning/shape_analysis.cpp:183] Expected ivalues_maps.count(input) to be true but got false
> Could not find torch::jit::Value* x.14 produced from %x.14 : Tensor = aten::embedding(%self.embedding.weight, %x.1, %padding_idx.42, %self.transformer_encoder.layers.0.training.21, %self.transformer_encoder.layers.0.training.21) # /usr/local/lib/python3.8/dist-packages/torch/nn/functional.py:2225:11 in lowering graph for mini graph input.
> ```
> 
> The above error is similar to #1815, #1834 - both models which I believe have nested `prim::If` substructure.



> Overall, the changes look good and this seems to resolve the main issue (pending #1823 verification). I just had some robustness questions about the nested `prim::If` check (mentioned below).
> 
> Also, I tested this PR out on #1756 and the first bug (`RuntimeError: Expected Tensor but got Uninitialized`) no longer appears, and it now shows:
> 
> ```python
> RuntimeError: [Error thrown at core/partitioning/shape_analysis.cpp:183] Expected ivalues_maps.count(input) to be true but got false
> Could not find torch::jit::Value* x.14 produced from %x.14 : Tensor = aten::embedding(%self.embedding.weight, %x.1, %padding_idx.42, %self.transformer_encoder.layers.0.training.21, %self.transformer_encoder.layers.0.training.21) # /usr/local/lib/python3.8/dist-packages/torch/nn/functional.py:2225:11 in lowering graph for mini graph input.
> ```
> 
> The above error is similar to #1815, #1834 - both models which I believe have nested `prim::If` substructure.

Yes, these models fails because of complicated embedded if blocks. 
We definitely need to optimize this handling for that.

gs-olive (commented):
Thanks for the replies and comments - this makes sense, and I think the bugfix looks good then. Could #1823 or similar be adapted into a test case for regression test use here?

bowang007:
> Thanks for the replies and comments - this makes sense, and I think the bugfix looks good then. Could #1823 or similar be adapted into a test case for regression test use here?

Sure, let me add that as a test case later.

core/partitioning/partitioning.cpp
```diff
@@ -118,7 +118,7 @@ void setNonTensorConnectedNodes(PartitioningCtx* ctx, std::vector<torch::jit::No
       if (!isTensor(output)) {
         for (auto use : output->uses()) {
           auto node = use.user;
-          if (node->kind() != torch::jit::prim::Constant && ctx->shouldNodeRunInTensorRT(node)) {
+          if (isConstantOrUninitialized(node) && ctx->shouldNodeRunInTensorRT(node)) {
```
gs-olive, line 121: Should this be `!isConstantOrUninitialized`?

gs-olive (commented):
Made a small comment on a conditional. Once that is addressed, it looks good.

core/partitioning/partitioning.cpp
bowang007: Yes, updated that just now. 
Thanks!

gs-olive (approved):
Looks good to me - just needs a rebase to get the test cases updated, then should be good to go.

PR #1944, opened 2023-05-22
fix/feat: Add Dynamo-only converter registry

### LINKED ISSUE #1942: Add Dynamo-Only Converter Registry

### LINKED ISSUE #2071: ✨[Feature] Augment `DYNAMO_CONVERTERS` to handle node arg checking
~75% of prototype detailed in RFC #2065

gs-olive:
# Description

- Add Dynamo converter registry which functions as a superset of the standard FX converter registry
- For use with new + experimental converters
- Uses custom decorator `dynamo_tensorrt_converter`
- Update references within Dynamo functions to use the converter registry `DYNAMO_CONVERTERS`

- Add custom class overriding default Dictionary class to access converters from various registries
- Add new dictionary type `Dict[Target, Sequence[ConverterSupport]]` as well as ConverterSupport class which stores a converter and its validation implementation
- Add unified `DYNAMO_CONVERTERS` dictionary which coalesces both the FX and Dynamo converter dictionaries and acts as a single unified dictionary

Fixes #1942 
Fixes #2071 

## Type of change

# Checklist:

  - Validated with CI + Converter Testing

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/converter_utils.py
+++ b/py/torch_tensorrt/dynamo/conversion/converter_utils.py
@@ -8,12 +8,12 @@ def dynamic_unsupported(node: torch.fx.Node) -> bool:
     ), "Inputs to validator functions must be FX Nodes"
 
     # Check node value itself
-    if node.meta["val"]._has_symbolic_sizes_strides:
+    if getattr(node.meta["val"], "_has_symbolic_sizes_strides", False):
         return False
 
     # Check node arguments individually
     if any(
-        arg.meta["val"]._has_symbolic_sizes_strides
+        getattr(arg.meta["val"], "_has_symbolic_sizes_strides", False)
         for arg in node.args
         if isinstance(arg, torch.fx.Node)
     ):
@@ -21,7 +21,7 @@ def dynamic_unsupported(node: torch.fx.Node) -> bool:
 
     # Check node keyword arguments individually
     if any(
-        kwarg.meta["val"]._has_symbolic_sizes_strides
+        getattr(kwarg.meta["val"], "_has_symbolic_sizes_strides", False)
         for kwarg in node.kwargs.values()
         if isinstance(kwarg, torch.fx.Node)
     ):
--- a/py/torch_tensorrt/dynamo/converter_registry.py
+++ b/py/torch_tensorrt/dynamo/converter_registry.py
@@ -1,0 +1,23 @@
+from typing import Any, Callable, Dict
+
+from torch.fx.node import Target
+from torch_tensorrt.fx.converter_registry import CONVERTERS
+
+DYNAMO_CONVERTERS: Dict[Target, Any] = dict(CONVERTERS)
+
+
+def dynamo_tensorrt_converter(
+    key: Target,
+    enabled: bool = True,
+) -> Callable[[Any], Any]:
+    def register_converter(converter):
+        DYNAMO_CONVERTERS[key] = converter
+        return converter
+
+    def disable_converter(converter):
+        return converter
+
+    if enabled:
+        return register_converter
+    else:
+        return disable_converter
--- a/py/torch_tensorrt/dynamo/converter_registry.py
+++ b/py/torch_tensorrt/dynamo/converter_registry.py
@@ -1,23 +1,327 @@
-from typing import Any, Callable, Dict
+from dataclasses import dataclass, field
+from typing import Any, Callable, Dict, Optional, Sequence, Union
+from enum import Enum, auto
 
-from torch.fx.node import Target
+from torch.fx.node import Target, Node, _get_qualified_name
 from torch_tensorrt.fx.converter_registry import CONVERTERS
 
-DYNAMO_CONVERTERS: Dict[Target, Any] = dict(CONVERTERS)
+
+class ConverterPriority(Enum):
+    """Enum to set a converter's priority in the registry"""
+
+    STANDARD = auto()
+    HIGH = auto()
+
+
+@dataclass(frozen=True)
+class ConverterSupport:
+    """Class representing a converter implementation and support function
+
+    Args:
+        converter_implementation: Function which converts said node to a TRT equivalent
+        capability_validator: Function which takes in a Node and returns a bool indicating
+            whether that node can be supported by its companion converter. Note that
+            this function must not modify the node or its graph
+    """
+
+    converter_implementation: Callable
+    capability_validator: Callable[[Node], bool] = field(default=lambda node: True)
+
+
+# Dictionary representing Dynamo aten-only converters
+# Each converter maps to a sequence of at least one ConverterSupport object(s)
+DYNAMO_ATEN_CONVERTERS: Dict[Target, Sequence[ConverterSupport]] = {}
 
 
 def dynamo_tensorrt_converter(
     key: Target,
     enabled: bool = True,
+    capability_validator: Optional[Callable[[Node], bool]] = None,
+    priority: ConverterPriority = ConverterPriority.STANDARD,
 ) -> Callable[[Any], Any]:
+    """Decorator for Dynamo TensorRT Converter
+
+    Registers the decorated function in the DYNAMO_ATEN_CONVERTERS registry
+
+    Args:
+        key: Node target for which the converter is implemented for
+            (for example, torch.ops.add.Tensor)
+        enabled: Whether the converter should be enabled/cached or not
+        capability_validator: Function which evaluates whether a node is valid for conversion
+            by the decorated converter. See ConverterSupport for more details.
+            Defaults to None, implying the capability_validator function is always true -
+            this means all nodes of "key" kind can be supported by this converter
+        priority: Converter's level of priority relative to other converters with the
+            same target
+    Returns:
+        The converter being decorated
+    """
+
     def register_converter(converter):
-        DYNAMO_CONVERTERS[key] = converter
+        """Helper function to register the converter, then return it"""
+        assert callable(converter), "Converter function must be callable"
+
+        # If no capability_validator function is specified, use the default function - always return true
+        if capability_validator is None:
+            converter_support = ConverterSupport(converter_implementation=converter)
+        else:
+            assert callable(
+                capability_validator
+            ), "Argument checking function must be callable"
+            converter_support = ConverterSupport(
+                converter_implementation=converter,
+                capability_validator=capability_validator,
+            )
+
+        # If a converter for this operator already exists, append the new converter to the list
+        # Otherwise, start a new list
+        if key in DYNAMO_ATEN_CONVERTERS:
+            # High priority converters are inserted at the front of the list,
+            # so they can be checked first by the registry
+            if priority is ConverterPriority.HIGH:
+                DYNAMO_ATEN_CONVERTERS[key].insert(0, converter_support)
+            else:
+                DYNAMO_ATEN_CONVERTERS[key].append(converter_support)
+        else:
+            DYNAMO_ATEN_CONVERTERS[key] = [converter_support]
+
         return converter
 
     def disable_converter(converter):
         return converter
 
+    # Select whether to cache/enable the converter
     if enabled:
         return register_converter
     else:
         return disable_converter
+
+
+class ConverterRegistry:
+    """Registry for storing multiple converter dictionaries
+
+    Capable of storing dictionaries with the following signature:
+    Dict[Target, Union[Callable, Sequence[ConverterSupport]]]
+
+    Also able to validate converter implementations against user-provided
+    argument-checking functions
+
+    Args:
+        registries: List of dictionaries representing converter registries.
+            The order of the provided dictionaries is the order in which they
+            will be traversed. This is only significant when using non-validated
+            methods.
+    """
+
+    def __init__(
+        self,
+        registries: Sequence[Dict[Target, Union[Callable, Sequence[ConverterSupport]]]],
+        registry_names: Optional[Sequence[str]] = None,
+    ):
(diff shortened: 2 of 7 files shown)


py/torch_tensorrt/dynamo/converter_registry.py
apbose, line 1: So this is the file where fx and dynamo converters are getting concatenated. The new converters could be registered using `dynamo_tensorrt_converter`, so wouldn't it be better to move the new converters to `py/torch_tensorrt/dynamo` instead of changing the existing converters in `py/torch_tensorrt/fx` from `tensorrt_converter` to `dynamo_tensorrt_converter`?

py/torch_tensorrt/dynamo/backend/conversion.py
apbose, line 1: Are these changes general? As in are they required for this PR?

py/torch_tensorrt/dynamo/converter_registry.py
gs-olive: Yes, I agree - no existing converters will need to be moved for this PR. New converters can be placed in the `py/torch_tensorrt/dynamo` directory and decorated with the `@dynamo_tensorrt_converter` decorator. If they are to be included in FX as well, we can just switch out the decorator for `@tensorrt_converter` (and optionally move the converter to `py/torch_tensorrt/fx`)

py/torch_tensorrt/dynamo/backend/conversion.py
gs-olive: Yes, these changes are required for this PR, because we need to use a version of `TRTInterpreter` which sources its converters from `DYNAMO_CONVERTERS`, which is not done by the FX implementation. For this reason, we need to use `torch_tensorrt.dynamo.fx_ts_compat.TRTInterpreter` and not `torch_tensorrt.fx.fx2trt.TRTInterpreter`

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,23 @@
...
+        DYNAMO_CONVERTERS[key] = converter
+        return converter
+
+    def disable_converter(converter):
+        return converter
+
+    if enabled:
+        return register_converter
+    else:
+        return disable_converter
```
gs-olive, line 23: Custom class which checks both dictionaries at runtime and incorporates constraint checks, etc. (RFC #2065). `__getitem__`/`__get__` overrides for class. Class only applies for `dynamo/` directory utilities.

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def __init__(
+        self,
+        registries: Sequence[Dict[Target, Union[Callable, Sequence[ConverterSupport]]]],
+    ):
+        # Copy reference to each dictionary object into attribute list
+        self.registries = [registry for registry in registries]
+        self.validate_invariants()
+
+    def validate_invariants(self):
```
narendasan, line 124: Can we throw a warning if there is a collision in converter registration? We should have semantics for shadowing (i.e. probably dynamo converters shadows fx converters) and this should take into account capabilities, i.e. `aten::max_pool(dyn_only) != aten::max_pool(static_only)`

gs-olive: For the current implementation, the distinction of `dynamic` and `static` is not yet present. We also need to consider multiple layers of shadowing. For example, there can be (and are, in #2070), FX and Dynamo converter registry entries with identical conversion capabilities, where one converter is simply more updated, etc.

If we are considering collisions within a single registry, I agree that it makes sense to differentiate on capabilities. I am wondering how we might combine this with the notion of a `check_args` function? For example, consider two converters for the same aten operation, like `torch.ops.aten._to_copy`. One might handle `device` casts, while the other handles `dtype` casts. What might be a good design to differentiate these from actual converter collisions?

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,266 @@
...
+
+# Dictionary representing Dynamo aten-only converters
+# Each converter maps to a sequence of at least one ConverterSupport object(s)
+DYNAMO_ATEN_CONVERTERS: Dict[Target, Sequence[ConverterSupport]] = {}
+
+
+def dynamo_tensorrt_converter(
+    key: Target,
+    enabled: bool = True,
+    check_args: Optional[Callable[[NodeKinds], bool]] = None,
```
gs-olive, line 50: Capability validator - rename the above (`capability_validator`)

There may be cases which your converter cannot handle, which we can validate at partitioning time. For instance, cannot support a specific setting.

gs-olive: `high_priority: bool` flag/argument addition, insert converter at the start of the sequence of converters to check + use first.

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,266 @@
...
+    Args:
+        converter_implementation: Function which converts said node to a TRT equivalent
+        check_args: Function which takes in a NodeKind and returns a bool indicating
+            whether that node can be supported by its companion converter. Note that
+            this function must only access the .target, .args, and .kwargs fields
+            of the node, and cannot modify the node or its graph
+    """
+
+    converter_implementation: Callable
+    check_args: Callable[[NodeKinds], bool] = field(default=lambda node: True)
```
gs-olive, line 39:
```python
dynamic: bool = False
static: bool = True
```
- Instead, use `validator` to check Dynamic shapes.
- `no_dynamic` function which rejects dynamically-shaped input Tensors.
- Enum or `rank`/`priority`/`performance` for hierarchically-grouping converters.
- Put user converters in a different/new registry.
- Prefer user converters to ours.

py/torch_tensorrt/dynamo/fx_ts_compat/fx2trt.py
```diff
@@ -141,14 +141,14 @@ def validate_conversion(self):
         missing_converter = set()
 
         for node in self.module.graph.nodes:
-            if node.op == "call_function" and not CONVERTERS.get(node.target):
+            if node.op == "call_function" and not CONVERTERS.get_validated(node):
```
gs-olive, line 144: Currently, we validate the nodes once in partitioning, and again during the `TRTInterpreter` phase. We could instead validate once in partitioning and pass the determined mapping from `Node` to converter to the interpreter. This could be a future feature, as the current setup enables `fx_ts_compat` to use a different partitioning system for the time being.

py/torch_tensorrt/dynamo/backend/lowering/_partition.py
```diff
@@ -55,6 +55,10 @@ def __init__(
         )
 
         self.min_block_size = min_block_size
+        logger.debug(
+            "Initialized Capability-Based Partitioner with available Converters:\n"
+            + f"{CONVERTERS.display_all_available_converters()}"
+        )
```
gs-olive, line 61: Added logging and debug functionality to print all available converters during partitioning.

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,305 @@
...
+# Dictionary representing Dynamo aten-only converters
+# Each converter maps to a sequence of at least one ConverterSupport object(s)
+DYNAMO_ATEN_CONVERTERS: Dict[Target, Sequence[ConverterSupport]] = {}
+
+
+def dynamo_tensorrt_converter(
+    key: Target,
+    enabled: bool = True,
+    capability_validator: Optional[Callable[[Node], bool]] = None,
+    high_priority: bool = False,
```
gs-olive, line 32: Added support for high-priority converters to be prioritized over others during compilation

gs-olive: Make `priority` flag an Enum `{HIGH, STANDARD}`

py/torch_tensorrt/dynamo/fx_ts_compat/fx2trt.py
gs-olive: Add feature issue

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,305 @@
...
+            for converters in registry.values():
+                if isinstance(converters, (list, tuple)):
+                    assert (
+                        all(isinstance(c, ConverterSupport) for c in converters)
+                        and len(converters) > 0
+                    )
+                else:
+                    assert callable(converters), "Converter function must be callable"
+
+    def __getitem__(self, key: Target):
```
gs-olive, line 147: `registry[Node]` --> call validated version

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,305 @@
...
+        try:
+            return self.__getitem__(key)
+        except KeyError:
+            return value
+
+    def __contains__(self, key: Target):
+        """Check whether a converter for input target exists"""
+        return any(key in registry for registry in self.registries)
+
+    def get_validated(self, node: Node, value=None):
```
gs-olive, line 207: Make default --> `get` and change syntax to `[]`

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,305 @@
...
+
+        raise KeyError(
+            f"None of the converter registries have a validated entry for {key}, with node {node}"
+        )
+
+    def keys(self):
+        """Get all unique targets across all dictionaries"""
+        return self.unique_targets()
+
+    def get(self, key: Target, value=None):
```
gs-olive, line 196: `get_unvalidated`

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,327 @@
...
+
+from torch.fx.node import Target, Node, _get_qualified_name
+from torch_tensorrt.fx.converter_registry import CONVERTERS
+
+
+class ConverterPriority(Enum):
+    """Enum to set a converter's priority in the registry"""
+
+    STANDARD = auto()
+    HIGH = auto()
```
gs-olive, line 13: Added Enum to track converter priority

py/torch_tensorrt/dynamo/converter_registry.py
```diff
@@ -0,0 +1,327 @@
...
+                if isinstance(converters, (list, tuple)):
+                    for candidate in converters:
+                        if candidate.capability_validator(node):
+                            return candidate.converter_implementation
+                else:
+                    return converters
+
+        raise KeyError(
+            f"None of the converter registries have a validated entry for {key}, with node {node}"
+        )
```
gs-olive, line 211: `__getitem__` and `get` access the registry with `Node` inputs only, while `__getitem_without_validation__` and `get_unvalidated` allow access with `Target` inputs only.

`__contains__` dispatches the access based on the type of input. `Node` inputs call a validated implementation while `Target` inputs call an unvalidated implementation.

py/torch_tensorrt/dynamo/backend/lowering/_partition.py
narendasan: This doesnt need to be printed every time a partitioner is created, IMO the best place is to actually print a debug message every time a converter is registered so at runtime as new ones get added it will get logged (so probs part of the converter decorator or whenever it gets added to the dictionary - imo should be fine to add this to fx.trt_converter as well as it is just a logging change

py/torch_tensorrt/fx/converter_registry.py
```diff
@@ -19,6 +23,13 @@ def register_converter(converter):
             NO_IMPLICIT_BATCH_DIM_SUPPORT[key] = converter
         if no_explicit_batch_dim:
             NO_EXPLICIT_BATCH_DIM_SUPPORT[key] = converter
+
+        logger.debug(
+            f"Converter for {key} added to FX Converter Registry "
+            + f"{'without' if no_explicit_batch_dim else 'with'} Explicit Batch Dim Support + "
+            + f"{'without' if no_implicit_batch_dim else 'with'} Implicit Batch Dim Support"
+        )
```
gs-olive, line 31: Added specific logging to FX (and Dynamo) used upon registration of converters

py/torch_tensorrt/dynamo/backend/lowering/_partition.py
```diff
@@ -146,15 +149,16 @@ def print_support_overview(self, num_trt_blocks: Optional[int] = None):
...
 
         logger.debug(supported_nodes_str)
 
-        if len(self.unsupported_operators) != 0:
+        if self.unsupported_operators:
             unsupported_nodes_str = "\nUnsupported or Excluded Nodes:\n"
-            for node_name in self.unsupported_operators:
-                unsupported_nodes_str += f"- {node_name}\n"
+            for node_name, count in self.unsupported_operators.items():
+                unsupported_nodes_str += f"- {node_name} + Operator Count: {count}\n"
```
gs-olive, line 160: Added support for printing operator count, in addition to name

py/torch_tensorrt/dynamo/converter_registry.py
gs-olive, line 1: In the new structure, does the `converter_registry.py` file belong in `conversion/` @peri044?

py/torch_tensorrt/dynamo/conversion/converter_utils.py
gs-olive, line 1: Sample implementation of a capability validator which disallows dynamic shapes. It could potentially also check the Dynamo config as an alternative way to verify no Dynamic shapes.

narendasan (approved):
LGTM

PR #1945, opened 2023-05-23
fix: Add lowering pass to remove output repacking in `convert_method_to_trt_engine` calls

### LINKED ISSUE #1938: ✨[Feature] Automatically remove output repacking for `convert_method_to_trt_engine` calls
# Context
For fully-supported models having multiple output Tensors, Torch often appends a Tuple/List packing function call to the scripted IR. This can cause `convert_method_to_trt_engine`, which would otherwise succeed due to converter coverage, to fail.

# Details
This proposed feature would automatically detect and remove unnecessary output packing schemes, such as `prim::TupleConstruct` and `prim::ListConstruct`, which are applied to the outputs of existing TRT engines, thereby allowing more models to be fully converted to their TRT engine counterparts directly.

### LINKED ISSUE #1939: ✨[Feature] Improved parity between `convert_method_to_trt_engine` and `require_full_compilation=True`
# Context
Currently in the TorchScript path, the `require_full_compilation=True` applies to conversion operators, or those performing some sort of mathematical operation or transformation. It does not apply to collection packing and unpacking operators. This can lead to confusion in later usage of `convert_method_to_trt_engine`, since some models which successfully compile with `require_full_compilation=True`, cannot be converted to fully-TRT engine objects with `convert_method_to_trt_engine`.

# Proposed Solution
The above is not fully a problem in and of itself, since a model which outputs the schema `(Tensor, (Tensor, Tensor))` could not be converted into a TRT-engine since its output is not a flat list of Tensors, but if its logic could be accelerated to a large extent by Torch-TensorRT, then this could reasonably be called "full compilation". Still, there are certain cases where the parity between `convert_method_to_trt_engine` and `require_full_compilation` could be improved - namely, the case of `(Tensor, Tensor, ...)` Tuple-type outputs. These cases can be addressed by removing the extraneous `prim::TupleConstruct` calls, and simply returning the `IValues` directly from the TRT engine.

**Note: ** A recent change to Torch/TorchScript prompted the above problem, as discussed in #1368, by changing the default behavior of traced multi-output-functions from `TensorList` to `Tuple`.

# Example
```python
import torch

class my_model(torch.nn.Module):
    def forward(self, x):
        a = 2 * x
        b = a + 2
        c = torch.relu(b)
        d = torch.sqrt(c)
        return c, d
```
On the above graph, `torch` currently by default considers the `c, d` output to be a `Tuple`, thereby prompting the insertion of a `prim::TupleConstruct` operator which is unnecessary. We would prefer if the following (lowered) graph instead were to have both `IValues` registered as outputs, so the entire graph could be converted to a TRT engine via `torch_tensorrt.convert_method_to_trt_engine`.

```python
Lowered Graph: graph(%x.1 : Tensor):
  %3 : int = prim::Constant[value=1]()
  %2 : int = prim::Constant[value=2]()
  %a.1 : Tensor = aten::mul(%x.1, %2)
  %b.1 : Tensor = aten::add(%a.1, %2, %3)
  %c.1 : Tensor = aten::relu(%b.1)
  %d.1 : Tensor = aten::sqrt(%c.1)
  %8 : (Tensor, Tensor) = prim::TupleConstruct(%c.1, %d.1)
  return (%8)

##### PREFERRED BELOW #####

Lowered Graph: graph(%x.1 : Tensor):
  %3 : int = prim::Constant[value=1]()
  %2 : int = prim::Constant[value=2]()
  %a.1 : Tensor = aten::mul(%x.1, %2)
  %b.1 : Tensor = aten::add(%a.1, %2, %3)
  %c.1 : Tensor = aten::relu(%b.1)
  %d.1 : Tensor = aten::sqrt(%c.1)
  return (%c.1, %d.1)
```

# Additional Context
Related to #1938

gs-olive:
# Description
- *Adds improved support for full-conversion **and** TRT engine export for a variety of models*
- Automatically remove output repacking for `convert_method_to_trt_engine` calls, to improve parity between models which can be converted directly to TRT engines, and models which can be fully compiled
- Add new internal `CompileSpec` argument for lowering which indicates whether the lowering passes originate from a `convert_method_to_trt_engine` call or a regular `compile` call, which affects whether the lowering pass is applied
- Regular TorchScript graphs cannot have this pass applied, as it can otherwise break the output graph. Newer versions of Torch disallow graph outputs with 0 or 2+ arguments which are not packed in a struct
- Current lowering pass detects outputs which are flat Lists or Tuples of Tensors and returns the outputs as-is (direct from the TRT Engine), so the entire model can be converted to a single TRT engine

Fixes #1938 
Fixes #1939 

## Type of change

# Checklist:

### DIFF
--- a/core/lowering/lowering.cpp
+++ b/core/lowering/lowering.cpp
@@ -144,6 +144,9 @@ void LowerGraph(std::shared_ptr<torch::jit::Graph>& g, std::vector<torch::jit::I
   passes::RemoveSingleUse0DTensors(g);
   passes::RemoveUnnecessaryCasts(g);
   passes::ReplaceAtenInt(g);
+  if (lower_info.converting_to_trt_engine) {
+    passes::RemoveCollectionCast(g);
+  }
   passes::UnpackAndCastMaskedFill(g, lower_info.getGPUDeviceString());
   passes::UnpackAndCastNumToTensor(g, lower_info.getGPUDeviceString());
   passes::UnpackAndCastFull(g, lower_info.getGPUDeviceString());
--- a/core/lowering/lowering.h
+++ b/core/lowering/lowering.h
@@ -16,6 +16,10 @@ struct LowerInfo {
   // Since these QDQ nodes will be identical as they share same input, one of them is eliminated due to CSE lowering
   // pass. Disable this in order to not disturb TensorRT's QAT optimizations.
   bool disable_cse = false;
+
+  // Whether the originating caller is `convert_method_to_trt_engine` (true) or `compile` (false)
+  bool converting_to_trt_engine = false;
+
   ir::Device target_device;
   std::vector<std::string> forced_fallback_modules;
   friend std::ostream& operator<<(std::ostream& os, const LowerInfo& l);
--- a/core/lowering/passes/passes.h
+++ b/core/lowering/passes/passes.h
@@ -33,6 +33,7 @@ void RemoveNOPs(std::shared_ptr<torch::jit::Graph> graph);
 void RemoveSingleUse0DTensors(std::shared_ptr<torch::jit::Graph>& g);
 void RemoveUnnecessaryCasts(std::shared_ptr<torch::jit::Graph>& graph);
 void ReplaceAtenInt(std::shared_ptr<torch::jit::Graph>& g);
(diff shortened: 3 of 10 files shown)


narendasan (approved):
LGTM

PR #1949, opened 2023-05-24
fix: Address `.numpy()` issue on fake tensors

gs-olive:
# Description

- Add `fake_tensor_unsupported` decorator to helper backend
- Refactor `conversion` implementation to use compilation settings object as well, to reduce code duplication and encourage reuse
- Improve debugger messages by pre-formatting support string

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/backend/backends.py
+++ b/py/torch_tensorrt/dynamo/backend/backends.py
@@ -52,6 +52,7 @@ def aot_torch_tensorrt_aten_backend(
     )
 
 
+@fake_tensor_unsupported
 def _pretraced_backend(
     gm: torch.fx.GraphModule,
     sample_inputs: Sequence[torch.Tensor],
@@ -120,9 +121,7 @@ def _compile_module(
         trt_mod = convert_module(
             submodule,
             submodule_inputs,
-            debug=settings.debug,
-            workspace_size=settings.workspace_size,
-            precision=settings.precision,
+            settings=settings,
         )
 
         # Replace FX Module with TRT Module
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -52,6 +52,7 @@ def aot_torch_tensorrt_aten_backend(
     )
 
 
+@fake_tensor_unsupported
```
narendasan, line 55: This is the ultimate backend everything goes through right? so does that mean we cant work on fake tensors? is this different than symbolic shapes?

narendasan: Also could we just turn a "fake_tensor" into an "ITensor" immediately. sounds like they are similar

gs-olive: Yes this is the backend everything goes through, and my understanding is that `FakeTensors` are for use at compile time, and are distinct from `SymInt`s which are the symbolic shape representations.

The challenge with fake tensors right now is that _any_ tensors instantiated during the call are fake, which means that the constant tensors which we need to provide to TensorRT, as in the code snippet below, are "fake" and thus contain no value to be parsed. I think a long-term solution could be to support Fake Tensors fully, but this solution temporarily resolves TRT/Torch compatibility issues.

https://github.com/pytorch/TensorRT/blob/c7e79b2f3f48a1bd47266077f414117e42ba2ff5/py/torch_tensorrt/fx/converters/converter_utils.py#L247-L255

gs-olive: Instantiated #1951 with a feature proposal and additional discussion

PR #1955, opened 2023-05-26
fix: Add support for fake tensors

### LINKED ISSUE #1951: ✨[Feature] Fully support `FakeTensor`s in Dynamo compile
# Context
Currently, the `torch.compile` backend for Torch-TensorRT requires `@fake_tensor_unsupported` due to a few functions which cannot handle fake tensors. This support needs to be updated in order to support fake tensors on the path to full-scale dynamic shape support.

**Key Consideration:** Is it possible to support fake tensors while preserving the notion of "constant" tensors in TRT?

# Proposed Changes
The `to_numpy` function will need to be updated to take a scalar:
https://github.com/pytorch/TensorRT/blob/c7e79b2f3f48a1bd47266077f414117e42ba2ff5/py/torch_tensorrt/fx/converters/converter_utils.py#L154

The `create_constant` function will need to be updated to not transform scalars into tensors:
https://github.com/pytorch/TensorRT/blob/c7e79b2f3f48a1bd47266077f414117e42ba2ff5/py/torch_tensorrt/fx/converters/converter_utils.py#L226

Special considerations must be made for constant tensors with more than one element.

gs-olive:
# Description
- Refactor `to_numpy` function to handle non-tensor inputs, avoiding fake tensor issue during compilation of constants
- Refactor multiple functions to allow for `numpy` data types and shapes, as a workaround for `torch.Tensor` instantiations being fake
- Develop new "unified" data type translation which takes any of `{TRT, Numpy, Torch}` data types and translates them to any other of those
- Add regression test case to elicit behavior

Fixes #1951 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/backend/test/test_specialized_models.py
+++ b/py/torch_tensorrt/dynamo/backend/test/test_specialized_models.py
@@ -54,6 +54,7 @@ class TestFakeTensors(TestCase):
             0,
             msg=f"MulInt TRT outputs don't match with the original model.",
         )
+        torch._dynamo.reset()
 
     def test_lowering_add_float(self):
         class AddFloat(torch.nn.Module):
@@ -105,6 +106,8 @@ class TestFakeTensors(TestCase):
             0,
             msg=f"AddFloat TRT outputs don't match with the original model.",
         )
+
+        torch._dynamo.reset()
 
 
 if __name__ == "__main__":
--- a/py/torch_tensorrt/dynamo/backend/test/test_specialized_models.py
+++ b/py/torch_tensorrt/dynamo/backend/test/test_specialized_models.py
@@ -1,0 +1,111 @@
+from utils import lower_graph_testing
+from torch.testing._internal.common_utils import run_tests, TestCase
+import torch
+from torch_tensorrt.dynamo import compile
+
+
+class TestFakeTensors(TestCase):
+    def test_lowering_mul_int(self):
+        class MulInt(torch.nn.Module):
+            def forward(self, x):
+                return x * 7
+
+        # Operations expected to be included in the traced graph after decompositions
+        expected_ops = {
+            torch.ops.aten.mul.Tensor,
+        }
+
+        inputs = [
+            torch.rand(
+                3,
+                5,
+                7,
+            ).cuda(),
+        ]
+
+        fx_graph = torch.fx.symbolic_trace(MulInt())
+        _, expected_ops_unseen = lower_graph_testing(
+            fx_graph,
+            inputs,
+            expected_ops=expected_ops,
+            min_block_size=1,
+        )
+
+        self.assertEquals(
+            len(expected_ops_unseen),
+            0,
+            f"The following expected ops were not encountered: {expected_ops_unseen}",
+        )
+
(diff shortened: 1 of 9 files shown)


py/torch_tensorrt/fx/utils.py
```diff
@@ -15,6 +16,45 @@
 from .types import Shape, TRTDataType
 
 
+class Frameworks(Enum):
+    NUMPY = "numpy"
+    TORCH = "torch"
+    TRT = "trt"
+
+
+DataTypeEquivalence: Dict[
```
gs-olive, line 25: Unified data type translator dictionary which takes a `TRTDataType` and translates it to Numpy, Torch, or TRT data types of the equivalent type and precision.

py/torch_tensorrt/dynamo/backend/test/test_specialized_models.py
```diff
@@ -0,0 +1,114 @@
+from utils import lower_graph_testing
+from torch.testing._internal.common_utils import run_tests, TestCase
+import torch
+from torch_tensorrt.dynamo import compile
+
+
+class TestFakeTensors(TestCase):
```
frank-wei, line 7: Are we enable the faketensor in this test?  I do not quite understand the purpose of this test file?

py/torch_tensorrt/fx/converters/converter_utils.py
```diff
@@ -151,28 +151,50 @@ def extend_mod_attr_to_tuple(mod: torch.nn.Module, name: str, size: int):
...
+            value = value.dequantize()
 
-    return tensor.cpu().detach().contiguous().numpy()
+        output = value.cpu().detach().contiguous().numpy()
+
+    elif isinstance(value, int):
+        output = np.array([value], dtype=np.int32)
+
+    elif isinstance(value, float):
+        output = np.array([value], dtype=np.float32)
```
frank-wei, line 186: add support for fp16?

py/torch_tensorrt/dynamo/backend/test/test_specialized_models.py
gs-olive: Based on the changes in `backends.py`, which remove all calls to `@fake_tensor_unsupported`, fake tensors will be enabled by default via Dynamo/AOT. The purpose of this test is to verify that utilities like `create_constant` do not instantiate Torch tensors when provided scalar inputs. For example, in the test `test_lowering_mul_int` below, the only op in the graph will be something like:
```python
call_function[target=torch.ops.aten.mul.Tensor](args=(%x, 7)...)
```
Without the changes in this PR, the above will fail at runtime because `create_constant` will make a `torch.Tensor` for the scalar 7, and this tensor will be fake (hold no values), so when TRT goes to extract the value to make a constant tensor, the script fails.

py/torch_tensorrt/fx/converters/converter_utils.py
gs-olive: For this particular instance, the intent was just to take Python builtins and translate them to numpy arrays. If the caller specifies an FP16 dtype in the function schema, that cast will be applied later, on line 196.

frank-wei (approved):

Issue #1565, opened 2022-12-27
🐛 [Bug] Unable to compile the model using torch tensorrt


IamExperimenting:
##  Bug Description

Hi team, I have built the object detection model using torchvision fasterrcnn model. I need to deploy this model in Nvidia server, so I’m trying to compile the model using torch_tensorrt but its failing.

## To Reproduce

import torch, tensorrt, torch_tensorrt,torchvision
device = 'cuda' if torch.cuda.is_available() else 'cpu'
model = torchvision.models.detection.fasterrcnn_mobilenet_v3_large_320_fpn().eval()

trt_module = torch_tensorrt.compile(model,
    inputs = [torch_tensorrt.Input((1, 3, 720, 1280))], # input shape   
    enabled_precisions = {torch.half} # Run with FP16
)
# save the TensorRT embedded Torchscript
torch.jit.save(trt_module, "trt_torchscript_module.ts")

## Expected behavior
pytorch model should be compiled using ```torch_tensorrt``` library

## Environment
OS : ubuntu 20.04
Python : 3.10.8
> Build information about Torch-TensorRT can be found by turning on debug messages

tensorrt version : 8.5.2.2

 - Torch-TensorRT Version (e.g. 1.0.0): 1.3.0
 - PyTorch Version (e.g. 1.0): 1.13.1
 - CPU Architecture:
 - OS (e.g., Linux): Linux - ubuntu 20.4
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source): conda
 - Build command you used (if compiling from source):
 - Are you using local sources or building from archives:
 - Python version: 3.10.8
 - CUDA version: 11.6
 - GPU models and configuration: no
 - Any other relevant information:

peri044:
Can you enable debug logging and provide the full log? 
```
with torch_tensorrt.logging.debug():
    trt_module = torch_tensorrt.compile(model,
        inputs = [torch_tensorrt.Input((1, 3, 720, 1280))], # input shape   
        enabled_precisions = {torch.half} # Run with FP16
    )
```

IamExperimenting:
@peri044 

please find the error log below,

```
INFO: [Torch-TensorRT] - ir was set to default, using TorchScript as ir
INFO: [Torch-TensorRT] - Module was provided as a torch.nn.Module, trying to script the module with torch.jit.script. In the event of a failure please preconvert your module to TorchScript
DEBUG: [Torch-TensorRT] - TensorRT Compile Spec: {
[...]

PR #1972, opened 2023-06-02
fix/feat: Move convolution core to `impl` + add feature (FX converter refactor) 

### LINKED ISSUE #1954: ✨[Feature] + 🐛 [Bug] Allow `ITensor` biases in `aten.convolution` converters
# Context
Currently, `aten.convolution` converters do not support `ITensor` biases, which can cause test failures in CI ([example](https://app.circleci.com/pipelines/github/pytorch/TensorRT/2907/workflows/e496ca44-0082-4373-a62c-bc716b0b3627/jobs/13460)), as the new Dynamo compile path primarily uses `ITensor` objects for general tensors throughout computation. 
https://github.com/pytorch/TensorRT/blob/d3a47c411ee9f8be9e8344c547b256ad852a407b/py/torch_tensorrt/fx/converters/acc_ops_converters.py#L202-L208

# Proposed Solution
Allow `ITensor` biases for `aten.convolution` ops in the same way that the kernel weights can be `ITensor` objects. See [`IConvolutionLayer`](https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Graph/Layers.html#tensorrt.IConvolutionLayer) for further information on TensorRT convolution layers.

gs-olive:
# Description

- Centralize `convolution` implementation in FX across all source IR variants, including support for `conv1d`, quantized, and other configurations
- Update reference implementations across the stack to use centralized utility and remove individual replicated implementations
- Allow conv layers to take bias inputs in FX, per new functionality from TRT
- Enable pass-through of build errors in Dynamo e2e tests to ensure errors are not being hidden (this PR fixes a bug which disallowed that pass-through)

Fixes #1954
Addresses first bug in #1565 

## Type of change

# Checklist:


py/torch_tensorrt/fx/converters/impl/convolution.py
```diff
@@ -0,0 +1,146 @@
...
+        )
+
+    # Process bias terms
+    if isinstance(bias, torch.Tensor):
+        # Transform the bias constant into a Numpy array
+        bias = to_numpy(bias)
+
+    elif isinstance(bias, TRTTensor):
+        bias = get_trt_tensor(network, bias, f"{name}_bias")
+
```
gs-olive, line 67: Did not add an unsqueeze operation to bias term since the requirement in TRT for the bias term is that it must have number of elements equal to the number of output features of the convolution, so the same bias as is used for Conv1D would work for Conv2D, with the number of output features being fixed.

py/torch_tensorrt/fx/converters/acc_ops_converters.py
```diff
@@ -268,32 +159,36 @@ def acc_ops_conv_transposend(
...
 
-    # for now we'll assume bias is constant Tensor or None,
-    # and bias being ITensor is not supported in TensorRT api
-    # right now
-    if kwargs["bias"] is not None and not isinstance(kwargs["bias"], torch.Tensor):
-        raise RuntimeError(
-            f"ConvTranspose {name} has bias of type {type(kwargs['bias'])}, Expect Optional[Tensor]"
-        )
-    bias = to_numpy(kwargs["bias"])  # type: ignore[arg-type]
+    if not isinstance(kwargs["bias"], TRTTensor):
```
narendasan, line 162: Do this in a separate PR and centralize conv^T

gs-olive: Instanced #1996 for this

py/torch_tensorrt/fx/converters/impl/convolution.py
frank-wei: just to clarify, you mean 1D or 2D conv? For 1D, we need bias to be unsqueezed.

gs-olive: I initially meant for all conv layers since [this documentation](https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Graph/Layers.html#iconvolutionlayer) seems to indicate we just need the number of elements in the bias Tensor to the be correct, and not necessarily the dimensions, but if bias needs to be unsqueezed for 1D, I can add that functionality back. I am wondering if the intended unsqueeze should be in the first dimension (`torch.unsqueeze(bias, 0)`) or the last dimension (`torch.unsqueeze(bias, -1)`)?

frank-wei: I am wondering if TRT has done the broadcast internally since the unit test for 1D works good even though you did not unsqueeze it.

gs-olive: I verified on a small sample that Conv1D with bias compiles + runs inference successfully without unsqueezing the bias term

wushirong (approved):

py/torch_tensorrt/fx/converters/convolution.py
```diff
@@ -1,212 +1,123 @@
...
-
-    if is_quantized:
-        mark_as_int8_layer(layer, conv_output.dynamic_range)
-
-    return layer.get_output(0)
+logger = logging.getLogger(__name__)
 
 
 @tensorrt_converter(torch.nn.modules.conv.Conv1d)
 def conv1d(network, submod, args, kwargs, layer_name):
```
narendasan, line 18: Move this to nn_converters

py/torch_tensorrt/fx/converters/convolution.py
```diff
@@ -1,212 +1,123 @@
...
+        bias=submod.bias,
+        stride=getattr(submod, "stride"),
+        padding=getattr(submod, "padding"),
+        dilation=getattr(submod, "dilation"),
+        groups=submod.groups,
     )
 
 
 @tensorrt_converter(torch.nn.modules.conv.Conv2d)
 def conv2d(network, submod, args, kwargs, layer_name):
```
narendasan, line 41: Same here, we should get rid of this file and just put the registrations for each opset in the same file

py/torch_tensorrt/fx/converters/impl/convolution.py
narendasan: Should we have a unit test for this just to catch if TRT behavior changes?

narendasan (commented):
Mostly organization stuff

py/torch_tensorrt/fx/converters/impl/convolution.py
gs-olive: I could add one, but it would likely be very similar to this case, which interprets, builds, and runs inference with a Conv1D model with TRT both with and without bias:
https://github.com/pytorch/TensorRT/blob/cae6b7c694db1b6a5f287f9fc4e6c40ba08e1079/py/torch_tensorrt/fx/test/converters/acc_op/test_convolution.py#L8-L45

py/torch_tensorrt/fx/converters/nn_ops_converters.py
gs-olive, line 1: Move `convolution.py` contents to `nn_ops_converters.py` for code cleanliness + organization. Remove unused imports.

gs-olive:
HI @wushirong - thank you for the review. I was wondering if you could have another look at the changes, as I've moved `convolution.py` contents to `nn_ops_converters.py` for code cleanliness + organization and removed unused imports.

wushirong (approved):

PR #1991, opened 2023-06-06
fix 2 level nested conditional missing value bug

bowang007:
# Description

For conditionals, blocks in prim::If will be segmented into different tensorrt engines. 
However, this is pointless and my introduce complexity when it comes to nested conditional. 
This fix will avoid such case and fix some missing value related bugs.


narendasan (commented):
Can you add a test for this case?

bowang007:
closing since the change is included in #1933

PR #2021, opened 2023-06-14
fix/feat: Add support for 64bit Tensor inputs FX aten [9 / x]

### LINKED ISSUE #2008: ✨[Feature] Automatic casting of Long/Double Tensor constants in FX
# Context
Certain Torch layers, including `torch.nn.Embedding` handle inputs, weights, and constant Tensors which are required by Torch to be 64-bit types. This causes issues when attempting to build TRT converters for such engines, since the inputs and constants will then be 64-bit, which is not supported by TRT.

# Feature Proposal
When users specify the `truncate_long_and_double` flag, it should have an effect in both FX and Dynamo. A similar feature was implemented for Dynamo compile in a broader context in PR #1983. For this feature, if the user specifies the `truncate_long_and_double` flag, we should automatically cast the 64-bit constant tensors (weights, constant indices, etc.) to their 32-bit equivalents for use in TRT.

gs-olive:
# Description

- Add `truncate_long_and_double` argument in FX settings to allow 64bit inputs
- Utilize existing Dynamo functionality to repair FX aten graphs with 64bit inputs
- Refactor imports in Dynamo to avoid circular import issue from new dependency issues
- Add test cases to validate new feature

**Note:** Breaks `python setup.py install --fx-only` by introducing Dynamo dependency in FX. Potential resolutions include moving Dynamo utilities out of the Dynamo directory, or changing C++ requiring imports in Dynamo, etc.

Fixes #2008

## Type of change

# Checklist:


peri044:
Closing this in favor of https://github.com/pytorch/TensorRT/pull/1983

PR #2052, opened 2023-06-22
fix: Improve logging and kwarg passing in Dynamo

gs-olive:
# Description

- Improve kwarg passing through Dynamo frontend
- Improve logging in case of TRT compilation failures

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/backend/__init__.py
+++ b/py/torch_tensorrt/dynamo/backend/__init__.py
@@ -8,7 +8,6 @@ from typing import Any, Sequence
 from torch_tensorrt import EngineCapability, Device
 from torch_tensorrt.fx.utils import LowerPrecision
 
-from torch_tensorrt.dynamo.backend._settings import CompilationSettings
 from torch_tensorrt.dynamo.backend.utils import prepare_inputs, prepare_device
 from torch_tensorrt.dynamo.backend.backends import torch_tensorrt_backend
 from torch_tensorrt.dynamo.backend._defaults import (
@@ -62,6 +61,10 @@ def compile(
 
     inputs = prepare_inputs(inputs, prepare_device(device))
 
+    if not isinstance(enabled_precisions, collections.abc.Collection):
+        enabled_precisions = [enabled_precisions]
+
+    # Parse user-specified enabled precisions
     if (
         torch.float16 in enabled_precisions
         or torch_tensorrt.dtype.half in enabled_precisions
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -77,13 +77,12 @@ def _pretraced_backend(
...
-            + "Returning GraphModule forward instead.",
-            exc_info=True,
-        )
-
         if not settings.pass_through_build_failures:
+            logger.error(
+                "FX2TRT conversion failed on the subgraph. See trace above. "
+                + "Returning GraphModule forward instead.",
+                exc_info=True,
+            )
```
gs-olive, line 85: Only log an error if the user is not requesting `pass_through_build_failures`. If they are requesting `pass_through_build_failures`, the error is redundant since the compilation will halt anyway.

py/torch_tensorrt/dynamo/backend/__init__.py
```diff
@@ -123,19 +123,12 @@ def create_backend(
     Returns:
         Backend for torch.compile
     """
-    if debug:
-        logger.setLevel(logging.DEBUG)
-
-    settings = CompilationSettings(
+    return partial(
+        torch_tensorrt_backend,
```
gs-olive, line 127: The default backend now takes `kwargs` instead of compilation settings.

py/torch_tensorrt/dynamo/backend/backends.py
narendasan: Drop uses of the term FX2TRT

py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -77,13 +77,12 @@ def _pretraced_backend(
...
         return trt_compiled
     except:
-        logger.error(
-            "FX2TRT conversion failed on the subgraph. See trace above. "
-            + "Returning GraphModule forward instead.",
-            exc_info=True,
-        )
-
         if not settings.pass_through_build_failures:
+            logger.error(
```
narendasan, line 81: What is typically the policy for compilation failures in torch.compile? should this be an error or a warning or info since this is part of expected behavior for torch.compile

gs-olive: I don't think there is a formal policy for compilation failures in `torch.compile`, it is more the choice of the backend. It seems that many other backends (such as `onnxrt`, `tvm`) generally do not have `try/except` clauses in their highest-level backends to catch internal errors, from what I could see. In this case, I think `logger.error` would be the correct level of error to report.

gs-olive: Switch to a `warning` + still display error from TRT. Fallback to PyTorch.

PR #2064, opened 2023-06-27
Dynamo export refactor

peri044:
# Description

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/backend/test/utils.py
```diff
@@ -13,6 +13,7 @@
 
 from torch._functorch.aot_autograd import aot_module_simplified, make_boxed_compiler
 
+DECIMALS_OF_AGREEMENT = 4
```
gs-olive, line 16: Would this constant not be necessary/helpful in the `export` testing? As in - could it instead go in `py/torch_tensorrt/dynamo/utils.py`?

py/torch_tensorrt/dynamo/utils.py
gs-olive, line 1: Should `dynamo/utils.py` be distinct from `dynamo/test_utils.py`? I don't think it makes a huge difference here, but if the number of utilities were to increase, it might help to differentiate compilation utilities from testing utilities.

gs-olive (commented):
Overall, the changes looks good - made some general organizational comments.

py/torch_tensorrt/dynamo/backend/test/utils.py
peri044: This parameter is used by tests which are currently written for `backend`. I don't have equivalent tests yet for export. I will work on those once this is evened out. After, that we can migrate this argument accordingly so that both can use it from a common place.

py/torch_tensorrt/dynamo/utils.py
peri044: Since there's only `cosine_sim` function in `test_utils.py`, I merged it with `dynamo/utils.py`. I'm fine with leaving it in a separate `test_utils.py` if there are more to come. In general, this is what I'm thinking 
We have a single `dynamo/test` directory which has something like 
1) `models`  subdirectory - has test_models.py  - tests 5 standard models with both `export` and `backend`
2) `export` sub directory - has tests specific to `export` workflow (maybe lowering and partitioning)
3) `backend` sub directory - has tests specific to `backend` workflow (maybe lowering and partitioning)

If lowering and partitioning are shared by `export` and `backend`, we could just have scripts instead of directories. But I think there might be other tests specific to these workflows which might be placed in their subdirectories.

py/torch_tensorrt/dynamo/__init__.py
```diff
@@ -1,2 +1,5 @@
+from ._TorchTensorRTModule import TorchTensorRTModule
 from torch_tensorrt.dynamo import fx_ts_compat
-from .backend import compile
+from ._settings import *
+from torch_tensorrt.dynamo import backend
+from torch_tensorrt.dynamo.export import compile
```
gs-olive, line 5: Maybe we could avoid directly promoting the `export.compile` function to the top level, as it could become quite confusing that `torch_tensorrt.dynamo.compile` invokes the `torch.export` path. The suggestion made here https://github.com/pytorch/TensorRT/discussions/1826#discussioncomment-6348562 could be helpful for an alternative.

PR #2080, opened 2023-07-06
fix: Repair null bindings issue in TRT Engines

### LINKED ISSUE #2072: 🐛 [Bug] Encountered TRT Error in Dynamo compile path
##  Bug Description
When compiling GPT-2 with Dynamo compile, the following error is encountered:
```python
[07/01/2023-00:07:12] [TRT] [E] 3: [executionContext.cpp::enqueueInternal::795] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::enqueueInternal::795, condition: bindings[x] || nullBindingOK
)
```
Additionally, this does not seem to cause failures in the Dynamo runtime (not caught by `pass_through_build_failures`), and it appears both with and without the experimental runtime.

## To Reproduce
```python
model = GPT2Model.from_pretrained("gpt2").eval().cuda()
input_ids      = torch.randint(0, 2, (1, 14), dtype=torch.int32).cuda()
attention_mask = torch.randint(0, 2, (1, 14), dtype=torch.int32).cuda()
traced = transformers_trace(model, input_names=["input_ids", "attention_mask"]).eval().cuda()
fx_trt_model = torch_tensorrt.compile(traced, ir="dynamo_compile", inputs=[input_ids, attention_mask], debug=True, pass_through_build_failures=True, min_block_size=10)
```

## Expected behavior

The model should not encounter TRT errors while compiling.

## Environment

 - Torch-TensorRT Version (e.g. 1.0.0): 2844630
 - PyTorch Version (e.g. 1.0): `2.1.0.dev20230620+cu118`
 - TensorRT Version: `8.6.1`

## Additional context

This error no longer appears when we apply the `@fake_tensor_unsupported` flag to the backends, as was removed in #1955. It is unclear whether this is the direct _cause_ of the bug, however.

Additionally, this bug seems to appear only sometimes, and not consistently.

gs-olive:
# Description

- Caused by passing Fake Tensor objects into TRT engines mid-compilation
- Resolved by replacing all FX modules with TRT equivalents after all TRT compilation is complete. This way, modules are not run on FakeTensors generated during compilation

Fixes #2072

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/backend/backends.py
+++ b/py/torch_tensorrt/dynamo/backend/backends.py
@@ -121,6 +121,9 @@ def _compile_module(
         torch_executed_ops=settings.torch_executed_ops,
     )
 
+    # Store TRT replicas of Torch subgraphs
+    trt_modules = {}
+
     # Iterate over all components that can be accelerated
     # Generate the corresponding TRT Module for those
     for name, _ in partitioned_module.named_children():
@@ -138,7 +141,10 @@ def _compile_module(
             settings=settings,
         )
 
-        # Replace FX Module with TRT Module
+        trt_modules[name] = trt_mod
+
+    # Replace all FX Modules with TRT Modules
+    for name, trt_mod in trt_modules.items():
         setattr(partitioned_module, name, trt_mod)
 
     return partitioned_module


py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -138,7 +141,10 @@ def _compile_module(
             settings=settings,
         )
 
-        # Replace FX Module with TRT Module
+        trt_modules[name] = trt_mod
+
+    # Replace all FX Modules with TRT Modules
+    for name, trt_mod in trt_modules.items():
         setattr(partitioned_module, name, trt_mod)
```
gs-olive, line 148: Set all TRT engine attributes _after_ completion of compilation. Forward passes during compilation are run on the Torch equivalents, _not_ the TRT graphs.

narendasan (approved):
LGTM

PR #2070, opened 2023-06-30
fix: Move all `aten` PRs to Dynamo converter registry

gs-olive:
# Description
**The below are now squashed**

The structure of this PR is as follows:
-  The first commit is a squashed set of 16 PRs which add `aten` improvements and converters, from existing PRs with the label [Story: ATen Op Support](https://github.com/pytorch/TensorRT/pulls?q=is%3Apr+is%3Aopen+label%3A%22Story%3A+ATen+Op+Support%22)
- The second commit **reverts** all of the changes to `py/torch_tensorrt/fx`
- The third commit contains **all** of the changes made to `py/torch_tensorrt/fx`, replayed on `py/torch_tensorrt/dynamo`, with required import, reference, and code updates. This sometimes includes code duplication where absolutely necessary to add some new functionality
- Subsequent commits add various improvements or modifications to these converters

Fixes #2068 

## Type of change

- New feature (non-breaking change which adds functionality)

# Checklist:

- [ x ] My code follows the style guidelines of this project (You can use the linters)
- [ x ] I have performed a self-review of my own code
- [ x ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ x ] I have made corresponding changes to the documentation
- [ - ] I have added tests to verify my fix or my feature
  - Each PR has its own tests
- [ x ] New and existing unit tests pass locally with my changes
- [ x ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #2084, opened 2023-07-07
Int cast of div converter- <Target: converter_reorg_elementwise>

apbose:
This PR addresses input casting for aten::div cases in which both the inputs are integer


apbose:
Closed in favor of #2070

PR #2104, opened 2023-07-12
feat: Dynamo refactor

peri044:
# Description

Refactor dynamo codebase

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/fx2trt.py
gs-olive, line 1: Consider moving `TRTInterpreter` into its own file named `TRTInterpreter.py`

gs-olive (commented):
New organization looks good. Added a few comments. Consider adding `lowering/` directory and `runtime/` directory as discussed in https://github.com/pytorch/TensorRT/discussions/1826#discussioncomment-6420812, to add `lower.py` and other key utilities.

py/torch_tensorrt/dynamo/conversion/fx2trt.py
peri044: done

py/torch_tensorrt/dynamo/backend/lowering/_fusers.py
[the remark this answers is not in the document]
peri044: this is currently  needed only for aten_tracer for dynamo export. This is coming from `fx` codebase. We should revisit their importance along with the substitutions you have with backend. As of now this is WIP.

.circleci/config.yml
gs-olive, line 1: Pending CI runs

py/torch_tensorrt/dynamo/compile.py
```diff
@@ -0,0 +1,240 @@
...
+            precision=lower_precision,
+            workspace_size=workspace_size,
+            min_block_size=min_block_size,
+            torch_executed_ops=torch_executed_ops,
+            pass_through_build_failures=pass_through_build_failures,
+            max_aux_streams=max_aux_streams,
+            version_compatible=version_compatible,
+            optimization_level=optimization_level,
+            use_python_runtime=use_python_runtime,
+        )
```
gs-olive, line 132: I think this could be refactored a bit to reduce code duplication, since some of the functionality is no longer critical. Specifically, I think `create_backend` is no longer a necessary utility so maybe we could remove that function altogether and replace the above with something like:
```python
compilation_options = {
            "precision"=lower_precision,
            "debug"=debug,
            "workspace_size"=workspace_size,
            "min_block_size"=min_block_size,
            "torch_executed_ops"=torch_executed_ops,
            "pass_through_build_failures"=pass_through_build_failures,
            "max_aux_streams"=max_aux_streams,
            "version_compatible"=version_compatible,
            "optimization_level"=optimization_level,
            "use_python_runtime"=use_python_runtime,
}

if kwargs.get("ir", "dynamo") == "torch_compile":
    model = torch.compile(gm, backend=torch_tensorrt_backend, options={**compilation_options , **kwargs})
    # Ensure compilation occurs by calling the function with provided inputs
    model(*inputs)
    return model

else:
    settings = CompilationSettings(**compilation_options)
```

py/torch_tensorrt/dynamo/compile.py
```diff
@@ -0,0 +1,240 @@
...
+            model = lower_model(model, inputs)
+            return _compile_module(model, inputs, settings)
+        else:
+            split_result = lower_model_using_trt_splitter(model, inputs)
+            trt_module = _compile_graph(split_result, inputs, settings)
+
+            return trt_module
+
+
+def create_backend(
```
gs-olive, line 146: See above comment - potentially to delete this function.

gs-olive (commented):
Overall the changes and testing updates look good, pending Circle CI verification. Made some comments about compilation settings.

py/torch_tensorrt/dynamo/compile.py
gs-olive: The main change that enables the above is the `torch.compile(..., options={...},...)` argument, which allows us to pass keywords to the backend directly.

peri044: That looks good. Added this change.

py/torch_tensorrt/dynamo/compile.py
peri044: Done

py/torch_tensorrt/_compile.py
```diff
@@ -47,31 +47,29 @@ def _get_target_ir(module_type: _ModuleType, ir: str) -> _IRType:
...
-                    logging.Level.Info, "ir was set to default, using TorchScript as ir"
+                    logging.Level.Info, "ir was set to default, using dynamo as ir"
                 )
-                return _IRType.ts
-            elif module_is_fxable:
+                return _IRType.dynamo
+            elif module_is_tsable:
                 raise ValueError(
-                    "Was given a torch.fx.GraphModule, fx is not currently supported by Torch-TensorRT"
+                    "Input graph is a Torchscript module but the ir provided is default (dynamo). Please set ir=torchscript to compile."
```
narendasan, line 71: Instead of value error print a warning you are using the torchscript backend to suppress the warning explictly set ir=torchscript but continue to compile

py/torch_tensorrt/_compile.py
```diff
@@ -156,13 +154,13 @@ def compile(
             dynamic_batch=False,
             **kwargs,
         )
-    elif target_ir == _IRType.dynamo_compile:
+    elif target_ir == _IRType.dynamo or target_ir == _IRType.torch_compile:
         return torch_tensorrt.dynamo.compile(
```
narendasan, line 158: torch_trt.dynamo.compile should only handle export (i.e no ir flag). Directly call torch.compile from here for ir=“torch_compile”

py/torch_tensorrt/dynamo/aten_tracer.py
```diff
@@ -0,0 +1,157 @@
...
+                "Please try torchdynamo.explain() to get possible the reasons",
+            ) from exc
+        except Exception as exc:
+            raise RuntimeError(
+                "torchdynamo internal error occured. Please see above stacktrace"
+            ) from exc
+
+
+@req_torch_version("2.dev")
+def trace(model, inputs, **kwargs):
```
narendasan, line 129: Thoughts on exposing this to users  under `torch_tensorrt.dynamo.trace`?

py/torch_tensorrt/dynamo/compile.py
```diff
@@ -0,0 +1,183 @@
...
+        "min_block_size": min_block_size,
+        "torch_executed_ops": torch_executed_ops,
+        "pass_through_build_failures": pass_through_build_failures,
+        "max_aux_streams": max_aux_streams,
+        "version_compatible": version_compatible,
+        "optimization_level": optimization_level,
+        "use_python_runtime": use_python_runtime,
+    }
+
+    if kwargs.get("ir", "dynamo") == "torch_compile":
```
narendasan, line 113: torch_tensorrt.dynamo.compile should be only for export

py/torch_tensorrt/dynamo/runtime/_PythonTRTModule.py
```diff
@@ -0,0 +1,244 @@
+from typing import Any, List, Sequence
+
+# @manual=//deeplearning/trt/python:py_tensorrt
+import tensorrt as trt
+import torch
+from torch_tensorrt.fx.utils import unified_dtype_converter, Frameworks
+
+
+class TRTModule(torch.nn.Module):
```
narendasan, line 9: Lets call this PythonTorchTRT module. Basically the name of the other other one +python. Also can you note in the docstring that serializing to torchscript isnt not supported for this module?

py/torch_tensorrt/dynamo/runtime/__init__.py
```diff
@@ -0,0 +1,2 @@
+from ._PythonTRTModule import TRTModule
```
narendasan, line 1: Use the same naming scheme

tests/py/dynamo/models/test_models.py
gs-olive, line 1: If this file is intended to only test `torch.compile`, then could it be renamed to `test_models_compile.py`? Or is the intent to merge this with `test_models_export.py`?

.circleci/config.yml
```diff
@@ -733,49 +733,49 @@ commands:
...
       - run:
-          name: Run Dynamo fx_ts_compat core tests
+          name: Run Dynamo torch_compile tests
           command: |
-            cd py/torch_tensorrt/dynamo/fx_ts_compat/test
-            pushd core/
-            pytest --junitxml=/tmp/artifacts/test_results/dynamo/fx_ts_compat/test_results.xml
+            cd tests/py/dynamo/backend/
+            pytest --junitxml=/tmp/artifacts/test_results/dynamo/torch_compile/test_results.xml
             popd
```
gs-olive, line 745: Remove `popd` call since no directory was pushed to the stack

tests/py/dynamo/models/test_models.py
peri044: Yeah this is intended to merge with test_models (and calling it with --ir flag correspondingly). At the moment they are separate (since BERT isn't supported and the output of export workflow is a list as opposed to tensors (coming from torch.compile workflow)

py/torch_tensorrt/_compile.py
peri044: done

py/torch_tensorrt/dynamo/aten_tracer.py
peri044: I added it now. The final version is likely going to be same as calling `torch._dynamo.export` in which case we might not need to expose it.

py/torch_tensorrt/dynamo/compile.py
peri044: done

py/torch_tensorrt/dynamo/runtime/_PythonTRTModule.py
peri044: done

py/torch_tensorrt/dynamo/runtime/__init__.py
peri044: done

.circleci/config.yml
peri044: done

py/torch_tensorrt/_compile.py
peri044: done. Can you verify now ?

py/torch_tensorrt/dynamo/__init__.py
```diff
@@ -1,6 +1,3 @@
-import torch
-from packaging import version
-
-if version.parse(torch.__version__) >= version.parse("2.1.dev"):
-    from torch_tensorrt.dynamo import fx_ts_compat
-    from .backend import compile
+from ._settings import *
+from .compile import compile
+from .aten_tracer import trace
```
gs-olive, line 3: Consider wrapping in `if version.parse(torch.__version__) >= version.parse("2.1.dev"):`, since certain functionality and imports could break with earlier versions

py/torch_tensorrt/dynamo/compile.py
```diff
@@ -0,0 +1,172 @@
...
+        "optimization_level": optimization_level,
+        "use_python_runtime": use_python_runtime,
+    }
+
+    settings = CompilationSettings(**compilation_options)
+    model = trace(gm, torch_inputs, **kwargs)
+
+    if kwargs.get("use_capability_partitioner", None):
+        model = lower_model(model, torch_inputs)
+        return _compile_module(model, torch_inputs, settings)
```
gs-olive, line 118: It seems `_compile_module` is no longer defined in this scope.

py/torch_tensorrt/dynamo/conversion/conversion.py
```diff
@@ -1,12 +1,11 @@
...
 import io
-from torch_tensorrt.fx.trt_module import TRTModule
-from torch_tensorrt.dynamo.backend._settings import CompilationSettings
-from torch_tensorrt.dynamo.fx_ts_compat.fx2trt import (
-    InputTensorSpec,
-    TRTInterpreter,
-)
+from torch_tensorrt.dynamo.runtime import TRTModule
+from torch_tensorrt.dynamo import CompilationSettings
+from torch_tensorrt import Input
```
gs-olive, line 6: This change breaks some functionality in `torch.compile`. Specifically, the `Input.from_tensors(inputs)` function is different from the `InputTensorSpec` variant in that it applies checks for aspects like contiguous memory, for which some of the `FakeTensors` Dynamo provides to us at compile-time fail the check.

py/torch_tensorrt/dynamo/runtime/__init__.py
gs-olive, line 1: Ensure imports of these utilities (for instance in `tests/py/ts/api/test_classes.py`) are updated to reference the new directory structure.

tests/py/dynamo/backend/test_compiler_utils.py
gs-olive, line 1: Update the test cases in this file to reflect the changes to the `prepare_inputs.py` function. For instance, can disregard the first output of the function.

py/torch_tensorrt/dynamo/conversion/conversion.py
gs-olive: `Input` could use an argument such as `disable_memory_format_check` to make it more like `InputTensorSpec`. Additionally, a method similar to `InputTensorSpec.create_inputs_from_specs` would be helpful for the converters featured in #2070.

narendasan: @gs-olive can you link to the code where this check occurs and where it causes failures in your code?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -0,0 +1,162 @@
...
+            from torch_tensorrt.dynamo.runtime import TorchTensorRTModule
+
+            using_python_runtime = False
+            reason = "since C++ dependency was detected as present"
+        except ImportError:
+            using_python_runtime = True
+            reason = "since import failed, C++ dependency not installed"
+
+    logger.info(
+        f"Using {'Python' if using_python_runtime else 'C++'} {reason} TRT Runtime"
```
narendasan, line 39: Imo should be something like: 

```py
f"Using {'Python-only' if using_python_runtime else 'Default'} Torch-TRT Runtime ({reason})"
```

No need for the `as`.

py/torch_tensorrt/dynamo/conversion/conversion.py
gs-olive: This is where the code is referenced:
https://github.com/pytorch/TensorRT/blob/d4c9c06213602c6e8d2ce568154ed287ee67dafd/py/torch_tensorrt/dynamo/backend/conversion.py#L38-L43
This is where the failure occurs:
https://github.com/pytorch/TensorRT/blob/d4c9c06213602c6e8d2ce568154ed287ee67dafd/py/torch_tensorrt/_Input.py#L315-L323
After looking into the issue (which appears on Resnet18, but potentially other models as well), it seems that some of the intermediate `FakeTensors` do not have `is_contiguous` with either of those memory formats (though, as `FakeTensors`, they store no values). When I disable the check, there is no such issue at runtime.

peri044: `disable_memory_format_check ` argument to `from_tensors` seems ok to me. 
`.example_tensor()` seems to be the equivalent to `create_inputs_from_specs` call. Would that be sufficient ?

gs-olive: Added `disable_memory_format_check` in newest commit + the `.example_tensor` would likely work here

py/torch_tensorrt/dynamo/compile.py
```diff
@@ -0,0 +1,172 @@
...
+    else:
+        split_result = lower_model_using_trt_splitter(model, torch_inputs)
+        trt_module = _compile_graph(split_result, torch_inputs, settings)
+
+        return trt_module
+
+
+def _compile_graph(
+    split_result: TRTSplitter,
```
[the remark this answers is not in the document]
peri044: modified it to `SplitResult`

py/torch_tensorrt/dynamo/__init__.py
peri044: you mean the aten_tracer ?

py/torch_tensorrt/dynamo/compile.py
```diff
@@ -0,0 +1,174 @@
...
+        "torch_executed_ops": torch_executed_ops,
+        "pass_through_build_failures": pass_through_build_failures,
+        "max_aux_streams": max_aux_streams,
+        "version_compatible": version_compatible,
+        "optimization_level": optimization_level,
+        "use_python_runtime": use_python_runtime,
+    }
+
+    settings = CompilationSettings(**compilation_options)
+    model = trace(gm, torch_inputs, **kwargs)
```
narendasan, line 116: This should ideally be in `torch_tensorrt.compile` and the output is passed to `torch_tensorrt.dynamo.compile`

peri044: done

py/torch_tensorrt/_compile.py
```diff
@@ -156,18 +156,40 @@ def compile(
...
+    device = kwargs.get("device", Device._current_device())
+    torchtrt_inputs, torch_inputs = prepare_inputs(inputs, prepare_device(device))
+    model = torch.compile(module, backend=torch_tensorrt_backend, options={**kwargs})
+    # Ensure compilation occurs by calling the function with provided inputs
+    model(*torch_inputs)
+
+    return model
+
+
 def convert_method_to_trt_engine(
```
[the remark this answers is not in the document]
peri044: done

py/torch_tensorrt/_compile.py
```diff
@@ -156,18 +156,40 @@ def compile(
...
+    import collections.abc
+
+    if not isinstance(inputs, collections.abc.Sequence):
+        inputs = [inputs]
+
+    device = kwargs.get("device", Device._current_device())
+    torchtrt_inputs, torch_inputs = prepare_inputs(inputs, prepare_device(device))
+    model = torch.compile(module, backend=torch_tensorrt_backend, options={**kwargs})
+    # Ensure compilation occurs by calling the function with provided inputs
+    model(*torch_inputs)
```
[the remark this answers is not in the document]
peri044: done

narendasan (approved):
LGTM, @peri044 can you quickly open another PR to change the default to dynamo compile and to bump the version to 2.0.0.dev in the python package and the C++ header?

py/torch_tensorrt/_compile.py
```diff
@@ -61,17 +61,17 @@ def _get_target_ir(module_type: _ModuleType, ir: str) -> _IRType:
...
-            if module_is_fxable:
-                logging.log(
-                    logging.Level.Info, "ir was set to default, using dynamo as ir"
-                )
-                return _IRType.dynamo
-            elif module_is_tsable:
+            if module_is_tsable:
                 logging.log(
                     logging.Level.Warning,
                     "Input graph is a Torchscript module but the ir provided is default (dynamo). Please set ir=torchscript to suppress the warning. Compiling the module with ir=ts",
```
gs-olive: This warning will show even when the input is an `nn.Module`, effectively making the default `ts`, since passing in `nn.Module` types is the common way to use Dynamo. Consider changing `module_is_tsable = any([module_type == t for t in [_ModuleType.nn, _ModuleType.ts]])` to `module_is_tsable = (module_type == _ModuleType.ts)` or similar.

py/torch_tensorrt/_compile.py
```diff
@@ -47,33 +47,33 @@ def _get_target_ir(module_type: _ModuleType, ir: str) -> _IRType:
...
     else:
         if ir == "default":
             # Options are listed in order of preference
             if module_is_tsable:
                 logging.log(
-                    logging.Level.Info, "ir was set to default, using TorchScript as ir"
+                    logging.Level.Warning,
+                    "Input graph is a Torchscript module but the ir provided is default (dynamo). Please set ir=torchscript to suppress the warning. Compiling the module with ir=ts",
                 )
                 return _IRType.ts
```
gs-olive, line 69: See above comment (applies here to newest commit)

peri044: fixed the messaging

py/torch_tensorrt/_compile.py
peri044: fixed the messaging

PR #2105, opened 2023-07-12
feat: Added support for aten::tile converter

andi4191:
# Description

Added support for aten::tile using aten::repeat mechanism. Refer torch implementation for torch.tile

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/expand.cpp
+++ b/core/conversion/converters/impl/expand.cpp
@@ -194,60 +194,6 @@ bool add_expand_dynamic(
   return true;
 }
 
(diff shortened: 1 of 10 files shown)


peri044 (commented):
Can we express aten::tile as aten::repeat ? Seems like both are similar. In that case, we should have a lowering pass which converters aten::tile to aten::repeat. This helps with not maintaining testcases and code.

andi4191:
> Can we express aten::tile as aten::repeat ? Seems like both are similar. In that case, we should have a lowering pass which converters aten::tile to aten::repeat. This helps with not maintaining testcases and code.

@peri044 : PTAL, thanks

peri044 (approved):
LGTM

PR #2126, opened 2023-07-20
Improve Python tooling

narendasan:
# Description

Installing a number of standard tools to make the codebase more maintainable longterm 

Adds: 
- [x] Mypy: Static type checking 
- [x] Ruff: Flake8 linting 
- [x] isort: sort includes

## Type of change

   - Some types have slightly been adjusted to make sense in a static type context

# Checklist:


py/torch_tensorrt/dynamo/conversion/impl/shape.py
```diff
@@ -1,29 +1,24 @@
...
 
 
 def get_shape_with_dynamic_shape(
     network: TRTNetwork,
     target: Target,
-    source_ir: SourceIR,
+    source_ir: Optional[SourceIR],
     name: str,
-    shape: Union[list, tuple, torch.Tensor],
+    shape: Union[List[int], Tuple[int, ...], torch.Tensor],
```
narendasan, line 21: Use 3.11 syntax

py/torch_tensorrt/dynamo/lowering/substitutions/einsum.py
```diff
@@ -34,15 +33,16 @@ def einsum_generic(
...
 
 # TODO: @gs-olive Port to dynamo converter
-@tensorrt_converter(torch.ops.tensorrt.einsum.default)
+@tensorrt_converter(torch.ops.tensorrt.einsum.default)  # type: ignore[misc]
 def aten_ops_einsum(
     network: TRTNetwork,
     target: Target,
     args: Tuple[Argument, ...],
-    kwargs: Dict[str, Argument],
+    _kwargs: Dict[str, Argument],
```
narendasan, line 41: need to switch this back

py/torch_tensorrt/dynamo/conversion/impl/activation.py
```diff
@@ -53,7 +49,9 @@ def gelu(
 
     layer = network.add_plugin_v2([input_val], plugin)
 
-    def gelu_dyn_range_fn(dyn_range):
+    def gelu_dyn_range_fn(
```
narendasan, line 52: @apbose @gs-olive ^

gs-olive: This would likely fail in the `FakeTensor` context we use in the `torch.compile` path. I believe the correct type signature for `dyn_range` should be `Tuple[float, float]`, [as mentioned here](https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Graph/LayerBase.html?highlight=dynamic_range#tensorrt.ITensor). Then, the fix would be to change the `return` to be:

```python
import math

...

        return (
            dyn_range[0] * 0.5 * (1.0 + math.erf(dyn_range[0] / math.sqrt(2.0)))
        ), (dyn_range[1] * 0.5 * (1.0 + math.erf(dyn_range[0] / math.sqrt(2.0))))
```
The above avoids the unnecessary cast to `torch.Tensor`. Additionally, the correct return type would then also be `Tuple[float, float]`.

narendasan: Is it supposed to be an op on a scalar or on a tensor?

gs-olive: I think this nested function `gelu_dyn_range_function` is supposed to be a function on a Tuple of two scalars.

peri044 (approved):
LGTM pending CI

PR #2138, opened 2023-07-25
feat: Add support for `require_full_compilation` in Dynamo

### LINKED ISSUE #2131: ✨[Feature] Add `min_block_size` exemption for fully-supported models
# Context
Minimum block size can be disregarded in the partitioner if the entire model is end-to-end supported.

# Proposal
Add a special case to the partitioner for end-to-end support and automatically enable any block size in that scenario.

gs-olive:
# Description

- When a graph is fully supported, we can ignore the minimum block size argument, which is primarily helpful in reducing segmentation. If the minimum block size is above the number of total operators in the graph, and we support all of those, the whole graph will run in Torch regardless. As a result, we can exempt fully supported graphs from the min block size requirement

Fixes #2131

## Type of change

# Checklist:

  - CI verification


py/torch_tensorrt/dynamo/lowering/_partition.py
```diff
@@ -85,7 +90,11 @@ def propose_partitions(self) -> List[Partition]:
                 ):
                     compute_node_count += 1
 
-            if compute_node_count < self.min_block_size and not exempted_partition:
+            if (
+                compute_node_count < self.min_block_size
+                and not exempted_partition
+                and not full_support
+            ):
```
gs-olive, line 97: Alternatively, if preferable, we can display a warning in such a case, but still respect the minimum block size argument

py/torch_tensorrt/dynamo/lowering/_partition.py
```diff
@@ -55,12 +56,23 @@ def __init__(
...
         # Propose partitions using the default, then refine the results
         initial_proposed_partitions = super().propose_partitions()
         partitions = {i: part for i, part in enumerate(initial_proposed_partitions)}
 
+        # A graph is fully supported if there is a single partition and all operators are supported/convertible
+        full_support = len(partitions) == 1 and not getattr(
+            self.operator_support, "unsupported_operators", True
+        )
+
+        if not full_support and self.require_full_compilation:
```
gs-olive, line 71: If `min_block_size` different from default, add warning to user.

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
```diff
@@ -185,6 +196,27 @@ def partition_graph(self) -> torch.fx.GraphModule:
...
+
+        if not full_support and self.require_full_compilation:
+            raise AssertionError(
+                "require_full_compilation=True was specified, but model is not fully supported"
+            )
+
+        if (
+            full_support
+            and self.require_full_compilation
+            and self.settings.min_acc_module_size != MIN_BLOCK_SIZE
```
peri044, line 212: would the default min_acc_module_size be 0 ?

tests/py/dynamo/backend/test_partitioning.py
```diff
@@ -30,6 +30,52 @@ def forward(self, x, y):
             "Single operators should not be segmented",
         )
 
+    def test_partition_fully_supported_one_op_require_full_compilation(self):
```
peri044, line 33: How about we move test_partitioning out of backend directory and test with both export and compile ?

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
gs-olive: The default would be whatever `MIN_BLOCK_SIZE` is, since that is how we specify `settings.min_acc_module_size` in the constructor. `min_acc_module_size` is effectively identical to `min_block_size`, just with a different name.

tests/py/dynamo/backend/test_partitioning.py
gs-olive: Ok that sounds good. These test cases already test just the partitioners and not the `torch.compile` path explicitly, so the tests will work for both paths as-is. Should I make a new test directory `shared_utilities`, containing `test_decompositions`, `test_partitioning`, and `test_compiler_utils`?

peri044: sounds good

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
peri044: sounds good

peri044 (approved):
LGTM

tests/py/dynamo/shared_utilities/test_partitioning.py
gs-olive, line 1: mirror structure of `/dynamo/`

tests/py/dynamo/models/__init__.py
gs-olive, line 1: move model files here

PR #2161, opened 2023-08-01
feat: Add `_to_copy`, `operator.get` and `clone` ATen converters

### LINKED ISSUE #2058: ↔ [Converter] Add support for Granular Op-Support Checks in Dynamo
## Add support for `operator.get` and `torch.ops.aten._to_copy` and `torch.ops.aten.clone`

For `operator.get` we need an evaluator-style implementation which simply dispatches the call out to the Python implementation. For `torch.ops.aten._to_copy`, the implementation will be a bit more challenging, since we can only support a subset of these operations. Currently, a simple way to perform the differentiation of which `_to_copy` ops we can handle is to modify the support logic to check if the `to` converter has a cast which can be performed in TRT. An example of an allowed cast might be `torch.float -> torch.int`, whereas `torch.int ->torch.long` is not allowed:
https://github.com/pytorch/TensorRT/blob/6e4aa0b03b91ac56279aacd0d4b1877fcc6ccd7e/py/torch_tensorrt/dynamo/backend/lowering/_partition.py#L109-L131

## Long Term Solution
A long term solution to this issue of selective conversion is to alter the `CONVERSION` dictionary to also store an auxiliary function in addition to its converter. This function will take an input node of the converter's type and return whether that node can be converted. For most converters, this function will be: `lambda node: True`. For `_to_copy`, however, it might be something like:
```python
def converts_to_copy(node: torch.fx.Node):
    return "dtype" in kwargs and kwargs["dtype"] in APPROVED_CAST_DTYPES
```

gs-olive:
# Description
- Add ATen converters for key operators in the pipeline of multiple models
- Add robust testing and patch issues in interpreter
- Add evaluator and casting utilities to the converter utils

Fixes #2058

## Type of change

# Checklist:


tests/py/dynamo/converters/test_casts.py
```diff
@@ -0,0 +1,57 @@
...
+                return y
+
+        inputs = [torch.rand((1, 3, 10))]
+        self.run_test(
+            ToCopyHalf(),
+            inputs,
+            expected_ops={torch.ops.aten._to_copy.default},
+            precision=torch.half,
+            disable_passes=True,
+        )
```
apbose, line 22: torch.rand() returns float32. So is casting to float16 allowed?

tests/py/dynamo/converters/harness.py
```diff
@@ -234,6 +234,10 @@ def generate_graph(
         # Combine with customized passes specific to any model
         if customized_passes:
             passes_list.extend(customized_passes)
+
+        if disable_passes:
+            passes_list = []
+
         fx_module, _ = aten_tracer.trace(mod, original_inputs)
```
apbose, line 241: This is a good addition. So I am assuming we should ideally set `disable_passes` for the subsequent tests?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -380,3 +380,77 @@ def aten_ops_permute(
...
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.evaluators.getitem(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
```
gs-olive, line 439:
```python
@dynamo_tensorrt_converter(operator.getitem, capability_validator=fn)
...
```

Out of the box supported schema registry --> evaluatable schemas. ConverterRegistry performs check of node schema in evaluatable schemas.

Create `DYNAMO_EVALUATORS` schema registry. `torch.fx.Node` get Schema ([operatorSchemas](https://github.com/pytorch/pytorch/blob/656412f0cb21de9880180b3bc2053c376bf547f6/torch/fx/operator_schemas.py#L137))

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -380,3 +380,77 @@ def aten_ops_permute(
         args[0],
         args[1],
     )
+
+
+def to_copy_dtype_validator(to_copy_node: Node) -> bool:
```
gs-olive, line 385: Check if `dtype` accesses on `args` works at validation time.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -380,3 +380,77 @@ def aten_ops_permute(
         args[0],
         args[1],
     )
+
```
gs-olive, line 383: Schema validator --> validates a node is a part of a set of schemas

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
gs-olive:
```python
@dyn...
@dyn...
@dyn...
@dyn...
@dyn...
def convert_eval(...):
    return node.target(*args, **kwargs)
```

tests/py/dynamo/converters/test_casts.py
gs-olive: Yes, I think casting to float16 is allowed in TorchScript, so I have left it enabled here

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -380,3 +380,77 @@ def aten_ops_permute(
...
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        kwargs["dtype"],
+    )
+
+
+@dynamo_tensorrt_converter(operator.getitem)
+def operator_getitem(
```
gs-olive, line 425: `operator.getitem` is a companion-converter. It can only be converted if its parent node is also converted. This is important to get correct for partitioning purposes. Otherwise, there are numerous cases where the inputs to subsequent layers can break invariants.

apbose (approved):

PR #2175, opened 2023-08-04
feat: Improve Dynamo partitioning System Performance on Large Models

gs-olive:
# Description
## Problem Context
The Dynamo partitioning system was very slow for large models (>1000 Nodes) with segmentation. The existing partitioner was using an exhaustive partitioning mechanism which was more than quadratic in the number of nodes, and worsened with more segmentation. This new system uses a simpler adjacency-based partitioning system which is much more performant on large models.

- Upgrade Dynamo partitioning to use a custom version of the Torch _SplitterBase for efficiency and optimized usage in the Dynamo case
- Validate existing use cases are still functional, with the same partitioning schema as before
- Upgrade qualified name checking

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/lowering/_partition_old.py
gs-olive, line 1: This is the old partitioning system file, as-is.

py/torch_tensorrt/dynamo/lowering/_partition.py
```diff
@@ -164,11 +87,139 @@ def print_support_overview(self, num_trt_blocks: Optional[int] = None):
             logger.debug("\nAll Nodes Supported\n")
 
 
+class TRTPartitioner(_SplitterBase):
```
narendasan, line 90: What is the correctness of this Partitioner vs the Capability one? We could keep both behind a flag or something if there is a difference

narendasan: If we want to do this, Id suggest a structure similar to runtime in a submodule

narendasan: Since inductor is on Capability partitioner id expect better maintenance, correctness and improvements to performance

narendasan: Alternatives in pytorch: https://github.com/pytorch/pytorch/blob/df8abaaf5f7e19d897aa30896bdbbe416ca2f825/torch/fx/experimental/accelerator_partitioner.py#L287
https://github.com/pytorch/pytorch/blob/df8abaaf5f7e19d897aa30896bdbbe416ca2f825/torch/fx/passes/infra/partitioner.py#L34

gs-olive: Fast vs full partitioning --> for compile, fast fits best, set as default, give users the option to upgrade to full partitioner.

py/torch_tensorrt/dynamo/lowering/_partition.py
gs-olive, line 1: Add directory `partitioning/` - file for `_adjacency_partitioner.py` and `_global_partitioner.py`

gs-olive: New test cases for each partitioning system.

py/torch_tensorrt/dynamo/partitioning/__init__.py
gs-olive, line 1: Users can toggle between the `fast_partition` (adjacent-segment-merge), or `global_partition` (global-segment-merge) via the `use_fast_partition` flag, which defaults to `True`.

py/torch_tensorrt/dynamo/compile.py
```diff
@@ -111,17 +113,13 @@ def compile(
...
     settings = CompilationSettings(**compilation_options)
-    if kwargs.get("use_capability_partitioner", None):
-        model = lower_model(gm, torch_inputs)
-        return _compile_module(model, torch_inputs, settings)
-    else:
-        split_result = lower_model_using_trt_splitter(gm, torch_inputs)
-        trt_module = _compile_graph(split_result, torch_inputs, settings)
 
-        return trt_module
+    model = lower_model(gm, torch_inputs)
```
gs-olive, line 121: Can remove

gs-olive:
- Inform user if none of the nodes are supported/no valid partitions
- Automatically fall back to global partitioning if adjacency/fast partitioning fails
  - Alert user via warning
  - Show trace in debug logs

PR #2194, opened 2023-08-11
feat: Improve Logging in Dynamo

peri044:
# Description

* Move compile_module from backends to top level since it is used by export and compile.
* Use logging library in `torch_tensorrt/_compile.py`. The original logging.py is using C++ utilities which probably isn't needed for this file.
* Define a root level logger with namespace `torch_tensorrt.dynamo` defined in `torch_tensorrt/dynamo/__init__.py`. Setting debug=True now invokes all the module level loggers to follow this level since they are children of the root level logger. This fixes partitioning log in the cases where it doesn't get logged sometimes.

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/aten_tracer.py
```diff
@@ -85,7 +87,6 @@ def setting_python_recursive_limit(limit: int = 10000) -> Generator[None, None,
         sys.setrecursionlimit(default)
 
 
-@req_torch_version("2.dev")
```
gs-olive, line 88: These function wrappers are intended to warn users compiling with the legacy library that these functions will not work with Torch 1.13. Are they removed here to be replaced by alternative log messaging?

gs-olive (commented):
Unification of the `compile_module` function and the logging bugfix look great! Looks like one of the ts tests has a logging conflict since it expects the C++ version, so the import could be specialized more for that test case. Left a comment regarding version requirements for legacy.

py/torch_tensorrt/dynamo/aten_tracer.py
peri044: So mypy complains about these decorators. I've changed the decorators according to mypy conventions. This should be resolved now.

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -169,7 +174,7 @@ def parse_dynamo_kwargs(kwargs: Any) -> CompilationSettings:
     return settings
 
 
-def req_torch_version(min_torch_version: str = "2.dev") -> Callable[..., Any]:
+def req_torch_version(func: Callable[P, T]) -> Callable[P, T]:
```
narendasan, line 177: Why was this changed?

narendasan: The arguments dont seem right

peri044: reverted back. Seems like mypy passes now with no changes. Although I'm not sure why it was giving issues before

PR #2198, opened 2023-08-14
Type mismatch for dynamo aten::where converter

### LINKED ISSUE #2179: Unclear comparisons in conditional operator implementation
https://github.com/pytorch/TensorRT/blob/918e9832207d462c2a3aa42f9e7d3ab7aa7415aa/py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py#L68
https://github.com/pytorch/TensorRT/blob/918e9832207d462c2a3aa42f9e7d3ab7aa7415aa/py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py#L75-L76
https://github.com/pytorch/TensorRT/blob/918e9832207d462c2a3aa42f9e7d3ab7aa7415aa/py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py

The above lines are being flagged by mypy as the following:

```sh
py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py:68: error: Non-overlapping equality check (left operand type: "list[Any]", right operand type: "int")  [comparison-overlap]
py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py:76: error: Non-overlapping equality check (left operand type: "list[Any]", right operand type: "int")  [comparison-overlap]
py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py:98: error: Non-overlapping equality check (left operand type: "list[Any]", right operand type: "int")  [comparison-overlap]
```

I cant really figure out what is being checked here but it is likely a bug.

apbose:
# Description

Fixes #2179 
Adds test case for the case where y_other dimension will not be equal to the broadcasted dimension
Further to do: Remove the evaluator section, since as discussed, this aten::where.self overload would have an unique schema having Tensor inputs, and by the time we get to partitioning/compilation the tracer already pre evaluates the constants

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
@@ -65,7 +65,7 @@ def where(
         condition_val = condition_layer.get_output(0)
     else:
         assert condition.dtype == trt.bool, "mask dtype is not bool!"
-        if condition_shape != condition_dim:  # TODO: What is this checking?
+        if len(condition_shape) != condition_dim:
             condition_val = expand(
                 network, target, source_ir, f"{name}_expand", condition, output_shape
             )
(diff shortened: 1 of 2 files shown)


narendasan (approved):
LGTM

PR #2234, opened 2023-08-16
fix: Allow rank differences in `aten.expand`

### LINKED ISSUE #2183: 🐛 [Bug] `aten.expand` fails when rank disagrees with tensor shape
##  Bug Description

When the rank and tensor shape disagree, the `torch.aten.ops.expand` operator fails due to this portion of the code:
https://github.com/pytorch/TensorRT/blob/8c62fca7649ac2e8e74ab7d9d66a20a164e2af51/py/torch_tensorrt/fx/converters/acc_ops_converters.py#L2457-L2475
This is not in agreement with Torch behavior, where calling `.expand` on a Tensor does not require that the expanded size have the same rank as the original Tensor. [See documentation here](https://pytorch.org/docs/stable/generated/torch.Tensor.expand.html).

```python
import torch
x = torch.ones(2, 2)
y = x.expand([5, 5, 5, 5, -1, -1])
print(y.shape)
>>> torch.Size([5, 5, 5, 5, 2, 2])
```

This is the error message in the converter:
```python
File "~/TensorRT/py/torch_tensorrt/fx/converters/acc_ops_converters.py", line 2475, in acc_ops_expand_tensor
     assert len(shape) == ranks
 torch._dynamo.exc.BackendCompilerFailed: backend='torch_tensorrt' raised:
 AssertionError: While executing %expand_1 : [num_users=1] = call_function[target=torch.ops.aten.expand.default](args = (%permute_3, [1, 512, 512]), kwargs =...
```
## To Reproduce

See above code snippet for desired behavior from converter.

## Expected behavior

Converter should succeed in this case.

## Environment

 - Torch-TensorRT Version (e.g. 1.0.0): 8c62fca7649ac2e8e74ab7d9d66a20a164e2af51
 - PyTorch Version (e.g. 1.0): `2.1.0.dev20230803+cu121`

gs-olive:
# Description

- Add support for `aten.expand.default` in Dynamo converter registry
- Build converter to support rank-padding for input Tensors, in line with the existing Torch behavior
- Add test case to validate new behavior, in addition to existing cases validating old behavior

Fixes #2183

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -420,3 +420,21 @@ def aten_ops_clone(
         name,
         args[0],
     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.expand.default)
+def aten_ops_expand(
+    network: TRTNetwork,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.slice.expand(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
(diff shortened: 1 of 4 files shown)


apbose (commented):
As such the changes look good.  But I think this was already taken care in the `slice/ops.py`. This condition was getting encountered in `aten::where` cases where the two inputs were not of equal dimensions. However since there was no `aten.ops.expand` in dynamo, it was picking up the `acc_ops` implementation. Its good that a test case for this in `expand` has been added.
We should either keep this or the slice one, and accordingly change the call to expand elsewhere.

gs-olive:
Thanks for the comments and suggestions. I have merged the implementations into one `expand` implementation now, which is in `slice/ops`.

narendasan (approved):
LGTM

PR #2238, opened 2023-08-17
chore: Fix Logging in torch_compile path 

peri044:
# Description

Set the log level in torch_compile path at the start of compilation

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -20,6 +20,12 @@
 def torch_tensorrt_backend(
     gm: torch.fx.GraphModule, sample_inputs: Sequence[torch.Tensor], **kwargs: Any
 ) -> torch.nn.Module:
+    # Set log level at the top of compilation (torch_tensorrt.dynamo)
+    if "options" in kwargs:
+        if "debug" in kwargs["options"]:
+            if logger.parent:
+                logger.parent.setLevel(logging.DEBUG)
```
gs-olive, line 27: Could this be moved into the `parse_dynamo_kwargs` function in `dynamo/utils`, under the `if settings.debug` section? If not, could the logic be changed to:
```python
if ((("options" in kwargs and "debug" in kwargs["options"] and kwargs["options"]["debug"])
   or ("debug" in kwargs and kwargs["debug"]))
   and logger.parent):
    logger.parent.setLevel(logging.DEBUG)
```

gs-olive (commented):
Changes look good - just made a conditional logic suggestion and question.

py/torch_tensorrt/dynamo/backend/backends.py
peri044: This is much cleaner. Thanks. I'd like to set the debug level on the entry points of compilation (for both export and compile), that way we don't have to set it anywhere else. Also, made the `CompilationSettings` object logging to info (similar to export).

gs-olive (approved):
Looks good to me!

Issue #2095, opened 2023-07-10
↔ [Converter] Add support for my_op in Torch-TensorRT


narendasan:
##

- **Function Schema**:

> torch.ops.aten.amax.default

- **Original PyTorch API**:

https://pytorch.org/docs/stable/generated/torch.amax.html

- **Relevant TensorRT Documentation**:

cc: @zewenli98

PR #2241, opened 2023-08-17
feat: support amax dynamo converter

zewenli98:
# Description

Support amax dynamo converter.

- Function Schema:
`torch.ops.aten.amax.default`

- Original PyTorch API:
https://pytorch.org/docs/stable/generated/torch.amax.html

Fixes issue #2095 

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -420,3 +420,28 @@ def aten_ops_clone(
...
+        input_val = cast_trt_tensor(network, input_val, trt.float32, name)
+
+    return impl.reduce.amax(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        input_val,
+        args[1],
+        args_bounds_check(args, 2, replacement=False),
```
narendasan, line 446: @gs-olive can this check be done in a validator?

gs-olive: This check can be done in a validator, but in this context it would only make sense for it to be done here, since we can support cases where this argument is both present and absent.

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -0,0 +1,36 @@
...
+
+def amax(
+    network: TRTNetwork,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: TRTTensor,
+    dim: Union[int, Tuple[int]],
+    keep_dims: Optional[bool] = False,
+    out: Optional[Any] = None,
```
gs-olive, line 19: This argument can be removed

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def amax(
+    network: TRTNetwork,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: TRTTensor,
+    dim: Union[int, Tuple[int]],
+    keep_dims: Optional[bool] = False,
```
gs-olive, line 18: The `Optional` here can be removed, since `keepdim` is a mandatory argument, as per [this documentation](https://github.com/pytorch/pytorch/blob/a408920817ac3852d27fc96ba85109ff4c981d52/aten/src/ATen/native/native_functions.yaml#L3643-L3646)

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -0,0 +1,38 @@
...
+    if dim is None:
+        raise ValueError("amax requires specifying dimension(s) (dim).")
+
+    layer = network.add_reduce(
+        input_val,
+        trt.ReduceOperation.MAX,
+        axes=get_axes_for_reduce_op(dim),
+        keep_dims=keepdim,
+    )
+    set_layer_name(layer, target, name)
```
gs-olive, line 37: Add the `source_ir` as the fourth argument in this function call, to make the layer name more descriptive

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -0,0 +1,38 @@
...
+    source_ir: Optional[SourceIR],
+    name: str,
+    input_val: TRTTensor,
+    dim: Union[int, Tuple[int]],
+    keepdim: bool = False,
+) -> TRTTensor:
+    if (isinstance(input_val, TRTTensor)) and (
+        input_val.dtype == trt.int8 or input_val.dtype == trt.int32
+    ):
+        input_val = cast_trt_tensor(network, input_val, trt.float32, name)
```
gs-olive, line 26: Is this required for the `IReduceLayer`?

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -0,0 +1,38 @@
...
+
+    if dim is None:
+        raise ValueError("amax requires specifying dimension(s) (dim).")
```
gs-olive, line 29: This check should not be necessary, since the top-level converter contract enforces that Torch will provide an integer or list of integers.

gs-olive (commented):
Converter looks great, and the testing is very robust! Made a few small comments

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
zewenli98: Yes, it's required because `TypeError: Provided an unsupported data type as an input data type (support: bool, int32, long, half, float), got: torch.int8` if I test int8. Even though the [doc](https://docs.nvidia.com/deeplearning/tensorrt/operators/docs/Reduce.html#data-types) says Reduce Layer supports int32, but I tested it doesn't support.

zewenli98:
> Converter looks great, and the testing is very robust! Made a few small comments

Thanks for your review! I've solved all the issues and pushed.

zewenli98:
It looks like failed on C++ tests that's nothing to do with this commit?

gs-olive:
I reran that C++ test [here](https://app.circleci.com/pipelines/github/pytorch/TensorRT/4320/workflows/fe76b3a2-5025-426a-be18-87f146a2a583/jobs/20208)

zewenli98:
Passed! please let me know if anything need to fix in this PR.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -440,6 +440,25 @@ def aten_ops_expand(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.amax.default)
```
gs-olive, line 443: Based on [this schema](https://github.com/pytorch/pytorch/blob/1ef4bd169d4ea7729496e5734e87f26f8ff6f640/aten/src/ATen/native/native_functions.yaml#L3647-L3650), it seems that the dimension (`dim`) can also be not present. For instance, the following model creates the subsequent graph:
```python
class argmax(torch.nn.Module):
   def forward(self, x):
        return torch.argmax(x)

"""
graph():
    %argmax : [num_users=1] = call_function[target=torch.ops.aten.argmax.default](args = (%arg0_1,), kwargs = {})
    return (argmax,)
"""
```
If we cannot support this case, you can add a `capability_validator` function to this decorator, which will note that case as unsupported. Roughly, that could be something like:

```python
def amax_param_validator(amax_node: Node) -> bool:
    return len(amax_node.args) >= 2

@dynamo_tensorrt_converter(
    torch.ops.aten.amax.default, capability_validator=amax_param_validator
)  # type: ignore[misc]
```

zewenli98: @gs-olive Thank you so much George! That's a good catch. I fixed and learned a lot from these details recently :smiley:

gs-olive: No problem - glad to hear it!

gs-olive (approved):
Looks great!

PR #2246, opened 2023-08-19
feat: support many unary dynamo converters

### LINKED ISSUE #2199: Get IUnaryLayer Ops exposed in dynamo.conversion.impl

zewenli98:
# Description

Implemented all 22 unary dynamo converters, including `EXP`, `LOG`, `SQRT`, `RECIP`, `ABS`, `SIN`, `COS`, `TAN`, `SINH`, `COSH`, `ASIN`, `ACOS`, `ATAN`, `ASINH`, `ACOSH`, `ATANH`, `CEIL`, `FLOOR`, `NOT`, `SIGN`, `ROUND`, `ISINF`.

Fixes #2199 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -24,7 +24,7 @@ def args_bounds_check(
     return args[i] if len(args) > i else replacement
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.batch_norm)
+@dynamo_tensorrt_converter(torch.ops.aten.batch_norm)  # type: ignore[misc]
 def aten_ops_batch_norm(
     network: TRTNetwork,
     target: Target,
@@ -48,9 +48,9 @@ def aten_ops_batch_norm(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.div.default)
-@dynamo_tensorrt_converter(torch.ops.aten.div.Tensor_mode)
-@dynamo_tensorrt_converter(torch.ops.aten.div.Tensor)
+@dynamo_tensorrt_converter(torch.ops.aten.div.default)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.div.Tensor_mode)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.div.Tensor)  # type: ignore[misc]
 def aten_ops_div(
     network: TRTNetwork,
     target: Target,
@@ -119,7 +119,7 @@ def embedding_param_validator(embedding_node: Node) -> bool:
 
 @dynamo_tensorrt_converter(
     torch.ops.aten.embedding.default, capability_validator=embedding_param_validator
-)
+)  # type: ignore[misc]
 def aten_ops_embedding(
     network: TRTNetwork,
     target: Target,
@@ -140,8 +140,8 @@ def aten_ops_embedding(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.fmod.Scalar)
-@dynamo_tensorrt_converter(torch.ops.aten.fmod.Tensor)
+@dynamo_tensorrt_converter(torch.ops.aten.fmod.Scalar)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.fmod.Tensor)  # type: ignore[misc]
 def aten_ops_fmod(
     network: TRTNetwork,
     target: Target,
@@ -152,7 +152,7 @@ def aten_ops_fmod(
     return impl.elementwise.fmod(network, target, SourceIR.ATEN, name, args[0], args[1])
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.gelu.default)
+@dynamo_tensorrt_converter(torch.ops.aten.gelu.default)  # type: ignore[misc]
 def aten_ops_gelu(
     network: TRTNetwork,
     target: Target,
@@ -169,8 +169,8 @@ def aten_ops_gelu(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.matmul)
-@dynamo_tensorrt_converter(torch.ops.aten.mm.default)
+@dynamo_tensorrt_converter(torch.ops.aten.matmul)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.mm.default)  # type: ignore[misc]
 def aten_ops_matmul(
     network: TRTNetwork,
     target: Target,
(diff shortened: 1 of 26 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -405,7 +405,7 @@ def aten_ops_to_copy_dtype(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.clone.default)
+@dynamo_tensorrt_converter(torch.ops.aten.clone.default)  # type: ignore[misc]
```
narendasan, line 408: Before this is committed check to see if theres a way to fix this mypy error @gs-olive can you look at the decorator as well?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -420,3 +420,523 @@ def aten_ops_clone(
...
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    input_val = args[0]
+    if (isinstance(input_val, TRTTensor)) and (
+        input_val.dtype == trt.int8 or input_val.dtype == trt.int32
+    ):
+        input_val = cast_trt_tensor(network, input_val, trt.float32, name)
+
+    return impl.unary.unary_op(
```
narendasan, line 463: Expose these as dedicated functions in impl

narendasan (commented):
@zewenli98 We need a layer of indirection between aten and impl which is why functions like sign had dedicated implementations.

For each op we need:
1. aten registration for the actual converter in aten_converters.py
2. a dedicated impl function 


So for `aten::cos`
We need: 
1. aten_ops_cos (converison.aten_ops_converters)
2. impl.cos (conversion.impl.unary.ops)

zewenli98:
> @zewenli98 We need a layer of indirection between aten and impl which is why functions like sign had dedicated implementations.
> 
> For each op we need:
> 
> 1. aten registration for the actual converter in aten_converters.py
> 2. a dedicated impl function
> 
> So for `aten::cos` We need:
> 
> 1. aten_ops_cos (converison.aten_ops_converters)
> 2. impl.cos (conversion.impl.unary.ops)


I thought there's almost same code for these converters supported by TensorRT, so I created `unary_op` in  `unary/ops.py`, like `elementwise_op` in `elementwise/ops.py`. I was wondering why dedicated implementations are better?

narendasan:
Dedicated implementations are better because they use a standardized interface for all ops. So if I want to make a new converter for say LogSumExp, instead of my code looking like this:

```py
layer = impl.unary_op(in, ..., trt.IUnaryOperation.LOG)
layer = impl.reduce_op(layer.output, ..., trt.IReduceOperation.SUM)
return impl.unary_op(layer.output, ..., trt.IUnaryOperation.EXP)
```

Peoples code would look closer to this:

```py
x = impl.log(in,...)
x = impl.sum(x,...)
return impl.exp(x,...)
```

So a converter writer does not need to have in depth knowledge about tensorrt or its APIs, they can write code more similar to pytorch instead.

zewenli98:
> 

Oh I see. It makes sense. Actually at first I implemented dedicated implementations but after seeing `elementwise_op`, I changed to the current code.  I'll change back and also for elementwise.

narendasan:
So there can be a core function which is centralized. So `unary_op` is good, but effectively we just need an alias for specific ops so users dont need to do that themselves

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -420,3 +420,523 @@ def aten_ops_clone(
...
+@dynamo_tensorrt_converter(torch.ops.aten.tan.default)  # type: ignore[misc]
+def aten_ops_tan(
+    network: TRTNetwork,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    input_val = args[0]
+    if (isinstance(input_val, TRTTensor)) and (
```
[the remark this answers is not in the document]
zewenli98: I don't think it's unary specific. This depends on the data types TensorRT supported. For unary ops, the supported types are: https://docs.nvidia.com/deeplearning/tensorrt/operators/docs/Unary.html#data-types

zewenli98:
OK, I'll change to something like `aten_ops_cos -> impl.cos -> impl.unary.unary_op(..., trt.IUnaryOperation.COS)`

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
gs-olive: I recently noticed that `mypy` does not accurately show errors for these decorators. I will look more into this.

The decorator is already as strongly-typed as it can be though, as here:
https://github.com/pytorch/TensorRT/blob/585b2b66e50a9900f513b3aa495ae3ed210351d4/py/torch_tensorrt/dynamo/conversion/converter_registry.py#L64-L69

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -420,3 +420,501 @@ def aten_ops_clone(
...
+@dynamo_tensorrt_converter(torch.ops.aten.exp.default)  # type: ignore[misc]
+def aten_ops_exp(
+    network: TRTNetwork,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    input_val = args[0]
+    if (isinstance(input_val, TRTTensor)) and (
```
narendasan, line 434: Lets put the input validation in either impl.unary.exp or impl.unary.unary_op,  probably in the op specific part since it seems to vary between ops

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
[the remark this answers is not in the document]
narendasan: Lets put the input validation in either impl.unary.exp or impl.unary.unary_op, probably in the op specific part since it seems to vary between ops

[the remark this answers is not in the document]
zewenli98: OK, I moved input validation into ops, like inside of `impl.unary.exp`

narendasan (approved):
LGTM @zewenli98 rebase and should be good to merge

zewenli98:
Do we still need `# type: ignore[misc]`?

narendasan:
Its fine for now we can resolve it later

zewenli98:
rebased, but I don't have access to merge.

PR #2263, opened 2023-08-25
feat: support many elementwise dynamo converters

### LINKED ISSUE #2208: Expose IElementWiseLayer in dynamo.conversion.impl

zewenli98:
# Description

Support many elementwise dynamo converters, including `add`, `mul`, `maximum`, `minimum`, `sub`, `div`(already implemented), `pow`, `floor_divide`, `logical_and`, `logical_or`, `logical_xor`, `eq`, `gt`, `lt`

Fixes #2208 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1,16 +1,10 @@
 import logging
 from typing import Any, Dict, Optional, Sequence, Tuple, Union
 
-import tensorrt as trt
 import torch
 from torch.fx.node import Argument, Node, Target
 from torch_tensorrt.dynamo._SourceIR import SourceIR
 from torch_tensorrt.dynamo.conversion import impl
-from torch_tensorrt.dynamo.conversion.converter_utils import (
-    cast_int_int_div_trt_tensor,
-    cast_trt_tensor,
-)
-from torch_tensorrt.fx.converters import acc_ops_converters
 from torch_tensorrt.fx.types import TRTNetwork, TRTTensor
 
 from .converter_registry import dynamo_tensorrt_converter
@@ -46,58 +40,6 @@ def aten_ops_batch_norm(
         args[6],
         args[7],
     )
-
-
-@dynamo_tensorrt_converter(torch.ops.aten.div.default)  # type: ignore[misc]
-@dynamo_tensorrt_converter(torch.ops.aten.div.Tensor_mode)  # type: ignore[misc]
-@dynamo_tensorrt_converter(torch.ops.aten.div.Tensor)  # type: ignore[misc]
-def aten_ops_div(
-    network: TRTNetwork,
-    target: Target,
-    args: Tuple[Argument, ...],
-    kwargs: Dict[str, Argument],
-    name: str,
-) -> Union[TRTTensor, Sequence[TRTTensor]]:
-    kwargs_new = {
-        "input": args[0],
-        "other": args[1],
-    }
-    # If both are TRTTensor, both are cast to float32
-    if isinstance(args[0], TRTTensor) and isinstance(args[1], TRTTensor):
-        kwargs_new["input"], kwargs_new["other"] = cast_int_int_div_trt_tensor(
-            network,
-            kwargs_new["input"],
-            kwargs_new["other"],
-            name,
-        )
-    # If one is TRTTensor, it is cast to float32
-    elif isinstance(args[0], TRTTensor) and (
-        kwargs_new["input"].dtype == trt.int8 or kwargs_new["input"].dtype == trt.int32
-    ):
-        kwargs_new["input"] = cast_trt_tensor(
-            network, kwargs_new["input"], trt.float32, name, target
-        )
-    elif isinstance(args[1], TRTTensor) and (
-        kwargs_new["other"].dtype == trt.int8 or kwargs_new["other"].dtype == trt.int32
-    ):
-        kwargs_new["other"] = cast_trt_tensor(
-            network, kwargs_new["other"], trt.float32, name, target
-        )
-    rounding_mode = kwargs.get("rounding_mode")
-    if rounding_mode is None:
-        return acc_ops_converters.acc_ops_div(network, target, None, kwargs_new, name)
-    elif rounding_mode == "floor":
-        return acc_ops_converters.acc_ops_floor_div(
-            network, target, None, kwargs_new, name
-        )
-    elif rounding_mode == "trunc":
-        return impl.elementwise.trunc_div(
-            network, target, SourceIR.ATEN, name, args[0], args[1]
-        )
-    else:
-        raise RuntimeError(
-            f"Target {target} does not support rounding mode {rounding_mode}"
-        )
 
 
 def embedding_param_validator(embedding_node: Node) -> bool:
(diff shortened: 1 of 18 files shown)


zewenli98:
The error seems not to be related to this PR.

gs-olive:
@zewenli98 - taking a look now! Also, could you rebase this one to `main` to resolve the merge conflicts?

zewenli98:
Rebased! thanks!

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
```diff
@@ -239,3 +240,299 @@ def _add_layer(
...
+    source_ir: Optional[SourceIR],
+    name: str,
+    lhs_val: TRTTensor,
+    rhs_val: TRTTensor,
+) -> TRTTensor:
+    if lhs_val.dtype == trt.int8 or lhs_val.dtype == trt.int32:
+        lhs_val = cast_trt_tensor(network, lhs_val, trt.float32, name)
+
+    if rhs_val.dtype == trt.int8 or rhs_val.dtype == trt.int32:
+        rhs_val = cast_trt_tensor(network, rhs_val, trt.float32, name)
```
gs-olive, line 257: The Dynamo error encountered is related to this change, because the type for inputs to all of the `elementwise` layers is generally `Union[TRTTensor, int, float]`, since we can encounter cases where floats get added to `TRTTensor`s for example. As a result, the code here fails, since it tries to access the `dtype` of a Python float.

gs-olive: The above code seems to show up a lot. Consider adding a function in `converter_utils.py`, which performs this check and cast automatically. Something like:

```python
def trt_cast_int_to_float(tensor: TRTTensor):
    if tensor.dtype == trt.int8 or tensor.dtype == trt.int32:
        return cast_trt_tensor(network, tensor, trt.float32, name)

    return tensor
```

gs-olive: Then, you can refactor the code like this, for instance:
```python
    if lhs_val.dtype == trt.int8 or lhs_val.dtype == trt.int32:
        lhs_val = cast_trt_tensor(network, lhs_val, trt.float32, name)

##### REPLACED BY

    if isinstance(lhs_tensor, TRTTensor):
        lhs_tensor = trt_cast_int_to_float(lhs_tensor)
```

gs-olive (changes requested):
See the comments below to fix the Dynamo errors appearing for this PR.

gs-olive (commented):
Two additional changes are needed to fix this CI issue. In `convert_binary_elementwise`, we need to change `Frameworks.TORCH` to `Frameworks.NUMPY`:
https://github.com/pytorch/TensorRT/blob/e49ef6d1de58f80bf51e37428d4a2287ed462871/py/torch_tensorrt/dynamo/conversion/impl/elementwise/base.py#L77-L82
As well as `torch.tensor` to `np.array`:
https://github.com/pytorch/TensorRT/blob/e49ef6d1de58f80bf51e37428d4a2287ed462871/py/torch_tensorrt/dynamo/conversion/impl/elementwise/base.py#L105-L108

The above are fixed in #2265, so let me try to merge that + rebase to resolve the above

zewenli98:
@gs-olive Thanks for the suggestions and detailed explanations! I updated and rebased!

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+#     name: str,
+# ) -> Union[TRTTensor, Sequence[TRTTensor]]:
+#     return impl.elementwise.div(
+#         network,
+#         target,
+#         SourceIR.ATEN,
+#         name,
+#         args[0],
+#         args[1],
+#     )
```
gs-olive, line 954: Refactor this implementation, and then remove the one on line 54, since that one uses the `acc_ops` which are not preferable.

For instance, you can replace this with your `floor_div` implementation:
https://github.com/pytorch/TensorRT/blob/da66673e380fe117fd9a279cb80f5a3197c38f84/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py#L90-L92

Just ensure all of the three variants (`default`, `Tensor`, and `Tensor_mode`) are supported. You can find their schemas [here](https://github.com/pytorch/pytorch/blob/eafc05887f44ed47e0631506d2654c429bb91fb3/aten/src/ATen/native/native_functions.yaml#L2078-L2084), with some rounding mode explanations [here](https://pytorch.org/docs/stable/generated/torch.div.html).

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.floor_divide.default)
```
gs-olive, line 975: I believe this also supports [`torch.ops.aten.floor_divide.Scalar`](https://github.com/pytorch/pytorch/blob/eafc05887f44ed47e0631506d2654c429bb91fb3/aten/src/ATen/native/native_functions.yaml#L2686-L2688). You can stack that decorator here as well.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.eq.Tensor)
```
gs-olive, line 1047: This also may support [`torch.ops.aten.eq.Scalar`](https://github.com/pytorch/pytorch/blob/eafc05887f44ed47e0631506d2654c429bb91fb3/aten/src/ATen/native/native_functions.yaml#L8504-L8511) (not sure about Scalar-Tensor equivalence checking in TRT).

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.gt.Tensor)
```
gs-olive, line 1065: Potentially can support Scalar variant as well, see above.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.lt.Tensor)
```
gs-olive, line 1083: Potentially can support Scalar variant as well, see above.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+#         network,
+#         target,
+#         SourceIR.ATEN,
+#         name,
+#         args[0],
+#         args[1],
+#     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.pow.Tensor_Tensor)
```
gs-olive, line 957: This can likely support [`torch.ops.aten.pow.Tensor_Scalar`](https://github.com/pytorch/pytorch/blob/eafc05887f44ed47e0631506d2654c429bb91fb3/aten/src/ATen/native/native_functions.yaml#L9831-L9837) as well.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.sub.Tensor)
```
gs-olive, line 920: Consider adding [`torch.ops.aten.sub.Scalar`](https://github.com/pytorch/pytorch/blob/eafc05887f44ed47e0631506d2654c429bb91fb3/aten/src/ATen/native/native_functions.yaml#L6546-L6551)

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.mul.Tensor)
```
gs-olive, line 866: Add [`mul.Scalar`](https://github.com/pytorch/pytorch/blob/eafc05887f44ed47e0631506d2654c429bb91fb3/aten/src/ATen/native/native_functions.yaml#L4011-L4018)

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.elementwise.add(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
```
gs-olive, line 863: The Torch [`add` operator](https://pytorch.org/docs/stable/generated/torch.add.html) has a mandatory final argument which is `alpha`, [see here](https://github.com/pytorch/pytorch/blob/eafc05887f44ed47e0631506d2654c429bb91fb3/aten/src/ATen/native/native_functions.yaml#L533-L543). You can support this in this converter by checking if `args_bounds_check(args, 2, replacement=1) == 1`, and if so, return what you return here, otherwise scale `other` by `alpha` using `impl.elementwise.mul`.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.add.Tensor)
```
gs-olive, line 848: Add [`add.Scalar`](https://github.com/pytorch/pytorch/blob/eafc05887f44ed47e0631506d2654c429bb91fb3/aten/src/ATen/native/native_functions.yaml#L599-L604)

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -845,6 +845,259 @@ def aten_ops_isinf(
...
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.elementwise.sub(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
```
gs-olive, line 935: The Torch [`sub` operator](https://pytorch.org/docs/stable/generated/torch.sub.html) has a mandatory final argument which is `alpha`, [see here](https://github.com/pytorch/pytorch/blob/eafc05887f44ed47e0631506d2654c429bb91fb3/aten/src/ATen/native/native_functions.yaml#L6527-L6535). You can support this in this converter by checking if `args_bounds_check(args, 2, replacement=1) == 1`, and if so, return what you return here, otherwise scale `other` by `alpha` using `impl.elementwise.mul`.

gs-olive (changes requested):
Changes look great - added some comments on schemas, converter support, and the `div`, `add`, and `sub` operators, which have special cases

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: Thanks! but when writing tests for `aten.floor_divide.Scalar`, I tried:
```
@parameterized.expand(
        [
            ("2d", (2, 1), 1.0),
            ("3d", (2, 1, 2), 2.0),
        ]
    )
    def test_floor_div_scalar(self, _, shape, scalar):
        class floor_div(nn.Module):
            def forward(self, lhs_val):
                return torch.floor_divide(lhs_val, torch.tensor(scalar))  # also tried directly using `scalar`

        inputs = [torch.randn(shape)]
        self.run_test(
            floor_div(),
            inputs,
            expected_ops={torch.ops.aten.floor_divide.Scalar},
        )
```
but got error: `AssertionError: False is not true : expected ops {<OpOverload(op='aten.floor_divide', overload='Scalar')>}, actuall ops {<OpOverload(op='aten.floor_divide', overload='default')>}`. Do you know in what scenarios we use `aten.floor_divide.Scalar`?

gs-olive: I see - that's interesting - it seems Torch traces the scalars to `default` as well. Since it is a valid schema based on that link, I think it is okay to keep the Scalar overload, but use `expected_ops={torch.ops.aten.floor_divide.default},` in the test case.

zewenli98:
@gs-olive Thanks for the review! Resolved issues above.

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -97,7 +97,7 @@ def cast_trt_tensor(
 
     if input_val.dtype != trt_dtype:
         source_ir = source_ir if source_ir is not None else SourceIR.UNKNOWN
-        target_str = ConverterRegistry.qualified_name_or_str(target)
+        target_str = ConverterRegistry.qualified_name_or_str(ConverterRegistry, target)
```
gs-olive, line 100: This function takes one argument - should just be `target_str = ConverterRegistry.qualified_name_or_str(target)`

gs-olive (changes requested):

py/torch_tensorrt/dynamo/conversion/converter_utils.py
zewenli98: Ohh I did this because the old `qualified_name_or_str()` reported errors if not changed. nvm, fixed!

gs-olive:
@zewenli98 - Seeing the following error in both test failures for this PR:
```python
[TRT] [E] 4: [network.cpp::inferOutputTypes::2063] Error Code 4: Internal Error (Output tensor output0 of type Int32 produced from output of incompatible type Float)
```
I assume it is because the forward function in those tests is like so:
```python
        class Tensor0DInput(torch.nn.Module):
            def forward(self, x):
                return x * 7
```
The input tensor `x` is an integer Tensor, so Torch provides an `Int32` output (meaning we also expect an Int32 output), but due to the new converters, we are casting all integers for these elementwise ops to floats. I know that some elementwise converters require float inputs, but do all of the converters require this? As in - could `aten.mul` work without this cast?

gs-olive:
For instance - in TorchScript, I don't see float casts for `mul` or `add`, only `div`.

gs-olive (changes requested):
New changes look great! See comments above on narrowing the usage of float casting.

py/torch_tensorrt/dynamo/conversion/impl/elementwise/base.py
zewenli98, line 1: @gs-olive Thanks for your review! I just found a similar error. For some ops, such as `eq`, we have to specify `output_dtypes=[torch.bool]` in tests, otherwise, we will get error: `[TRT] [E] 4: [network.cpp::inferOutputTypes::2063] Error Code 4: Internal Error (Output tensor output0 of type Int32 produced from output of incompatible type Bool)`.

However, the weird is that if we don't specify `output_dtypes=[torch.bool]`, instead, just add `output.dtype` (do nothing) before `return output` in `convert_binary_elementwise`. The error disappeared and it can pass the test. So, I was wondering if this is kind of like lazy mode, requiring running `output.dtype` to get right type?

gs-olive: That is definitely strange, calling `output.dtype` shouldn't have an effect on its own, to my knowledge. The way we currently determine output data types in our _torch.compile_ backend is we run the graph in Torch and see the output types, then we set the TRT engine outputs according to this. This becomes problematic if TRT requires a float cast where Torch does not (for instance, on Int, Int adds or multiplies). For this reason, we do not need float casting in add or mul, as there is currently not float casting in our TorchScript path. On the other hand, we do need bool casting for eq, as you observed, since the output will otherwise be Int32.

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
```diff
@@ -239,3 +243,299 @@ def _add_layer(
...
+    source_ir: Optional[SourceIR],
+    name: str,
+    lhs_val: Union[TRTTensor, int, float],
+    rhs_val: Union[TRTTensor, int, float],
+) -> TRTTensor:
+    if isinstance(lhs_val, TRTTensor):
+        lhs_val = trt_cast_int_to_float(network, name, lhs_val)
+
+    if isinstance(rhs_val, TRTTensor):
+        rhs_val = trt_cast_int_to_float(network, name, rhs_val)
```
gs-olive, line 260: This can be removed, unless there is reason to cast the `add` operator to a float.

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
```diff
@@ -239,3 +243,299 @@ def _add_layer(
...
```
gs-olive, line 279: This can also be removed

gs-olive (changes requested):
See the [`cast_int_int_div_trt_tensor`](https://github.com/pytorch/TensorRT/blob/bb5bf0069a4198498d27f1ab827c2f80476768b9/py/torch_tensorrt/dynamo/conversion/converter_utils.py#L111) function and cross-validate with the [C++ implementations](https://github.com/pytorch/TensorRT/blob/bb5bf0069a4198498d27f1ab827c2f80476768b9/core/conversion/converters/impl/element_wise.cpp) to verify where casts are needed for the elementwise operators. I think very few of these need `trt_cast_int_to_float`. Please let me know if there are any resources that indicate otherwise.

zewenli98:
yes! Thanks for the details! I just wanted to check if adding `output.dtype` is useful (control variables) in this commit. I'll let you know after resolving all the issues!

zewenli98:
@gs-olive Let me give some explanations here. At first, I referred to this doc https://docs.nvidia.com/deeplearning/tensorrt/operators/docs/ElementWise.html#data-types where it says `int32` is not supported for all elementwise ops. However, after some tests, I found only `div` and `pow` don't support int32. Accordingly, I made these changes in this commit. But it looks like there's an error about `pip` in the smoke test.

gs-olive:
Understood - thanks for this resource. In general, I would recommend using the C++ TRT documentation for input restrictions. [This link](https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/namespacenvinfer1.html#aacd7b0a3338921b2a0c87570295beef4) for the C++ API suggests that `POW` is restricted to `float`, `half`, and `bool`, whereas all other ops can also have `int32` (though `div` is an outlier, as you mentioned). Regarding the `pip` error, it is resolved in #2298 which will be merged soon.

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -188,3 +188,17 @@ def extend_attr_to_tuple(
     if isinstance(val, list):
         val = tuple(val)
     return val
+
+
+def trt_cast_int_to_float(network: TRTNetwork, name: str, tensor: TRTTensor):
+    if tensor.dtype == trt.int8 or tensor.dtype == trt.int32:
```
gs-olive, line 194: I think this should only apply to `trt.int32`, since `trt.int8` is a special case (quantization) which seems to be supported by all elementwise ops except the boolean ones.

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
```diff
@@ -239,3 +243,245 @@ def _add_layer(
...
```
gs-olive, line 331: Consider using `cast_int_int_div_trt_tensor` here instead of these calls. Based on C++ and TorchScript code, it seems TRT will only complain if _both_ inputs are integers, and not if either individual input is an integer. Additionally, consider updating `cast_int_int_div_trt_tensor` to not check for `int8`, which is a special case, as mentioned above.

gs-olive (changes requested):

gs-olive:
Thanks for the changes - #2298 was just merged, so could you rebase to main

zewenli98:
Rebased! Thanks George!

gs-olive (approved):
Looks great to me! Well tested and very useful utilities + converters added. This will greatly improve our operator coverage and correctness - thanks @zewenli98!

gs-olive:
@zewenli98 - Could you rebase to the latest `main` for TorchScript testing

PR #2265, opened 2023-08-25
fix: Add special cases for `clone` and `to_copy` where input of graph is output

gs-olive:
# Description

- TRT does not allow inputs of graphs to be outputs as well, however many of the scenarios encountered in real models can have this situation come up, especially in cases where the input is cloned or copied and then returned
- The current converters will register these operators as a no-op, causing TRT engine building to fail on such inputs
- Instead of requiring creation of an identity layer for every case of a clone or copy node, we instead check if that node is the only operator on a placeholder (input) and then insert the identity layer or not, accordingly
- Coalesce implementations of clone and to_copy, which are effectively the same operator
- Add test cases to validate new behavior
- Add new boilerplate converter validator utility to support this case

Addresses bug in #1565

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1,10 +1,13 @@
 import logging
-from typing import Any, Dict, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Union
 
 import torch
 from torch.fx.node import Argument, Node, Target
 from torch_tensorrt.dynamo._SourceIR import SourceIR
 from torch_tensorrt.dynamo.conversion import impl
+from torch_tensorrt.dynamo.conversion.converter_utils import (
+    is_only_operator_on_placeholder,
+)
 from torch_tensorrt.fx.types import TRTNetwork, TRTTensor
 
 from .converter_registry import dynamo_tensorrt_converter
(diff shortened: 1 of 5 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -401,28 +432,41 @@ def aten_ops_to_copy_dtype(
...
+    # we need to force cast to ensure a layer is added to the TRT engine
+    # since TRT engine inputs cannot also be TRT engine outputs
+    return impl.cast.to_copy(
         network,
         target,
         SourceIR.ATEN,
         name,
         args[0],
+        kwargs.get("dtype", args[0].dtype),
+        force_cast=True,
```
gs-olive, line 465: Rename to `force_layer`

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -401,28 +432,41 @@ def aten_ops_to_copy_dtype(
...
     target: Target,
     args: Tuple[Argument, ...],
     kwargs: Dict[str, Argument],
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
-    return impl.cast.clone(
+    # For clone or copy nodes where the input is also the output,
+    # we need to force cast to ensure a layer is added to the TRT engine
+    # since TRT engine inputs cannot also be TRT engine outputs
+    return impl.cast.to_copy(
```
gs-olive, line 458: Add logic to `impl.cast` to perform differentiation on identity layers

narendasan:
@gs-olive is this only for graphs like: 
```
graph(x: Tensor):
  return x
```

Is it worth even converting these into engines?

gs-olive:
@narendasan - This change is primarily for cases encountered in detectron-style models where inputs are also outputs of the engine. It might be something like:
```python
graph(x: Tensor, y: Tensor):
  # A lot of logic/operations applied to y
  ...
  return x, y_new
```

py/torch_tensorrt/dynamo/conversion/impl/elementwise/base.py
```diff
@@ -116,9 +121,9 @@ def convert_binary_elementwise(
     # tensor. This is safe because broadcast will pad dimensions on the left
     # (prepend) to make lhs and rhs shape compatible.
     if network.has_implicit_batch_dimension:
-        if isinstance(lhs_val, torch.Tensor):
```
apbose, line 119: Can we remove the `network.has_implicit_batch_dimension` altogether here?

py/torch_tensorrt/dynamo/conversion/impl/elementwise/base.py
```diff
@@ -103,9 +104,13 @@ def convert_binary_elementwise(
     # dtype but we don't have a way to detect whether it makes sense for the
     # scalar to be float or half. Hence we go with the lhs dtype.
     if is_lhs_trt_tensor and isinstance(rhs_val, (float, int)):
-        rhs_val = torch.tensor([rhs_val], dtype=lhs_dtype)
```
apbose, line 106: Why is this getting converted to `np` array in place of the `torch.Tensor`. Is that because the `torch.Tensor` was leading to clones getting produced in the graph?

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -43,27 +43,49 @@ def get_node_name(node: torch.fx.Node) -> str:
...
+        node.op == "call_function"
+        and any(
+            arg.op == "placeholder"
+            for arg in node.args
+            if isinstance(arg, torch.fx.Node)
+        )
+        and any(user.op == "output" for user in list(node.users.keys()))
+    )
+
+
```
apbose, line 59: Why is there `any` used here? I believe that the above is for `to_copy` and` clone` cases having one node arg only? Or are there different cases for this?

py/torch_tensorrt/dynamo/conversion/impl/elementwise/base.py
gs-olive: Creating `torch.Tensor` objects within the `torch.compile` scope causes them to be "Fake-ified" which removes the data contained within them.

py/torch_tensorrt/dynamo/conversion/converter_utils.py
gs-olive: `any` is used here because we are checking if any of the users of the node are outputs, meaning that the node is the only function between a placeholder and an output. We are effectively search for subgraphs where an input is followed by a function is followed by an output.

py/torch_tensorrt/dynamo/conversion/impl/elementwise/base.py
gs-olive: Sure, I will make this change

apbose (approved):

PR #2271, opened 2023-08-28
feat: Implement Dynamic shapes + fallback support for export path

peri044:
# Description

Implement Dynamic shapes + fallback support for export path

TLDR: 
1) We need torch inputs at a few places in the codebase and to perform shape analysis. We also need `torch_tensorrt.Input` objects to hold the dynamic shape information accordingly for graph/subgraph inputs. `Input.torch_tensor` attribute is created to handle this and utility functions `get_torch_inputs` can be used. `prepare_inputs` can only return torchtrt_inputs now. Ideally, I think it would be convenient if `Input` objects are used through out the code and `get_torch_inputs` can be used whenever necessary.
2) I added support for kDouble enums. We support truncation and accept graphs which have torch.float64 as inputs and casts are added internally. But our Input object errors out if the dtype is torch.float64. 
3) Updated aten_tracer to use constraints to set bounds on Input dimensions. These bounds are dictated by TensorRT. The dynamic dimensions cannot be unbounded.

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/aten_tracer.py
```diff
@@ -20,13 +21,51 @@ def trace(
...
+            for dim in constraint_dims:
+                if min_shape[dim] > 1:
+                    constraints.append(min_shape[dim] <= dynamic_dim(trace_input, dim))
+                if max_shape[dim] > 1:
+                    constraints.append(dynamic_dim(trace_input, dim) <= max_shape[dim])
+            trace_inputs.append(trace_input)
+        else:
+            trace_inputs.append(torch_inputs[idx])
+
     experimental_decompositions = kwargs.get(
```
narendasan, line 60: Are you overriding the decomposition settings here? We should always respect the user setting

peri044: No. I'm not overriding. Based on the user setting, either get_decompositions(experimental_decompositions=True) or get_decompositions(False) (default case) is passed. Anything I'm missing ?

py/torch_tensorrt/dynamo/aten_tracer.py
```diff
@@ -20,13 +23,54 @@ def trace(
...
+
+            constraint_dims = []
+            new_shape = []
+            for dim in range(len(min_shape)):
+                if min_shape[dim] == opt_shape[dim] == max_shape[dim]:
+                    new_shape.append(torch_inputs[idx].shape[dim])
+                else:
+                    constraint_dims.append(dim)
+                    if torch_inputs[idx].shape[dim] == 1:
+                        new_shape.append(torch_inputs[idx].shape[dim] + 1)
```
gs-olive, line 49: Given that we are not using the `SymInt` approach here and a trace is made for each shape, why do we need to consider Torch's 0/1 specialization restriction in this case?

gs-olive (commented):
Added some questions and comments. Overall the updates look great for adding Dynamic shape support to export.

Additionally, I am not sure if this is an issue but currently we support Double inputs in Dynamo, but not in the C++ core. I think modifications are needed to the [autocast](https://github.com/pytorch/TensorRT/blob/6836773701afb445b9064c8a2e75bdc736cb8150/core/lowering/lowering.cpp#L28-L31) functionality to enable this.

py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -87,9 +87,10 @@ def _pretraced_backend(
 
             constant_fold(graph_module)
 
+            torchtrt_inputs = prepare_inputs(sample_inputs)
```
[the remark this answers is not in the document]
peri044: discussed in slack and no action for now.

py/torch_tensorrt/dynamo/aten_tracer.py
peri044: Not sure, I understood the question completely but here's what's going on.
1) torch.export(model, inputs) returns a model which only works for given input shape (model(inputs) will fail if there's a different input shape eg: different batch too). So, we use contraints API provided by Torch which inserts `sym_int` ops in the graph which operate on the dynamic dimension. 
2) Torch's 0/1 specialization is due to their implementation on dynamo. In the case of dynamic shapes, we can't have the tracing inputs to have `1` in the dynamic dimension and this is a restriction (and not a bug) as per Angela's comments from Meta.
Does that answer your question ?

gs-olive: This does answer my question, yes - thank you for the detailed response. How do we "correct" for this off-by-one issue then, if the user specified a minimum size of 1?

peri044: I detect the dynamic dimension to be 1 and in this case, I just add 1 and create a new tensor with it. https://github.com/pytorch/TensorRT/blob/16d19e80e89a9538d2f404bc39b3a88b1717da1b/py/torch_tensorrt/dynamo/aten_tracer.py#L49

py/torch_tensorrt/dynamo/aten_tracer.py
gs-olive: While not required, this could be switched to:
```python
from torch_tensorrt.dynamo._defaults import ENABLE_EXPERIMENTAL_DECOMPOSITIONS
...
experimental_decompositions = kwargs.get("enable_experimental_decompositions", ENABLE_EXPERIMENTAL_DECOMPOSITIONS)
```
The above would reflect changes in the default automatically

py/torch_tensorrt/_Input.py
```diff
@@ -171,6 +172,14 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
 
         self.tensor_domain = Input._parse_tensor_domain(domain)
 
+        if "torch_tensor" in kwargs:
+            self.torch_tensor = kwargs["torch_tensor"]
+        else:
+            if self.shape_mode == Input._ShapeMode.DYNAMIC:
+                self.torch_tensor = self.example_tensor("opt_shape")
+            else:
+                self.torch_tensor = self.example_tensor()
```
gs-olive, line 181: What if `"torch_tensor"` is provided, but the shape is also Dynamic - would it be an issue if this Tensor had the shape of `min` or `max` instead of `opt`? This could be refactored to override or validate the specified tensor if the shape is dynamic.

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -63,6 +63,35 @@ def cosine_similarity(gt_tensor: torch.Tensor, pred_tensor: torch.Tensor) -> flo
     return res
 
 
+def input_is_dynamic(inputs: Sequence[Input]) -> bool:
```
gs-olive, line 66: Should the schema be switched to `Sequence[Union[Input, torch.Tensor]]`, given the below check?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -63,6 +63,35 @@ def cosine_similarity(gt_tensor: torch.Tensor, pred_tensor: torch.Tensor) -> flo
...
+    """
+    Return true if the provided inputs are dynamic.
+    """
+    if any(isinstance(input, torch.Tensor) for input in inputs):
+        return False
+    return any(input.shape_mode == Input._ShapeMode.DYNAMIC for input in inputs)
```
gs-olive, line 72: Consider shortening to `return not any(isinstance(input, torch.Tensor) for input in inputs) and any(input.shape_mode == Input._ShapeMode.DYNAMIC for input in inputs)`

py/torch_tensorrt/_Input.py
peri044: >  would it be an issue if this Tensor had the shape of min or max instead of opt?
So the usecase you are saying is if the user provides 
`x = torch_tensorrt.Input(min_shape=<>, opt_shape=<>, max_shape=<>, torch_tensor=<random_tensor>)`
This wouldn't be an issue, as we ignore torch_tensor in this case of dynamic shape compilation. https://github.com/pytorch/TensorRT/blob/695dc9b5f4a2a5f22b40416c7c83ed0eddc6c39a/py/torch_tensorrt/dynamo/partitioning/common.py#L48 would generate input tensors for all the modes.
We could maybe pass a warning to users that torch_tensor is not being used in such cases.

py/torch_tensorrt/dynamo/utils.py
peri044: Yes, done.

py/torch_tensorrt/dynamo/utils.py
peri044: done

py/torch_tensorrt/_Input.py
gs-olive: I see - I believe the feature ask from #2323 is to allow providing tensors for each of `min`, `opt`, and `max`. Could `torch_tensor` be instead either a list/tuple of 3 tensors (`min`, `opt`, `max`) _or_ a single tensor?

gs-olive (commented):
Added a small comment on documentation. Otherwise looks good to me!

docsrc/user_guide/dynamic_shapes.rst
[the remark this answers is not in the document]
peri044: done. thanks

PR #2273, opened 2023-08-28
Enabling var_mean decomposition

apbose:
This PR enables the var_mean and rsub decomposition in `torch_enabled_decompositions`


gs-olive (commented):
Change looks good! Consider adding `aten.rsub`, `aten.rsqrt`, `aten.sqrt`, or any other decompositions which might be preferable/have recent PRs.

For instance, the following can just be replaced with `aten.rsub`:
https://github.com/pytorch/TensorRT/blob/1ad2aeba6e7255abe65f1ba1ffc856ea20cc6a63/py/torch_tensorrt/dynamo/lowering/_decomposition_groups.py#L128-L129

py/torch_tensorrt/dynamo/lowering/_decomposition_groups.py
```diff
@@ -175,6 +175,7 @@
     aten.linalg_vector_norm,
     aten.full,
     aten.repeat,
+    aten.var_mean,
```
apbose, line 178: @gs-olive you mentioned the `rsub`. I will add that.
Regarding `sqrt` and `rsqrt`, sqrt is not of lowering type so I do not think it should be included here.
For `rsqrt`, should I include it here? Since `rsqrt` is already present in ``py/torch_tensorrt/dynamo/lowering/_decompositions.py`` and that would take precedence over the enabled ones, right? And we would not need to add it to the disabled ops then.

gs-olive: That is true, I agree with your points. I think the `rsub` would be the main one to change; we can leave the others out, and the `rsqrt` implementation in `_decompositions.py` will take precedence, yes.

gs-olive (approved):
Looks good to me!

PR #2253, opened 2023-08-21
feat: support linear (fully connected layer) dynamo converter

zewenli98:
# Description

Support linear (fully connected layer) dynamo converter.

Fixes #2251 

## Type of change

- New feature (non-breaking change which adds functionality)

# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests to verify my fix or my feature
- [x] New and existing unit tests pass locally with my changes
- [x] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #2279, opened 2023-08-30
fix: add an arg in matmul

### LINKED ISSUE #2203: Expose IMatrixMultiplyLayer in dynamo.conversion.impl

zewenli98:
# Description

`set_layer_name(layer, target, name)` -> `set_layer_name(layer, target, name, source_ir)`

Fixes #2203 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -171,6 +171,7 @@ def aten_ops_gelu(
 
 @dynamo_tensorrt_converter(torch.ops.aten.matmul)  # type: ignore[misc]
 @dynamo_tensorrt_converter(torch.ops.aten.mm.default)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.mv.default)  # type: ignore[misc]
 def aten_ops_matmul(
     network: TRTNetwork,
     target: Target,
@@ -179,7 +180,12 @@ def aten_ops_matmul(
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return impl.matmul.matrix_multiply(
-        network, target, SourceIR.ATEN, name, args[0], args[1]
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
     )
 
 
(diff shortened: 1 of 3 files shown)


gs-olive (commented):
Looks good - could you also add two arguments to `matrix_multiply`, which both default to `trt.MatrixOperation.NONE`, which allow us to control whether the matrix being multiplied gets transposed as well, so we can reduce operations required overall? They would control the variables `input_matrix_op` and `other_matrix_op`

zewenli98:
Done! Thanks for the suggestion!

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -179,7 +180,14 @@ def aten_ops_matmul(
...
     return impl.matmul.matrix_multiply(
-        network, target, SourceIR.ATEN, name, args[0], args[1]
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+        args_bounds_check(args, 2, trt.MatrixOperation.NONE),
+        args_bounds_check(args, 3, trt.MatrixOperation.NONE),
```
gs-olive, line 190: You can remove these lines, since none of the converted Torch ops will have a third or fourth argument, based on their schemas.

The primary reason for adding that argument to `matrix_multiply` below is that later, for `aten.linear`, you can pass in `other_matrix_op=trt.MatrixOperation.TRANSPOSE` for instance, and then reduce the number of layers (no need for a permute)

zewenli98: It makes sense! I've updated this PR and `aten.linear` in #2253.

gs-olive (approved):
Looks good to me!

gs-olive:
@zewenli98 - could you re-enable these tests for this PR, since it adds a converter for `mv`? It should just require uncommenting here:
https://github.com/pytorch/TensorRT/blob/e49ef6d1de58f80bf51e37428d4a2287ed462871/tests/py/dynamo/conversion/test_matmul_aten.py#L14-L15
As well as here:
https://github.com/pytorch/TensorRT/blob/e49ef6d1de58f80bf51e37428d4a2287ed462871/tests/py/dynamo/conversion/test_matmul_aten.py#L44-L45

zewenli98:
Done! lmk if more changes are needed! Thanks!

gs-olive (approved):
Looks good to me!

PR #2292, opened 2023-09-06
tools(opset_coverage): Map default ops to unoverloaded ops

narendasan:
# Description

Updates the coverage tool to treat converters for default overloads to the root op name

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/tools/opset_coverage.py
+++ b/py/torch_tensorrt/dynamo/tools/opset_coverage.py
@@ -122,14 +122,20 @@ def opset_coverage(
         converter_registry if converter_registry is not None else DYNAMO_CONVERTERS
     )
     converter_registry_targets = {
-        c_registry.qualified_name_or_str(target).removeprefix("torch.ops.")
+        c_registry.qualified_name_or_str(target)
+        .removeprefix("torch.ops.")
+        .replace(".default", "")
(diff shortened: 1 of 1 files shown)


gs-olive (approved):
In line with the syntax used in [native_functions.yaml](https://github.com/pytorch/pytorch/blob/cd6a332bc500b70e92bc05af06c65fcbcbf4f123/aten/src/ATen/native/native_functions.yaml), this modification gives more clarity to the operator coverage set. A suggestion for a future improvement might be to provide coverage percentages for _just_ the default operators, since this could give a more accurate view of what's covered.

narendasan:
> In line with the syntax used in [native_functions.yaml](https://github.com/pytorch/pytorch/blob/cd6a332bc500b70e92bc05af06c65fcbcbf4f123/aten/src/ATen/native/native_functions.yaml), this modification gives more clarity to the operator coverage set. A suggestion for a future improvement might be to provide coverage percentages for _just_ the default operators, since this could give a more accurate view of what's covered.

Why do we need to distinguish default operators? Because certain overloads are also part of core_aten so I'd think we would need to treat them equally?

gs-olive:
The only reason was to indicate partial support for an operator, even if we don't support every case of it. I assumed we would almost always have the `default` operator for each unique operation.

PR #2293, opened 2023-09-06
Update _Input.py

phyboy:
# Description

NHCW should be NCHW

## Type of change


gs-olive (approved):
Looks good to me!

PR #2307, opened 2023-09-11
neg converter correction

apbose:
The` cast_trt_tensor` has been moved to the `torch_tensorrt/dynamo/conversion/impl/unary/ops.py`. Removing this from `aten_ops_converters`


gs-olive (approved):
Looks good to me! Just pending CI verification.

PR #2310, opened 2023-09-11
fix: Move aten.neg test case

gs-olive:
# Description

## Type of change

# Checklist:


tests/py/dynamo/conversion/test_neg_aten.py
```diff
@@ -43,8 +44,8 @@ def forward(self, input):
         self.run_test(
             neg(),
             inputs,
-            output_dtypes=[torch.int32],
```
gs-olive, line 46: The output dtype will not agree with the casted output dtype unless another cast is inserted to "undo" the initial cast. Here, for example, if the input is `torch.int32`, the output will be `torch.float`.

apbose (approved):

PR #2315, opened 2023-09-12
small fix: Packaging version switch

gs-olive:
- Versioning does not recognize `2.1.0+cu118` as equal to `2.1.0`. Switch implementation to use less-than comparison


peri044 (approved):
LGTM

PR #2324, opened 2023-09-18
feat: Add tensor type enforcement for converters

### LINKED ISSUE #2306: ✨[Feature] Add `GetITensor` constant management system to Dynamo
# Context
In the TorchScript path of the codebase, tensors are accessed via the `ITensorOrFreeze` function, as below:
https://github.com/pytorch/TensorRT/blob/ea5a28963a6f931aa0dc433e37b3a0809a515eec/core/conversion/var/Var.cpp#L89
This function serves to freeze any input constants before returning them. Such a utility would be very useful in Dynamo, to handle cases where converters cannot process Numpy array or Torch Tensor inputs (like `aten.cat`, for example).

Converters could use this function when a certain input is required to be an ITensor, and omit it when this is not required.

gs-olive:
# Description

- Add utilities to enforce tensor types for aten converters via direct reference
- Add robust testing for utilities
- Update decorator schemas throughout code

Fixes #2306

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -6,6 +6,7 @@ import torch
 from torch.fx.node import Argument, Node, Target
 from torch_tensorrt.dynamo._SourceIR import SourceIR
 from torch_tensorrt.dynamo.conversion import impl
+from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
 from torch_tensorrt.dynamo.conversion.converter_registry import (
     dynamo_tensorrt_converter,
 )
@@ -13,7 +14,7 @@ from torch_tensorrt.dynamo.conversion.converter_utils import (
     enforce_tensor_types,
     is_only_operator_on_placeholder,
 )
-from torch_tensorrt.fx.types import TRTNetwork, TRTTensor
+from torch_tensorrt.fx.types import TRTTensor
 
 from .converter_utils import dynamic_unsupported_with_args
 
@@ -28,14 +29,14 @@ def args_bounds_check(
 
 @dynamo_tensorrt_converter(torch.ops.aten.batch_norm)
 def aten_ops_batch_norm(
-    network: TRTNetwork,
+    ctx: ConversionContext,
     target: Target,
     args: Tuple[Argument, ...],
     kwargs: Dict[str, Argument],
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return impl.normalization.batch_norm(
-        network,
+        ctx,
         target,
         SourceIR.ATEN,
         name,
@@ -71,14 +72,14 @@ def embedding_param_validator(embedding_node: Node) -> bool:
     torch.ops.aten.embedding.default, capability_validator=embedding_param_validator
 )
 def aten_ops_embedding(
-    network: TRTNetwork,
+    ctx: ConversionContext,
     target: Target,
     args: Tuple[Argument, ...],
     kwargs: Dict[str, Argument],
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return impl.embedding.embedding(
-        network,
+        ctx,
         target,
         SourceIR.ATEN,
         name,
@@ -93,25 +94,25 @@ def aten_ops_embedding(
 @dynamo_tensorrt_converter(torch.ops.aten.fmod.Scalar)
 @dynamo_tensorrt_converter(torch.ops.aten.fmod.Tensor)
 def aten_ops_fmod(
-    network: TRTNetwork,
+    ctx: ConversionContext,
     target: Target,
     args: Tuple[Argument, ...],
     kwargs: Dict[str, Argument],
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
-    return impl.elementwise.fmod(network, target, SourceIR.ATEN, name, args[0], args[1])
+    return impl.elementwise.fmod(ctx, target, SourceIR.ATEN, name, args[0], args[1])
 
 
 @dynamo_tensorrt_converter(torch.ops.aten.relu.default)
 def aten_ops_relu(
-    network: TRTNetwork,
+    ctx: ConversionContext,
     target: Target,
     args: Tuple[Argument, ...],
     kwargs: Dict[str, Argument],
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return impl.activation.relu(
-        network,
+        ctx,
         target,
         SourceIR.ATEN,
         name,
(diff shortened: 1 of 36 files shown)


py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -269,3 +271,95 @@ def get_trt_tensor(
         return input_val
     else:
         raise AssertionError(f"Cannot convert {input_val} to TRT constant")
+
+
+def enforce_tensor_types(
+    type_dictionary: Dict[Union[int, str], Tuple[Union[TRTTensor, np.ndarray], ...]],
+    promote: bool = True,
+) -> Callable[[ConverterImplSignature], ConverterImplSignature]:
```
gs-olive, line 279: Investigate type enforcement for tensors.

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -250,22 +252,113 @@ def get_trt_tensor(
...
+    if isinstance(input_val, torch.Tensor):
+        if input_val.dtype == torch.int64:
+            input_val = input_val.to(torch.int32)
+        elif input_val.dtype == torch.float64:
+            input_val = input_val.to(torch.float32)
+    elif isinstance(input_val, np.ndarray):
+        if input_val.dtype == np.int64:
+            input_val = input_val.astype(np.int32)
+        elif input_val.dtype == np.float64:
+            input_val = input_val.astype(np.float32)
```
gs-olive, line 265: Adds automatic type-casting for `get_attr` operators (cast to 32-bit). This function currently does not check the `truncate_long_and_double` argument since it applies to weights and additionally cannot easily access compiler-context variables without many code changes.

Additionally, this PR _removes_ autocasting of boolean values to integers for constant freezing, since TRT documentation ([link 1](https://docs.nvidia.com/deeplearning/tensorrt/operators/docs/Constant.html), [link 2](https://docs.nvidia.com/deeplearning/tensorrt/operators/index.html#layers-precision-matrix)) indicates that boolean values are supported in this layer.

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -250,22 +252,113 @@ def get_trt_tensor(
...
-    ):
-        input_val = input_val.astype(np.int32)
-
-    if isinstance(input_val, (torch.Tensor, np.ndarray, int, float)):
+    # If the input is 64-bit, cast it to 32-bit for TRT freezing
+    if isinstance(input_val, torch.Tensor):
+        if input_val.dtype == torch.int64:
+            input_val = input_val.to(torch.int32)
+        elif input_val.dtype == torch.float64:
+            input_val = input_val.to(torch.float32)
```
gs-olive, line 260: Augment `TRTNetwork` to store user settings

- Conversion Context class --> TRTNetwork, CompileSettings
- INetwork + Context Class, Network inputs // CTX inputs
- `ctx.net` replace network calls in converters

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -295,22 +298,119 @@ def get_trt_tensor(
...
-        return create_constant(network, input_val, name, dtype)
+    if isinstance(input_val, (torch.Tensor, np.ndarray, int, float, bool)):
+        return create_constant(ctx, input_val, name, dtype)
     elif isinstance(input_val, TRTTensor):
         return input_val
     else:
         raise AssertionError(f"Cannot convert {input_val} to TRT constant")
+
+
+def enforce_tensor_types(
```
narendasan, line 327: Where did we land on using this to enforce scalar types as well?

gs-olive: I did not add it, since I wasn't fully clear on the specification. For instance, should a user be able to specify `3: (int, float)`? Additionally, if they specify `3: (int, np,ndarray)`, would it be the `enforce_tensor_types` responsibility to cast that numpy array to integer if it was a float type? The latter seems out-of-scope for this function.

gs-olive: Alternatively, if the specification was `3: (int, float)`, and we see an `np.ndarray`, should we attempt a cast (`int(array)`)? Or should it throw an error?

narendasan: If promote is true then enforce_types should try to cast it, for 

> Alternatively, if the specification was 3: (int, float), and we see an np.ndarray, should we attempt a cast (int(array))? Or should it throw an error?

Here we just run `.item()` and if it fails we thrown an error. 

> I did not add it, since I wasn't fully clear on the specification. For instance, should a user be able to specify 3: (int, float)? Additionally, if they specify 3: (int, np.ndarray), would it be the enforce_tensor_types responsibility to cast that numpy array to integer if it was a float type? The latter seems out-of-scope for this function

Here yes, you should be able to say, I want either a float32 or an int32. If the converter is provided a type that is in the set no conversion should happen, otherwise if promote_types is true then it should try to convert to the most preferred (first) type I think.

We could probably do this in another PR to get converter folks this feature sooner, but we should probably rename this function `enforce_types`

gs-olive: Added #2349 with the proposed new features. This may need some additional discussion because there is a key distinction between Schema-equivalent-types, Tensor-equivalent-types, and data types. It seems that all three definitions are coming together here.

The intent of this function was to translate between Tensor-equivalent types, without consideration for the data types within those tensors. For instance, a converter could specify that it cannot handle `np.ndarray`, but it can handle `TRTTensor`, with no regard for the actual data types within. One step further would be to allow Schema-equivalent-types. That is - translate between Tensor types as needed, and Scalar types as needed, but no modification of data types within the Tensors. Finally is data types, which would probably require a new function header/schema altogether, since it requires that we also check the data type within the tensor and cast appropriately.

For now, I can leave this PR as-is and work on the other feature separately, to unblock any converter implementations that might depend on it. What do you think?

narendasan: Yeah lets merge this and work towards a schema type level enforcement mechanism

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1354,9 +1364,16 @@ def conv_param_validator(conv_node: Node) -> bool:
 
 @dynamo_tensorrt_converter(
     torch.ops.aten.convolution.default, capability_validator=conv_param_validator
-)  # type: ignore[misc]
+)
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
+        1: (TRTTensor, np.ndarray, torch.Tensor),
+        2: (TRTTensor, np.ndarray, torch.Tensor),
```
gs-olive, line 1372: Consider tuple ordering as priority for the converter

gs-olive: `(np.ndarray, TRTTensor)` implies np array preferred

PR #2330, opened 2023-09-20
feat: support group_norm, batch_norm, and layer_norm

### LINKED ISSUE #2225: Expose INormalizationLayer in dynamo.conversion.impl

zewenli98:
# Description

Update `batch_norm` and `layer_norm`

Fixes #2225 

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -31,14 +31,37 @@ def aten_ops_batch_norm(
...
         name,
-        args[0],
-        args[1],
-        args[2],
-        args[3],
-        args[4],
-        args[5],
-        args[6],
-        args[7],
+        input=args[0],
```
gs-olive, line 34: Consider adding support/decorator for [`native_batch_norm`](https://github.com/pytorch/pytorch/blob/1df14f1bf8661f14e3ebaaf441844ab0cd8534b3/aten/src/ATen/native/native_functions.yaml#L4162) as well.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -31,14 +31,37 @@ def aten_ops_batch_norm(
...
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.normalization.layer_norm(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        input=args[0],
+        normalized_shape=args[1],
+        weight=args_bounds_check(args, 2, replacement=1),
+        bias=args_bounds_check(args, 3, replacement=0),
```
gs-olive, line 62: Weight and bias default to None, so they should do the same here

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -36,6 +36,7 @@ def batch_norm(
     training: torch.Tensor,
     momentum: torch.Tensor,
     eps: List[float],
+    cudnn_enabled: bool,
```
gs-olive, line 39: Since `None` is a valid input to the layer and batch norm weight and biases, could you add support for those inputs being optional/None and update the types in the function header here? Specifically, it could just be something like:

```python
if weight is None:
    weight = 1.0

if bias is None:
    bias = 0.0
```

gs-olive (commented):
Updates look great - added some suggestions to better follow the Torch schemas for these functions

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -31,14 +31,37 @@ def aten_ops_batch_norm(
...
-        args[7],
+        input=args[0],
+        weight=args_bounds_check(args, 1, replacement=1),
+        bias=args_bounds_check(args, 2, replacement=0),
+        running_mean=args_bounds_check(args, 3),
+        running_var=args_bounds_check(args, 4),
+        training=args_bounds_check(args, 5),
+        momentum=args_bounds_check(args, 6, replacement=0.1),
+        eps=args_bounds_check(args, 7, replacement=1e-05),
+        cudnn_enabled=args_bounds_check(args, 8, replacement=False),
```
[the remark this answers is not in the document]
zewenli98: Thanks! For some args with type `Tensor?`. Does this mean the arg is optional?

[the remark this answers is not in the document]
gs-olive: My understanding is that this means the arg can be a `Tensor` or `None`, as in Python typing: `Optional[Tensor]`. I am not sure whether it means that the argument can be missing entirely, but I do not think it allows for the argument to be missing entirely.

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -47,8 +47,20 @@ def batch_norm(
...
+        weight = np.array(1.0)
+
+    if bias is None:
+        bias = np.array(0.0)
+
+    if running_mean is None:
+        running_mean = np.array(0.0)
+
+    if running_var is None:
+        running_var = np.array(1.0)
```
gs-olive, line 60: For these, it should be okay to not cast to `np.array` in the converter (instead leave them as ints or floats), since `to_numpy` should dictate this casting behavior for ints and floats. Specifically, one small difference is that I think `np.array(1.0)` has shape `()` (0D), but `to_numpy` generally adds a dimension, to make it 1D.

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -102,6 +114,12 @@ def layer_norm(
             "of the TensorRT region!"
         )
 
+    if weight is None:
+        weight = np.array(1.0)
+
+    if bias is None:
+        bias = np.array(0.0)
```
gs-olive, line 121: See above comment

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -152,16 +170,22 @@ def layer_norm_no_plugin(
...
         raise RuntimeError(
             f"LayerNorm received input {input} that is not part "
             "of the TensorRT region!"
         )
 
+    if weight is None:
+        weight = np.array(1.0)
+
+    if bias is None:
+        bias = np.array(0.0)
```
gs-olive, line 187: See above comment

zewenli98: Since line 189 is `shape = weight.shape` and lines 191 and 192 call `weight.reshape` and `bias.reshape`, I think `weight` and `bias` shouldn't be scalars.

gs-olive: I see - in that case, it might be preferable to use `to_numpy(0.0)`, for instance, to get back a default-formatted numpy array for the float default. Additionally, I noticed the code below has some issues:

```python
gamma = to_numpy(weight.reshape(*shape))
##### Above is invalid, since the reshape should apply to the numpy output. It should instead be:
gamma = to_numpy(weight).reshape(shape)
```
The same as the above applies for `beta`.

Additionally, lines 194 - 196 should be using `get_axes_for_reduce_op`, as here:
https://github.com/pytorch/TensorRT/blob/ecdc040aec3902cbc396648043f54c96e6a91ee7/py/torch_tensorrt/dynamo/conversion/converter_utils.py#L209

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -152,16 +170,22 @@ def layer_norm_no_plugin(
     name: str,
     input: TRTTensor,
     normalized_shape: List[int],
-    weight: torch.Tensor,
-    bias: torch.Tensor,
-    eps: List[float],
+    weight: Optional[Union[TRTTensor, torch.Tensor, np.ndarray]],
+    bias: Optional[Union[TRTTensor, torch.Tensor, np.ndarray]],
```
gs-olive, line 174: `TRTTensor` would not be a valid input here, for the scale layer

gs-olive:
As discussed, add `group_norm` implementation here. Additionally, for any converters added, remove those converters from `torch_tensorrt.dynamo.lowering._decomposition_groups`

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
zewenli98: Do you mean the type of `weight` and `bias` in all the three functions should be `Optional[Union[torch.Tensor, np.ndarray]]`?  I see its native function: 
```
func: layer_norm(Tensor input, SymInt[] normalized_shape, Tensor? weight=None, Tensor? bias=None, float eps=1e-05, bool cudnn_enable=True) -> Tensor
```

gs-olive: Yes, I think it should be `ptional[Union[torch.Tensor, np.ndarray]]`, because if either of those is a `TRTTensor`, the computation below would not work (`to_numpy` can't be called on a `TRTTensor`)

zewenli98:
@gs-olive group_norm was added!

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -6,13 +6,12 @@
...
-from torch_tensorrt.dynamo.conversion.impl.elementwise.base import (
-    convert_binary_elementwise,
-)
-from torch_tensorrt.dynamo.conversion.impl.unary.base import convert_unary
+from torch_tensorrt.dynamo.conversion import impl
+from torch_tensorrt.dynamo.conversion.converter_utils import get_axes_for_reduce_op
 from torch_tensorrt.fx.converters.converter_utils import (
     get_positive_dim,
     get_trt_plugin,
+    get_trt_tensor,
```
gs-olive, line 14: Switch to using the `dynamo.conversion.converter_utils` implementation of this

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -46,8 +46,20 @@ def batch_norm(
...
+
+    if running_mean is None:
+        running_mean = 0.0
+
+    if running_var is None:
+        running_var = 1.0
+
     scale = cast(torch.Tensor, to_numpy(weight)) / np.sqrt(
-        cast(torch.Tensor, to_numpy(running_var)) + cast(float, eps)
+        cast(torch.Tensor, to_numpy(running_var)) + eps
```
gs-olive, line 62: The `torch.Tensor` cast can be removed, because `to_numpy` will return an `np.ndarray`, so this typing would be incorrect.

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -244,32 +267,113 @@ def layer_norm_no_plugin(
...
+        _LOGGER.error(
+            "Unable to find group norm plugin, fall back to TensorRT implementation."
+        )
+
+    layer = network.add_plugin_v2([input, scale, bias], plugin)
+    set_layer_name(layer, target, f"{name}_GroupNormalizationPlugin", source_ir)
+
+    # PyTorch requires three return values: (out, mean, rstd)
+    dummy_tensor = torch.tensor(0)
+    return layer.get_output(0), dummy_tensor, dummy_tensor
```
gs-olive, line 374: Is it possible to avoid invoking the plugin here, and instead use the full implementation, adapting from here: https://github.com/NVIDIA-AI-IOT/torch2trt/blob/36656b614f3fbc067ac673932e2200d7afdae712/torch2trt/converters/group_norm.py#L7-L73? The plugin is not preferable for use in new converters unless it cannot be otherwise supported.

gs-olive, line 374: The returned values here should be correct intermediate tensors from during the computation unless we explicitly remove support for nodes which need the other two values

gs-olive (changes requested):
Added a few comments. Additionally, if the dynamic shape version of this converter is not passing, that is okay since it is not required for the first pass of support

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
gs-olive: Alternatively, the TRT layer-based implementation can be the backup for the plugin, etc.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -31,14 +32,84 @@ def aten_ops_batch_norm(
...
+        running_mean=args[3],
+        running_var=args[4],
+        training=args[5],
+        momentum=args[6],
+        eps=args[7],
+        cudnn_enabled=args_bounds_check(args, 8, replacement=True),
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.native_layer_norm.default)  # type: ignore[misc]
```
gs-olive, line 47: Based on the [schema](https://github.com/pytorch/pytorch/blob/d1a13129bbe237853ae8859c2664d551b450a47b/aten/src/ATen/native/native_functions.yaml#L3177) of `native_layer_norm`, it looks like it requires 3 outputs much like `native_group_norm`. As a comment on both of those - if you want to support it with essentially the same converter as the regular layer norm, you can do the following:

Add this validator
```python
def validator(layer_norm: Node) -> bool:
    # Validate only one user, which is a getitem node that accesses the first element in the list
    return (len(layer_norm.users) == 1 and
            list(node.users)[0].target == operator.getitem and
            list(node.users)[0].args[1] == 0))
```

Add this converter
```python
@dynamo_tensorrt_converter(torch.ops.aten.native_layer_norm.default, capability_validator=validator)
def converter(...):
    return (regular_layer_norm, )
```

It is important that the above converter returns a tuple, because it will be accessed by `getitem`, but as you have validated, it will only access the first element. This should also work for group norm.

zewenli98:
This PR depends on #2354 and #2355

gs-olive:
@zewenli98 - when you have the chance, please rebase this PR to the latest `main`. Additionally, to follow up on the discussion from [this comment](https://github.com/pytorch/TensorRT/pull/2330/files/a4634f75ba9f472f7bc962a9bb7ca0542c8fcbab#r1340763891), the individual functions for `layer_norm`, `batch_norm`, etc. should likely return their intermediate values too, so that we can convert those `native_layer_norm`-style functions.

zewenli98:
Yes! It's still in progress. Thanks for the reminder!

gs-olive (commented):
Looks good to me - will update again pending a manual check against SD

gs-olive (approved):
Works on SD - looks good to me!

PR #2333, opened 2023-09-21
Torch version check change for imports

apbose:
The snippet throws error for torch= 2.1.0.dev20230905+cu121
This PR changes the python version check for the inductor import.

### DIFF
--- a/py/torch_tensorrt/dynamo/backend/backends.py
+++ b/py/torch_tensorrt/dynamo/backend/backends.py
@@ -20,7 +20,7 @@ from torch_tensorrt.dynamo.utils import parse_dynamo_kwargs, set_log_level
 from packaging import version
 
 # Modify import location of utilities based on Torch version
-if version.parse(sanitized_torch_version()) < version.parse("2.1.1"):
+if version.parse(sanitized_torch_version()) <= version.parse("2.1.0.dev20230902"):
     from torch._inductor.freezing import ConstantFolder, replace_node_with_constant
 else:
     from torch._inductor.constant_folding import (


apbose:
I was curious that this error maybe comes up because I am on a dev version of 2.1.0.*. Would this be the correct way of handling this?

gs-olive (changes requested):
I am not sure whether this change will achieve the desired effect, since the following two boolean expressions are both true:
```python
version.parse("2.1.0.dev20230905+cu121") < version.parse("2.1.1")  # True
version.parse("2.1.0.dev20230905+cu121") <= version.parse("2.1.0")  # True
```
Maybe this could be changed to something like:
```python
if version.parse(sanitized_torch_version()) < version.parse("2.1.0.dev20230902")
```
It looks like the change which switched the file location of these functions is https://github.com/pytorch/pytorch/pull/108450, on September 1st, so the September 2nd nightly would be the first instance where the change shows up.

apbose:
Ok yes that would work. I earlier thought the order is of the type 2.1.0 < 2.1.0.dev*<2.1.1, but looks like that is not the case

py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -20,7 +20,7 @@
 from packaging import version
 
 # Modify import location of utilities based on Torch version
-if version.parse(sanitized_torch_version()) < version.parse("2.1.1"):
+if version.parse(sanitized_torch_version()) <= version.parse("2.1.0.dev20230902"):
```
gs-olive, line 23: I think it should be `<` and not `<=`, since the September 2 nightly should not enter this code block.

narendasan: I think 2.1.0 will pass this check however since all dev builds are considered < any stable build iirc

gs-olive: Add `or` to conditional

gs-olive (changes requested):
@narendasan, @apbose - After looking into this, it seems that the commit which switched the location of these utilities did not make it into [`release/2.1`](https://github.com/pytorch/pytorch/blob/release/2.1/torch/_inductor/freezing.py) (PyTorch), despite the fact that it was added on [September 1st](https://github.com/pytorch/pytorch/commits/main/torch/_inductor/constant_folding.py), in one of the `0901` dev nightlies.

I think this does need to be an `or` condition, perhaps something like:

```python
if version.parse(sanitized_torch_version()) < version.parse("2.1.0.dev20230902") or version.parse(sanitized_torch_version()) == version.parse("2.1.0"):
```

PR #2337, opened 2023-09-22
feat: support deconv (1d, 2d, and Nd) dynamo converter

### LINKED ISSUE #2207: Expose IDeconvolutionLayer in dynamo.conversion.impl

zewenli98:
# Description

Support deconv (1d, 2d, and Nd) dynamo converter.

Fixes #2207 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -357,14 +357,14 @@ def aten_ops_softmax(
 
 @dynamo_tensorrt_converter(
     torch.ops.aten.split.Tensor, capability_validator=dynamic_unsupported_with_args([1])
-)
+)  # type: ignore[misc]
 @dynamo_tensorrt_converter(
     torch.ops.aten.split.sizes, capability_validator=dynamic_unsupported_with_args([1])
-)
+)  # type: ignore[misc]
 @dynamo_tensorrt_converter(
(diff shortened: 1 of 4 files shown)


gs-olive (commented):
This looks great! Could you also add decorators for `conv_transposeNd`, [as here](https://github.com/pytorch/pytorch/blob/d0c8e8240dc04ca85dd071d99be49454a9138de6/aten/src/ATen/native/native_functions.yaml#L1679-L1689), since these seem to be supported with the new implementation.

zewenli98:
Sure! Actually I've implemented `conv_transposeNd` but when I did test I found they were not used while converting. For testing, do you know which PyTorch ops corresponds to these converters?

gs-olive:
I would guess the `ConvTransposeND` like [this operator](https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html), but these also may just end up as `convolution` with the boolean `transpose` specified. If it is not worthwhile with the argument schemas to add `conv_transposeNd`, then we can add a lowering pass for these

zewenli98:
> these also may just end up as `convolution` with the boolean `transpose` specified. 

Yes, I found it ends up as `convolution` with the boolean `transpose` specified. That's why I didn't add `conv_transposeNd`. If adding `conv_transposeNd`, since the args are different, we need to add a new function instead of decorators. Or, just like you said, do it in lowering phase.

gs-olive (approved):
Looks good to me!

PR #2343, opened 2023-09-25
Dynamo converter cat

apbose:
This PR moves the aten::cat implementation to the dynamo impl library.

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -65,7 +65,7 @@ def aten_ops_cat(
         SourceIR.ATEN,
         name,
         tensors=args[0],
-        dim = args_bounds_check(args, 2, 1),
+        dim=args_bounds_check(args, 2, 1),
     )
 
 
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -53,19 +53,19 @@ def aten_ops_batch_norm(
 
 @dynamo_tensorrt_converter(torch.ops.aten.cat.default)
 def aten_ops_cat(
-    network: TRTNetwork,
+    ctx: ConversionContext,
     target: Target,
     args: Tuple[Argument, ...],
     kwargs: Dict[str, Argument],
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return impl.cat.cat(
-        network,
+        ctx,
         target,
         SourceIR.ATEN,
         name,
-        tensors=args[0],
-        dim=args_bounds_check(args, 2, 1),
+        input=args[0],
+        dim=args_bounds_check(args, 1, 0),
     )
 
 
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -51,6 +51,24 @@ def aten_ops_batch_norm(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.cat.default)
+def aten_ops_cat(
+    network: TRTNetwork,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.cat.cat(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        tensors=args[0],
+        dim = args_bounds_check(args, 2, 1),
+    )
+
+
 def embedding_param_validator(embedding_node: Node) -> bool:
     scale_grad_by_freq = args_bounds_check(embedding_node.args, 3)
     sparse = args_bounds_check(embedding_node.args, 4)
--- a/py/torch_tensorrt/dynamo/conversion/impl/cat.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/cat.py
@@ -1,11 +1,16 @@
 from typing import Dict, Optional, Sequence, Union
 
+import numpy as np
 import torch
 from torch.fx.node import Target
 from torch_tensorrt.dynamo._SourceIR import SourceIR
-from torch_tensorrt.fx.converters.converter_utils import set_layer_name
 from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
-from torch_tensorrt.dynamo.conversion.converter_utils import get_trt_tensor
+from torch_tensorrt.dynamo.conversion.converter_utils import (
+    SourceIR,
+    get_positive_dim,
+    get_trt_tensor,
+)
+from torch_tensorrt.fx.converters.converter_utils import set_layer_name
 from torch_tensorrt.fx.types import TRTNetwork, TRTTensor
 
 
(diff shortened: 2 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -46,6 +46,24 @@ def aten_ops_batch_norm(
...
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.cat.cat(
+        network,
+        target,
+        SourceIR.ATEN,
+        name,
+        tensors=args[0],
+        dim=args_bounds_check(args, 2, 1),
```
gs-olive, line 63: Switch to `dim=args_bounds_check(args, 1, 0)`, since the default should be 0 at the `args[1]` position, [as here](https://github.com/pytorch/pytorch/blob/a4dec8d306d96637aa4dc1ee9deba289b128c148/aten/src/ATen/native/native_functions.yaml#L1324).

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -0,0 +1,28 @@
...
+    input: TRTNetwork,
+    dim: int,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    if any(not isinstance(t, TRTTensor) for t in input):  # type: ignore[union-attr]
+        raise RuntimeError(
+            f"cat received inputs {input} that is not part " "of the TensorRT region!"
+        )
+    concat_layer = network.add_concatenation(input)
+    if dim < 0:
+        dim = len(input[0].shape) + dim
```
gs-olive, line 24: Switch to `get_positive_dim`, as here:
https://github.com/pytorch/TensorRT/blob/ecdc040aec3902cbc396648043f54c96e6a91ee7/py/torch_tensorrt/fx/converters/converter_utils.py#L83

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -0,0 +1,28 @@
...
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: TRTNetwork,
+    dim: int,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    if any(not isinstance(t, TRTTensor) for t in input):  # type: ignore[union-attr]
+        raise RuntimeError(
+            f"cat received inputs {input} that is not part " "of the TensorRT region!"
+        )
```
gs-olive, line 21: Instead of throwing an error, instead check the inputs and use `get_trt_tensor` for any inputs which are not already `TRTTensor` objects.

gs-olive: Alternatively, #2324 can help with this

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -0,0 +1,28 @@
...
+from torch_tensorrt.fx.converters.converter_utils import set_layer_name
+from torch_tensorrt.fx.types import TRTNetwork, TRTTensor
+
+
+def cat(
+    network: TRTNetwork,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: TRTNetwork,
```
gs-olive, line 15: This should likely be something like `Sequence[TRTTensor]` or `Sequence[Union[TRTTensor, torch.Tensor, np.ndarray]]`

py/torch_tensorrt/dynamo/conversion/impl/cat.py
gs-olive: Consider using `@enforce_tensor_types` here

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -0,0 +1,29 @@
...
+from torch_tensorrt.dynamo.conversion.converter_utils import get_trt_tensor
+from torch_tensorrt.fx.types import TRTNetwork, TRTTensor
+
+
+def cat(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: Union[TRTTensor, Sequence[TRTTensor]],
```
gs-olive, line 17: Switch to `Sequence[Union[TRTTensor, np.ndarray, torch.Tensor]]`

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -0,0 +1,29 @@
...
+    dim: int,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    for each_input in input:
+        if(not isinstance(each_input, TRTTensor)):
+            each_input = get_trt_tensor(each_input)
```
gs-olive, line 22: Switch to:
```python
trt_tensor_input = [get_trt_tensor(ctx, each_input, name + "_tensor{i}" for i, each_input in enumerate(input)]
```

py/torch_tensorrt/dynamo/conversion/impl/cat.py
gs-olive: See above comment

gs-olive (changes requested):

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -0,0 +1,34 @@
...
+    name: str,
+    input: Sequence[Union[TRTTensor, torch.Tensor, np.ndarray]],
+    dim: int,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    for each_input in input:
+        if not isinstance(each_input, TRTTensor):
+            each_input = get_trt_tensor(each_input)
+    concat_layer = ctx.net.add_concatenation(input)
+    if dim < 0:
+        dim = get_positive_dim(dim, len(input[0].shape))
```
zewenli98, line 30: Since `get_positive_dim` will internally check if `dim` is less than 0, the condition could be deleted.

zewenli98 (commented):
The current change looks good overall. A small issue could be fixed, as I commented below.

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -0,0 +1,34 @@
...
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: Sequence[Union[TRTTensor, torch.Tensor, np.ndarray]],
+    dim: int,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    for each_input in input:
+        if not isinstance(each_input, TRTTensor):
+            each_input = get_trt_tensor(each_input)
```
gs-olive, line 27: This loop will not save the `get_trt_tensor` objects anywhere, since the `each_input` object gets reset on every loop iteration. Consider replacing with something like:
```python
    trt_inputs = []
    for i, each_input in enumerate(input):
        each_input = get_trt_tensor(ctx, each_input, name + "_tensor{i}")
        trt_inputs.append(each_input)
...
# Replace all uses of input with trt_inputs
```

gs-olive (approved):
Looks good to me!

zewenli98 (approved):
Thanks! LGTM.

gs-olive:
Verified tests passing locally.

PR #2354, opened 2023-09-29
feat: support flatten and reshape via shuffle_layer

### LINKED ISSUE #2214: Expose IShuffleLayer in dynamo.conversion.impl

zewenli98:
# Description

Support `flatten` and `reshape` via shuffle_layer

Fixes #2214

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1555,3 +1555,27 @@ def tensorrt_scaled_dot_product_attention(
     return impl.attention.scaled_dot_product_attention(
         ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2]
     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.reshape.default)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.view.default)  # type: ignore[misc]
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
+    }
+)  # type: ignore[misc]
+def aten_ops_reshape(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.shuffle.reshape(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        input=args[0],
+        shape=args[1],
+    )
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -506,7 +506,7 @@ def aten_ops_slice(
     {
         0: (TRTTensor,),
     }
-)  # type: ignore[misc]
+)
 def aten_ops_permute(
     ctx: ConversionContext,
     target: Target,
@@ -1394,7 +1394,7 @@ def conv_param_validator(conv_node: Node) -> bool:
         1: (np.ndarray, torch.Tensor, TRTTensor),
         2: (np.ndarray, torch.Tensor, TRTTensor),
     }
-)  # type: ignore[misc]
+)
 def aten_ops_convolution(
     ctx: ConversionContext,
     target: Target,
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -506,7 +506,7 @@ def aten_ops_slice(
     {
         0: (TRTTensor,),
     }
-)
+)  # type: ignore[misc]
 def aten_ops_permute(
     ctx: ConversionContext,
     target: Target,
@@ -1394,7 +1394,7 @@ def conv_param_validator(conv_node: Node) -> bool:
         1: (np.ndarray, torch.Tensor, TRTTensor),
         2: (np.ndarray, torch.Tensor, TRTTensor),
     }
-)
+)  # type: ignore[misc]
 def aten_ops_convolution(
     ctx: ConversionContext,
     target: Target,
--- a/py/torch_tensorrt/dynamo/conversion/converter_utils.py
+++ b/py/torch_tensorrt/dynamo/conversion/converter_utils.py
@@ -17,6 +17,7 @@ from torch_tensorrt.dynamo.conversion.converter_registry import (
 from torch_tensorrt.fx.converters.converter_utils import (
     Frameworks,
     get_axes_for_reduce_op,
+    set_layer_name,
     unified_dtype_converter,
 )
 from torch_tensorrt.fx.types import TRTDataType, TRTTensor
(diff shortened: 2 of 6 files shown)


zewenli98:
@bowang007

zewenli98:
When I test `flatten` via `torch.flatten(inputs, start_dim, end_dim)`, got error `AssertionError: False is not true : expected ops {<OpOverload(op='aten.flatten', overload='using_ints')>}, actuall ops {<OpOverload(op='aten.view', overload='default')>, <OpOverloadPacket(op='aten.sym_size')>, <built-in function mul>}`

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1532,3 +1532,41 @@ def tensorrt_scaled_dot_product_attention(
     return impl.attention.scaled_dot_product_attention(
         ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2]
     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.reshape.default)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.view.default)  # type: ignore[misc]
+def aten_ops_reshape(
+    network: TRTNetwork,
```
gs-olive, line 1540: This should be a `ctx: ConversionContext` now

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1532,3 +1532,41 @@ def tensorrt_scaled_dot_product_attention(
...
+        SourceIR.ATEN,
+        name,
+        input=args[0],
+        shape=args[1],
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.flatten.using_ints)  # type: ignore[misc]
+def aten_ops_flatten(
+    network: TRTNetwork,
```
gs-olive, line 1558: See above

tests/py/dynamo/conversion/test_reshape_aten.py
gs-olive, line 1: Add testing via reshape, as discussed

gs-olive (changes requested):
Is there a specific reason - performance or otherwise, why `flatten` should need a different implementation than `reshape`, when using static shapes? Specifically, we can comment out the `flatten` implementation for now, and for any converters needing `flatten` for static shapes, they can just use a reshape and flatten the dimensions themselves.

As an alternative, @zewenli98, you can add a utility `flatten_dims`, which will flatten the dimensions of an input tensor into a reshape-usable form, then you can have @bowang007's converter test that utility.

zewenli98:
> Is there a specific reason - performance or otherwise, why `flatten` should need a different implementation than `reshape`, when using static shapes? 

Thanks for the advice! I did this because I noticed there's a `flatten` op in this [schema](https://github.com/pytorch/pytorch/blob/bdecdfd202df3fa25fd9998070fd19fee4b14971/aten/src/ATen/native/native_functions.yaml#L2571). I thought our goal is to support these native_functions as much as possible. But anyways, I'll comment out the `flatten` converter and then wrap the reshape to do flatten op.

gs-olive:
Generally, the focus is to cover as much of this operation set as possible: https://pytorch.org/docs/stable/ir.html#core-aten-ir, though if there are operators that show up which we can directly convert as opposed to lowering, that is certainly a good thing to have.

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
```diff
@@ -0,0 +1,59 @@
...
+        input = get_trt_tensor(ctx, input, f"{name}_flatten")
+
+    num_elements = 1
+    for i in range(start_dim, end_dim + 1):
+        num_elements *= shape[i]
+
+    new_shape = (
+        tuple(shape[:start_dim])
+        + (num_elements,)
+        + (tuple(shape[end_dim + 1 :]) if end_dim + 1 < dim_size else tuple())
```
gs-olive, line 56: This can just be `tuple(shape[end_dim + 1 :])` - Python will return an empty tuple if the slice range is past the end of the tuple

gs-olive (commented):
I do still think `flatten_dims` can be a utility which gives the shape to pass to `reshape`. That way, it can get tested as a utility and not as a converter (see `tests/py/dynamo/conversion/test_converter_utils.py`). Added a suggestion on syntax.

zewenli98:
> I do still think `flatten_dims` can be a utility which gives the shape to pass to `reshape`.

I tried implementing `flatten_dims` in `converter_utils.py`. Since `flatten_dims` needs to call `reshape`, it caused circular import.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1532,3 +1532,42 @@ def tensorrt_scaled_dot_product_attention(
...
+# ) -> Union[TRTTensor, Sequence[TRTTensor]]:
+#     return impl.shuffle.flatten(
+#         ctx,
+#         target,
+#         SourceIR.ATEN,
+#         name,
+#         input=args[0],
+#         start_dim=args_bounds_check(args, 1, 0),
+#         end_dim=args_bounds_check(args, 2, -1),
+#     )
```
gs-olive, line 1573: This can be removed for now

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -511,3 +511,35 @@ def to_numpy(
         raise AssertionError(
             f"to_numpy can only be called on None, bool, int, float, np.ndarray, or torch.Tensor, got: {value}"
         )
+
+
+def flatten_dims(
```
gs-olive, line 516: Add a test to `test_converter_utils.py`, which tests this function on a simple Numpy array or Torch Tensor, to validate the output shape is correct given the input shape.

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
```diff
@@ -0,0 +1,25 @@
...
+from torch_tensorrt.fx.converters.converter_utils import set_layer_name
+from torch_tensorrt.fx.types import TRTTensor
+
+
+def reshape(
+    ctx: ConversionContext,
+    target: Union[Target, str],
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: Sequence[Union[TRTTensor, torch.Tensor, np.ndarray]],
```
gs-olive, line 17: Instead of this, you can restrict the input further to be `input: TRTTensor`, then use `@enforce_tensor_types({0: (TRTTensor,)})` on the converter

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
```diff
@@ -0,0 +1,25 @@
...
+    shape: Sequence[int],
+) -> TRTTensor:
+    if not isinstance(input, TRTTensor):
+        input = get_trt_tensor(ctx, input, f"{name}_input")
```
gs-olive, line 21: If the above change is made, this line can be removed

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
zewenli98: Ok. So, for all converters with single-value input, do you mean we should use `@enforce_tensor_types({0: (TRTTensor,)})`, making the `input` convert to `TRTTensor` automatically?

gs-olive: Yes, only if that converter _requires_ a `TRTTensor`, though. There are some converters like conv which can use numpy arrays or torch Tensors to reduce runtime computations by performing operations on the weights at compile time

zewenli98: However, for `input` with a sequence, like [this case](https://github.com/pytorch/TensorRT/pull/2343#discussion_r1346509449), we'd better cast to `TRTTensor` manually?

gs-olive: Yes, there is currently not support for sequence inputs, which is why that PR has to call `get_trt_tensor` directly for each input

zewenli98: BTW, if we put the type conversion part to decorators, in some scenarios, we may want to directly call a function like `impl.shuffle.reshape(...)`. Is there any potential error?

gs-olive: I don't understand the use case which could cause the issue exactly here. This decorator is only intended for converters in the `aten_ops_converters.py` file, and not for `impl` implementations. It is the `aten_ops_converters.py` converter's responsibility to ensure that all called converter implementations are valid for the specified tensor types.

tests/py/dynamo/conversion/test_reshape_aten.py
```diff
@@ -65,6 +68,7 @@ def forward(self, x):
         self.run_test_with_dynamic_shape(
             TestModule(target_shape),
             input_specs,
+            expected_ops={torch.ops.aten.view.default},
```
gs-olive, line 71: This line can be removed, in accordance with the new testing PR

gs-olive (approved):
Looks good to me, pending CI pass.

gs-olive:
Relevant tests pass locally on Torch 2.1.0. Merging to `main`.

zewenli98:
@bowang007 Please consult this PR for the `shuffle` op.

bowang007:
@zewenli98  Thanks! Let me update PR accordingly

PR #2355, opened 2023-09-30
feat: support prod, max, min, and mean via reduce layer

### LINKED ISSUE #2205: Expose IReduceLayer in dynamo.conversion.impl

zewenli98:
# Description

Support prod, max, min, and mean via reduce layer

Fixes #2205 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -668,7 +668,7 @@ def aten_ops_amax(
         SourceIR.ATEN,
         name,
         args[0],
-        args[1],
+        args_bounds_check(args, 1, replacement=[]),
         args_bounds_check(args, 2, replacement=False),
     )
 
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -692,6 +692,86 @@ def aten_ops_sum(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.prod.default)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.prod.dim_int)  # type: ignore[misc]
+def aten_ops_prod(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.reduce.prod(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args_bounds_check(args, 1, replacement=None),
+        args_bounds_check(args, 2, replacement=False),
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.max.default)  # type: ignore[misc]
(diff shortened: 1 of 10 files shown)


py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -61,3 +62,121 @@ def sum(
...
+    input_val: TRTTensor,
+    dim: Optional[Union[int, Sequence[int]]],
+    keepdim: bool,
+) -> TRTTensor:
+    if (isinstance(input_val, TRTTensor)) and (
+        input_val.dtype == trt.int8 or input_val.dtype == trt.int32
+    ):
+        input_val = cast_trt_tensor(ctx, input_val, trt.float32, name)
+
+    if dim is None:
```
gs-olive, line 172: Based on `torch` behavior, it seems that `torch.mean` can also accept an empty list as an indicator to reduce over all dimensions (but `min`, `max`, and `prod` don't seem to accept this). Consider switching this to: `if dim is None or (isinstance(dim, (tuple, list)) and len(dim) == 0):`

zewenli98: Good catch! And then I also checked docs for `amax` and `sum`. They do accept empty sequence as well, so I made changes accordingly.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -669,6 +670,135 @@ def aten_ops_sum(
...
+    return impl.reduce.min(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+        args_bounds_check(args, 2, replacement=False),
+        return_indices=True,
+    )
```
gs-olive, line 779: Consider coalescing these by using something like:
```python
    return impl.reduce.min(
        ctx,
        target,
        SourceIR.ATEN,
        name,
        args[0],
        args_bounds_check(args, 1, replacement=None),
        args_bounds_check(args, 2, replacement=False),
        return_indices=(target==torch.ops.aten.min.dim),
    )
```

The same could apply for `max`

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -61,3 +65,121 @@ def sum(
...
+    layer = ctx.net.add_reduce(
+        input_val,
+        trt.ReduceOperation.MAX,
+        axes=get_axes_for_reduce_op(get_positive_dim(dim, len(input_val.shape))),
+        keep_dims=keepdim,
+    )
+    set_layer_name(layer, target, name, source_ir)
+
+    if return_indices:
+        return layer.get_output(0), None
```
gs-olive, line 124: @narendasan, @apbose - there are certain converter cases where indices are returned from an operator but never used nor accessed in the graph (confirmed via validator). In these cases, we wouldn't want to use extra computation time to add layers for an unused tensor. Which of these seems best?

- Return None for the unused tensor, as here - `(data, None)`
- Return only the data, since the unused tensor should never be accessed, as in: `(data,)`

gs-olive:
```
max.dim(Tensor self, int dim, bool keepdim=False) -> (Tensor values, Tensor indices)
```

gs-olive (approved):
Looks good to me, pending CI pass.

gs-olive:
Tests passing - verified locally

PR #2376, opened 2023-10-09
chore: Switch to new export apis

peri044:
# Description

* Move to new dynamic shapes API which accepts dynamic shapes for each Input via names.
* Add name attribute for `torch_tensorrt.Input` which is only used for dynamo tracing.
* Remove constraints API calls
* Remove unittest patching and use `run_decompositions` API.

## Type of change

# Checklist:


gs-olive (approved):
This looks good to me, pending CI

py/torch_tensorrt/_Input.py
```diff
@@ -180,6 +182,9 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             else:
                 self.torch_tensor = self.example_tensor()
 
+        if "name" in kwargs:
+            self.name = kwargs["name"]
```
gs-olive, line 186: If the name is not specified, or the user does not know the name (such as in a pretrained model from torchvision, etc.), do we have a way to extract the name or otherwise match the tensor to a name?

gs-olive: Additionally, since the `self.name` field is used below, and most users likely will not fill in this field, even when using Dynamic shapes, the script could error out if no default is provided here.

gs-olive (commented):
Overall looks good - added some questions on the `name` parameter

py/torch_tensorrt/_Input.py
peri044: > If the name is not specified .....

I had a similar thought. I digged into if we can get input names from the nn.Module but couldn't find one. We can bring this up with Meta and will check the latest source. 

> Additionally, since the self.name ..... 
I added an assertion error now if user has a dynamic input but doesn't provide a non-null name. However, we cannot verify if the name is actually valid since I didn't find anything to compare with.

gs-olive: Thanks for the clarification. This sounds good. Additionally, since the `self.name` check happens in the tracer and `self.name` is only defined here if `"name" in kwargs`, then could you also add an `else` here, to set `self.name` to None in the event no name is specified, so that the attribute is still defined? Then the tracer check can just be `if input.name is None`.

peri044: The name is pre-defined to be an empty string here https://github.com/pytorch/TensorRT/blob/d58bc5df85c49982e6d6c559647ba585541bd38d/py/torch_tensorrt/_Input.py#L50

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -228,7 +228,10 @@ def create_trt_exp_program(
     and constructs an Exported Program object with the new IO node names and state_dict
     """
     input_nodes = [node for node in gm.graph.nodes if node.op == "placeholder"]
-    output_nodes = [node for node in gm.graph.nodes if node.op == "output"]
+    # output_nodes = [node for node in gm.graph.nodes if node.op == "output"]
+    # graph = [placeholder, conv1, relu, trt_node, getitem, output]
+    # output_nodes[0].args[0]
+    output_nodes = list(gm.graph.nodes)[-1].args[0]
```
gs-olive, line 234: This will now contain a list of nodes which output Tensors, as opposed to the single output `Node`. Is this the intended behavior here?

Additionally, do we need to consider or include an assertion for graphs which have no outputs?

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -292,12 +296,16 @@ def inline_trt_modules(
             # Insert getitem nodes as outputs (for export serialization to work)
             with gm.graph.inserting_after(trt_node):
                 getitem_output = gm.graph.call_function(operator.getitem, (trt_node, 0))
+                getitem_output.meta["val"] = trt_node.meta["val"]
```
gs-olive, line 299: BERT appears to be having an issue with its `dict` output in the exporter. Is this something that could be resolved in the `output_spec` section?

py/torch_tensorrt/dynamo/_exporter.py
peri044: Yes, that's the intended behavior. 
I added an assert output_nodes as a safe check, but I haven't found graphs without an output node implicitly added. Did you encounter them?

py/torch_tensorrt/dynamo/_exporter.py
peri044: Are you referring to BERT failing test in test_models_export ?

gs-olive: Yes, that is the test case I am referring to

py/torch_tensorrt/dynamo/_exporter.py
gs-olive: No, I have not encountered such a case. I think every graph likely needs an output node, and any graph without an output node or with an empty `args` for its output node would be invalid, so I think the assert should suffice, thanks for the clarification.

py/torch_tensorrt/dynamo/_exporter.py
peri044: That's different. BERT wasn't being exported properly (torch.export.export() was failing due to some nightly torch changes). Looks like they got fixed as of today's nightly

py/torch_tensorrt/_Input.py
```diff
@@ -68,7 +69,8 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             format (torch.memory_format or torch_tensorrt.TensorFormat): The expected format of the input tensor (default: torch_tensorrt.TensorFormat.NCHW)
             tensor_domain (Tuple(float, float), optional): The domain of allowed values for the tensor, as interval notation: [tensor_domain[0], tensor_domain[1]).
                 Note: Entering "None" (or not specifying) will set the bound to [0, 2)
-
+            torch_tensor (torch.Tensor): Holds a corresponding torch tensor with this Input.
+            name (str, optional): Name of this input in the pytorch graph. Used to specify dynamic shapes in dynamo tracer.
```
gs-olive, line 73: Would it be preferred to say `Name of this input in the "forward" function`? Just to make it easier to specify the naming when using Dynamic shapes, pending a way of automatically getting these names from the graph or elsewhere.

gs-olive (approved):
Looks good to me - added a minor documentation comment/question

py/torch_tensorrt/_Input.py
narendasan: What is the behavior if the user does not provide a name but it is necessary for dynamo?

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -229,6 +229,8 @@ def create_trt_exp_program(
     """
     input_nodes = [node for node in gm.graph.nodes if node.op == "placeholder"]
     output_nodes = [node for node in gm.graph.nodes if node.op == "output"]
+    assert output_nodes
+    output_nodes = output_nodes[0].args[0]
```
narendasan, line 233: Is this always the output_nodes?

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -276,6 +278,7 @@ def inline_trt_modules(
                 (trt_module_node.args, trt_module.engine),
             )
             trt_node.meta["val"] = []
+            assert num_outputs
```
narendasan, line 281: Can you make this assert more clear? like `assert num_outputs > 0` or something?

py/torch_tensorrt/dynamo/_tracer.py
```diff
@@ -81,46 +65,38 @@ def trace(
...
+
+            dynamic_shapes[input.name] = dynamic_dims
+
+    experimental_decompositions = kwargs.get(
+        "enable_experimental_decompositions", ENABLE_EXPERIMENTAL_DECOMPOSITIONS
+    )
+
+    exp_program = export(
+        mod, tuple(torch_inputs), dynamic_shapes=dynamic_shapes
+    ).run_decompositions(get_decompositions(experimental_decompositions))
```
narendasan, line 100: I thought we were moving decomposition to the compile side so this function could eventually be not required.

py/torch_tensorrt/dynamo/_exporter.py
peri044: done

py/torch_tensorrt/dynamo/_tracer.py
peri044: Modified it now. Previously, I thought we wanted it in another PR.

py/torch_tensorrt/dynamo/_exporter.py
peri044: Yeah

Issue #2394, opened 2023-10-12
Add functionality to aten::index converter to handle numpy indices


apbose:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

apbose:
#2396

PR #2396, opened 2023-10-13
Numpy changes for aten::index converter

apbose:
This PR addresses the changes for the cases where indices are list of numpy arrays or numpy for aten::index. #2394

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/select.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/select.py
@@ -81,30 +81,20 @@ def index(
     source_ir: Optional[SourceIR],
     name: str,
     input: TRTTensor,
-    index: Union[
-        TRTTensor,
-        Sequence[TRTTensor],
-        np.ndarray,
-        Sequence[np.ndarray],
-        torch.Tensor,
-        Sequence[torch.Tensor],
-    ],
+    index: Sequence[Union[TRTTensor, np.ndarray, torch.Tensor]],
 ) -> TRTTensor:
     adv_indx_indices = []
     tensor_indices = []
-    # _LOGGER.debug(f"The index shape is {index.shape}")
     # check if the input is dynamic
     dynamic_shape = has_dynamic_shape(input.shape)
     # is_numpy is a flag to specify if all the indices are numpy or torchTensor.
     # If any is not this flag will be set to False
-    is_numpy = True
-    _LOGGER.debug(f"Checking for the is_numpy flag")
-    for i, ind in enumerate(index):
-        if ind is None:
-            continue
-        if not (isinstance(ind, torch.Tensor) or isinstance(ind, np.ndarray)):
-            is_numpy = False
-            break
+    _LOGGER.debug(
+        f"Determining whether aten.index constant-index optimization can be invoked"
+    )
+    is_numpy = all(
+        isinstance(ind, (torch.Tensor, np.ndarray)) for ind in index if ind is not None
+    )
     # here we need to check if all the index are broadcastable
     # if no, then we need to broadcast
     last_index = None
@@ -117,7 +107,6 @@ def index(
             # other cases are kept as TRTTensor
             if is_numpy:
                 ind = to_numpy(ind)
-                is_numpy = True
             else:
                 ind = get_trt_tensor(ctx, ind, name + f"_parameter_to_fp32_tensor_{i}")
             if last_index is not None:
@@ -156,9 +145,7 @@ def index(
         for i in range(rank):
             dim = input_shape[i]
             dim_tensor = get_trt_tensor(ctx, dim, name + f"_individual_dim_{i}")
-            # dim_tensor_list is a list of tensors or numpy
-            if is_numpy:
-                dim_list.append(dim)
+            # dim_tensor_list is a list of tensors
             dim_tensor_list.append(dim_tensor)
 
         # for cases like
@@ -211,12 +198,12 @@ def index(
         # tensor index = \sum_{i=1}^m (ind_i * \prod_{j=i+1}^m (x_j)),  ind_i is input indices[i], x_j is the
         # // j dimension of input x.
         if is_numpy:
-            multiplier = dim_list[adv_indx_indices[adv_indx_count - 1]]
+            multiplier = input_shape[adv_indx_indices[adv_indx_count - 1]]
             cum_adv_index = tensor_indices[adv_indx_count - 1]
             for i in range(adv_indx_count - 2, -1, -1):
                 adv_index = multiplier * tensor_indices[i]
                 cum_adv_index = cum_adv_index + adv_index
-                multiplier = multiplier * dim_list[adv_indx_indices[i]]
+                multiplier = multiplier * input_shape[adv_indx_indices[i]]
             cum_adv_index = get_trt_tensor(
                 ctx, cum_adv_index, name + f"_index_sum_intermediate"
             )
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -80,23 +81,45 @@ def index(
...
     input: TRTTensor,
-    index: Union[TRTTensor, Sequence[TRTTensor]],
+    index: Union[
+        TRTTensor,
+        Sequence[TRTTensor],
+        np.ndarray,
+        Sequence[np.ndarray],
+        torch.Tensor,
+        Sequence[torch.Tensor],
+    ],
```
gs-olive, line 91: This can be shortened to `index: Sequence[Union[TRTTensor, np.ndarray, torch.Tensor]]`, since the subsequent code assumes `index` is a sequence of tensors.

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -80,23 +81,45 @@ def index(
...
     adv_indx_indices = []
     tensor_indices = []
     # _LOGGER.debug(f"The index shape is {index.shape}")
     # check if the input is dynamic
     dynamic_shape = has_dynamic_shape(input.shape)
-
+    # is_numpy is a flag to specify if all the indices are numpy or torchTensor.
+    # If any is not this flag will be set to False
+    is_numpy = True
+    _LOGGER.debug(f"Checking for the is_numpy flag")
```
gs-olive, line 101: Consider switching to `"Determining whether aten.index constant-index optimization can be invoked"`

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -80,23 +81,45 @@ def index(
...
+    for i, ind in enumerate(index):
+        if ind is None:
+            continue
+        if not (isinstance(ind, torch.Tensor) or isinstance(ind, np.ndarray)):
+            is_numpy = False
+            break
```
gs-olive, line 107: Consider shortening to
```python
is_numpy = all(isinstance(ind, (torch.Tensor, np.ndarray)) for ind in index if index is not None)
```

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -80,23 +81,45 @@ def index(
...
             _LOGGER.debug(f"Shape of {i} index is {ind.shape}")
             adv_indx_indices.append(i)
-            # torch.nn.parameter.Parameter=> torch.Tensor
-            ind = get_trt_tensor(ctx, ind, name + f"_parameter_to_fp32_tensor_{i}")
+            # torch.nn.parameter.Parameter=> numpy array
+            # numpy array is kept as numpy
+            # other cases are kept as TRTTensor
+            if is_numpy:
+                ind = to_numpy(ind)
+                is_numpy = True
```
gs-olive, line 120: This line can be removed

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -80,23 +81,45 @@ def index(
...
 ) -> TRTTensor:
     adv_indx_indices = []
     tensor_indices = []
     # _LOGGER.debug(f"The index shape is {index.shape}")
     # check if the input is dynamic
     dynamic_shape = has_dynamic_shape(input.shape)
-
+    # is_numpy is a flag to specify if all the indices are numpy or torchTensor.
+    # If any is not this flag will be set to False
+    is_numpy = True
```
gs-olive, line 100: This line can be replaced if the above suggestion is used

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -175,47 +203,58 @@ def index(
...
+            multiplier = dim_list[adv_indx_indices[adv_indx_count - 1]]
+            cum_adv_index = tensor_indices[adv_indx_count - 1]
+            for i in range(adv_indx_count - 2, -1, -1):
+                adv_index = multiplier * tensor_indices[i]
+                cum_adv_index = cum_adv_index + adv_index
+                multiplier = multiplier * dim_list[adv_indx_indices[i]]
+            cum_adv_index = get_trt_tensor(
+                ctx, cum_adv_index, name + f"_index_sum_intermediate"
             )
-            multiplier = convert_binary_elementwise(
```
gs-olive, line 210: Why did the implementation of the `multiplier` change?

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -127,11 +151,14 @@ def index(
...
         dim_tensor_list = []
+        dim_list = []
 
         for i in range(rank):
             dim = input_shape[i]
             dim_tensor = get_trt_tensor(ctx, dim, name + f"_individual_dim_{i}")
-            # dim_tensor_list is a list of tensors
+            # dim_tensor_list is a list of tensors or numpy
+            if is_numpy:
+                dim_list.append(dim)
```
gs-olive, line 161: Why is a different list needed here if we have the `is_numpy` flag?

gs-olive: As in - could `dim_list` just hold the `ITensors` _or_ the `numpy` arrays, or do you need both? If both are needed, can you just use `input_shape` in place of `dim_list`, as they should be the same

gs-olive (changes requested):
Overall, optimization looks great. Added comments on key parts in the new code.

py/torch_tensorrt/dynamo/conversion/impl/select.py
gs-olive: I meant the line `is_numpy = True` can be replaced with `is_numpy = all(isinstance(ind, (torch.Tensor, np.ndarray)) for ind in index if index is not None)`

py/torch_tensorrt/dynamo/conversion/impl/select.py
gs-olive: Nevermind - I believe I misread the code. Looks good to me

py/torch_tensorrt/dynamo/conversion/impl/select.py
gs-olive: The `is_numpy = True` line would no longer be needed

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -127,6 +140,7 @@ def index(
         rank = len(input_shape)
         adv_indx_count = len(adv_indx_indices)
         dim_tensor_list = []
+        dim_list = []
```
gs-olive, line 143: If this is unused, `dim_list = []` can be removed

gs-olive (approved):
Pending the above comment, it looks good to me!

gs-olive (changes requested):
Works well on the SD usecase, and tests pass locally - could a test case be added specifically for this addition, which tests the functionality when all inputs are constants?

apbose (commented):
Hi @gs-olive. Regarding adding of the test for this feature, the earlier tests were all numpy indices, where is_numpy  is set as True. So those tests should be good for this. Is there any way so that tests can be added so that the indices are ITensors?

gs-olive:
Yes, there is. You would just need to pass the indices as inputs to the `forward` function, instead of having them as constants in the model/graph

tests/py/dynamo/conversion/test_index_aten.py
```diff
@@ -2,11 +2,10 @@
 
 import torch
 import torch.nn as nn
+from harness import DispatchTestCase
```
gs-olive, line 5: Requires switch back to `.harness` to pass CI

apbose: Ok corrected, wil wait for CI then will merge.

gs-olive (approved):
Looks good to me!

PR #2402, opened 2023-10-17
feat: support tile dynamo converter

### LINKED ISSUE #2215: Expose ISliceLayer in dynamo.conversion.impl

zewenli98:
# Description

Support `tile` dynamo converter. This is a part of `slice` converter.

Fixes part of #2215 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
@@ -1,7 +1,6 @@
 import math
-from typing import Optional
+from typing import Optional, Sequence
 
-import numpy as np
 import tensorrt as trt
 from torch.fx.node import Target
 from torch_tensorrt.dynamo._SourceIR import SourceIR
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
```diff
@@ -157,3 +159,39 @@ def chunk(
...
+        # prepend 1 to dims
+        dims = (1,) * -diff + tuple(dims)
+
+    if all(isinstance(d, int) for d in dims):
+        shapes = [i * j for i, j in zip(input.shape, dims)]
+    else:
+        shapes = []
+        for i, (s, d) in enumerate(zip(input.shape, dims)):
+            shapes.append(
+                impl.elementwise.mul(ctx, target, source_ir, f"{name}_mul_{i}", s, d)
```
apbose, line 189: Why do we need this part, if in the function definition the dims is Sequence[int]? 
I am assuming it is for the case where the dims is a sequence of ITensors? Then accordingly maybe the function definition needs to be changed?

zewenli98: Yes, the `else` part is not needed. Thanks for comments!

apbose (commented):
Looks good to me! Just a minor clarification I wanted, in [this](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml) I see that dims are SymInt[] so that should be fine for int based arithmetic operations right?

zewenli98:
Yes, I think `symInt` is kind of like `int` per the [doc](https://pytorch.org/docs/stable/torch.html#torch.SymInt). @gs-olive Could you help us confirm we can use `int` when seeing `symInt` in the schema?

gs-olive:
Hi - generally `SymInt` and `int` are not interchangeable, since `int` is a concrete value whereas `SymInt` does not generally have a fixed or concrete value. In the test cases, are the values you encounter in `args[1]` `int`s or `SymInt`s when calling this function?

zewenli98:
Hi, in the tests we pass in `int[]` and also get `int[]` for `args[1]`

gs-olive:
In that case, we should be typing everything with `int` and not `SymInt`, since we are not able to support Dynamic shape through `SymInt` now anyway.

apbose (approved):

PR #2411, opened 2023-10-23
feat: a lowering pass to re-compose ops into aten.linear

### LINKED ISSUE #2398: ✨[Feature] Writing a lowering pass to re-compose ops into `aten.linear`
The discussion can be found here: https://github.com/pytorch/TensorRT/discussions/2296

zewenli98:
# Description

A lowering pass to re-compose ops into `aten.linear`.

A discussion about why we need this lowering pass can be found here: #2296 

Fixes #2398 

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/lowering/passes/lower_linear.py
```diff
@@ -0,0 +1,67 @@
...
+        out = orig.graph.call_function(
+            torch.ops.aten.addmm.default,
+            args=(bias, input, W_T),
+        )
+
+    # Assign the output of the graph to be the single getitem output
+    output.args = (out,)
+
+    orig.graph.lint()
+    orig.recompile()
```
gs-olive, line 59: The `linear` operator may not necessarily have the same restrictions as the attention operator did, so it may not be necessary to build the fx graph from scratch. For instance, this whole block could instead be:
```python
    def orig(
        input: torch.Tensor, weight: torch.Tensor, bias: torch.Tensor
    ) -> torch.Tensor:
        W_T = torch.ops.aten.permute.default(weight, [1, 0])
        out = torch.ops.aten.addmm.default(bias, input, W_T)
        return out
```
The above would need verification to validate functionality.

tests/py/dynamo/lowering/test_aten_lowering_passes.py
```diff
@@ -267,5 +267,110 @@ def forward(self, q, k, v):
...
+
+        inputs = [
+            torch.rand((3, 32)).cuda(),
+            torch.rand((64, 32)).cuda(),
+            torch.rand((64,)).cuda(),
+        ]
+
+        fx_graph = torch.fx.symbolic_trace(Linear())
+        expected_ops = {torch.ops.aten.linear.default}
+        unexpected_ops = {}
```
gs-olive, line 285: Unexpected ops include `addmm` and `permute`

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -105,22 +105,6 @@ def alias_replacement(x: torch.Tensor) -> torch.Tensor:
...
-    input_: torch.Tensor,
-    mat1: torch.Tensor,
-    mat2: torch.Tensor,
-    *,
-    beta: int = 1,
-    alpha: int = 1,
-) -> torch.Tensor:
-    return torch.add(
-        torch.mul(input_, beta), torch.mul(torch.matmul(mat1, mat2), alpha)
-    )
```
gs-olive, line 121: This PR also requires a converter for `aten.addmm`, since we are no longer decomposing it, so it would be a decrease in coverage otherwise

gs-olive: Additionally, you can remove the `addmm` lowering pass tests, since the functionality is no longer present

gs-olive (changes requested):
Added comments detailing suggested improvements. Overall the design looks good!

py/torch_tensorrt/dynamo/lowering/passes/lower_linear.py
zewenli98: Thanks for the details. I'm thinking why we need these restrictions for `attention` but not for `linear`. Is there anything different in `attention`? Can we just define the `orig` like:
```python
def orig(
        q: torch.Tensor, k: torch.Tensor, v: torch.Tensor
    ) -> torch.Tensor:
        outputs = torch.ops.aten._scaled_dot_product_efficient_attention.default(q, k, v, None, False)
        out = operator.getitem(outputs, 0)
        return out
```

zewenli98: I just tried to use the `orig` function above, and it also works for `attention`.

gs-olive: Ok that's great - the above is definitely cleaner and the preferred usage. I think I had overlooked it or was working on a Torch nightly where that wasn't the case. If you could add the change to the attention pass as well for this PR, that would be much appreciated.

zewenli98:
@gs-olive Thanks for your review and constructive comments. I have addressed all the issues above. Please take a look when you get the chance. Thanks a lot!

zewenli98:
@gs-olive Can we merge into the main branch if everything looks good?

gs-olive:
@zewenli98 - this one looks good, but would like to wait until later this week to merge this particular change

gs-olive (approved):
Looks good to me!

PR #2421, opened 2023-10-30
Gather implementation in impl

apbose:
This has the "gather" layer implementation in impl


apbose:
Closed in favor of #2457.

PR #2429, opened 2023-11-01
feat: support more elementwise and unary dynamo converters

### LINKED ISSUE #2208: Expose IElementWiseLayer in dynamo.conversion.impl

zewenli98:
# Description

Support more elementwise and unary dynamo converters.

For `bitwise_op` converters, TensorRT doesn't support bitwise ops. Thus, if inputs are `int` values, they will fall back to pytorch. If inputs are `bool`, they will be converted with the help of `logical_op`.

Fixes #2208 #2199 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1701,6 +1701,120 @@ def aten_ops_logical_xor(
     )
 
 
+def bitwise_type_validator(node: Node) -> bool:
+    targets = [
+        torch.ops.aten.bitwise_and.Tensor,
+        torch.ops.aten.bitwise_or.Tensor,
+        torch.ops.aten.bitwise_xor.Tensor,
+    ]
+    if node.target not in targets:
+        return False
+
+    lhs_val = node.args[0]
+    rhs_val = node.args[1]
+    lhs_meta = lhs_val.meta.get("tensor_meta")
+    rhs_meta = rhs_val.meta.get("tensor_meta")
+
+    if lhs_meta is None or rhs_meta is None:
+        return False
+
+    supported_type = [torch.bool, bool]
+    return lhs_meta.dtype in supported_type and rhs_meta.dtype in supported_type
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_and.Tensor, capability_validator=bitwise_type_validator)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_and.Scalar)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_and.Scalar_Tensor)  # type: ignore[misc]
+def aten_ops_bitwise_and(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.elementwise.bitwise_and(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_or.Tensor, capability_validator=bitwise_type_validator)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_or.Scalar)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_or.Scalar_Tensor)  # type: ignore[misc]
+def aten_ops_bitwise_or(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.elementwise.bitwise_or(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_xor.Tensor, capability_validator=bitwise_type_validator)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_xor.Scalar)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_xor.Scalar_Tensor)  # type: ignore[misc]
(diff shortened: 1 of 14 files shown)


py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
```diff
@@ -422,13 +423,46 @@ def logical_xor(
...
+
 def eq(
     ctx: ConversionContext,
     target: Target,
     source_ir: Optional[SourceIR],
     name: str,
-    lhs_val: Union[TRTTensor, int, float],
-    rhs_val: Union[TRTTensor, int, float],
+    lhs_val: TRTTensor,
+    rhs_val: Union[TRTTensor, int, float, bool, Sequence[Union[int, float, bool]]],
```
gs-olive, line 465: In what cases could this be a sequence of integers, floats, or bools? Could the `convert_binary_elementwise` function handle such situations?

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
```diff
@@ -441,13 +475,30 @@ def eq(
...
+
 def gt(
     ctx: ConversionContext,
     target: Target,
     source_ir: Optional[SourceIR],
     name: str,
-    lhs_val: Union[TRTTensor, int, float],
-    rhs_val: Union[TRTTensor, int, float],
+    lhs_val: TRTTensor,
+    rhs_val: Union[TRTTensor, int, float, Sequence[Union[int, float]]],
```
gs-olive, line 501: As mentioned above, is it necessary to have this sequence case? It does not appear in the schemas, [as here](https://github.com/pytorch/pytorch/blob/d977f118ad5422bec0b57f5b98d7124a6e2366a3/aten/src/ATen/native/native_functions.yaml#L8817).

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1647,9 +1647,128 @@ def aten_ops_logical_xor(
...
+
+    if lhs_meta is None or rhs_meta is None:
+        return False
+
+    supported_type = [torch.bool, bool]
+    return lhs_meta.dtype in supported_type and rhs_meta.dtype in supported_type
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_and.Tensor, capability_validator=bitwise_type_validator)  # type: ignore[misc]
+@dynamo_tensorrt_converter(torch.ops.aten.bitwise_and.Scalar)  # type: ignore[misc]
```
gs-olive, line 1672: The `# type: ignore[misc]` can be removed as discussed

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1647,9 +1647,128 @@ def aten_ops_logical_xor(
...
+        torch.ops.aten.bitwise_or.Tensor,
+        torch.ops.aten.bitwise_xor.Tensor,
+    ]
+    if node.target not in targets:
+        return False
+
+    lhs_val = node.args[0]
+    rhs_val = node.args[1]
+    lhs_meta = lhs_val.meta.get("tensor_meta")
+    rhs_meta = rhs_val.meta.get("tensor_meta")
```
gs-olive, line 1662: If `rhs` is a Scalar, does this error out?

gs-olive (commented):
Overall looks good - had a few questions and suggestions as mentioned in the comments

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
zewenli98: Thanks for the comments. At first I thought it could be a sequence of integers, floats, or bools, but I checked the doc it should not be.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: I added support for Scalar with bool type, and related tests.

tests/py/dynamo/conversion/test_eq_aten.py
```diff
@@ -2,64 +2,65 @@
...
     @parameterized.expand(
         [
-            ("2d", (2, 1), 1),
-            ("3d", (2, 1, 2), 2.0),
+            ("2d", (5, 3), 1),
+            ("3d", (5, 3, 2), 2.0),
         ]
     )
-    def test_equal_tensor_scalar(self, _, shape, scalar):
-        class equal(nn.Module):
```
apbose, line 36: Is this case different from the previous `torch.ops.aten.eq.Tensor` case? As in in terms of the trace? Same application for the other `ge`, `gt`, `le`, `lt`, `ne`?

tests/py/dynamo/conversion/test_bitwise_and_aten.py
apbose, line 1: Overall everything and the test cases look good to me. Just one suggestion, can all the similar test cases files be consolidated into one, using lambda functions like `test_binary_ops_aten.py`? Since the structure of the test classes are similar for `and`, `or` and `xor`. Similarly for` eq`, `ge`, `gt`,` le`,` lt`,` ne`

tests/py/dynamo/conversion/test_eq_aten.py
zewenli98: I think `torch.ops.aten.eq.Tensor` and `torch.ops.aten.eq.Scalar` are different. The main difference is the second argument. In `torch.ops.aten.eq.Tensor`, it is `Tensor` while in `torch.ops.aten.eq.Scalar`, it's `Scalar`.

tests/py/dynamo/conversion/test_bitwise_and_aten.py
zewenli98: Well, that's a good point. Since the previous tests are separated into many files even though they share the similarity (like `unary` or `elementwise`), as you said, I just follow the convention. But I do agree with you that the current tests are kind of redundant, so maybe we can refactor them later?

gs-olive (commented):
Looks good to me overall! One question - should `convert_binary_elementwise` also be allowed to handle `bool` types?

zewenli98:
> Looks good to me overall! One question - should `convert_binary_elementwise` also be allowed to handle `bool` types?

Yes, I agree with you. Fixed! Thanks for the comments.

apbose (approved):

gs-olive (commented):
Just a minor comment, `bool` should likely be added here, to lines 123 + 127 since this PR adds support for `bool` in `convert_binary_elementwise`:
https://github.com/pytorch/TensorRT/blob/867dc7bf23b34f8d3809a1d64841b0013a9a0054/py/torch_tensorrt/dynamo/conversion/impl/elementwise/base.py#L123-L130
Otherwise, seems good to go!

PR #2432, opened 2023-11-02
fix: Segfault fix for Benchmarks

gs-olive:
# Description

- Segfault fix for benchmarking on Docker container with CUDNN 8.8
- Likely due to Torch 2.1.0 based on CUDNN 8.9

## Type of change

# Checklist:

  - Tested manually in Docker

### DIFF
--- a/tools/perf/perf_run.py
+++ b/tools/perf/perf_run.py
@@ -7,6 +7,7 @@ import os
 import time
 import timeit
 import warnings
+from functools import wraps
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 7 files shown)


tools/perf/perf_run.py
```diff
@@ -527,7 +528,6 @@ def recordStats(backend, timings, precision, batch_size=1, compile_time_s=None):
     )
     args = arg_parser.parse_args()
 
-    cudnn.benchmark = True
```
gs-olive, line 530: @narendasan - this line causes a segfault at inference time, when we compile the Docker container with CUDNN 8.8, and Torch 2.1.0 uses the CUDNN 8.9 Python distributions. When removed, inference works as expected.

Do you think it would be necessary/important to upgrade the build stack to CUDNN 8.9 for the upcoming release?

narendasan (approved):
LGTM

Issue #1888, opened 2023-05-05
🐛 [Bug] Loading Torch-TensorRT models (.ts) on multiple GPUs (in TorchServe)


emilwallner:
##  Bug Description

Everything works well when I'm using 1 GPU, but as soon as I try to load a model on 4 separate GPUs, I get this error:

 MODEL_LOG - RuntimeError: [Error thrown at core/runtime/TRTEngine.cpp:42] Expected most_compatible_device to be true but got false
MODEL_LOG - No compatible device was found for instantiating TensorRT engine

## To Reproduce

Create a (.ts) model and load it on 4 different GPUs. I don't know if this is specific to TorchServe, or a general issue.

Here's the simple version (TorchServe Handler):

```
def initialize(self, ctx):
        properties = ctx.system_properties
        self.device = torch.device("cuda:" + str(properties.get("gpu_id")) if torch.cuda.is_available() else "cpu")
        self.model = torch.jit.load('model.ts')
```

I'm not sure if it relates to this [issue](https://github.com/NVIDIA/TensorRT/issues/219#issuecomment-559249117). From what I can tell it seems like I need to restrict the CUDA context, however, the GPU is assigned in the handler. I tried these things, but it's still giving me the same problem. 

```
def initialize(self, ctx):
        properties = ctx.system_properties
        self.device = torch.device("cuda:" + str(properties.get("gpu_id")) if torch.cuda.is_available() else "cpu")
        torch.cuda.set_device(self.device)
        torch_tensorrt.set_device(int(properties.get("gpu_id")))

        with torch.cuda.device(int(properties.get("gpu_id"))):
              self.model = torch.jit.load('model.ts')
              self.model.to(self.device)
              self.model.eval()
```
I also tried mapping the model straight to the GPU on load, but with the same problem. 

## Expected behavior

Load a .ts model by specifying the GPU Id without any issues.

## Environment

> Build information about Torch-TensorRT can be found by turning on debug messages

Official PyTorch image: nvcr.io/nvidia/pytorch:22.12-py3
GPUs: 4x  NVIDIA A10G
Pytorch: [1.14.0a0+410ce96](https://github.com/pytorch/pytorch/commit/410ce96)
[NVIDIA CUDA 11.8.0](https://docs.nvidia.com/cuda/archive/11.8.0/cuda-toolkit-release-notes/index.html)
[TensorRT 8.5.1](https://docs.nvidia.com/deeplearning/tensorrt/release-notes/index.html)
[...]

PR #2445, opened 2023-11-08
feat: Add hardware compatibility option in Dynamo

### LINKED ISSUE #1929: ✨[Feature] Add Support for TRT 8.6 Hardware Compatibility Mode
Add support for newly-released TRT 8.6 hardware compatibility feature for GPUs with related architectures.

**Relevant Release Notes:** https://docs.nvidia.com/deeplearning/tensorrt/release-notes/index.html#rel-8-6-1

gs-olive:
# Description

- Add support for hardware compatibility for Ampere and later architectures
- Add necessary functions to support the modification throughout the stack, including C++ and Python components
- Update ABI version to address new metadata format for TRT Engines
- Update engine serialization schema accordingly
- Add test cases to validate feature

Fixes #1929 
Addresses #1888

## Type of change

- New feature (ABI-Breaking)

# Checklist:


core/runtime/TRTEngine.cpp
```diff
@@ -32,24 +32,35 @@ TRTEngine::TRTEngine(
...
 
 TRTEngine::TRTEngine(std::vector<std::string> serialized_info)
     : TRTEngine(
           serialized_info[NAME_IDX],
           serialized_info[ENGINE_IDX],
           RTDevice(serialized_info[DEVICE_IDX]),
           split(serialized_info[INPUT_BINDING_NAMES_IDX], BINDING_DELIM),
-          split(serialized_info[OUTPUT_BINDING_NAMES_IDX], BINDING_DELIM)) {}
+          split(serialized_info[OUTPUT_BINDING_NAMES_IDX], BINDING_DELIM),
+          static_cast<bool>(std::stoi(serialized_info[HARDWARE_COMPATIBLE]))) {}
```
narendasan, line 52: Call this `HW_COMPATIBLE_IDX` since it should refer to the location in the serialized format where the setting will be found

narendasan (commented):
Changes look fine, but have we tested this with torchscript to make sure nothing breaks there?

gs-olive:
Currently, there is a Torchscript test which loads a hardware-compatible `.ts` file which has been pre-compiled and attempts to run inference with it.

narendasan:
Will torchscript be able to create new engines using this API since the signature changed right? Is is just default not compatible?

gs-olive:
Yes, the signature changed and the ABI version was bumped. For compatibility with older compiled engines (like ABI 4), would there be a way to just assume `hardware_compatible=False`, if the field is missing, or does the ABI mismatch forbid any parsing of the serialized object?

narendasan:
Im not talking about old engines since they will not work with the bumped ABI. But like if I do torch_tensorrt.ts.compile() will this work properly?

gs-olive:
I see - currently, `hardware_compatible` is not enabled for Torchscript - all TS compilations will work, but will be with `hardware_compatible=False`. Should I add this feature to TS as part of this PR?

PR #2457, opened 2023-11-13
Gather Implementation

### LINKED ISSUE #2202: Expose IGridSampleLayer in dynamo.conversion.impl

apbose:
Fixes #2202


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -183,6 +183,31 @@ def aten_ops_native_group_norm(
...
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.select.gather(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        input=args[0],
+        dim=args[1],
+        index=args[2],
+        sparse_grad=args_bounds_check(args, 4, False),
```
gs-olive, line 207: The `impl.select.gather` function cannot take this argument as input, and since it seems to not be used, it can be omitted here entirely.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -183,6 +183,31 @@ def aten_ops_native_group_norm(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.gather.default)
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
+    }
+)  # type: ignore[misc]
```
gs-olive, line 191: The `# type: ignore` can be removed.

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -68,13 +68,26 @@ def select(
     indices_tensor = ctx.net.add_constant(
         index_value.shape, to_numpy(index_value)
     ).get_output(0)
-    layer = ctx.net.add_gather(input, indices_tensor, dim)
-    out = layer.get_output(0)
+    out = gather(input, indices_tensor, dim)
     if len(out.shape) != 1:
         layer = ctx.net.add_shuffle(out)
     return layer.get_output(0)
```
gs-olive, line 74: This line (`return layer.get_output(0)`) will fail since `layer` is no longer defined. It should now return `out`, and line 73 above it would need to be changed to have something like `out = layer.get_output(0)`

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -127,9 +140,7 @@ def index(
         )
         index = adv_indx_indices[0]
         _LOGGER.debug(f"The advanced index indices is {adv_indx_indices}")
-        gather_layer = ctx.net.add_gather(input, indices_tensor, index)
-        set_layer_name(gather_layer, target, name + "_index_gather", source_ir)
-        return gather_layer.get_output(0)
+        return gather(input, index, indices_tensor)
```
gs-olive, line 143: This call is missing inputs (`ctx`, `target`, etc.)

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -242,11 +253,7 @@ def index(
                     dim_tensor_list[adv_indx_indices[i]],
                 )
 
-        gather_layer_element = ctx.net.add_gather(flatten_tensor, cum_adv_index, 0)
-        set_layer_name(
-            gather_layer_element, target, name + "_index_gather_element", source_ir
-        )
-        gather_out = gather_layer_element.get_output(0)
+        gather_out = gather(flatten_tensor, cum_adv_index, 0)
```
gs-olive, line 256: This call is missing inputs (`ctx`, `target`, etc.)

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -68,13 +68,26 @@ def select(
     indices_tensor = ctx.net.add_constant(
         index_value.shape, to_numpy(index_value)
     ).get_output(0)
-    layer = ctx.net.add_gather(input, indices_tensor, dim)
-    out = layer.get_output(0)
+    out = gather(input, indices_tensor, dim)
```
gs-olive, line 71: This call is missing inputs (`ctx`, `target`, etc.)

gs-olive (commented):
Added a few comments, and just needs some test cases for the new `gather` converter

apbose:
Had a doubt on this PR. In the gather implementation we explicitly convert the torch input to Tensors since the gather layer TensorRT would accept TRT Tensors. Is that needed as such? Also another question would be, suppose we feed in int64 ITensor for the test case directly, how do we deal with those cases? Does there need to be explicitly casting there? Eg cases like this -

gs-olive:
The Dynamo paths both use a truncation mechanism for int64 inputs which takes effect prior to the `TRTInterpreter` to avoid that issue. Does Int32 not work for the test cases?

apbose:
@gs-olive int32 in the test case leads to this error- RuntimeError: gather(): Expected dtype int64 for index

gs-olive:
Ok got it - could the test then mimic the scheme used here:
https://github.com/pytorch/TensorRT/blob/16c031349c6a1af5a8408a817f2ef8542aa6f176/tests/py/dynamo/backend/test_backend_compiler.py#L193
This uses `torch.compile` directly to allow the 64-bit repair code to run.

apbose:
Hi @gs-olive 
In the test case below:

gs-olive:
2 Test Cases [ITensor + constant input types]

- Indices as input [Differing Semantics between PyTorch and TensorRT --> PyTorch gets Int64, TRT gets Int32]
  - Known operator to assemble indices (i.e. `torch.arange`, or other constant-producing layer)
- Inlined constant indices test case as [Int64 Input Type]

```python
"""
graph(x, indices):
    --> inserted cast int64 to int32
    # If running in PyTorch cast back to int64
    out = aten.gather(x, 0, indices)

    return out
"""
```

## For a Later PR/Issue:
- Avoid running PyTorch graphs with invalid casts
- Refactor `repair_trunca...` to consume output of type inference
- Reorder truncation to come as late as possible in the compilation process

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -183,6 +183,31 @@ def aten_ops_native_group_norm(
...
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.select.gather(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        input=args[0],
+        dim=args[1],
+        index=args[2],
+        sparse_grad=args_bounds_check(args, 3, False),
```
gs-olive, line 207: Based on the schema [here](https://github.com/pytorch/pytorch/blob/4cdc2d7096ee030b35b666a70d48aeeee47388d0/aten/src/ATen/native/native_functions.yaml#L9095), `sparse_grad` would be in `kwargs`. Additionally, since it seems to have no effect in the `gather` converter below, it can be removed/ignored, or a validator can be used to ensure it is `False`.

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -127,9 +148,7 @@ def index(
         )
         index = adv_indx_indices[0]
         _LOGGER.debug(f"The advanced index indices is {adv_indx_indices}")
-        gather_layer = ctx.net.add_gather(input, indices_tensor, index)
-        set_layer_name(gather_layer, target, name + "_index_gather", source_ir)
-        return gather_layer.get_output(0)
+        return gather(ctx, target, source_ir, name, input, index, indices_tensor)
```
gs-olive, line 151: Line `126`/`147` needs to be renamed since it uses `name + f"_parameter_to_fp32_tensor"`, which also appears in the `gather` function. This could cause a duplicate name error in edge cases

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -68,11 +69,31 @@ def select(
...
+
+def gather(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: TRTTensor,
+    dim: int,
+    index: Union[TRTTensor, np.ndarray, torch.Tensor],
+    sparse_grad: bool = False,
```
gs-olive, line 83: See above, can remove if never used.

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -68,11 +69,31 @@ def select(
...
+) -> TRTTensor:
+    if not isinstance(index, TRTTensor):
+        index = get_trt_tensor(ctx, index, name + f"_parameter_to_fp32_tensor")
+    # This is for the case where torch.ops.aten.gather requires torch.int64
+    # However TRTInterpreter complains that torch.int64 is not a supported type
+    # So the below cast does not help
+    # index = cast_trt_tensor(ctx, input, trt.int32, name, target, source_ir)
```
gs-olive, line 90: Does this issue still occur in the test cases?

gs-olive (changes requested):

py/torch_tensorrt/dynamo/conversion/impl/select.py
apbose: Yes it does. aten.scatter has similar use cases, so I am working on that. The casting of nodes in the TRT test infrastructure can be used here to get over. This is the PR- https://github.com/pytorch/TensorRT/pull/2664.

Issue #2415, opened 2023-10-26
❓ [Question] Examples not working in nvcr.io/nvidia/pytorch:23.09-py3.


sayakpaul:
## ❓ Question

I am within the `nvcr.io/nvidia/pytorch:23.09-py3` container. Trying out some snippets from:
https://youtu.be/eGDMJ3MY4zk?si=MhkbgwAPVQSFZEha. 

Both JIT and AoT examples failed. For JIT, it complained that "tensorrt" backend isn't available, for AoT, it complained that "The user code is using a feature we don't support. Please try torchdynamo.explain() to get possible the reasons". 

I am on an A100. What's going on?

gs-olive:
Hello - for the `backend="tensorrt"` issue, please ensure the line `import torch_tensorrt` is in the script (even though it may not be used directly). This is important because the import itself registers the backend, as here:
https://github.com/pytorch/TensorRT/blob/0112bb4996d464f46ae8027f7d7146eb045d6364/py/torch_tensorrt/dynamo/backend/backends.py#L27
If this does not address the issue, could you please also share the version of Torch-TensorRT being used?

Regarding the AoT approach, @peri044 may be able to help further, but based on the error it sounds like the model may not be natively trace-able. Which specific example from the presentation are you referencing for the AoT sample?

sayakpaul:
This one: 
<img width="416" alt="image" src="https://github.com/pytorch/TensorRT/assets/22957388/ba6bea05-fb31-4121-a836-3e360c15b977">

peri044:
Hello @sayakpaul , I would suggest you to try our latest main branch. 23.09 is old now. Let us know if you see any issues with the main branch. 
Alternatively, you can try our nightly container if you don't want to build from source : https://github.com/pytorch/TensorRT/pkgs/container/tensorrt%2Ftorch_tensorrt

sayakpaul:
With the latest container (https://github.com/pytorch/TensorRT/pkgs/container/tensorrt%2Ftorch_tensorrt), it works. Have a few questions. So, let me use this thread to ask them. Maybe the community might find them to be useful.

sayakpaul:
I used the following code snippet I am using to benchmark the JIT workflow:

```python
import torch
import torch_tensorrt
import torch.utils.benchmark as benchmark

class MyModel(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.conv = torch.nn.Conv2d(64, 128, 3)
[...]

PR #2464, opened 2023-11-14
fix: Error with `aten.view` across Tensor memory

gs-olive:
# Description

- Address error where `aten.view` is called on TRT output Tensors, which can be in a different memory format than Torch expects
- Specifically, TRT can modify tensor memory to optimize certain layers, but Torch's view operator depends on specific configurations which can be violated at runtime (but not at compile time, since Torch itself would run these configurations correctly)
- Add a custom lowering pass to replace `view` with `reshape`, avoiding this issue. Reshape will make a copy of the underlying Tensor if necessary
- Torch-TRT's `aten.view` implementation is the same as that for `aten.reshape`, and they share a schema so no changes are needed on the converter side
- Add test case to validate new lowering pass

Addresses #2415

## Type of change

# Checklist:


zewenli98 (approved):
This PR looks good to me. Just curious, is this problem that Pytorch would do something after converting ops or after creating TRT Engines (because I think `aten.reshape.default` and `aten.view.default` are exactly same)?

gs-olive:
Thanks for the review @zewenli98 - the main issue here is that when an `aten.view.default` operator falls into a PyTorch segment due to partitioning restrictions (like a high `min_block_size`), the requirements for the memory layout of the input tensor can be invalidate by TensorRT. Specifically, the PyTorch `aten.view` op can make some assumptions about the memory format (contiguous,  strided, etc.) based on the flow of the tensor in the graph preceding it, but when that graph gets run in TensorRT, those assumptions can be invalidated, causing the error:
```python
RuntimeError: view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead.
```

zewenli98:
Aha I see, so does this mean we can omit `view` because `reshape` can deal with more scenarios than `view`? or say whether the scenarios where `view` can be used could replace with `reshape`?

gs-olive:
@zewenli98 - I believe the cases that `aten.view` can handle are a strict subset of those `aten.reshape` can handle, because reshape should be copying the Tensor memory only if needed (otherwise, should behave like `aten.view`)

PR #2466, opened 2023-11-14
fix: Wrap perf benchmarks with no_grad

gs-olive:
- Benchmark script fix
- `torch.compile` requires `no_grad` even during the initial compilation phase to avoid adding backward-pass instrumentation to the model

### DIFF
--- a/tools/perf/perf_run.py
+++ b/tools/perf/perf_run.py
@@ -593,8 +593,10 @@ if __name__ == "__main__":
     # Generate report
     print("Model Summary: ", model_name)
     summary = pd.DataFrame(results)
-    summary["model_name"] = (
-        model_name_torch if model_name_torch is not None else model_name
+    summary.insert(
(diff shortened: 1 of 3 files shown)


tools/perf/perf_run.py
```diff
@@ -578,23 +578,24 @@ def recordStats(backend, timings, precision, batch_size=1, compile_time_s=None):
...
+                precision,
+                batch_size,
+                is_trt_engine,
+                model_torch=model_torch,
+            )
 
     # Generate report
     print("Model Summary: ", model_name)
     summary = pd.DataFrame(results)
+    summary["model_name"] = (
```
peri044, line 596: a minor comment - Does model_name get appended as the last column here? 
Instead can we have 
model_name, Batch size, Precision .....  ?

peri044 (commented):
LGTM

PR #2467, opened 2023-11-14
feat: add `convert_method_to_trt_engine()` for dynamo

### LINKED ISSUE #2410: Implenting torch_tensorrt.dynamo.convert_method_to_trt_engine

zewenli98:
# Description

Add a function `convert_method_to_trt_engine()` for dynamo.

I'm not familiar with this part, so detailed reviews and suggestions are greatly welcome.

Fixes #2410 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/_compile.py
+++ b/py/torch_tensorrt/_compile.py
@@ -319,7 +319,7 @@ def convert_method_to_trt_engine(
             "convert_method_to_trt_engine call is not supported for ir=fx"
         )
     elif target_ir == _IRType.dynamo:
-        return torch_tensorrt.dynamo.convert_method_to_trt_engine(  # type: ignore[no-any-return]
+        return torch_tensorrt.dynamo.convert_module_to_trt_engine(  # type: ignore[no-any-return]
             module,
             inputs=inputs,
             method_name=method_name,
--- a/py/torch_tensorrt/_compile.py
+++ b/py/torch_tensorrt/_compile.py
@@ -319,8 +319,12 @@ def convert_method_to_trt_engine(
             "convert_method_to_trt_engine call is not supported for ir=fx"
         )
     elif target_ir == _IRType.dynamo:
-        raise RuntimeError(
-            "convert_method_to_trt_engine call is not supported for ir=dynamo."
+        return torch_tensorrt.dynamo.convert_method_to_trt_engine(  # type: ignore[no-any-return]
+            module,
+            inputs=inputs,
+            method_name=method_name,
+            enabled_precisions=enabled_precisions_set,
+            **kwargs,
         )
     elif target_ir == _IRType.torch_compile:
         raise RuntimeError(
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -50,7 +50,7 @@ from torch_tensorrt.dynamo.conversion import (
     CompilationSettings,
     UnsupportedOperatorException,
     convert_module,
-    interpret_module,
+    interpret_module_to_result,
     repair_long_or_double_inputs,
 )
 from torch_tensorrt.dynamo.conversion._ConverterRegistry import (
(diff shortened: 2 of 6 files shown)


py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -319,3 +325,122 @@ def compile_module(
...
+    compilation_options = {
+        "precision": precision,
+        "debug": debug,
+        "device": device,
+        "workspace_size": workspace_size,
+        "truncate_long_and_double": truncate_long_and_double,
+        "max_aux_streams": MAX_AUX_STREAMS,
+        "version_compatible": VERSION_COMPATIBLE,
+        "optimization_level": OPTIMIZATION_LEVEL,
+    }
```
gs-olive, line 434: This should use the same set of compilation options as here, perhaps omitting `use_python_runtime` and `use_fast_partitioner` which would not be necessary in the full-graph conversion case:
https://github.com/pytorch/TensorRT/blob/867dc7bf23b34f8d3809a1d64841b0013a9a0054/py/torch_tensorrt/dynamo/_compiler.py#L177-L195

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -319,3 +325,122 @@ def compile_module(
...
+
+    settings = CompilationSettings(**compilation_options)
+    logger.info("Compilation Settings: %s\n", settings)
+    interpreter_result = interpreter(module, input_list, settings, method_name)
```
gs-olive, line 438: On this line, use a `try`/`except` seeking this error:
https://github.com/pytorch/TensorRT/blob/867dc7bf23b34f8d3809a1d64841b0013a9a0054/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py#L37-L38
Then, you can return the user a message indicating the model was not fully supported or convertible, along with the exception info containing the reasoning

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -319,3 +325,122 @@ def compile_module(
         settings.use_fast_partitioner = True
 
     return partitioned_module
+
+
+def interpreter(
```
narendasan, line 330: Chose a better name for this function, this is not very descriptive. Function names should be a "verb" that conveys the action being performed by the function

py/torch_tensorrt/dynamo/_compiler.py
zewenli98: Thanks for the suggestion! Yes I did like this at first, but many arguments like `min_block_size, torch_executed_ops, pass_through_build_failures, etc.` don't show up in the function. Naren told me to use the function declaration from torchscript. So do you think we need to add more args in the declaration and remove those useless args?

gs-olive: I see - the main concern is that `max_aux_streams`, `optimization_level`, and `version_compatible` options are all attributes of the TRT Engine itself (set in the `TRTInterpreter`), so the user should be able to customize these instead of just using the defaults.

zewenli98: I agree with you! Maybe we can expose some/all args in the function declaration for users allowing them to specify values?

zewenli98:
@gs-olive @narendasan Another question - Regarding testing, how to invoke this function like directly calling `torchtrt.dynamo.convert_method_to_trt_engine`? and where should I put the test cases?

py/torch_tensorrt/dynamo/_compiler.py
gs-olive: I think just those three arguments suffice to expose in the function header

gs-olive:
I think it could potentially go in `tests/py/dynamo/runtime`? Then, you can directly call the function to generate the TRT engine which can be run directly in Python for inference purposes.

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -3,34 +3,26 @@
...
-        settings: Compilation settings
-        name: TRT engine name
-    Returns:
-        _PythonTorchTensorRTModule or TorchTensorRTModule
-    """
-    # Specify module output data types to ensure TRT output types agree with
-    # that of the equivalent Torch module
+) -> TRTInterpreterResult:
     torch_inputs = get_torch_inputs(inputs, settings.device)
+    module.to(to_torch_device(settings.device))
```
gs-olive, line 25: Is this line necessary? If it is only required for `convert_method_to_trt_engine`, then could it instead be called in `py/torch_tensorrt/_compile.py`?

zewenli98: It's necessary because module and inputs should stay in the same device. I think we can let this line stay here because `convert_module` also needs to call it, and the `device` is read from `settings`. Same reason for `torch_inputs = get_torch_inputs(inputs, settings.device)`

zewenli98:
Got a Python library import error:
```
==================================== ERRORS ====================================
_ ERROR collecting tests/py/dynamo/runtime/test_convert_method_to_trt_engine.py _
ImportError while importing test module '/__w/TensorRT/TensorRT/pytorch/tensorrt/tests/py/dynamo/runtime/test_convert_method_to_trt_engine.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/python/cp38-cp38/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
runtime/test_convert_method_to_trt_engine.py:4: in <module>
    import pycuda.driver as cuda
E   ModuleNotFoundError: No module named 'pycuda'
```
Can we add `pycuda` in CI?

zewenli98:
> Hi @zewenli98 - sure, I think we can add this dependency at the path [`tests/modules/requirements.txt`](https://github.com/pytorch/TensorRT/blob/867dc7bf23b34f8d3809a1d64841b0013a9a0054/tests/modules/requirements.txt), and it should be reflected in the CI afterward.

Thanks! Naren suggested that `PythonTorchTensorRTModule` should be used instead of directly calling native functions, so I modified accordingly.

py/torch_tensorrt/_compile.py
```diff
@@ -319,8 +319,12 @@ def convert_method_to_trt_engine(
             "convert_method_to_trt_engine call is not supported for ir=fx"
         )
     elif target_ir == _IRType.dynamo:
-        raise RuntimeError(
-            "convert_method_to_trt_engine call is not supported for ir=dynamo."
+        return torch_tensorrt.dynamo.convert_method_to_trt_engine(  # type: ignore[no-any-return]
```
narendasan, line 322: Can we get the type annotations to work here?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -319,3 +322,91 @@ def compile_module(
...
+    dla_sram_size: int = 1048576,
+    dla_local_dram_size: int = 1073741824,
+    dla_global_dram_size: int = 536870912,
+    truncate_long_and_double: int = False,
+    calibrator: object = None,
+    allow_shape_tensors: bool = False,
+    max_aux_streams: Optional[int] = MAX_AUX_STREAMS,
+    version_compatible: bool = VERSION_COMPATIBLE,
+    optimization_level: Optional[int] = OPTIMIZATION_LEVEL,
+) -> bytes:
```
narendasan, line 349: We need docstrings for this API

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -3,34 +3,26 @@
...
+)
 from torch_tensorrt.dynamo.runtime import PythonTorchTensorRTModule, TorchTensorRTModule
-from torch_tensorrt.dynamo.utils import get_torch_inputs
-
-import tensorrt as trt
+from torch_tensorrt.dynamo.utils import get_torch_inputs, to_torch_device
 
 
-def convert_module(
+def interpret_module(
```
narendasan, line 18: Call this convert_module_to_engine or something, I am not sure what the difference between convert and interpret is.

py/torch_tensorrt/dynamo/_compiler.py
zewenli98: OK sounds good, and do you think we need to remove the arguments that are not used within the function?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -319,3 +322,91 @@ def compile_module(
         settings.use_fast_partitioner = True
 
     return partitioned_module
+
+
+def convert_method_to_trt_engine(
```
gs-olive, line 327: - Rename to `convert_module_to_trt_engine`
- In `torch_tensorrt._compile` - limit usage of `ir="dynamo"` for non-forward method names
- `convert_module_to_trt_engine` in `._compile`, `convert_module_to_trt_engine` in `dynamo._compiler.py`, and `convert_method_to_trt_engine`

gs-olive: Should this still be `convert_module_to_trt_engine` @narendasan?

gs-olive: Use `dynamo.convert_module_to_trt_engine`, but keep top-level function as `convert_method_to_trt_engine` above Dynamo/TS scope (`_compiler.py`)

zewenli98: > Use `dynamo.convert_module_to_trt_engine`, but keep top-level function as `convert_method_to_trt_engine` above Dynamo/TS scope (`_compiler.py`)

Do we need to change `ts.convert_method_to_trt_engine` to `ts.convert_module_to_trt_engine` accordingly?

gs-olive: I think that should not be necessary, since the `ts` variant can convert arbitrary methods.

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -342,3 +346,189 @@ def compile_module(
...
+
+def convert_module_to_trt_engine(
+    module: torch.fx.GraphModule,
+    method_name: str = "forward",
+    inputs: Optional[Sequence[Input | torch.Tensor]] = None,
+    enabled_precisions: Optional[Set[torch.dtype | _enums.dtype]] = None,
+    debug: bool = DEBUG,
+    workspace_size: int = WORKSPACE_SIZE,
+    min_block_size: int = MIN_BLOCK_SIZE,
+    torch_executed_ops: Set[str] = field(default_factory=set),
```
gs-olive, line 359: I am not sure if `field` is used outside of the context of an explicit `dataclass`, so this could alternatively just be `set()` or adapt from what is currently used in the other functions with this argument.

gs-olive (commented):
Overall looks good. Added one suggestion and needs a rebase.

PR #2469, opened 2023-11-15
feat: introduce type propagation infrastructure

bowang007:
# Description

This PR introduces type propagation infrastructure through pytorch inference.
Inferred output types are used to set the TensorRT engines output types.

Related PR: #1853 

Previous solution has several limitations. 

- Bug fix 
- New feature 

# Checklist:

### DIFF
--- a/core/conversion/conversion.cpp
+++ b/core/conversion/conversion.cpp
@@ -211,7 +211,11 @@ void AddInputs(ConversionCtx* ctx, c10::ArrayRef<const torch::jit::Value*> input
 #endif
 }
 
-void MarkOutputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> outputs) {
+void MarkOutputs(
+    ConversionCtx* ctx,
+    at::ArrayRef<const torch::jit::Value*> outputs,
+    const std::vector<at::ScalarType>& out_types) {
+  size_t out_idx = 0;
   for (auto out : outputs) {
     auto it = ctx->value_tensor_map.find(out);
     if (it == ctx->value_tensor_map.end()) {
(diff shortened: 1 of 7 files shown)


core/partitioning/shape_analysis.cpp
```diff
@@ -340,6 +340,19 @@ void getSegmentsOutputByRunning(
 
   seg_block.register_inshapes(input_shapes, shape_mode);
   seg_block.register_intypes(input_types);
+
+  // get output type for each segmented block so this can be used in conversion process
+  std::vector<at::ScalarType> output_types;
```
mfeliz-cruise, line 345: One thing to check: Because we're running this at the end of shape propagation does this work for fully convertible modules?

From memory I think that shape prop is only called form BuildHybridGraph->partition() which I don't think is called if all ops are convertible.

core/conversion/conversion.cpp
```diff
@@ -265,6 +269,7 @@ void MarkOutputs(ConversionCtx* ctx, at::ArrayRef<const torch::jit::Value*> outp
       }
 
       if (!setOutput) {
+        out_tensor->setType(util::ScalarTypeToTRTDataType(out_types[out_idx++]));
```
mfeliz-cruise, line 272: The out_idx needs to be incremented even if one of the outputs is an input (see where setOutput may be set above)

gs-olive (commented):
Overall, looks good, just needs a few small fixes. Needs a test case for verification. Additionally, it seems some of the Torchscript test cases are failing due to this change:
```
ERROR conda.cli.main_run:execute(41): `conda run python -m pytest --junitxml=/tmp/test_results/ts_api_test_results.xml api/` failed. (See above for error)
Fatal Python error: Segmentation fault

Current thread 0x00007f4711145740 (most recent call first):
  File "/opt/python/cp38-cp38/lib/python3.8/site-packages/torch_tensorrt/ts/_compiler.py", line 266 in convert_method_to_trt_engine
  File "/__w/TensorRT/TensorRT/pytorch/tensorrt/tests/py/ts/api/test_classes.py", line 300 in test_detect_invalid_input_binding
```
This may be due to the reason in the comment above https://github.com/pytorch/TensorRT/pull/2469#discussion_r1393529898

PR #2473, opened 2023-11-16
feat: Implement symbolic shape propagation, sym_size converter

peri044:
# Description

This PR replaces our old shape propagation system which relied on dummy runs of inputs. If the inputs were dynamic, we used to run dummy inference on min, opt, max shapes to capture output shapes/dtypes. 

For data-dependant shapes (with static input), our shape propagation system might not work and would give incorrect info about subgraph input shapes. This PR uses symbolic shape propagation from PyTorch. 

Eg:

```py
class DDS(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x, mask):
        return x[mask]

model = DDS().eval().cuda()
x = torch.randn(1, 3, 4, 4).cuda()
y = torch.rand((1, 3, 4, 4), device="cuda") < 0.9

trt_model = torch_tensorrt.compile(model, inputs=[x, y], 
                                    ir="dynamo",
                                    min_block_size=1, 
                                    debug=True)
```
Example 2: 
```py
class Sample(torch.nn.Module):
    def __init__(self):
        super(Sample, self).__init__()

    def forward(self, x):
        input_shape = x.size()
        y = x.view(input_shape[0], -1)
        return y
```

RFC : https://github.com/pytorch/TensorRT/discussions/2409

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -39,6 +39,12 @@ def convert_module(
     # such as aten.sum - such outputs can be truncated
     output_dtypes = []
     for output in module_outputs:
+        if not isinstance(output, torch.Tensor):
+            output = torch.tensor(output)
+            if isinstance(output, int):
+                output = output.to(torch.int32)
+            elif isinstance(output, float):
+                output = output.to(torch.float32)
```
gs-olive, line 47: These conditionals would not be entered since `output = torch.tensor(output)` runs immediately before them, so they would be an instance of `torch.Tensor`. Consider switching to:
```python
for output in module_outputs:
    if not isinstance(output, torch.Tensor):
        output = torch.tensor(output, dtype=(torch.float if isinstance(output, float) else torch.int32))
```

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -392,6 +392,25 @@ def aten_ops_sigmoid(
...
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.shape.shape(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args_bounds_check(args, 1, None),
```
gs-olive, line 409: This should be `args[1]`, since the schema, [here](https://github.com/pytorch/pytorch/blob/0978482afa69118da1a986a4edec3acea01d2c6d/aten/src/ATen/native/native_functions.yaml#L5324), requires that this argument be specified and an integer.

py/torch_tensorrt/dynamo/conversion/impl/shape.py
```diff
@@ -8,14 +8,37 @@
...
+) -> TRTTensor:
+    """
+    This is the general shape layer implementation in TensorRT.
+    sym_size.int ops map to addShape layer in TensorRT and returns
+    the dynamic shape of the tensor optionally taking in a dim argument.
+    """
+    input_shape = ctx.net.add_shape(input_val).get_output(0)
+    if not dim:
+        max_dim = len(input_val.shape)
+        dim = dim if dim > 0 else dim + max_dim
```
gs-olive, line 35: This can be removed, since the converter should not allow `None` dimension specification for `sym_size.int`

py/torch_tensorrt/dynamo/conversion/impl/shape.py
```diff
@@ -8,14 +8,37 @@
...
+    indices = get_trt_tensor(ctx, dim, name + "_dim")
+    gather_dim = ctx.net.add_gather(input_shape, indices, axis=0).get_output(0)
+
```
gs-olive, line 38: Consider adding `set_layer_name(gather_dim, target, name + "_gather", source_ir)` here.

py/torch_tensorrt/dynamo/partitioning/common.py
```diff
@@ -9,6 +9,78 @@
...
+        if isinstance(dim, torch.SymInt):
+            node = dim.node
+            expr = node.expr
+            shape_env = node.shape_env
+            var_range = shape_env.var_to_range.get(expr, None)
+            var_val = shape_env.var_to_val.get(expr, None)
+            assert var_range, var_val
+            # Torchdynamo 0/1 specialization outlier
+            if var_range.lower == 2:
+                min_shape.append(1)
```
gs-olive, line 44: Does this imply that we cannot differentiate between a dynamic dimension being 1 or 2, as specified by the user? Specifically, if the user specifies the min/opt/max shape ranges `(1, 3, 224, 224)`, `(1, 3, 224, 224)`, `(5, 3, 224, 224)`, is this indistinguishable from `(2, 3, 224, 224)`, `(2, 3, 224, 224)`, `(5, 3, 224, 224)`?

gs-olive (commented):
Overall looks good - needs some test cases for validation. I added a few comments and suggestions throughout.

py/torch_tensorrt/dynamo/conversion/impl/shape.py
peri044: fixed in the latest update

py/torch_tensorrt/dynamo/conversion/impl/shape.py
peri044: fixed in the latest update

py/torch_tensorrt/dynamo/conversion/impl/shape.py
```diff
@@ -8,14 +8,37 @@
...
+) -> TRTTensor:
+    """
+    This is the general shape layer implementation in TensorRT.
+    sym_size.int ops map to addShape layer in TensorRT and returns
+    the dynamic shape of the tensor optionally taking in a dim argument.
+    """
+    input_shape = ctx.net.add_shape(input_val).get_output(0)
+    if not dim:
+        max_dim = len(input_val.shape)
+        dim = dim if dim >= 0 else dim + max_dim
```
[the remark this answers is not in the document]
peri044: fixed in the latest update

py/torch_tensorrt/dynamo/partitioning/common.py
peri044: Yes

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -402,6 +413,7 @@ def index_dtype_validator(node: Node) -> bool:
     return True
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.index.Tensor)
```
gs-olive, line 416: If this PR removes the `index_dtype_validator` constraint on the `index` operator, remove the below line to avoid duplicate registrations of the same operator.

py/torch_tensorrt/dynamo/partitioning/common.py
```diff
@@ -9,6 +9,80 @@
...
+            input_meta = input.meta["val"]
+            input_shape = input_meta.size()
+            if contains_sym_int(input_shape):
+                torchtrt_inputs.append(construct_dynamic_input(input_meta))
+            else:
+                torchtrt_inputs.append(Input(shape=input_shape, dtype=input_meta.dtype))
+        else:
+            raise AssertionError(
+                f"Input {input.name} does not contain metadata. Please ensure you have exported the graph correctly"
+            )
```
gs-olive, line 81: Is it true that `input.meta and "val" in input.meta` is only needed in the Dynamic shape case? If so, consider restructuring this code to _only_ throw an error if the inputs are dynamic _and_ the `"val"` key is missing, to avoid throwing errors in the static shape case if not necessary.

gs-olive (commented):
Overall looks good - added some small questions and considerations

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -302,6 +301,24 @@ def compile_module(
             f"Detected support for {num_supported_ops} operators out of {total_ops} in subgraph."
         )
 
+    def contains_metadata(gm: torch.fx.GraphModule) -> bool:
+        for node in gm.graph.nodes:
+            if node.op != "output" and (not node.meta) and "val" not in node.meta:
```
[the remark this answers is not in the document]
peri044: currently I'm relying on `"val"` key in the node.meta which has the fake tensors with symbolic shape/static shapes.
There's also `example_tensor` key sometimes but since I'm not using it, `and "val" not in node.meta` condition will make it strict.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -392,6 +392,17 @@ def aten_ops_sigmoid(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.sym_size.int)
```
gs-olive, line 395: Consider adding the following above:
https://github.com/pytorch/TensorRT/blob/6b8ccbc69f05de3b6275cac4613735934fb559b0/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py#L62-L66

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -270,7 +270,7 @@ def create_constant(
     """
     numpy_value = to_numpy(value, dtype)
     constant = ctx.net.add_constant(
-        (1,) if isinstance(value, (int, float, bool)) else value.shape,
+        trt.Dims() if isinstance(value, (int, float, bool)) else value.shape,
```
gs-olive, line 273: Is it problematic for a single-scalar tensor to have a dimension `(1,)`?

py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
```diff
@@ -99,9 +98,13 @@ def expand(
...
         [int(i == o) for i, o in zip(input_tensor_shape, shape)]
     )  # stride == 1 if dimensions match, 0 otherwise
-    layer = ctx.net.add_slice(input_t, start=start, shape=shape, stride=stride)
-    set_layer_name(layer, target, name, source_ir)
-    return layer.get_output(0)
+    # layer = ctx.net.add_slice(input_t, start=start, shape=shape, stride=stride)
+
+    # set_layer_name(layer, target, name, source_ir)
+    # return layer.get_output(0)
+    breakpoint()
```
gs-olive, line 105: Debug statement to remove

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -46,4 +46,14 @@ def aten_ops_arange_start_step(
...
+    # breakpoint()
+    fill_layer = ctx.net.add_fill(trt.Dims(), trt.FillOperation.LINSPACE)
+    fill_layer.set_input(0, args[1])
+    fill_layer.set_output_type(0, trt.DataType.INT32)
+    # fill_layer.set_input(1, 0)
+    # fill_layer.set_input(2, 1)
+    # start_tensor = get_trt_tensor(ctx, 0, "_start_tensor")
+    # fill_layer.set_input(1, start_tensor)
+    # delta_tensor = get_trt_tensor(ctx, torch.tensor([0], dtype=torch.int32), "_delta_tensor")
+    # fill_layer.set_input(2, delta_tensor)
```
gs-olive, line 58: Debug logs to remove

PR #2482, opened 2023-11-23
feat: support for many padding dynamo converters

### LINKED ISSUE #2461: Add support for padding aten operators
```[tasklist]
### Tasks
- [x] torch.ops.aten.pad.default
- [x] torch.ops.aten.constant_pad_nd.default
- [x] torch.ops.aten.reflection_pad1d.default
- [x] torch.ops.aten.reflection_pad2d.default
- [x] torch.ops.aten.reflection_pad3d.default
- [x] torch.ops.aten.replication_pad1d.default
- [x] torch.ops.aten.replication_pad2d.default
- [x] torch.ops.aten.replication_pad3d.default
- [x] torch.ops.aten._pad_circular.default
```

zewenli98:
# Description

Support for many padding dynamo converters, including:

```
torch.ops.aten.pad.default
torch.ops.aten.constant_pad_nd.default
torch.ops.aten.reflection_pad1d.default
torch.ops.aten.reflection_pad2d.default
torch.ops.aten.reflection_pad3d.default
torch.ops.aten.replication_pad1d.default
torch.ops.aten.replication_pad2d.default
torch.ops.aten.replication_pad3d.default
torch.ops.aten._pad_circular.default
```

Note: 
1. Not sure why there're no similar names for `torch.ops.aten._pad_circular.default` like `torch.ops.aten.circular_pad1d.default`, but according to [PyTorch doc](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html#torch.nn.functional.pad), it does have the similar functionality as `reflection` and `replication`, targeting for 2D, 3D, 4D, and 5D inputs.
2. `torch.ops.aten.pad.default` is a higher-level function to call the other padding functions.

Fixes #2461 #2211 

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
```diff
@@ -37,10 +37,6 @@ def slice_op(  # TODO: This should be slice not whatever is in base
             "of the TensorRT region!"
         )
 
-    dim = get_positive_dim(dim, len(input.shape))
-    start = get_positive_dim(start, input.shape[dim])
-    stop = get_positive_dim(stop, input.shape[dim])
-
```
zewenli98, line 43: I did this change because in some cases I want to use `slice_op` to get a reserved tensor but the `stop` arg cannot be negative. For example, when doing:
```
impl.slice.slice_op(
                ctx,
                target,
                source_ir,
                f"{name}_slice_post{i}",
                input,
                dim=dim,
                start=2,
                stop=-1,
                step=-1,
            )
```
if has these lines, `stop` would be a positive value (i.e. `length - 1`), but since `step` is still negative, it will never stop, however, my expected sequence is `[2, 1, 0]`. 

The reason why I don't use positive `stop` and `step` is I wanna get a reversed tensor.

The unit tests proved these lines are not necessary.

py/torch_tensorrt/dynamo/conversion/impl/pad.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # Implement constant padding via concat
+    curr_dim = len(input.shape) - 1
+
+    for i in range(0, len(pad), 2):
+        input_shape = list(input.shape)
+
+        pre_pad = pad[i]
+        post_pad = pad[i + 1]
+        pre_pad_shape = copy.deepcopy(input_shape)
```
gs-olive, line 38: Is the `list` deepcopy required here? Could we instead use `tuple`s to avoid the need for `copy.deepcopy`?

py/torch_tensorrt/dynamo/conversion/impl/pad.py
```diff
@@ -0,0 +1,285 @@
...
+        pre_pad_shape[curr_dim] = pre_pad
+        pre_pad_tensor = torch.full(pre_pad_shape, float(value))
```
gs-olive, line 40: Due to the `FakeTensor` issue in `torch.compile`, could this instead be a Numpy array?

py/torch_tensorrt/dynamo/conversion/impl/pad.py
```diff
@@ -0,0 +1,285 @@
...
+    name: str,
+    input: TRTTensor,
+    padding: Sequence[int],
+) -> TRTTensor:
+    if has_dynamic_shape(input.shape):
+        assert input.shape[1] != -1, "Channel dim can't be dynamic for padding."
+
+    padding_dims = len(padding) // 2
+
+    if padding_dims == 1 or padding_dims == 2 or padding_dims == 3:
```
gs-olive, line 74: Consider condensing to `if padding_dims in (1, 2, 3)`

py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
gs-olive: The issue with removing this is that the line `output_shape[dim] = math.ceil((stop - start) / step)` below will be incorrect when certain values are negative. This change would have to be accompanied by a fix for that equation, since these test cases should otherwise fail:
https://github.com/pytorch/TensorRT/blob/867dc7bf23b34f8d3809a1d64841b0013a9a0054/tests/py/dynamo/conversion/test_slice_aten.py#L16-L17

zewenli98: Thanks for the comments! I undid the change due to better implementation.

zewenli98:
Hi @gs-olive, thanks for your review! I just found out a better way to implement paddings, with the help of `ISliceLayer` with different `SliceMode`. So I undid the change to `slice_op`.

py/torch_tensorrt/dynamo/conversion/impl/pad.py
```diff
@@ -1,14 +1,22 @@
...
 
-import torch
-import torch_tensorrt.dynamo.conversion.impl as impl
+import tensorrt as trt
 from torch.fx.node import Target
 from torch_tensorrt.dynamo._SourceIR import SourceIR
 from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
-from torch_tensorrt.fx.converters.converter_utils import has_dynamic_shape
+from torch_tensorrt.fx.converters.converter_utils import (
+    get_trt_tensor,
```
gs-olive, line 8: This should use the `dynamo.converters.converter_utils` implementation of `get_trt_tensor`

py/torch_tensorrt/dynamo/conversion/impl/pad.py
```diff
@@ -19,43 +27,36 @@ def constant_padNd(
...
-        output = impl.cat.cat(
-            ctx,
-            target,
-            source_ir,
-            f"{name}_concat{curr_dim}",
-            input=(pre_pad_tensor, input, post_pad_tensor),
-            dim=curr_dim,
+    rank = len(input.shape)
+
+    if len(pad) / 2 > rank:
```
gs-olive, line 35: Should this be a `//` instead of `/`?

zewenli98: Both are fine because `len(pad)` should be even. BTW, if PyTorch checks some restrictions like `len(pad) should be even`, do we need to check again within our converters?

gs-olive: It shouldn't be necessary if PyTorch checks it, but can't hurt to add the check

gs-olive: Additionally, `//` is preferred if the results shouldn't matter, since `//` will output an integer

py/torch_tensorrt/dynamo/conversion/impl/pad.py
```diff
@@ -0,0 +1,205 @@
...
+
+    rank = len(input.shape)
+
+    if len(pad) // 2 > rank:
+        raise RuntimeError(
+            f"Trying to pad last {len(pad) // 2} dimension but the input only has {rank} dimension."
+        )
+
+    start_list = [0] * len(input.shape)
+    new_shape = input.shape
```
gs-olive, line 157: Since `new_shape` gets assigned to below, I think this might need to be `list(input.shape)`, to ensure it is casted to a mutable list.

gs-olive: This applies to the other cases above as well. It may not be required with TRT's current implementation of `input.shape`, but it would be preferable in general to have this cast to be safe.

py/torch_tensorrt/dynamo/conversion/impl/pad.py
```diff
@@ -0,0 +1,205 @@
...
+
+    for i in range(0, len(padding) // 2):
+        start_list[-i - 1] = -padding[i * 2]
+        new_shape[-i - 1] += padding[i * 2] + padding[i * 2 + 1]
+
+    stride_list = [1] * len(new_shape)
```
gs-olive, line 125: For clarity, could `len(input.shape)` on line 118 and `len(new_shape)` on line 125 both be replaced with `rank`, or would this make the code incorrect?

The above applies to a few other functions as well I believe.

gs-olive (commented):
Overall looks great, added a few style comments and suggested modifications.

py/torch_tensorrt/dynamo/conversion/impl/pad.py
zewenli98: That makes sense, thanks for the suggestion!

gs-olive (approved):
Looks good to me!

PR #2487, opened 2023-11-28
fix: aten.index converter

### LINKED ISSUE #2480: 🐛 [Bug] aten.index converter doesn't work when index is a TRTTensor
##  Bug Description

aten index converter fails for data dependent shape operator. If we try to get the  output shape of aten.index converter, it gives an error message as follows invalidating the INetwork. 

To reproduce, use the https://github.com/pytorch/TensorRT/tree/dyn_2.2 branch as it has the implementation for sym_int converter which is required to run the reproducer.

Error: 
```
gather_layer.get_output(0).shape # Print the output shape of this index converter
[11/21/2023-00:33:07] [TRT] [E] 3: [GATHER]-[aten_ops.index.Tensor]-[__/index_index_gather]: only kINT32 allowed for input 1 to this layer.
(1, 3, 4, 4, 3, 4, 4)
```

## To Reproduce

```py
import torch
import torch_tensorrt

class DDS(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.relu = torch.nn.ReLU()

    def forward(self, x, mask):
        out = x[mask]
        out = self.relu(out)
        return out

model = DDS().eval().cuda()
x = torch.randn(1, 3, 4, 4).cuda()
y = torch.rand((1, 3, 4, 4), device="cuda") < 0.9

trt_model = torch_tensorrt.compile(model, inputs=[x, y], 
                                    ir="dynamo",
                                    min_block_size=1, 
                                    debug=True)
```

1.
2.
3.

## Environment

 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - Python version:
 - CUDA version:

zewenli98:
# Description

See details in the issue below.

Fixes #2480 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -366,7 +366,12 @@ def aten_ops_sigmoid(
 
 def index_dtype_validator(node: Node) -> bool:
     index = node.args[1]
-    return all(ind.meta["val"].dtype == torch.int32 for ind in index if ind is not None)
+    for ind in index:
+        if ind is not None:
+            val = ind.meta.get("val")
+            if val is not None and val.dtype != torch.int32:
+                return False
+    return True
 
 
 @dynamo_tensorrt_converter(
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -364,7 +364,14 @@ def aten_ops_sigmoid(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.index.Tensor)
+def index_dtype_validator(node: Node) -> bool:
+    index = node.args[1]
+    return all(ind.meta["val"].dtype == torch.int32 for ind in index if ind is not None)
+
+
+@dynamo_tensorrt_converter(
+    torch.ops.aten.index.Tensor, capability_validator=index_dtype_validator
+)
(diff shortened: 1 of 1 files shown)


apbose (approved):
Looks good!

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -364,7 +364,14 @@ def aten_ops_sigmoid(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.index.Tensor)
+def index_dtype_validator(node: Node) -> bool:
+    index = node.args[1]
+    return all(ind.meta["val"].dtype == torch.int32 for ind in index if ind is not None)
```
gs-olive, line 369: Use `get` for dictionary

gs-olive (approved):
It seems that the bug in #2480 is only related to the fact that our `index` converter cannot handle boolean mask indices. The fact that the example has data-dependent shapes does not seem to be the root cause of the initial error.

This is a good temporary fix. Could you also file an issue which specifies that we need to expand `index` support to boolean masks, since this is a valid usage in Torch and we already support something very similar in `aten.where`, we just need to add the functionality here. See the sample case below:
```python
>>> import torch
>>> y = torch.rand(2, 2)
>>> y
tensor([[0.3105, 0.9580],
        [0.4824, 0.0796]])
>>> y[torch.Tensor([[True, True], [False, False]]).bool()]
tensor([0.3105, 0.9580])
```

Issue #2219, opened 2023-08-14
Expose IResizeLayer in dynamo.conversion.impl


narendasan:

bowang007:
`torch.ops.aten.resize.default` is doing something like resizing the data's memory instead of what we expect for resize a image. 
For resizing an input from `[4, 4]` to `[2, 2]`:
```input:  [tensor([[-0.0766,  0.3599, -0.7820,  0.0715],
        [ 0.6648, -0.2868,  1.6206, -1.5967],
        [-0.0517, -0.3060,  0.2485, -0.2226],
        [ 0.9132,  0.2043,  0.5740,  0.4163]])]
trt:  [tensor([[-0.0766,  0.0715],
        [ 0.9132,  0.4163]], device='cuda:0')]
torch:  [tensor([[-0.0766,  0.3599],
        [-0.7820,  0.0715]])]
```

Similarly, if we resize an input from `[4, 4]` to `[8, 8]`, we have: 
```        [ 0.0000e+00,  0.0000e+00,  0.0000e+00,  0.0000e+00,  0.0000e+00,
          0.0000e+00,  0.0000e+00,  0.0000e+00],
```
for filling output. 

TensorRT's IResizeLayer is more likely to be `torch.nn.functional.interpolate`, which emphasize more on interpolating.

peri044:
PR linked : https://github.com/pytorch/TensorRT/pull/2488

PR #2488, opened 2023-11-28
feat: expose IResizeLayer in dynamo.conversion.impl

bowang007:
# Description

#2219 

## Type of change
Expose IResizeLayer in dynamo.conversion.impl

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2463,3 +2463,45 @@ def aten_ops_pad(
         mode=args_bounds_check(args, 2, "constant"),
         value=args_bounds_check(args, 3, None),
     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.upsample_nearest2d.vec)
+def upsample_nearest2d(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.upsample.upsample(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        input=args[0],
+        out_shape=args_bounds_check(args, 1),
+        scale_factors=args_bounds_check(args, 2),
+        resize_mode="nearest",
+        align_corners=False,
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.upsample_bilinear2d.vec)
+def upsample_bilinear2d(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.upsample.upsample(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        input=args[0],
+        out_shape=args_bounds_check(args, 1),
+        scale_factors=args_bounds_check(args, 3),
+        resize_mode="bilinear",
+        align_corners=args_bounds_check(args, 2),
+    )
--- a/py/torch_tensorrt/dynamo/conversion/impl/upsample.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/upsample.py
@@ -1,0 +1,67 @@
+from typing import Optional, Sequence
+
+import tensorrt as trt
+from torch.fx.node import Target
+from torch_tensorrt.dynamo._SourceIR import SourceIR
+from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
+from torch_tensorrt.fx.converters.converter_utils import set_layer_name
+from torch_tensorrt.fx.types import TRTTensor
+
+
+def upsample(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: TRTTensor,
+    out_shape: Optional[Sequence[int]],
+    scale_factors: Optional[Sequence[float]],
+    resize_mode: str,
+    align_corners: bool,
+) -> TRTTensor:
+    resize_layer = ctx.net.add_resize(input)
+    # output size calculation
+    # Pytorch assumes that one of out_shape/scale_factor is None
+    # Pytorch assumes that dimensions match for out_shape/scale factor
+    if out_shape is not None:
+        resize_layer.shape = list(input.shape)[:2] + list(out_shape)
+    elif scale_factors is not None:
+        resize_layer.scales = [1.0, 1.0] + list(scale_factors)
+    else:
+        raise RuntimeError(
+            f"At least one of out_shape and scale_factors should be specified."
+        )
+
+    # interpolate mode
+    if resize_mode == "nearest" or None:
+        resize_layer.resize_mode = trt.ResizeMode.NEAREST
+    elif resize_mode == "bilinear":
+        resize_layer.resize_mode = trt.ResizeMode.LINEAR
+        if align_corners is None or not align_corners:
+            raise RuntimeError(
+                f"Interpolation works differently is align_corners is False for {resize_mode} mode in PyTorch and TensorRT."
+            )
+    else:
+        raise RuntimeError(
+            f"Interpolation mode is {resize_mode} which is not supported by TensorRT."
+        )
+
+    if resize_mode == "nearest":
+        resize_layer.coordinate_transformation = (
+            trt.ResizeCoordinateTransformation.ASYMMETRIC
+        )
+    elif resize_mode == "bilinear":
+        # align corners
+        if align_corners is not None and align_corners:
+            resize_layer.coordinate_transformation = (
+                trt.ResizeCoordinateTransformation.ALIGN_CORNERS
+            )
+        else:
+            resize_layer.coordinate_transformation = (
+                trt.ResizeCoordinateTransformation.ASYMMETRIC
+            )
+
+    set_layer_name(resize_layer, target, name, source_ir)
+
+    out = resize_layer.get_output(0)
+    return out
--- a/py/torch_tensorrt/dynamo/conversion/impl/__init__.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/__init__.py
@@ -28,4 +28,5 @@ from . import (
     topk,
     unary,
     unsqueeze,
+    upsample,
 )
--- a/tests/py/dynamo/conversion/test_upsample.py
+++ b/tests/py/dynamo/conversion/test_upsample.py
@@ -1,0 +1,97 @@
+import torch
+from parameterized import parameterized
+from torch.testing._internal.common_utils import run_tests
+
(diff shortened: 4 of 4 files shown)


py/torch_tensorrt/dynamo/conversion/impl/interpolate.py
```diff
@@ -0,0 +1,60 @@
...
+    align_corners: Optional[bool],
+) -> TRTTensor:
+    resize_layer = ctx.net.add_resize(input)
+    # output size calculation
+    # Pytorch assures that only one of out_shape/scale_factor is not None
+    # Pytorch assures that dimensions match for out_shape/scale factor
+    if out_shape is not None:
+        resize_layer.shape = list(input.shape)[:2] + list(out_shape)
+    else:
+        resize_layer.scales = [1, 1] + list(scale_factor)
```
gs-olive, line 29: Should both branches of the conditional ensure that both `shape` and `scales` are set, or can TRT handle cases where only one is set?

py/torch_tensorrt/dynamo/conversion/impl/interpolate.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # interpolate mode
+    if interpolate_mode == "nearest" or None:
+        resize_layer.resize_mode = trt.ResizeMode.NEAREST
+    elif interpolate_mode == "bilinear" or "linear":
```
gs-olive, line 34: Rewrite as:
```python
elif interpolate_mode in ("bilinear", "linear")
```

py/torch_tensorrt/dynamo/conversion/impl/interpolate.py
```diff
@@ -0,0 +1,60 @@
...
+    # output size calculation
+    # Pytorch assures that only one of out_shape/scale_factor is not None
+    # Pytorch assures that dimensions match for out_shape/scale factor
+    if out_shape is not None:
+        resize_layer.shape = list(input.shape)[:2] + list(out_shape)
+    else:
+        resize_layer.scales = [1, 1] + list(scale_factor)
+
+    # interpolate mode
+    if interpolate_mode == "nearest" or None:
```
gs-olive, line 32: Rewrite as:
```python
if interpolate_mode == "nearest" or interpolate_mode is None
```

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2026,3 +2026,24 @@ def aten_ops_addmm(
...
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        input=args[0],
+        out_shape=kwargs.get("size", 1),
+        scale_factor=kwargs.get("scale_factor", 1),
+        interpolate_mode=kwargs.get("mode", 1),
+        align_corners=kwargs.get("align_corners", 1),
+    )
```
zewenli98, line 2049: The decorator's `key` shoud exactly match what the schema has, which typically is something like `torch.ops.aten.xxxxx`. And the type of arguments should also exactly match with the schema. After the corresponding changes, consider using `args_bounds_check(args, ...)` that takes arguments out of `args` instead of `kwargs`.

tests/py/dynamo/conversion/test_interpolate.py
```diff
@@ -0,0 +1,162 @@
+import torch
+from parameterized import parameterized
+from torch.testing._internal.common_utils import run_tests
+
+from .harness import DispatchTestCase
+
+
+class TestNearestInterpolateConverter(DispatchTestCase):
```
zewenli98, line 8: Consider adding more diverse test cases, like data with different dimensions.

zewenli98 (dismissed):
Our converters come from the [schema](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml). The decorator's `key` should be one of the scheme's functions.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
bowang007: from my understanding, when we have some function schema as: 
`
call_function[target=torch.ops.aten.upsample_bilinear2d.vec](args = (%input_1,), kwargs = {output_size: (4, 4), align_corners: True, scale_factors: None})
`
we should use `kwargs` since arguments like `output_size` are in kwargs instead, right?

zewenli98: The reason why `output_size` and so on appeared in `kwargs` is that you passed in arguments in tests by: 
```
return torch.ops.aten.upsample_nearest2d.vec(
    input, output_size=output_shape, scale_factors=None
)
```
If you remove `=`, as you can see we don't use `=` in other tests, they should be in `args`. 

For the underlying reason, let's go back to the schema:
```
func: upsample_bilinear2d.vec(Tensor input, SymInt[]? output_size, bool align_corners, float[]? scale_factors) -> Tensor
```
They should belong to `args`, instead of `kwargs`. 
Only those arguments following `*` are considered as `kwargs`. For example,
```
replication_pad3d.out(Tensor self, SymInt[6] padding, *, Tensor(a!) out) -> Tensor(a!)
```
Based on the schema above, `self` and `padding` should be considered as `args` while `out` should be `kwargs`. The use of `*` in the function signature enforces this syntax.

I encountered this problem before as well, @gs-olive told me **"The current tracer being used by default in testing is a very barebones tracer, so it knows nothing about `aten` ops specifically. It seems that if an argument is specified with `=`, it will consider that a `kwarg`"**

bowang007: cool, thanks!

py/torch_tensorrt/dynamo/conversion/impl/upsample.py
```diff
@@ -0,0 +1,67 @@
...
+        raise RuntimeError(
+            f"At least one of out_shape and scale_factors should be specified."
+        )
+
+    # interpolate mode
+    if resize_mode == "nearest" or None:
+        resize_layer.resize_mode = trt.ResizeMode.NEAREST
+    elif resize_mode == "bilinear":
+        resize_layer.resize_mode = trt.ResizeMode.LINEAR
+        if align_corners is None or align_corners is False:
```
gs-olive, line 40: Switch `align_corners is False` to `not align_corners`

py/torch_tensorrt/dynamo/conversion/impl/upsample.py
```diff
@@ -0,0 +1,67 @@
...
+def upsample(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: TRTTensor,
+    out_shape: Optional[Sequence[int]],
+    scale_factors: Optional[Sequence[float]],
+    resize_mode: str,
+    align_corners: Optional[bool],
```
gs-olive, line 20: This argument could be switched to just `bool`, since `None` and `False` seem to have the same meaning in the function

py/torch_tensorrt/dynamo/conversion/impl/upsample.py
```diff
@@ -0,0 +1,67 @@
...
+) -> TRTTensor:
+    resize_layer = ctx.net.add_resize(input)
+    # output size calculation
+    # Pytorch assumes that one of out_shape/scale_factor is None
+    # Pytorch assumes that dimensions match for out_shape/scale factor
+    if out_shape is not None:
+        resize_layer.shape = list(input.shape)[:2] + list(out_shape)
+    elif scale_factors is not None:
+        resize_layer.scales = [1.0, 1.0] + list(scale_factors)
```
gs-olive, line 29: Are these specific to the 2D case? If so, could they be generalized for 1D or 3D as well?

gs-olive (commented):
Added a few suggestions, but overall looks good! Could you also add support for the non-`vec` variants, such as [`upsample_bilinear2d`](https://github.com/pytorch/pytorch/blob/4c5fe66880031f9acee4622e1d085f5ca4a470f6/aten/src/ATen/native/native_functions.yaml#L12368).

py/torch_tensorrt/dynamo/conversion/impl/upsample.py
bowang007: `upsample` or `interpolate` functions in pytorch assumes that the input is an image with (n, c, h, w ...), in other words, the input of these functions should at least be `dim == 3` since the first 2 dimensions are `batchsize` and `channels`.

gs-olive: Understood - thanks for the clarification, that makes sense

py/torch_tensorrt/dynamo/conversion/impl/interpolate.py
bowang007: @gs-olive 
Looks like there are some internal assertions in Pytorch that checks the arguments for `shape` and `scales`. 
So pytorch will throws out an error if none of them is set or both of them are set. 
Since this is a converter so we assumes that the arguments are valid or pytorch will throw out error right? 
Please correct me if I'm wrong, thanks!

py/torch_tensorrt/dynamo/conversion/impl/upsample.py
bowang007: Thanks this makes sense!

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2026,3 +2026,45 @@ def aten_ops_addmm(
...
+        name,
+        input=args[0],
+        out_shape=args_bounds_check(args, 1),
+        scale_factors=args_bounds_check(args, 2),
+        resize_mode="nearest",
+        align_corners=None,
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.upsample_bilinear2d.vec)  # type:ignore[misc]
```
gs-olive, line 2052: The `# type:ignore[misc]` can be removed for these decorators. If linting shows issues with this, adding `--no-verify` to `git commit` will allow you to bypass the issues.

gs-olive (commented):
Looks good overall - added minor comment about `mypy` linting. Also needs a rebase onto the latest `main` to resolve merge conflicts

bowang007:
> @bowang007 - actually, I realized the [`upsample_bilinear2d`](https://github.com/pytorch/pytorch/blob/4c5fe66880031f9acee4622e1d085f5ca4a470f6/aten/src/ATen/native/native_functions.yaml#L12368) op is not in the Core ATen IR set, so it is okay to omit it. That error arises when the decorator specified for the converter is something like `aten.upsample_bilinear2d` instead of `aten.upsample_bilinear2d.default`

In that case, how can we add converter for `aten.upsample_bilinear2d`?

gs-olive:
We can add this converter using the decorator `@torch.ops.aten.upsample_bilinear2d.default` as opposed to `@torch.ops.aten.upsample_bilinear2d`.

bowang007:
> Our converters come from the [schema](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml). The decorator's `key` should be one of the scheme's functions.

yeah that makes sense! Thanks!

gs-olive (approved):
Just needs a rebase, then looks good to me

PR #2490, opened 2023-11-28
Exposing select layer

### LINKED ISSUE #2213: Expose ISelectLayer in dynamo.conversion.impl

apbose:
This PR exposes ISelect layer. Fixes #2213. 
Note it does not have test cases since `aten.select` uses TensorRT gather layer. TensorRT select layer is used by `aten.where`. So the functionality will be tested by where test cases.

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
@@ -96,7 +96,7 @@ def where(
                 ctx, target, source_ir, f"{name}_y_expand", y_val, output_shape
             )
 
-    return select(ctx, target, source_ir, name, x_val, y_val, condition)
+    return select(ctx, target, source_ir, name, x_val, y_val, condition_val)
 
 
 def select(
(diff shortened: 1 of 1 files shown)


py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
```diff
@@ -96,8 +96,18 @@ def where(
...
-    set_layer_name(select_layer, target, f"{name}_select")
 
+def select(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: Union[TRTTensor, np.ndarray, torch.Tensor],
+    other: Union[TRTTensor, np.ndarray, torch.Tensor],
+    condition: Union[TRTTensor, np.ndarray, torch.Tensor],
```
gs-olive, line 109: It seems that all three of these would need to be `TRTTensor` since they are all directly fed into the `add_select` function.

gs-olive (commented):
Added a minor comment, otherwise looks good!

gs-olive (approved):
Looks good to me

PR #2504, opened 2023-11-29
feat: support aten.amin dynamo converter

### LINKED ISSUE #2493: `aten.amin`

zewenli98:
# Description

Support `aten.amin` dynamo converter

Fixes #2493 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -956,6 +956,11 @@ def aten_ops_expand(
 
 
 @dynamo_tensorrt_converter(torch.ops.aten.amax.default)
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
+    }
+)
 def aten_ops_amax(
     ctx: ConversionContext,
     target: Target,
@@ -975,6 +980,11 @@ def aten_ops_amax(
 
 
 @dynamo_tensorrt_converter(torch.ops.aten.amin.default)
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
+    }
+)
 def aten_ops_amin(
     ctx: ConversionContext,
     target: Target,
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -955,19 +955,7 @@ def aten_ops_expand(
     )
 
 
-def amax_param_validator(amax_node: Node) -> bool:
-    if len(amax_node.args) < 2:
-        _LOGGER.debug(
-            f"At least two args input and dim should be provided, but only got {len(amax_node.args)} args."
-        )
-        return False
-
-    return True
-
-
-@dynamo_tensorrt_converter(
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -955,19 +955,7 @@ def aten_ops_expand(
...
 
 
-def amax_param_validator(amax_node: Node) -> bool:
-    if len(amax_node.args) < 2:
-        _LOGGER.debug(
-            f"At least two args input and dim should be provided, but only got {len(amax_node.args)} args."
-        )
-        return False
-
-    return True
```
zewenli98, line 965: According to the [schema](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L3769), this validator seems not necessary.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -986,6 +974,25 @@ def aten_ops_amax(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.amin.default)
```
gs-olive, line 977: This converter (as well as `amax` above) seem to require `ITensor` for `args[0]`, so the `@enforce_tensor_types` decorator can be added.

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -19,15 +19,15 @@ def amax(
     source_ir: Optional[SourceIR],
     name: str,
     input_val: TRTTensor,
-    dim: Union[int, Sequence[int]],
+    dim: Sequence[int] = [],
```
gs-olive, line 22: Minor suggestion/comment - should the `amin` and `amax` converters have any default arguments? It seemed most of the `impl` implementations do not have any default arguments, and instead let the converter in `aten_ops_converters.py` dictate the defaults.

zewenli98: Well, that's a good question. I also thought about it. At first I didn't think too much about it, just letting `aten_ops_converters.py` dictate default arguments, but later, I started to add default arguments that the schema provides in the function definition, because I was thinking maybe some functions need to call basic functions from `impl`. In this case, we can directly use the default arguments if needed. Do you think which one is better? adding default arguments in the function definition in `impl` or not?

apbose: I think the default arg could be included in the `impl` as well because of the point which you mentioned. In that case don't you think we should just keep the case in which the `dim` arg is `None`?

zewenli98: @apbose Thanks for the comment! I know it's kind of unintuitive but based on the [schema](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml), the defination of `amin` is:
```
- func: amin(Tensor self, int[1] dim=[], bool keepdim=False) -> Tensor
  variants: function, method
  structured_delegate: amin.out
  tags: core
```
So I think it would be better to keep consistency by using `dim=[]`

gs-olive (approved):
Looks good to me

PR #2505, opened 2023-11-30
feat: support aten.arange.start_step dynamo converter

### LINKED ISSUE #2492: `aten.arange.start_step`

zewenli98:
# Description

Support `aten.arange.start_step` dynamo converter.

Fixes #2492 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
+++ b/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
@@ -3,6 +3,7 @@ import logging
 import operator
 from typing import Dict, Sequence, Tuple, Union
 
+import numpy as np
 import torch
 from torch.fx.node import Argument, Node, Target
 from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
@@ -26,7 +27,6 @@ def getitem_validator(getitem_node: Node) -> bool:
 @dynamo_tensorrt_converter(operator.getitem, capability_validator=getitem_validator)
 @dynamo_tensorrt_converter(builtins.getattr)
 @dynamo_tensorrt_converter(torch.ops.aten.detach.default)
-@dynamo_tensorrt_converter(torch.ops.aten.arange.start_step)
 def generic_evaluator(
     ctx: ConversionContext,
     target: Target,
@@ -38,3 +38,14 @@ def generic_evaluator(
         f"Evaluating {ConverterRegistry.qualified_name_or_str(target)} on object with name: {name}"
     )
     return target(*args)
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.arange.start_step)
+def aten_ops_arange_start_step(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return np.arange(*args)
--- a/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
+++ b/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
@@ -1,3 +1,4 @@
+import builtins
 import logging
 import operator
 from typing import Dict, Sequence, Tuple, Union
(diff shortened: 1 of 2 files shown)


apbose (commented):
Can this be implemented in the evaluator instead?

py/torch_tensorrt/dynamo/conversion/impl/sequence.py
```diff
@@ -0,0 +1,21 @@
...
+def arange(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    start: Union[int, float],
+    end: Union[int, float],
+    step: Union[int, float] = 1,
+) -> TRTTensor:
+    values = np.arange(start, end, step)
```
gs-olive, line 20: Consider switching output to `np.array` type + making evaluator

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -23,7 +24,9 @@ def getitem_validator(getitem_node: Node) -> bool:
 
 # TODO: Subsequent evaluators should be registered here with their own validators
 @dynamo_tensorrt_converter(operator.getitem, capability_validator=getitem_validator)
+@dynamo_tensorrt_converter(builtins.getattr)
 @dynamo_tensorrt_converter(torch.ops.aten.detach.default)
+@dynamo_tensorrt_converter(torch.ops.aten.arange.start_step)
```
gs-olive, line 29: This may not work out of the box in the `torch.compile` workflow, since it will invoke `torch.ops.aten.arange.start_step()`, which can create `FakeTensor` objects. A temporary workaround is to add a new converter in this file just for `arange`.

zewenli98: @gs-olive Thanks for the explanation! I fixed it with `np.arange`

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -23,6 +25,7 @@ def getitem_validator(getitem_node: Node) -> bool:
 
 # TODO: Subsequent evaluators should be registered here with their own validators
 @dynamo_tensorrt_converter(operator.getitem, capability_validator=getitem_validator)
+@dynamo_tensorrt_converter(builtins.getattr)
```
gs-olive, line 28: What does the `getattr` builtin do in the FX graph? Are there any sample uses of it in models? It may need a validator like `operator.getitem`

zewenli98: Thanks for the comment! Yes, `builtins.getattr` is used in the test here: 
```python
class Arange(nn.Module):
    def forward(self, x):
        return torch.ops.aten.arange.start_step(start, x.shape[0], step)
```
where `x.shape` calls `builtins.getattr`, as the graph below: 
```
.INFO:harness:FX graph= graph():
    %getattr_1 : [num_users=1] = call_function[target=builtins.getattr](args = (%x, shape), kwargs = {})
    %getitem : [num_users=1] = call_function[target=operator.getitem](args = (%getattr_1, 0), kwargs = {})
    %arange_start_step : [num_users=1] = call_function[target=torch.ops.aten.arange.start_step](args = (3, %getitem, 3), kwargs = {})
    return arange_start_step
```

I thought about adding `validator`, but it depends on the attributes that the caller has. It seems not general to only allow `shape` attr I think? Do you have any idea about validator?

gs-olive: I don't think `builtins.getattr` is a standard op, and should not appear in Dynamo graphs after ATen tracing. Could you try running the test case with `use_dynamo_tracer=True`?

zewenli98: Cool, it disappeared, thanks!

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -35,3 +36,14 @@ def generic_evaluator(
...
+
+@dynamo_tensorrt_converter(torch.ops.aten.arange.start_step)
+def aten_ops_arange_start_step(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return np.arange(*args)
```
gs-olive, line 49: A very minor comment - using `np.arange(*args)` requires the schemas of `arange.start_step` and `np.arange` to be the same position-wise. In this case they are the same and are unlikely to change, so I think this is fine, just wanted to mention this scheme may not hold between Torch and Numpy in general.

gs-olive (approved):
Looks good to me. Made a small comment with no action required.

PR #2515, opened 2023-12-05
select_scatter decomp

### LINKED ISSUE #2436: Add support for `aten.select_scatter`

apbose:
Fixes #2436
This PR would be dependant on https://github.com/pytorch/TensorRT/pull/2519, https://github.com/pytorch/TensorRT/pull/2664 and https://github.com/pytorch/TensorRT/pull/2669. Major changes

2519- Decomposition of aten::slice_scatter
2664- Implementation makes use of aten::scatter.src
2669- Constants getting converted to fake tensors in get_attr call due to which different device location meta and cpu in torch

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -213,6 +213,19 @@ def slice_scatter_decomposition(
     return output_tensor
 
 
+@register_torch_trt_decomposition(
+    torch.ops.aten.select_scatter.default, registry=TORCH_TRT_DECOMPOSITIONS
+)
+def select_scatter_decomposition(
+    input_tensor: torch.Tensor,
+    src_tensor: torch.Tensor,
+    dim: int,
+    index: int,
+) -> torch.Tensor:
+    src_tensor = torch.unsqueeze(src_tensor, dim)
+    return torch.slice_scatter(input_tensor, src_tensor, dim, index, index + 1, 1)
+
+
 def get_decompositions(
     enable_experimental_decompositions: bool = False,
 ) -> Dict[OpOverload, Callable[[Any], Any]]:
--- a/tests/py/dynamo/lowering/test_decompositions.py
+++ b/tests/py/dynamo/lowering/test_decompositions.py
@@ -530,7 +530,7 @@ class TestLowering(TestCase):
             "torch_compile",
             inputs,
             min_block_size=1,
-            truncate_long_and_double=True,
+            truncate_double=True,
             pass_through_build_failures=True,
         )
         optimized_model_results = optimized_model(*inputs).detach().cpu()
@@ -593,7 +593,7 @@ class TestLowering(TestCase):
             "torch_compile",
             inputs,
             min_block_size=1,
-            truncate_long_and_double=True,
+            truncate_double=True,
             pass_through_build_failures=True,
         )
         optimized_model_results = optimized_model(*inputs).detach().cpu()
@@ -663,7 +663,7 @@ class TestLowering(TestCase):
             "torch_compile",
             inputs,
             min_block_size=1,
-            truncate_long_and_double=True,
+            truncate_double=True,
             pass_through_build_failures=True,
         )
         optimized_model_results = optimized_model(*inputs).detach().cpu()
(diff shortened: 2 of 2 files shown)


py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +162,35 @@ def var_decomposition(
...
+        raise AssertionError("The index should not be greater than dim")
+
+    # expanding the src_tensor to have the same dimension as input_tensor
+    # check if the dimension of the src tensor is same as slice tensor
+    select_tensor = torch.select(input_tensor, dim, index)
+
+    if select_tensor.shape != src_tensor.shape:
+        raise AssertionError(
+            "The slice tensor shape should be equal to the src tensor shape"
+        )
```
gs-olive, line 186: Are the `AssertionError` cases invalid in Torch, or just cases we can't support? Having `AssertionError`s in lowering passes can cause models to inexplicably fail for users, so it is not preferable.

If these are invalid cases in Torch itself, then we do not need these assertions. If they are not supported by TRT, then we can instead return the original op and not lower.

gs-olive (changes requested):

py/torch_tensorrt/dynamo/lowering/_decompositions.py
apbose: Yes it is invalid test case in torch itself. So I think the right thing would be to do away with the assertion.
When you say return the original op, that would mean in those cases, we just do
```
if(condition == True):
     return <unlowered_original_op>
```

gs-olive: Ok thanks for that clarification. I meant that if it would be invalid in Torch, you can assume it will not be the case that the condition would ever be encountered - otherwise the model should have failed earlier. Specifically, if `select_tensor.shape != src_tensor.shape` is a _requirement_ of `select_scatter`, then it is safe to assume the inputs are valid inputs to that function, otherwise we can let Torch throw the error.

apbose: Ok got it. Thanks for the clarification!

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +162,21 @@ def var_decomposition(
...
+def select_scatter_decomposition(
+    input_tensor: torch.Tensor,
+    src_tensor: torch.Tensor,
+    dim: int,
+    index: int,
+) -> torch.Tensor:
+    unbind_tensors = torch.unbind(input_tensor, dim)
+    unbind_tensors_list = list(unbind_tensors)
+    unbind_tensors_list[index] = src_tensor
+    return torch.stack(tuple(unbind_tensors_list), dim)
```
gs-olive, line 177: What operators in the graph does this generate after tracing? Is there a before/after sample that could be shared

gs-olive (commented):
Overall, looks good - just had one question, added below

py/torch_tensorrt/dynamo/lowering/_decompositions.py
apbose: @gs-olive, these were the graphs-
Pre-AOT Autograd graph:
```
graph():
    %l_x_ : torch.Tensor [num_users=1] = placeholder[target=L_x_]
    %l_src_ : torch.Tensor [num_users=1] = placeholder[target=L_src_]
    %select_scatter_default : [num_users=1] = call_function[target=torch.ops.aten.select_scatter.default](args = (%l_x_, %l_src_, 0, 0), kwargs = {})
    return (select_scatter_default,)
```
Post-AOT Autograd graph:
```
graph():
    %clone : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%arg1_1,), kwargs = {})
    %clone_1 : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%arg0_1,), kwargs = {})
    %slice_2 : [num_users=1] = call_function[target=torch.ops.aten.slice.Tensor](args = (%clone_1, 0, 1, 2), kwargs = {})
    %squeeze_1 : [num_users=1] = call_function[target=torch.ops.aten.squeeze.dim](args = (%slice_2, 0), kwargs = {})
    %cat : [num_users=1] = call_function[target=torch.ops.aten.cat.default](args = ([%clone, %squeeze_1],), kwargs = {})
    %view : [num_users=1] = call_function[target=torch.ops.aten.view.default](args = (%cat, [2, 2]), kwargs = {})
    return (view,)
Graph after constant folding:
graph():
    %slice_2 : [num_users=1] = call_function[target=torch.ops.aten.slice.Tensor](args = (%arg0_1, 0, 1, 2), kwargs = {})
    %squeeze_1 : [num_users=1] = call_function[target=torch.ops.aten.squeeze.dim](args = (%slice_2, 0), kwargs = {})
    %cat : [num_users=1] = call_function[target=torch.ops.aten.cat.default](args = ([%arg1_1, %squeeze_1],), kwargs = {})
    %view : [num_users=1] = call_function[target=torch.ops.aten.view.default](args = (%cat, [2, 2]), kwargs = {})
    return (view,)
```
Post-lowering passes Autograd graph:
```
graph():
    %slice_2 : [num_users=1] = call_function[target=torch.ops.aten.slice.Tensor](args = (%arg0_1, 0, 1, 2), kwargs = {})
    %squeeze_1 : [num_users=1] = call_function[target=torch.ops.aten.squeeze.dim](args = (%slice_2, 0), kwargs = {})
    %cat : [num_users=1] = call_function[target=torch.ops.aten.cat.default](args = ([%arg1_1, %squeeze_1],), kwargs = {})
    %reshape_default : [num_users=1] = call_function[target=torch.ops.aten.reshape.default](args = (%cat, [2, 2]), kwargs = {})
    return (reshape_default,)
```

However I have changed the implementation now to make use of `slice_scatter` implementation which I have updated in the description.

gs-olive (changes requested):
I think the old implementation was valid, but the new one does not seem to work in some cases, for example:

```python
>>> import torch
>>> a = torch.zeros(2, 2)
>>> b = torch.ones(2)
>>> torch.select_scatter(a, b, 0, 0)
tensor([[1., 1.],
        [0., 0.]])
>>> torch.slice_scatter(a, b.unsqueeze(0), 0, 1, 1)
RuntimeError: expected src to have a size equal to the slice of self. src size = [1, 2], slice size = [0, 2]
```

gs-olive:
See [this decomposition](https://github.com/pytorch/pytorch/blob/e5e0685f614171ab3e3a22c85dd4ee8cd0cf09ae/aten/src/ATen/functorch/BatchRulesScatterOps.cpp#L881-L890) for an alternative approach.

apbose:
Thanks @gs-olive for pointing the above.
But I think the implementation using `slice_scatter` decomposition should also work in our case.
For eg: in the above case the unsqueeze dimension with `src` would lead to` src_tensor` being` [1,2]` (`torch.slice_scatter` would expect it to be `[0,2]`. But since in `slice_scatter` decomposition, we do away with dimension at `dim (dim=0) `in this case the above error would not come there. Also in more than two dimensions this case should never be encountered

apbose:
I misread the earlier case:

apbose:
```
>>> import torch
>>> a = torch.zeros(2, 2)
>>> b = torch.ones(2)
>>> torch.select_scatter(a, b, 0, 0)
tensor([[1., 1.],
        [0., 0.]])
>>> torch.slice_scatter(a, b.unsqueeze(0), 0, 1, 1)
RuntimeError: expected src to have a size equal to the slice of self. src size = [1, 2], slice size = [0, 2]
```

apbose:
I misread the case pointed by you.
```
>>> import torch
>>> a = torch.zeros(2, 2)
>>> b = torch.ones(2)
>>> torch.select_scatter(a, b, 0, 0)
tensor([[1., 1.],
        [0., 0.]])
>>> torch.slice_scatter(a, b.unsqueeze(0), 0, 1, 1)
RuntimeError: expected src to have a size equal to the slice of self. src size = [1, 2], slice size = [0, 2]
```
In the above according to the implementation above, the `slice_scatter `op would be
` torch.slice_scatter(a, b.unsqueeze(0), 0, 1, 2)`
Which would lead to the slice dimension being [1,2]
The difference between `slice_scatter` and `select_scatter` is that since `select_scatter` inserts a single dimensional tensor, while inputting the `src` tensor we generally just provide  single dimension along the dim.
Example: for 
`torch.select_scatter=>`
for input tensor torch.zeros((2,2)) with shape = [2,2] ,the src tensor should be torch.ones(2) with shape [2] instead of [1,2] for dim = 1 
for input tensor torch.zeros((2,3,4)) with shape = [2,3,4], the src tensor should be torch.ones(2,4) with shape [2,4] for dim = 1
The op would be `torch.select_scatter(input, src, dim, index)`

`torch.slice_scatter=> `
for input tensor torch.zeros((2,2)) with shape = [2,2], the src tensor should be torch.ones(1,2) with shape [1,2] for dim = 1
for input tensor torch.zeros((2,3,4)) with shape = [2,3,4], the src tensor should be torch.ones(2, 1, 4) with shape [2,4] for dim = 1
The op would be `torch.slice_scatter(input, src, dim, index, index+1)`

To answer the above question-
1. No `slice_scatter` decomposition does not alter the torch slice_scatter behavior. It is 1:1 behavior
2. My earlier comment was because I misunderstood. The `torch.slice_scatter` would also expect it to be `[1,2]`, since the slice_scatter is for `torch.slice_scatter(input_tensor, src_tensor, dim, index, index + 1, 1)  `(index to index+1).  But as mentioned in the above comment, since we do away with the unsqueezed dimension in `slice_scatter`, I partially misunderstood the example of the` slice_scatter` op and thought that would work (basically the `slice_scatter` op would not come in the first place).

gs-olive (approved):
Looks good to me, pending rebase + CI passing

PR #2519, opened 2023-12-06
slice_scatter decomposition

### LINKED ISSUE #2434: Add support for `aten.slice_scatter`

apbose:
Fixes #2434 
This PR would be dependant on #2664 and #2669. Major changes
1. 2664- Implementation makes use of `aten::scatter.src`
2. 2669- Constants getting converted to fake tensors in `get_attr` call due to which different device location `meta` and` cpu` in torch

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -4,6 +4,7 @@ from typing import Any, Callable, Dict, List, Optional
 import torch
 from torch._decomp import register_decomposition
 from torch._ops import OpOverload
+from torch_tensorrt.dynamo.conversion.converter_utils import get_positive_dim
 
 from ._decomposition_groups import (
     ENABLED_TORCH_DECOMPOSITIONS,
@@ -172,6 +173,44 @@ def empty_permuted_decomposition(*args, **kwargs) -> torch.Tensor:
     for permute_index, permute_element in enumerate(empty_permute):
         perm[permute_element] = permute_index
     return torch.empty([empty_size[l] for l in empty_permute], **kwargs).permute(perm)
+
+
+@register_torch_trt_decomposition(
+    torch.ops.aten.slice_scatter.default, registry=TORCH_TRT_DECOMPOSITIONS
+)
+def slice_scatter_decomposition(
+    input_tensor: torch.Tensor,
+    src_tensor: torch.Tensor,
+    dim: int,
+    start: Optional[int] = None,
+    end: Optional[int] = None,
+    step: Optional[int] = None,
+):
+    dim_size = input_tensor.shape[dim]
+    start = get_positive_dim(start, input_tensor.shape[dim])
+    if end is None:
+        end = dim_size
+    end = get_positive_dim(end, input_tensor.shape[dim])
+    if step is None:
+        step = 1
+
+    src_dim = src_tensor.shape
+    # step == 0 is not a valid torch case
+    # also src_dim should be equal to slice dimension
+
+    if start == 0 and end == dim_size and step == 1:
+        return src_tensor
+
+    cat_tensors = []
+    index_tensor_shape = []
+    for i, src_each_dim in enumerate(list(src_dim)):
+        if i != dim:
+            index_tensor_shape.append(src_each_dim)
+    for index in range(start, end, step):
+        cat_tensors.append(index * torch.ones(index_tensor_shape, dtype=torch.long))
+    index_tensor = torch.stack(cat_tensors, dim).cuda()
+    output_tensor = torch.scatter(input_tensor, dim, index_tensor, src_tensor)
+    return output_tensor
 
 
 def get_decompositions(
--- a/tests/py/dynamo/lowering/test_decompositions.py
+++ b/tests/py/dynamo/lowering/test_decompositions.py
@@ -484,6 +484,201 @@ class TestLowering(TestCase):
             f"The optimized model results shape and torch model results shape should be equal in empty_like",
         )
 
+    def test_lowering_slice_scatter_dimOne_module(self):
+        class sliceScatter(torch.nn.Module):
+            def __init__(self, *args, **kwargs) -> None:
+                super().__init__(*args, **kwargs)
+
+            def forward(self, x, src, dim, start=None, end=None, step=1):
+                y = torch.ops.aten.slice_scatter(x, src, dim, start, end, step)
+                return y
+
+        # Operations expected to be removed in the traced graph after decompositions
+        expected_ops = {
+            torch.ops.aten.scatter.src,
+        }
+        unexpected_ops = {torch.ops.aten.select_scatter}
+
+        inputs = [torch.zeros(8, 8).cuda(), torch.ones(8, 2).cuda(), 1, 6, None, 1]
+
+        fx_graph = torch.fx.symbolic_trace(sliceScatter())
+        unexpected_ops_seen, expected_ops_unseen = lower_graph_testing(
+            fx_graph,
+            inputs,
+            expected_ops=expected_ops,
+            unexpected_ops=unexpected_ops,
+            min_block_size=1,
+        )
+
+        self.assertEqual(
+            len(unexpected_ops_seen),
+            0,
+            f"The following unexpected ops were encountered: {unexpected_ops_seen}",
+        )
+
+        self.assertEqual(
+            len(expected_ops_unseen),
+            0,
+            f"The following expected ops were not encountered: {expected_ops_unseen}",
+        )
+
+        torch._dynamo.reset()
+
+        # Validate that the results between Torch and Torch-TRT are similar
+        optimized_model = torch_tensorrt.compile(
+            fx_graph,
+            "torch_compile",
+            inputs,
+            min_block_size=1,
+            truncate_long_and_double=True,
+            pass_through_build_failures=True,
+        )
+        optimized_model_results = optimized_model(*inputs).detach().cpu()
+        torch_model_results = fx_graph(*inputs).detach().cpu()
+
+        max_diff = float(
+            torch.max(torch.abs(optimized_model_results - torch_model_results))
+        )
+        self.assertAlmostEqual(
+            max_diff,
+            0,
+            DECIMALS_OF_AGREEMENT,
+            f"Slice_scatter TRT outputs don't match with the original model.",
+        )
+
+    def test_lowering_slice_scatter_dimZero_StepTwo_module(self):
+        class sliceScatter(torch.nn.Module):
+            def __init__(self, *args, **kwargs) -> None:
+                super().__init__(*args, **kwargs)
+
+            def forward(self, x, src, dim, start, end, step):
+                y = torch.ops.aten.slice_scatter.default(x, src, dim, start, end, step)
+                return y
+
+        # Operations expected to be removed in the traced graph after decompositions
+        expected_ops = {
+            torch.ops.aten.scatter.src,
+        }
+        unexpected_ops = {torch.ops.aten.slice_scatter}
+
+        inputs = [torch.zeros(8, 8).cuda(), torch.ones(2, 8).cuda(), 0, 2, 6, 2]
+
+        fx_graph = torch.fx.symbolic_trace(sliceScatter())
+
+        unexpected_ops_seen, expected_ops_unseen = lower_graph_testing(
+            fx_graph,
+            inputs,
+            expected_ops=expected_ops,
+            unexpected_ops=unexpected_ops,
+            min_block_size=1,
+        )
+
+        self.assertEqual(
+            len(unexpected_ops_seen),
+            0,
+            f"The following unexpected ops were encountered: {unexpected_ops_seen}",
+        )
+
+        self.assertEqual(
+            len(expected_ops_unseen),
+            0,
+            f"The following expected ops were not encountered: {expected_ops_unseen}",
+        )
+
+        torch._dynamo.reset()
+
+        # Validate that the results between Torch and Torch-TRT are similar
+        optimized_model = torch_tensorrt.compile(
+            fx_graph,
+            "torch_compile",
+            inputs,
+            min_block_size=1,
+            truncate_long_and_double=True,
+            pass_through_build_failures=True,
+        )
+        optimized_model_results = optimized_model(*inputs).detach().cpu()
+        torch_model_results = fx_graph(*inputs).detach().cpu()
+
+        max_diff = float(
+            torch.max(torch.abs(optimized_model_results - torch_model_results))
+        )
+        self.assertAlmostEqual(
+            max_diff,
+            0,
+            DECIMALS_OF_AGREEMENT,
+            f"Slice_scatter TRT outputs don't match with the original model.",
+        )
+
+    def test_lowering_slice_scatter_dimOne_3d_module(self):
+        class sliceScatter(torch.nn.Module):
+            def __init__(self, *args, **kwargs) -> None:
+                super().__init__(*args, **kwargs)
+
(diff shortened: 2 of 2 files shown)


py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +163,52 @@ def var_decomposition(
...
+):
+    dim_size = input_tensor.shape[dim]
+    if start is not None and start < 0:
+        start = start + dim_size
+    if end is not None and end < 0:
+        end = end + dim_size
+    if start is None:
+        start = 0
+    if end is None:
+        end = dim_size
```
gs-olive, line 185: Consider switching to use `get_positive_dim` utility.

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +163,52 @@ def var_decomposition(
...
+    if step_dim > src_dim[dim]:
+        end_dim = src_dim[dim]
+    else:
+        indices = torch.Tensor(np.arange(0, step_dim))
+        indices = indices.to(torch.int32)
+        src = torch.index_select(src, dim, indices)
+
+    if start == 0 and end == dim_size and step == 0:
+        return input_tensor
+    index_tensor = np.arange(start, end_dim, step)
```
gs-olive, line 199: Does this work with `torch.arange`?

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +163,52 @@ def var_decomposition(
...
+        start = start + dim_size
+    if end is not None and end < 0:
+        end = end + dim_size
+    if start is None:
+        start = 0
+    if end is None:
+        end = dim_size
+
+    src_dim = src_tensor.shape
+    step_dim = torch.floor_divide(end - start, step)
```
gs-olive, line 188: `(end - start) // step`

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +163,52 @@ def var_decomposition(
...
+    end_dim = end
+    if step_dim > src_dim[dim]:
+        end_dim = src_dim[dim]
+    else:
+        indices = torch.Tensor(np.arange(0, step_dim))
```
gs-olive, line 193: `torch.arange`

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +163,52 @@ def var_decomposition(
...
+    if start == 0 and end == dim_size and step == 0:
+        return input_tensor
+    index_tensor = np.arange(start, end_dim, step)
+
+    unbind_input_tensors = torch.unbind(input_tensor, dim)
+    unbind_input_tensors_list = list(unbind_input_tensors)
+    unbind_source_tensors = torch.unbind(src, dim)
+    unbind_source_tensors_list = list(unbind_source_tensors)
+
+    for i, index in enumerate(index_tensor):
```
gs-olive, line 206: `range(start, end_dim, step)` instead of index tensor

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +166,46 @@ def var_decomposition(
...
+    if step_dim > src_dim[dim]:
+        end_dim = src_dim[dim]
+    else:
+        indices = torch.Tensor(torch.arange(0, step_dim))
+        indices = indices.to(torch.int32)
+        src = torch.index_select(src, dim, indices)
+
+    if start == 0 and end == dim_size and step == 0:
+        return input_tensor
+    index_tensor = torch.arange(start, end_dim, step)
```
gs-olive, line 196: Is this tensor needed; could it be replaced with `range`, as below?

py/torch_tensorrt/dynamo/lowering/_decompositions.py
gs-olive: See other comment

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +166,46 @@ def var_decomposition(
...
+    start = get_positive_dim(start, input_tensor.shape)
+    end = get_positive_dim(end, input_tensor.shape)
+
+    src_dim = src_tensor.shape
+    step_dim = (end - start) // step
+    end_dim = end
+    if step_dim > src_dim[dim]:
+        end_dim = src_dim[dim]
+    else:
+        indices = torch.Tensor(torch.arange(0, step_dim))
```
gs-olive, line 190: `torch.arange` should already return a Tensor, so the cast should not be needed

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +166,46 @@ def var_decomposition(
...
+        indices = indices.to(torch.int32)
+        src = torch.index_select(src, dim, indices)
+
+    if start == 0 and end == dim_size and step == 0:
```
gs-olive, line 194: Should this be `step == 1`?

gs-olive (changes requested):

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +164,47 @@ def var_decomposition(
...
+    end = get_positive_dim(end, input_tensor.shape)
+
+    src_dim = src_tensor.shape
+    step_dim = (end - start) // step
+    end_dim = end
+    if step_dim > src_dim[dim]:
+        end_dim = src_dim[dim]
+    else:
+        indices = torch.arange(0, step_dim)
+        indices = indices.to(torch.int32)
```
gs-olive, line 189: If the indices are `int64`, it is fine to leave them as-is and not change the data type, since later operators may expect or require `int64`

py/torch_tensorrt/dynamo/lowering/_decompositions.py
apbose: I think this should be `step == 0` since `step == 1` would result in tensors being inserted in source tensor at step 1 interval.

apbose:
Monitoring the CI to see if this error comes in the test- 
```
torch._dynamo.exc.BackendCompilerFailed: backend='functools.partial(<function fx_dynamo_testing_backend at 0x7f19514f7af0>, store_intermediate_graphs=[], min_
block_size=1, torch_executed_ops=set(), use_fast_partitioner=True)' raised:
RuntimeError: Attempted to set the storage of a tensor on device "meta" to a storage on different device "cpu".  This is no longer allowed; the devices must match.
```

py/torch_tensorrt/dynamo/lowering/_decompositions.py
gs-olive: Does `torch` expect `int64` for the indices in `index_select`, or TensorRT? If it is TensorRT, then there is no need to perform the cast, because the outputs of the above operation will already have been handled in the `TRTInterpreter`

apbose: Yes torch would expect int64 input for the indices. Since this is a constant, I think TRTInterpretor should be able to handle it. Yes I will remove this.

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +163,45 @@ def var_decomposition(
...
+    index_tensor_shape = []
+    for i, src_each_dim in enumerate(list(src_dim)):
+        if i != dim:
+            index_tensor_shape.append(src_each_dim)
+    for index in range(start, end, step):
+        cat_tensors.append(index * torch.ones(index_tensor_shape))
+    index_tensor = torch.stack(cat_tensors, dim)
+    index_tensor = index_tensor.to(torch.int64).cuda()
+    output_tensor = torch.scatter(input_tensor, dim, index_tensor, src)
+    return output_tensor
```
gs-olive, line 202: Could this potentially be simplified to avoid `for`-loops using `torch.arange`? For instance, see this [implementation](https://github.com/pytorch/pytorch/blob/e5e0685f614171ab3e3a22c85dd4ee8cd0cf09ae/aten/src/ATen/functorch/BatchRulesScatterOps.cpp#L872-L879)

apbose: Hi @gs-olive I tried the above implementation.
I am not sure how `get_expanded_index` works, but I think it will be difficult to achieve the above behavior without for loops.
I tried two alternate
```
indices = torch.arange(start,stop, step)
cat_tensors = torch.unsqueeze(indices,1) * torch.ones(index_tensor_shape)).flatten(end_dim=1)
#or
cat_tensors = indices(:, None) * torch.ones(index_tensor_shape)).flatten(end_dim=1)
```
The thing is we need to unsqueeze indices n no of times, where n is the dimension of index_tensor_shape. While the above would work for cases
```
input = torch.ones(8,8)
src = torch.ones(8,2)
out  = torch.slice_scatter(input, src, 1, 6, 8, 1) 
```
or 
```
input = torch.ones(8,8)
src = torch.ones(8,1)
out  = torch.slice_scatter(input, src, 1, 6, 7, 1) 
```
it would start failing for input and src with sizes `torch.ones(8,8,8)` and `torch.zeros(8,2,8)` or `torch.zeros(8,1,8) `respectively. We would have to unsqueeze n no of times, eg: torch.unsqueeze(indices,1,1) or indices[:,None,None] would work, but then that would again be a for loop.

I cannot think of another way on top of my mind, if you have any suggestion you could let me know. 
For now the test cases pass with for loop so I have reverted back to that,

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -162,6 +163,45 @@ def var_decomposition(
...
+
+    cat_tensors = []
+    index_tensor_shape = []
+    for i, src_each_dim in enumerate(list(src_dim)):
+        if i != dim:
+            index_tensor_shape.append(src_each_dim)
+    for index in range(start, end, step):
+        cat_tensors.append(index * torch.ones(index_tensor_shape))
+    index_tensor = torch.stack(cat_tensors, dim)
+    index_tensor = index_tensor.to(torch.int64).cuda()
```
gs-olive, line 200: This will cause a graph break if it inserts a cast in the graph representation, since TRT cannot support Int64 casts. What is the resultant output graph in this case?

gs-olive: This operation might be avoidable by specifying `dtype=torch.long` in the `torch.ones(...)` call, though if the index tensor is a constant and not an `ITensor`, it may not be necessary.

tests/py/dynamo/lowering/test_decompositions.py
```diff
@@ -420,6 +420,197 @@ def forward(self, x):
...
+                return y
+
+        # Operations expected to be removed in the traced graph after decompositions
+        expected_ops = {
+            torch.ops.aten.index.Tensor,
+            torch.ops.aten.scatter.src,
+        }
+        unexpected_ops = {torch.ops.aten.select_scatter}
+
+        inputs = [torch.zeros(8, 8).cuda(), torch.ones(8, 2).cuda(), 1, 6, None, 1]
```
gs-olive, line 503: Could this case be modified to be 3D, as in your comment above.

gs-olive (commented):
Overall looks good - added a few comments/questions

py/torch_tensorrt/dynamo/lowering/_decompositions.py
apbose: @gs-olive the `torch.long` was present since otherwise torch would have complained that torch requires int64 input for torch.scatter in this [line](https://github.com/pytorch/TensorRT/blob/main/tests/py/dynamo/testing_utilities.py#L186).

For the case torch.slice_scatter(torch.zeros(8,8), torch.ones(8,2), 1, 6, None, 1) leads to this with the cast to index_tensor = index_tensor.to(torch.int64).cuda() -
Pre-AOT Autograd graph:=============
```
graph():
    %l_x_ : torch.Tensor [num_users=1] = placeholder[target=L_x_]
    %l_src_ : torch.Tensor [num_users=1] = placeholder[target=L_src_]
    %clone_default_1 : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%l_src_,), kwargs = {})
    %clone_default : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%l_x_,), kwargs = {})
    %slice_scatter : [num_users=1] = call_function[target=torch.ops.aten.slice_scatter](args = (%clone_default, %clone_default_1, 1, 6,
 None, 1), kwargs = {})
    return (slice_scatter,)
```
Post AOT Autograd graph:=============
```
graph():
    %clone : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%arg1_1,), kwargs = {})
    %clone_1 : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%arg0_1,), kwargs = {})
    %empty_strided : [num_users=1] = call_function[target=torch.ops.aten.empty_strided.default](args = ([8], [1]), kwargs = {dtype: to$
ch.int64, layout: torch.strided, device: cpu, pin_memory: False})
    %full_like : [num_users=1] = call_function[target=torch.ops.aten.full_like.default](args = (%empty_strided, 1), kwargs = {pin_memo$
y: False})
    %mul : [num_users=1] = call_function[target=torch.ops.aten.mul.Tensor](args = (%full_like, 6), kwargs = {})
    %empty_strided_1 : [num_users=1] = call_function[target=torch.ops.aten.empty_strided.default](args = ([8], [1]), kwargs = {dtype: t
orch.int64, layout: torch.strided, device: cpu, pin_memory: False})
    %full_like_1 : [num_users=1] = call_function[target=torch.ops.aten.full_like.default](args = (%empty_strided_1, 1), kwargs = {pin_m
emory: False})
    %mul_1 : [num_users=1] = call_function[target=torch.ops.aten.mul.Tensor](args = (%full_like_1, 7), kwargs = {})
    %unsqueeze : [num_users=1] = call_function[target=torch.ops.aten.unsqueeze.default](args = (%mul, 1), kwargs = {})
    %unsqueeze_1 : [num_users=1] = call_function[target=torch.ops.aten.unsqueeze.default](args = (%mul_1, 1), kwargs = {})
    %cat : [num_users=1] = call_function[target=torch.ops.aten.cat.default](args = ([%unsqueeze, %unsqueeze_1], 1), kwargs = {})
    %_to_copy : [num_users=1] = call_function[target=torch.ops.aten._to_copy.default](args = (%cat,), kwargs = {dtype: torch.int64, lay
out: torch.strided, device: cuda:0})
    %scatter : [num_users=1] = call_function[target=torch.ops.aten.scatter.src](args = (%clone_1, 1, %_to_copy, %clone), kwargs = {})
    return (scatter,)
```
Post lowering Autograd graph:=============
```
graph():
    %_frozen_param0 : [num_users=1] = get_attr[target=_frozen_param0]
    %scatter : [num_users=1] = call_function[target=torch.ops.aten.scatter.src](args = (%arg0_1, 1, %_frozen_param0, %arg1_1), kwargs =
 {})
    return (scatter,)
```

As mentioned  by you since it is a frozen param and a constant, there are no graph breaks and not necessary. Not sure if this would be the case always though.
Hence I changed it to torch.ones() with dtype torch.long as suggested.
A side general question- Would the graph break lead to significant performance impact? That is the reason we should avoid them as far as possible?

tests/py/dynamo/lowering/test_decompositions.py
apbose: I kept the old test case and added another with the 3d.

PR #2522, opened 2023-12-08
feat: support aten.clamp.Tensor and update aten.clamp.default dynamo converters

### LINKED ISSUE #2517: `aten.clamp.Tensor`

zewenli98:
# Description

support `aten.clamp.Tensor` and update `aten.clamp.default` dynamo converters

Fixes #2517 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
@@ -1,6 +1,5 @@
 from typing import Optional, Union
 
-import numpy as np
 import tensorrt as trt
 import torch
 import torch_tensorrt.dynamo.conversion.impl as impl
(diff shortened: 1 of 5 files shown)


py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
```diff
@@ -186,63 +184,22 @@ def clamp(
...
-    if max_val is not None:
-        clamp_max_layer = _add_layer(
-            ctx, input_val, max_val, trt.ElementWiseOperation.MIN, name
-        )
-        set_layer_name(clamp_max_layer, target, f"{name}_clamp_max")
-        input_val = clamp_max_layer.get_output(0)
+    if min_val is None:
+        min_val = float("-inf")
+    if max_val is None:
+        max_val = float("inf")
```
gs-olive, line 193: In the case that one of these is not specified, can `convert_binary_elementwise` + TRT handle infinity as an argument? Would it be possible to instead omit that `min` or `max` layer from the computation entirely? As in:
```python
clamp_min = input_val
if min_val is not None:
    # Compute clamp min
    clamp_min = ...

clamp_max = clamp_min
if max_val is not None:
    # Compute clamp max
    clamp_max = ...

return clamp_max

zewenli98: That makes sense! Thanks for pointing out!

gs-olive (commented):
Overall looks good - it seems that a few of the `test_clip` converter tests are failing, however

gs-olive (approved):
Looks good to me!

PR #2528, opened 2023-12-12
fix: `release/2.1` CI Repair

gs-olive:
- Fix CI runs for release branch
- Ensure testing is completed on `torch==2.1.2` for alignment with release version


.github/scripts/install-torch-tensorrt.sh
```diff
@@ -2,8 +2,8 @@
 set -eou pipefail
 # Source conda so it's available to the script environment
 source ${BUILD_ENV_FILE}
-${CONDA_RUN} ${PIP_INSTALL_TORCH} torchvision pyyaml
+${CONDA_RUN} ${PIP_INSTALL_TORCH} torch==2.1.2 torchvision==0.16.2 pyyaml
```
narendasan, line 5: does this do anything? since theres already ${PIP_INSTALL_TORCH}?

.github/workflows/linux-test.yml
```diff
@@ -66,6 +70,7 @@ jobs:
       REF: ${{ inputs.ref }}
       CU_VERSION: ${{ matrix.desired_cuda }}
       SCRIPT: ${{ inputs.script }}
+      CHANNEL: ${{ inputs.channel }}
```
narendasan, line 73: Does this do anything?

py/ci/soname_excludes.params
```diff
@@ -36,4 +36,4 @@
 --exclude libcublas.so.12.1.3.1
 --exclude libcublasLt.so.12.1.3.1
 --exclude libcudart.so.11.8.89
---exclude libcudart.so.11
\ No newline at end of file
+--exclude libcudart.so.11.0
```
narendasan, line 39: This change is not relevant for tests in CI

.github/workflows/linux-test.yml
gs-olive: Initially, I thought it would affect the `PIP_INSTALL_TORCH` command generated, but I will try omitting it and see how the CI performs.

.github/scripts/install-torch-tensorrt.sh
gs-olive: This is necessary, since in `build-test.yml`, when building Torch-TRT, we have:
```
PIP_INSTALL_TORCH='pip install torch==2.1.2 --extra-index-url https://download.pytorch.org/whl/test/cu118'
```
In `linux-test.yml`, when testing Torch-TRT using the artifact from build, we have:
```
PIP_INSTALL_TORCH='pip install torch --extra-index-url https://download.pytorch.org/whl/test/cu118'
```
The second one installs the `2.2.0.dev` test build, which is the most up-to-date test build, but is incompatible with the version on which it was built. Specifying the version explicitly resolves this problem.

PR #2543, opened 2023-12-15
feat: support aten.trunc dynamo converter

### LINKED ISSUE #2536: aten.trunc

zewenli98:
# Description

Support `aten.trunc` dynamo converter

Fixes #2536 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/unary/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/unary/ops.py
@@ -444,6 +444,17 @@ def trunc(
     name: str,
     input_val: TRTTensor,
 ) -> TRTTensor:
+    if input_val.dtype not in (trt.float16, trt.float32):
+        return impl.cast.to_copy(
+            ctx,
+            target,
+            source_ir,
+            f"{name}_copy",
+            input_val,
+            input_val.dtype,
+            force_layer=True,
+        )
+
     dividend = get_trt_tensor(ctx, 1, f"{name}_dividend")
     return impl.elementwise.trunc_div(
         ctx, target, source_ir, f"{name}_trunc", input_val, dividend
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/impl/unary/ops.py
```diff
@@ -432,3 +435,16 @@ def erf(
...
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input_val: TRTTensor,
+) -> TRTTensor:
+    dividend = get_trt_tensor(ctx, 1, f"{name}_dividend")
+    return impl.elementwise.trunc_div(
+        ctx, target, source_ir, f"{name}_trunc", input_val, dividend
+    )
```
gs-olive, line 450: Based on the documentation of `torch.trunc`, it seems that the behavior differs based on the data type. Specifically, for `float` inputs, the output is also a float, and for `int` inputs, the output is also an int. Does this line up with the behavior of `trunc_div` as well?

gs-olive: Additionally, could a line be added such as
```python
if dtype not in (float16, float32):
   return input_val
```
This can help to avoid unnecessary layer insertion

zewenli98: Thanks for pointing this out! I added the code below since TRT engine inputs cannot be TRT engine outputs:
``` python
if input_val.dtype not in (trt.float16, trt.float32):
        return impl.cast.to_copy(
            ctx,
            target,
            source_ir,
            f"{name}_copy",
            input_val,
            input_val.dtype,
            force_layer=True,
        )
```

gs-olive (approved):
Looks good to me!

PR #2550, opened 2023-12-18
feat: support aten.copy dynamo converter

### LINKED ISSUE #2435: Add support for `aten.copy`

zewenli98:
# Description

Support `aten.copy` dynamo converter.

Fixes #2435 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2547,3 +2547,28 @@ def aten_ops_trunc(
         name,
         args[0],
     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.copy.default)
+@enforce_tensor_types(
+    {
+        1: (TRTTensor,),
+    }
+)
+def aten_ops_copy(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    src = args[1]
+    return impl.cast.to_copy(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        src,
+        src.dtype,
+        force_layer=True,
+    )
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2530,7 +2530,7 @@ def aten_ops_sort(
 @dynamo_tensorrt_converter(torch.ops.aten.trunc.default)
 @enforce_tensor_types(
     {
-        1: (TRTTensor,),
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2541,3 +2541,28 @@ def aten_ops_sort(
         dim=args_bounds_check(args, 1, -1),
         descending=args_bounds_check(args, 2, False),
     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.copy.default)
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
```
gs-olive, line 2549: Should this be `1: (TRTTensor,)`? We don't seem to use the first argument in the converter

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2541,3 +2541,28 @@ def aten_ops_sort(
...
+    src = args[1]
+    return impl.cast.to_copy(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        src,
+        src.dtype,
+        force_layer=True,
+    )
```
gs-olive, line 2568: The `force_layer=True` does seem useful for the purpose of the test case, but in a normal model we should prefer `force_layer=False`. This should be a future feature, but `to_copy` should have an intelligent `force_layer` mechanism which can select whether to insert the layer based on whether the node is an `input` node to the TRTEngine.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: Thanks for the good catch!

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: Yes, I agree with you, but for now TensorRT doesn't support the same input and output.

gs-olive: I added #2561 to track this proposed feature (in Torch-TRT)

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2530,7 +2530,7 @@ def aten_ops_sort(
 @dynamo_tensorrt_converter(torch.ops.aten.trunc.default)
 @enforce_tensor_types(
     {
-        0: (TRTTensor,),
+        1: (TRTTensor,),
```
gs-olive, line 2533: It doesn't seem that this modification should apply to `trunc` here

zewenli98: Oh good catch! That was accidentally involved by rebase.

gs-olive (approved):
Looks good to me!

PR #2575, opened 2024-01-05
chore: Set default return type to ExportedProgram

peri044:
# Description

This change adds the following

1) Returns ExportedProgram as the output of compilation by default
2) `output_format` flag to return desired output format (exported_program | torchscript | graph_module)
3) Modify export serde test cases to handle the above changes

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -45,19 +43,11 @@ def export(
                         format=torch.channel_last
                     ), # Dynamic input shape for input #2
                     torch.randn((1, 3, 224, 244)) # Use an example tensor and let torch_tensorrt infer settings
-        ir (str): torchscript | exported_program. Based on the provided ir, the output type would be a torchscript or exported program.
     """
-    if ir == "torchscript":
-        return torch.jit.trace(gm, inputs)
```
gs-olive, line 51: What are the options for users who want to `torchscript` the exported product for serialization?

gs-olive (commented):
Overall looks good. Added a question, and it seems 2 tests are failing currently. Consider redirecting the branch to the `staging_rel_2_2` branch to get the release test suite.

py/torch_tensorrt/dynamo/_exporter.py
narendasan: They could always post return trace

narendasan: Alternatively, we can have a string/enum, input for `output_format` with options: `exported_program/ep`, `torchscript/ts`, `graph_module/fx`

narendasan (commented):
We need documentation updates in this PR

peri044:
This PR has an open issue https://github.com/pytorch/pytorch/issues/116831 which should be fixed first

py/torch_tensorrt/dynamo/_exporter.py
peri044: ^ This looks good. Added `output_format` now

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -84,34 +72,78 @@ def transform(
...
+
+            # state_dict has these parameters/buffers as torch.Tensors. We override them as torch.nn.Parameter/torch.Tensors respectively.
+            for name, _ in gm.named_parameters():
+                if node.target == name:
+                    input_kind = InputKind.PARAMETER
+                    state_dict[name] = constant_tensor
+            for name, _ in gm.named_buffers():
+                if node.target == name:
+                    input_kind = InputKind.BUFFER
+                    state_dict[name] = constant_tensor
```
gs-olive, line 118: Consider using `break` once the name is matched to a parameter or buffer. Are the `get_parameter`/`get_buffer` functions useful for this case?

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -84,34 +72,78 @@ def transform(
...
+    # The input_specs should be of the form [params, buffers, constant_tensors, user_inputs]
+    non_user_input_idx = 0
+    for node in gm.graph.nodes:
         if node.op == "get_attr":
-            constant_tensor = getattr(trt_gm, node.target)
-            with trt_gm.graph.inserting_before(first_user_input):
-                const_placeholder_node = trt_gm.graph.placeholder(node.target)
-                const_placeholder_node.meta = copy.deepcopy(node.meta)
+            constant_tensor = getattr(gm, node.target)
+            input_kind = InputKind.CONSTANT_TENSOR
```
gs-olive, line 108: In the constant tensor case, does the `state_dict` not need to be updated?

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -84,34 +72,78 @@ def transform(
...
                     constant_tensor
                 )
                 node.replace_all_uses_with(const_placeholder_node)
-                trt_gm.graph.erase_node(node)
+                gm.graph.erase_node(node)
+
+                # Add these parameters/buffers/constants to the existing graph signature
+                # before user inputs. These specs are looked up in the state_dict during ExportedProgram creation.
+                graph_signature.input_specs.insert(
+                    non_user_input_idx,
```
gs-olive, line 134: Is it critical that the `[params, buffers, constant_tensors, user_inputs]` ordering above be strictly followed? It seems that in certain suboptimal node interleavings, params can come after buffers if the `get_attr` nodes are ordered differently.

gs-olive (commented):
Overall, looks great - just added a few questions on the exporter.

py/torch_tensorrt/dynamo/_exporter.py
peri044: Added a break. `get_parameter` apparently raises an `AttributeError`https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.get_parameter  if it's not found. We could wrap it in a try-except but I'll leave it like this for now unless there's a strong preference. All we want here is if the particular node is a parameter, assign `input_kind` appropriately and override the state_dict

py/torch_tensorrt/dynamo/_exporter.py
peri044: Yeah no need because they are of the type `torch.Tensor` which is correct

py/torch_tensorrt/dynamo/_exporter.py
peri044: `[params, buffers, constant_tensors, user_inputs]` is the convention that I saw being followed within Pytorch code. From my understanding, the ordering of input specs should follow the order of `get_attr` nodes which is what the code does.

py/torch_tensorrt/dynamo/_exporter.py
gs-olive: This sounds good to me - nested `try/except` sounds pretty messy

PR #2578, opened 2024-01-05
feat: support aten.any related converters in dynamo

bowang007:
# Description

support aten.any converter

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2605,6 +2605,27 @@ def aten_ops_remainder(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.any.default)
+@dynamo_tensorrt_converter(torch.ops.aten.any.dim)
+@dynamo_tensorrt_converter(torch.ops.aten.any.dims)
+def aten_ops_any(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.reduce.any(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args_bounds_check(args, 1, replacement=None),
+        args_bounds_check(args, 2, replacement=False),
+    )
+
+
 @dynamo_tensorrt_converter(torch.ops.aten._pdist_forward.default)
 @enforce_tensor_types(
     {
--- a/py/torch_tensorrt/dynamo/conversion/impl/reduce.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/reduce.py
@@ -3,6 +3,7 @@ from typing import Optional, Sequence, Tuple, Union
 import tensorrt as trt
 from torch.fx.node import Target
 from torch_tensorrt.dynamo._SourceIR import SourceIR
+from torch_tensorrt.dynamo.conversion import impl
 from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
 from torch_tensorrt.dynamo.conversion.converter_utils import (
     cast_trt_tensor,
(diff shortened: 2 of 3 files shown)


narendasan (commented):
Do you have a bool test case? Its probably a common variant of any

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2572,3 +2572,24 @@ def aten_ops_copy(
...
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.reduce.any(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args_bounds_check(args, 1, replacement=[]),
```
gs-olive, line 2593: Based on the `any.dims` schema [here](https://github.com/pytorch/pytorch/blob/534c73d478fc967cb3a6a29f5eada94bc4ce2c29/aten/src/ATen/native/native_functions.yaml#L727), the replacement should be `None`

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -208,3 +209,25 @@ def mean(
...
     return layer.get_output(0)
+
+
+def any(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input_val: TRTTensor,
+    dim: Sequence[int] = [],
```
gs-olive, line 220: Based on the supported schemas, specifically [`any.dim` and `any.dims`](https://github.com/pytorch/pytorch/blob/534c73d478fc967cb3a6a29f5eada94bc4ce2c29/aten/src/ATen/native/native_functions.yaml#L721-L727), this would need to be:
```python
dim: Union[int, Optional[Sequence[int]]] = None,
```

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -208,3 +209,25 @@ def mean(
...
+) -> TRTTensor:
+    num_out = cast_trt_tensor(ctx, input_val, trt.float32, f"{name}_cast")
+    abs_out = impl.unary.abs(
+        ctx,
+        target,
+        source_ir,
+        f"{name}_abs",
+        num_out,
+    )
+    max_out = amax(ctx, target, source_ir, f"{name}_amax", abs_out, dim, keepdim)
```
gs-olive, line 231: Consider using the Int32-Cast, Sum-Reduce approach to calculate the `any` operator:
https://github.com/pytorch/TensorRT/blob/16c031349c6a1af5a8408a817f2ef8542aa6f176/core/conversion/converters/impl/reduce.cpp#L12-L40

gs-olive (commented):
Added a few suggestions on the implementation. Also needs a rebase to resolve merge conflict.

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
bowang007: Hey @gs-olive, regarding to this, what if we have an input tensor as [-1, 1], then reduce sum would be 0 and the result would be `false`. However, it should return `true` right? 
Thanks!

gs-olive: Hi @bowang007 - thanks for the catch - this is a great point. I agree - the absolute value is helpful in light of that, and `Max` + `Sum` should have the same ultimate effect for this converter. I think the Int32 cast is likely still needed, however, since floating point discrepancies could affect the value of `0` when casting from Float to Bool.

py/torch_tensorrt/dynamo/conversion/impl/reduce.py
```diff
@@ -208,3 +209,32 @@ def mean(
...
+        input_val,
+    )
+    if dim is None:
+        dim = []
+    elif isinstance(dim, int):
+        dim = [dim]
+
+    max_out = amax(ctx, target, source_ir, f"{name}_amax", abs_out, dim, keepdim)
+
+    return cast_trt_tensor(ctx, max_out, trt.bool, f"{name}_cast")
```
gs-olive, line 240: The name in this function call should be changed to `f"{name}_cast_to_bool"`, since it will otherwise collide with the name on line 224 and throw an error in TRT.

gs-olive (approved):
Looks good to me; it seems that the test failures are unrelated to the PR.

PR #2580, opened 2024-01-05
Rand converter - evaluator

apbose:
Covers the converters- aten.rand, aten.randn, aten.randperm

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
+++ b/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
@@ -62,7 +62,9 @@ def rand_validator(rand_node: Node) -> bool:
         return False
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.rand.default)
+@dynamo_tensorrt_converter(
+    torch.ops.aten.rand.default, capability_validator=rand_validator
+)
 def aten_ops_rand(
     ctx: ConversionContext,
     target: Target,
@@ -71,7 +73,7 @@ def aten_ops_rand(
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     device = kwargs.get("device", None)
-    return np.random.rand(*args).to(device=device)
+    return np.random.rand(*args)
 
 
 def randn_validator(randn_node: Node) -> bool:
@@ -87,7 +89,9 @@ def randn_validator(randn_node: Node) -> bool:
         return False
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.randn.default)
+@dynamo_tensorrt_converter(
+    torch.ops.aten.randn.default, capability_validator=randn_validator
+)
 def aten_ops_randn(
     ctx: ConversionContext,
     target: Target,
@@ -96,7 +100,7 @@ def aten_ops_randn(
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     device = kwargs.get("device", None)
-    return np.random.randn(*args).to(device=device)
+    return np.random.randn(*args)
 
 
 def randperm_validator(randperm_node: Node) -> bool:
@@ -112,7 +116,9 @@ def randperm_validator(randperm_node: Node) -> bool:
         return False
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.randperm.default)
+@dynamo_tensorrt_converter(
+    torch.ops.aten.randperm.default, capability_validator=randperm_validator
+)
 def aten_ops_randperm(
     ctx: ConversionContext,
     target: Target,
@@ -124,4 +130,4 @@ def aten_ops_randperm(
     input = args[0]
     if not isinstance(input, int):
         raise RuntimeError(f"The input must be an integer")
-    return np.random.randperm(*args).to(device=device)
+    return np.random.permutation(*args)
--- a/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
+++ b/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
@@ -47,3 +47,83 @@ def aten_ops_arange_start_step(
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return np.arange(*args)
+
+
+def rand_validator(rand_node: Node) -> bool:
+    dtype = rand_node.kwargs.get("dtype", None)
+    layout = rand_node.kwargs.get("layout", None)
+    if dtype is not None:
+        _LOGGER.debug(
+            f"Currently we don't support specifying output dtype, got {dtype}."
+        )
+        return False
+    if layout is not None:
+        _LOGGER.debug(
+            f"Currently we don't support specifying layout, got {layout}."
+        )
+        return False    
+@dynamo_tensorrt_converter(torch.ops.aten.rand.default)
+def aten_ops_rand(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    device = kwargs.get("device", None)
+    return np.random.rand(*args).to(device=device)
+
+
+def randn_validator(randn_node: Node) -> bool:
+    dtype = randn_node.kwargs.get("dtype", None)
+    layout = randn_node.kwargs.get("layout", None)
+    if dtype is not None:
+        _LOGGER.debug(
+            f"Currently we don't support specifying output dtype, got {dtype}."
+        )
+        return False
+    if layout is not None:
+        _LOGGER.debug(
+            f"Currently we don't support specifying layout, got {layout}."
+        )
+        return False   
+@dynamo_tensorrt_converter(torch.ops.aten.randn.default)
+def aten_ops_randn(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    device = kwargs.get("device", None)
+    return np.random.randn(*args).to(device=device)
+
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -47,3 +47,81 @@ def aten_ops_arange_start_step(
...
+@dynamo_tensorrt_converter(torch.ops.aten.rand.default)
+def aten_ops_rand(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    device = kwargs.get("device", None)
+    return np.random.rand(*args).to(device=device)
```
gs-olive, line 74: As above, I think you would need to unpack the integers with something like `np.random.rand(*args[0])`. Additionally, `.to` would not have effect in Numpy

gs-olive (changes requested):

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
apbose: Thanks for the review. I think `np.random.rand(*args) ` should also work, since it passes locally. Let me cross check.

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -47,3 +47,91 @@ def aten_ops_arange_start_step(
...
+    dtype = randn_node.kwargs.get("dtype", None)
+    layout = randn_node.kwargs.get("layout", None)
+    if dtype is not None:
+        _LOGGER.debug(
+            f"Currently we don't support specifying output dtype, got {dtype}."
+        )
+        return False
+    if layout is not None:
+        _LOGGER.debug(f"Currently we don't support specifying layout, got {layout}.")
+        return False
```
gs-olive, line 89: Could use same validator as above, doesn't need a second if they are the same

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -47,3 +47,91 @@ def aten_ops_arange_start_step(
...
+        _LOGGER.error(f"Input should be of type int.")
+        return False
+    if dtype is not None:
+        _LOGGER.debug(
+            f"Currently we don't support specifying output dtype, got {dtype}."
+        )
+        return False
+    if layout is not None:
+        _LOGGER.debug(f"Currently we don't support specifying layout, got {layout}.")
+        return False
```
gs-olive, line 120: Needs to `return True` at end

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -47,3 +47,81 @@ def aten_ops_arange_start_step(
...
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    device = kwargs.get("device", None)
+    input = args[0]
+    if not isinstance(input, int):
+        raise RuntimeError(f"The input must be an integer")
```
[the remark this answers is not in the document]
gs-olive: To remove

gs-olive (changes requested):

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -47,3 +47,77 @@ def aten_ops_arange_start_step(
...
+    torch.ops.aten.rand.default, capability_validator=rand_validator
+)
+def aten_ops_rand(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    device = kwargs.get("device", None)
```
gs-olive, line 76: Can be removed since it is unused

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -47,3 +47,77 @@ def aten_ops_arange_start_step(
...
+    torch.ops.aten.randn.default, capability_validator=rand_validator
+)
+def aten_ops_randn(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    device = kwargs.get("device", None)
```
gs-olive, line 90: Same as above

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -47,3 +47,77 @@ def aten_ops_arange_start_step(
...
+    torch.ops.aten.randperm.default, capability_validator=randperm_validator
+)
+def aten_ops_randperm(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    device = kwargs.get("device", None)
```
gs-olive, line 122: Same as above

gs-olive (commented):
Added a small comment, otherwise looks good

gs-olive (approved):
See below

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -47,3 +47,74 @@ def aten_ops_arange_start_step(
...
+    torch.ops.aten.randn.default, capability_validator=rand_validator
+)
+def aten_ops_randn(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return np.random.randn(*args)
```
gs-olive, line 89: This fails on my machine, since `np.random.randn` does not accept tuples, and `args` is a tuple containing a tuple:
```python
>>> args = ((1, 2, 3,),)
>>> np.random.randn(*args)
  File "mtrand.pyx", line 1270, in numpy.random.mtrand.RandomState.randn
  File "mtrand.pyx", line 1431, in numpy.random.mtrand.RandomState.standard_normal
  File "_common.pyx", line 636, in numpy.random._common.cont
TypeError: 'tuple' object cannot be interpreted as an integer
```

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -47,3 +47,74 @@ def aten_ops_arange_start_step(
...
+    torch.ops.aten.rand.default, capability_validator=rand_validator
+)
+def aten_ops_rand(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return np.random.rand(*args)
```
gs-olive, line 76: Same error as above on my machine

gs-olive (changes requested):

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
apbose: @gs-olive Thanks for pointing this out.
I think this got missed in the tests, since in the tests the input is None, and the compilation is not invoked.
I changedthe above to have `np.random.rand(*args[0])`. Also in the test I modified it to have an input (hacky way, I can as well use the harness.py with the below model)
```
def test_rand(self, name, op, shape_or_input):
        class TestModule(nn.Module):
            def __init__(self, rand_op, size):
                super().__init__()
                self.rand_op = rand_op
                self.size = size
            def forward(self, x):
                b = x + 1
                self.size[0] = b
                return self.rand_op(self.size)
```
But right now I am running into segmentation fault. I am looking into this further.

tests/py/dynamo/conversion/harness.py
```diff
@@ -252,6 +251,40 @@ def run_test(
...
+        self,
+        mod,
+        inputs,
+        expected_ops,
+        comparators: List[Tuple[Callable, List]],
+        precision=torch.float32,
+        output_dtypes=None,
+        use_dynamo_tracer=False,
+        enable_passes=False,
+    ):
```
gs-olive, line 264: Does this only test the shapes of the results? If so, can the name be updated to `run_test_compare_shapes_only` or something similar?

apbose: In this case it is checking the shape of the Tensors. But it is supposed to be a generic function where the callable can check other attribute of the Tensor. For example, in the case of `rand` and `randn`  we compare the data type as well. So I think the name should be `run_test_comparator` but you can let me know if you think otherwise.

gs-olive: Ok that makes sense, I think the name is reasonable then. If possible, maybe `run_test_compare_tensor_attributes_only` just to delineate that the actual values in the tensors are ignored, but it is also okay as is

apbose: Ok makes sense, I will rename it to the above then.

gs-olive (approved):
Looks good to me!

peri044:
@apbose  Can you open a cherry pick PR to release/2.3 ?

apbose:
Sure I will do that today @peri044

zewenli98:
Hi @apbose I found a small bug/typo and had a fix for it here: https://github.com/pytorch/TensorRT/pull/2809 I'll merge it after passing CI. Can you cherry pick this PR with this fix by the way?

PR #2591, opened 2024-01-12
Grant write permission to token

huydhn:
The default write permission of the token has been revoked org-wide, so we need to set that explicitly here for these workflows that update PRs (write).

(Also update the labeler to v4 cause why not)


gs-olive (approved):
Looks good to me! @narendasan to verify

PR #2589, opened 2024-01-12
Clean up AWS credentials

huydhn:
The credential has been revoked to be replaced by OIDC https://github.com/pytorch/test-infra/pull/4865

PR #2592, opened 2024-01-14
Clean up AWS credentials

huydhn:
(Replace https://github.com/pytorch/TensorRT/pull/2589)

The credential has been revoked to be replaced by OIDC https://github.com/pytorch/test-infra/pull/4865

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -64,8 +64,7 @@ jobs:
           - repository: pytorch/tensorrt
             package-name: torch_tensorrt
             pre-script: packaging/pre_build_script.sh
-    # TODO: FOR TESTING, SWITCH BACK TO MAIN BEFORE LANDING
-    uses: pytorch/tensorrt/.github/workflows/linux-test.yml@cleanup-aws-credentials-v2
+    uses: pytorch/tensorrt/.github/workflows/linux-test.yml@main
     with:
       job-name: tests-py-torchscript-fe
       repository: "pytorch/tensorrt"
@@ -100,8 +99,7 @@ jobs:
           - repository: pytorch/tensorrt
             package-name: torch_tensorrt
             pre-script: packaging/pre_build_script.sh
-    # TODO: FOR TESTING, SWITCH BACK TO MAIN BEFORE LANDING
-    uses: pytorch/tensorrt/.github/workflows/linux-test.yml@cleanup-aws-credentials-v2
+    uses: pytorch/tensorrt/.github/workflows/linux-test.yml@main
     with:
       job-name: tests-py-dynamo-converters
       repository: "pytorch/tensorrt"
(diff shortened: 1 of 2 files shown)


gs-olive:
Hi @huydhn - it seems that all of the tests are failing with `Error: Unable to find an artifact with the name: pytorch_tensorrt__3.8__`, [as here](https://github.com/pytorch/TensorRT/actions/runs/7517603643/job/20543945258?pr=2592). I noticed that in earlier CI runs ([example](https://github.com/pytorch/TensorRT/actions/runs/7470735405/job/20402221767)), the name of the artifact was named `pytorch_tensorrt__3.8` - do you know of any recent test-infra PRs which would have changed the artifact naming scheme?

gs-olive:
Hi - thanks for the commit - it seems the same error appears. Another thing I noticed is that in the older CI runs, there are only two jobs within the build-package step, [as here](https://github.com/pytorch/TensorRT/actions/runs/7470735405/job/20402217929), but now there are four, [as here](https://github.com/pytorch/TensorRT/actions/runs/7548465959/job/20550583441?pr=2592). It looks like the artifact name, as below, is not correct:
https://github.com/pytorch/TensorRT/blob/fd19353159ac71860f783e8b00783053b94b5c36/.github/workflows/linux-test.yml#L112-L115

huydhn:
> but now there are four, [as here](https://github.com/pytorch/TensorRT/actions/runs/7548465959/job/20550583441?pr=2592)

This is expected because the job has been broken up into 2 part, building and uploading artifacts.  The latter has access to the credentials and its access is protected.

huydhn:
@gs-olive The test round seems to work correctly https://github.com/pytorch/TensorRT/actions/runs/7548790822, so landing this should be safe.  I'm rerunning some failures there https://github.com/pytorch/TensorRT/actions/runs/7548790822/job/20552955754 just to be sure, but they look like pre-existing failures and unrelated to this change?

gs-olive:
This looks good! The test failures are unrelated to this PR.

gs-olive (approved):
Looks good to me, pending the switch back to `main` on the workflow.

gs-olive:
Hi @huydhn - I am noticing the following credential error on our Docker workflow: [CI link](https://github.com/pytorch/TensorRT/actions/runs/7627012453/job/20774718883). This workflow uses the login action as here:https://github.com/pytorch/TensorRT/blob/e642f86018c42173a148cde81f2ed3082bea437e/.github/workflows/docker_builder.yml#L40-L42
Does this also need to be updated in light of the new authentication changes?

huydhn:
I think you're right that an update is need to grant the job the permission to push to ghcr.io.  From https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs, I think we need to add the follow snippet:

```
permissions:
  packages: write
```

gs-olive:
Ok thank you for the suggestion - just added it here: https://github.com/pytorch/TensorRT/pull/2619

PR #2627, opened 2024-01-30
feat: Add dynamic shapes support for torch.compile workflow

peri044:
# Description

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -74,10 +75,17 @@ def _pretraced_backend(
             fake_mode, "allow_non_fake_inputs", True
         ), fake_mode:
             repair_input_aliasing(gm)
+
+            # Remove sym_int placeholders and inputs
+            remove_sym_nodes(gm)
+            torch_inputs = [
+                input for input in sample_inputs if isinstance(input, torch.Tensor)
+            ]
```
gs-olive, line 83: Could `sample_inputs` in the function header be modified to `Sequence[Any]`, since symbolic ints would now also be valid inputs?

py/torch_tensorrt/dynamo/lowering/_remove_sym_nodes.py
```diff
@@ -0,0 +1,30 @@
...
+    dynamic=True behavior
+    """
+    # Extract SymInt placeholder Tensors
+    placeholders = [
+        node
+        for node in gm.graph.nodes
+        if (
+            node.op == "placeholder"
+            and isinstance(node.type, type)
+            and issubclass(node.type, torch.SymInt)
```
gs-olive, line 19: Just in case this were to change in the future, could this be updated to include other symbolic types, such as `issubclass(node.type, (torch.SymInt, torch.SymFloat, torch.SymBool))`

py/torch_tensorrt/dynamo/lowering/_remove_sym_nodes.py
```diff
@@ -0,0 +1,30 @@
...
+        )
+    ]
+
+    for node in placeholders:
+        gm.graph.erase_node(node)
```
gs-olive, line 24: How does this affect downstream consumers of the `SymInt` node, such as tensors having that node as a shape dimension?

gs-olive (commented):
Overall looks good. Added a few clarification questions and a suggestion

py/torch_tensorrt/dynamo/lowering/_remove_sym_nodes.py
peri044: From the cases that I validated, there seems to be no users of these `SymInt` nodes. If we encounter such case, we need to refactor this pass accordingly.

py/torch_tensorrt/dynamo/lowering/_remove_sym_nodes.py
peri044: `torch.SymInt` nodes here indicate the shape of that dimension and don't seem to be used by other nodes (in torch.compile's dynamic=True behavior). I'm not sure when `torch.SymFloat` and `torch.SymBool` nodes show up, so we remove only `torch.SymInt` nodes conservatively. We could change this in the future based on the behavior of `torch.SymFloat` and `torch.SymBool` nodes.

PR #2632, opened 2024-02-01
feat: support adaptive avg pool 2d and 3d dynamo converters

### LINKED ISSUE #2611: aten.adaptive_avg_pool2d

zewenli98:
# Description

Support `adaptive_avg_pool2d` and `adaptive_avg_pool3d`.

Fixes #2611 #2612 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2185,7 +2185,7 @@ def aten_ops_avg_pool(
 
 
 @dynamo_tensorrt_converter(torch.ops.aten.adaptive_avg_pool1d.default)
-def aten_ops_adaptive_avg_pool(
+def aten_ops_adaptive_avg_pool1d(
     ctx: ConversionContext,
     target: Target,
     args: Tuple[Argument, ...],
@@ -2193,6 +2193,44 @@ def aten_ops_adaptive_avg_pool(
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return impl.pool.adaptive_avg_pool1d(
+        ctx,
+        target,
+        source_ir=SourceIR.ATEN,
+        name=name,
+        input=args[0],
+        output_size=args[1],
+    )
+
+
+def adaptive_pool_static_input_validator(pool_node: Node) -> bool:
+    output_size = args_bounds_check(pool_node.args, 1)
+    return all([x > 0 for x in output_size])
+
+
+@dynamo_tensorrt_converter(
+    torch.ops.aten.adaptive_avg_pool2d.default,
+    capability_validator=adaptive_pool_static_input_validator,
+)
+@dynamo_tensorrt_converter(
+    torch.ops.aten._adaptive_avg_pool2d.default,
+    capability_validator=adaptive_pool_static_input_validator,
+)
+@dynamo_tensorrt_converter(
+    torch.ops.aten.adaptive_avg_pool3d.default,
+    capability_validator=adaptive_pool_static_input_validator,
+)
+@dynamo_tensorrt_converter(
+    torch.ops.aten._adaptive_avg_pool3d.default,
+    capability_validator=adaptive_pool_static_input_validator,
+)
+def aten_ops_adaptive_avg_poolNd(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.pool.adaptive_avg_poolNd(
         ctx,
         target,
         source_ir=SourceIR.ATEN,
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2202,35 +2202,35 @@ def aten_ops_adaptive_avg_pool1d(
     )
 
 
-def adaptive_pool_static_input_validator(pool_node: Node) -> bool:
-    output_size = args_bounds_check(pool_node.args, 1)
-    return all([x > 0 for x in output_size])
-
-
-@dynamo_tensorrt_converter(
-    torch.ops.aten.adaptive_avg_pool2d.default,
-    capability_validator=adaptive_pool_static_input_validator,
-)
-@dynamo_tensorrt_converter(
-    torch.ops.aten._adaptive_avg_pool2d.default,
-    capability_validator=adaptive_pool_static_input_validator,
-)
-@dynamo_tensorrt_converter(
-    torch.ops.aten.adaptive_avg_pool3d.default,
-    capability_validator=adaptive_pool_static_input_validator,
-)
-@dynamo_tensorrt_converter(
-    torch.ops.aten._adaptive_avg_pool3d.default,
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2184,6 +2184,27 @@ def aten_ops_avg_pool(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.adaptive_avg_pool2d.default)
+@dynamo_tensorrt_converter(torch.ops.aten._adaptive_avg_pool2d.default)
+@dynamo_tensorrt_converter(torch.ops.aten.adaptive_avg_pool3d.default)
+@dynamo_tensorrt_converter(torch.ops.aten._adaptive_avg_pool3d.default)
+def aten_ops_adaptive_avg_poolNd(
```
narendasan, line 2191: Add a validator to disable dynamic cases

py/torch_tensorrt/dynamo/conversion/impl/pool.py
```diff
@@ -104,3 +108,184 @@ def max_poolNd(
...
+    ctx: ConversionContext,
+    target: Union[Target, str],
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: TRTTensor,
+    output_size: Sequence[int],
+) -> TRTTensor:
+    input_rank = len(input.shape)
+
+    if input_rank == 3:  # TRT doesn't support 3D pooling
```
narendasan, line 123: This should also be part of a validator

zewenli98: I think this should be fine. The native TRT doesn't support 3D pooling, as I commented, so I manually reshape to 4D. Torch-TRT can handle this case without throwing any error.

narendasan (approved):
LGTM pending CI

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2184,6 +2184,44 @@ def aten_ops_avg_pool(
     )
 
 
+def adaptive_pool_static_input_validator(pool_node: Node) -> bool:
+    output_size = args_bounds_check(pool_node.args, 1)
+    return all([x > 0 for x in output_size])
```
narendasan, line 2189: Should this check input or output? since for adaptive pooling I think the output is constant?

zewenli98: According to the schema:
```
- func: adaptive_avg_pool2d(Tensor self, SymInt[2] output_size) -> Tensor
```
I think this should check `output_size`, because input is a Tensor instead of size, so its shape doesn't contain negative numbers. However, even though `output_size` is constant, but it is passed by users, so they may leave one dim as `-1`, but this is not supported because `output_size` is necessary for calculating kernel/stride sizes. 
Pytorch doesn't support negative `output_size` either, e.g., `RuntimeError: adaptive_avg_pool2d: elements of output_size must be greater than or equal to 0 but received {1, -2}`

py/torch_tensorrt/dynamo/conversion/impl/pool.py
gs-olive, line 1: The functions `adaptive_avg_pool2d` and `adaptive_avg_pool3d` look very similar - are there major differences which would make it hard to write a single `adaptive_avg_poolNd`, which has a toggle selecting the dimensionality?

py/torch_tensorrt/dynamo/conversion/impl/pool.py
```diff
@@ -169,3 +172,451 @@ def end_index(idx: int, out_dim: int, in_dim: int) -> int:
...
+    output_size = list(output_size)
+    original_input = input
+
+    # repeat_interleave the input if the dim of output is larger than input
+    insert_axises = []
+    for axis in range(1, extend_len + 1):
+        axis = -axis
+        positive_axis = get_positive_dim(
+            axis, input_rank
+        )  # convert to positive axis, which is for calculating new shapes below
```
gs-olive, line 207: Is this the same as `axis % input_rank`? Asked since the negation + re-cast to positive could be confusing here

py/torch_tensorrt/dynamo/conversion/impl/pool.py
```diff
@@ -169,3 +172,451 @@ def end_index(idx: int, out_dim: int, in_dim: int) -> int:
...
+            times = output_dim // input_dim
+            remainder = output_dim % input_dim
+            if (
+                diff == 2 and remainder == 2
+            ):  # case 1: output_dim - input_dim == 2 and is not an integral multiple
+                insert_axises.append(axis)
+                remainder -= 1
+                output_size[axis] -= 1
+
+            if (
```
gs-olive, line 221: Should this be `elif` / should there ever be a case where both this `if` and the above are both entered?

gs-olive (commented):
Overall looks good - added a few small suggestions

py/torch_tensorrt/dynamo/conversion/impl/pool.py
zewenli98: Thanks for the suggestion. I think it would be better if combine them together. Thanks!

py/torch_tensorrt/dynamo/conversion/impl/pool.py
zewenli98: Sorry for the confusion. This is because I want to reversely traverse the input. For example, the tuple is [0,1,2,3], for the first iteration, `axis = 1; axis = -axis = -1; positive_axis = get_positive_dim(-1, 4) = 3`. Both `axis` and `positive_axis` are recorded for the following use.

py/torch_tensorrt/dynamo/conversion/impl/pool.py
zewenli98: It would be possible for some cases to enter the two `if`

tests/py/dynamo/conversion/test_adaptive_avgpool_aten.py
```diff
@@ -80,6 +80,224 @@ def forward(self, x):
...
+    ):
+        class TestModule(torch.nn.Module):
+            def forward(self, x):
+                return torch.ops.aten.adaptive_avg_pool3d.default(x, output_size)
+
+        inputs = [torch.randn(input_shape)]
+        self.run_test(
+            TestModule(),
+            inputs,
+            # use_dynamo_tracer=True,
```
gs-olive, line 271: Can remove if unnecessary

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2202,6 +2202,27 @@ def aten_ops_adaptive_avg_pool(
...
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.pool.adaptive_avg_poolNd(
+        ctx,
+        target,
+        source_ir=SourceIR.ATEN,
+        name=name,
+        input=args[0],
```
gs-olive, line 2221: If this has to be an `ITensor`, consider adding a decorator to enforce the type

gs-olive (approved):
Added two small comments - looks good to me pending CI for the test cases and those fixes.

narendasan (approved):
LGTM

PR #2654, opened 2024-02-23
aten::empty_like 

apbose:


apbose:
Had a doubt on this one. Does  this require a test. In the following test:

gs-olive:
I'm not sure what the `empty_like` lowers to, but potentially you could add another operation in the `nn.Module` so that the graph is non-empty. It is likely the case that the graph is completely empty, so the partitioning fails. Since this decomposition is Torch-provided, we shouldn't need a test, however it is important to verify that whatever the operator is lowered to, is also supported by Torch-TRT

apbose:
I do not think that the graph would be empty since it would reduce to the lowering operations of `aten::size` and `torch.Tensor()` of the corresponding size getting created. So the graph once lowered should lead to these operations, though I need to confirm.
 Ok I will add another operation to the module and verify the lowering.

gs-olive:
Thanks for the analysis @apbose - this is very helpful. It looks like the `constant_folding` lowering pass is freezing the memory for the `empty_like` operator and storing it as an attribute of the model.

Regarding `empty_permuted` - it seems like it would be necessary in the dynamic shape case, since we would not be able to freeze the parameter in that case. It seems based on the [Core ATen IR](https://pytorch.org/docs/main/torch.compiler_ir.html) that `prims.empty_permuted` is a core op, so I do think the conversion/evaluation of that would be helpful here, but it could go in a separate PR.

apbose:
Ok I will go ahead and make a separate PR for `empty_permute`. For now this PR can be merged then?

gs-olive (approved):
Looks good to me

PR #2664, opened 2024-02-27
Aten scatter converter

apbose:
Dependency of PR- #2519

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -691,13 +691,14 @@ def aten_ops_clamp(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.scatter.src)
+@dynamo_tensorrt_converter(torch.ops.aten.scatter.value)
 @enforce_tensor_types(
     {
         0: (TRTTensor,),
+        2: (TRTTensor,),
     }
 )
-@dynamo_tensorrt_converter(torch.ops.aten.scatter.src)
-@dynamo_tensorrt_converter(torch.ops.aten.scatter.value)
 def aten_ops_scatter(
     ctx: ConversionContext,
     target: Target,
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -691,28 +691,21 @@ def aten_ops_clamp(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.scatter.value)
-def aten_ops_scatter_value(
-    ctx: ConversionContext,
-    target: Target,
-    args: Tuple[Argument, ...],
-    kwargs: Dict[str, Argument],
-    name: str,
-) -> Union[TRTTensor, Sequence[TRTTensor]]:
-    return impl.select.scatter_value(
-        ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2], args[3]
-    )
-
-
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
+    }
+)
 @dynamo_tensorrt_converter(torch.ops.aten.scatter.src)
-def aten_ops_scatter_src(
+@dynamo_tensorrt_converter(torch.ops.aten.scatter.value)
+def aten_ops_scatter(
     ctx: ConversionContext,
     target: Target,
     args: Tuple[Argument, ...],
     kwargs: Dict[str, Argument],
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
-    return impl.select.scatter_src(
+    return impl.select.scatter(
         ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2], args[3]
     )
 
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -700,7 +700,7 @@ def aten_ops_scatter_value(
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return impl.select.scatter_value(
-        ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2]
+        ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2], args[3]
     )
 
 
@@ -713,19 +713,7 @@ def aten_ops_scatter_src(
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return impl.select.scatter_src(
-        ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2]
-    )
-
-
-def aten_ops_select(
-    ctx: ConversionContext,
-    target: Target,
-    args: Tuple[Argument, ...],
-    kwargs: Dict[str, Argument],
-    name: str,
-) -> Union[TRTTensor, Sequence[TRTTensor]]:
-    return impl.select.select(
-        ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2]
+        ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2], args[3]
     )
 
 
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -691,6 +691,44 @@ def aten_ops_clamp(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.scatter.value)
+def aten_ops_scatter_value(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.select.scatter_value(
+        ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2]
+    )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.scatter.src)
+def aten_ops_scatter_src(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.select.scatter_src(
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -370,3 +372,108 @@ def index(
...
+    input: TRTTensor,
+    dim: int,
+    index: Union[TRTTensor, np.ndarray, torch.Tensor],
+    value: float,
+) -> TRTTensor:
+    if not isinstance(input, TRTTensor):
+        raise RuntimeError(
+            f"scatter_tensor received input {input} that is not part "
+            "of the TensorRT region!"
+        )
```
gs-olive, line 391: Can remove in favor of `enforce_tensor_types` decorator on converter

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -370,3 +372,108 @@ def index(
...
+    input_shape = input.shape
+    index_shape = index.shape
+    index_shape_list = list(index.shape)
+    if not (isinstance(index, TRTTensor)):
+        index = get_trt_tensor(ctx, index, f"_index_tensor")
```
gs-olive, line 396: Can also add this input to `enforce_tensor_types` and avoid this call

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -370,3 +372,108 @@ def index(
...
+    if len(input_shape) != len(index_shape):
+        raise RuntimeError(f"The no of dimensions of input and index should be equal")
```
gs-olive, line 398: If this would throw an error in Torch, it can be omitted here; Torch would have caught it already

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -370,3 +372,108 @@ def index(
...
+    dim = get_positive_dim(dim, len(input_shape))
+    dynamic_shape = has_dynamic_shape(input.shape)
+    if dynamic_shape:
+        # Check whether slice target dim is dynamic shape dim
+        assert input.shape[dim] != -1, "Can't scatter on negative shape dimension!"
```
gs-olive, line 403: This should be checked in a converter validator, since it would otherwise cause a model failure

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -370,3 +372,108 @@ def index(
...
+
+    input_dims = len(input_shape)
+    for i in range(0, input_dims):
+        if i != dim and (index_shape[i] >= input.shape[i]):
+            raise RuntimeError(
+                f"cannot have index size greater than the input size along dimension {dim}"
+            )
```
gs-olive, line 410: Does this cause an error in Torch? If so, it can be omitted, otherwise, it should go in the validator

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -697,6 +697,32 @@ def aten_ops_clamp(
...
+def aten_ops_scatter_src(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.select.scatter_src(
+        ctx, target, SourceIR.ATEN, name, args[0], args[1], args[2], args[3]
+    )
```
gs-olive, line 723: It seems the main difference between `impl.select.scatter_src` and `impl.select.scatter_value` is the casting of the `value` tensor. Could these functions (converters and implementation) be merged to a single function that can either accept `float` or `TRTTensor`, and differentiates accordingly?

py/torch_tensorrt/dynamo/conversion/impl/select.py
apbose: This will be  required, since `index` can be a constant too.

py/torch_tensorrt/dynamo/conversion/impl/select.py
apbose: Should it be a converter validator? Since a validator would be some feature of the converter such that it will be supported later? But this is not so in this case.

gs-olive: In this case, even if there is no way to support the feature later, the check should still be in the validator so that we can fall back to Torch if this behavior is not supported

apbose:
The `harness.py` file is changed in this for the cases in which the index is passed in the `forward` function.
torch scatter function takes only int64 inputs for index
Couple of points
1. When passed as a variable outside the `forward` function, the `get_trt_tensor` function will cast to int64
2. When int64 is passed in the `forward` function, it throws error in the `unified_dtype_converter` in `fx/utils.py`. This is in `TRTInterpretor.py` while adding the placeholder nodes - `self.ctx.net.add_input`
3. Passing `truncate_long_and_double` in `compilationsettings` does not help since this would repair inputs in `_compiler.py` outside the TRTInterpreter and it would not help in harness.py
4. Had the tests been in the form of lowering tests, the torch forward function is called after the call to TRTInterpreter to invoke compilation, due to which the testing fails.

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -390,3 +391,52 @@ def index_select(
...
+    input_shape = input.shape
+    index_shape = index.shape
+    index_shape_list = list(index.shape)
+    if not (isinstance(index, TRTTensor)):
+        if isinstance(index, torch.Tensor):
+            if index.dtype == torch.int64:
+                index = index.to(torch.int32)
+        elif isinstance(index, np.ndarray):
+            if index.dtype == np.int64:
+                index = index.astype(np.int32)
```
gs-olive, line 415: Since TRT supports `int64` now, is this still necessary?

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -390,3 +391,52 @@ def index_select(
...
+        index = get_trt_tensor(ctx, index, f"_index_tensor")
+    dim = get_positive_dim(dim, len(input_shape))
+    dynamic_shape = has_dynamic_shape(input.shape)
+    if dynamic_shape:
+        # Check whether slice target dim is dynamic shape dim
+        assert input.shape[dim] != -1, "Can't scatter on negative shape dimension!"
```
gs-olive, line 421: This can be removed, and should be addressed by #2796

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -390,3 +391,52 @@ def index_select(
...
+
+    src_tensor = src
+    # scatter.value
+    if isinstance(src, int) or isinstance(src, float):
+        src_tensor = get_trt_tensor(
+            ctx, src * torch.ones(index_shape_list), name + "_value_tensor"
```
gs-olive, line 427: It is preferred to use `np.ones` here to avoid FakeTensor issues

py/torch_tensorrt/dynamo/conversion/impl/select.py
gs-olive: If not, the `enforce_tensor_types` could be used to autocast any input to an `ITensor`

tests/py/dynamo/conversion/harness.py
```diff
@@ -1,5 +1,4 @@
-# type: ignore
-
+import copy
```
gs-olive, line 1: Is this import used?

gs-olive (commented):
Overall looks good - added a few suggestions

py/torch_tensorrt/dynamo/conversion/impl/select.py
apbose: Looks like in the scatter converter layer in TRT still needs int32 since it complains of ERROR:torch_tensorrt [TensorRT Conversion Context]:4: [SCATTER]-[aten_ops.scatter.value]-[scatter_value_scatter_layer]: Indices tensor must be Int32.

tests/py/dynamo/conversion/harness.py
apbose: No will remove

py/torch_tensorrt/dynamo/conversion/impl/select.py
apbose: So after this PR is merged the above would need to be addressed accordingly?

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -21,6 +21,7 @@
     set_layer_name,
 )
 from torch_tensorrt.fx.types import Shape, TRTTensor
+from torch_tensorrt.fx.utils import Frameworks, unified_dtype_converter
```
apbose, line 24: Will remove this.

py/torch_tensorrt/dynamo/conversion/impl/select.py
gs-olive: Yes, I believe the above would have to be addressed in a future PR which adds dynamic shape support for this converter. With #2796, this converter would be registered as static-only by default

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -390,3 +391,41 @@ def index_select(
...
+    input: TRTTensor,
+    dim: int,
+    index: Union[TRTTensor, np.ndarray, torch.Tensor],
+    src: Union[TRTTensor, int, float],
+) -> TRTTensor:
+    input_shape = input.shape
+    index_shape = index.shape
+    index_shape_list = list(index_shape)
+    if index.dtype == trt.int64:
+        index = cast_trt_tensor(ctx, index, trt.int32, name + "_cast_index_tensor")
```
gs-olive, line 410: If the index must be `trt.int32` and no other types (`trt.float32`, `trt.float16`, etc.) are acceptable, then it is fine to remove the `if` statement, as the `cast_trt_tensor` function will not insert a cast if the type is already `int32`, as here:

https://github.com/pytorch/TensorRT/blob/4b993f8ee30fd02b7ab9cff47114a0538562cf81/py/torch_tensorrt/dynamo/conversion/converter_utils.py#L147

gs-olive (commented):
Added a small comment, otherwise looks good pending CI

py/torch_tensorrt/dynamo/conversion/impl/select.py
apbose: It would be required for `trt.int64` cases, which will be the case in our test cases since torch requires int64 inputs.

PR #2677, opened 2024-03-08
feat: support 1d ITensor offsets for embedding_bag converter

### LINKED ISSUE #2345: Implement embedding bag convertor

zewenli98:
# Description

Support 1d ITensor offsets for embedding_bag converter.
Note that this is only for 1d inputs.

There's a bug that same offsets with different types (tensor or ITensor) when include_last_offset=True will give different results. I doubt this is a bug from PyTorch, I'm still investigating it.

Fixes #2345 

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -263,22 +244,21 @@ def aten_ops_embedding(
         name,
         input=args[1],
         weight=args[0],
-        # args[2] is the padding index, which is useful for training only
-        scale_grad_by_freq=args_bounds_check(args, 3),
-        sparse=args_bounds_check(args, 4),
+        # args[2, 3, 4] are useful for training only
+        padding_idx=args_bounds_check(args, 2, -1),
+        scale_grad_by_freq=args_bounds_check(args, 3, False),
+        sparse=args_bounds_check(args, 4, False),
```
gs-olive, line 250: If these are training-only parameters (no effect at inference time), it is safe to omit them entirely (ignore those args). If they do have an effect at inference time, but we don't support that effect, then a validator can be used. It seems the validator is removed so I assume it is the former case.

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -541,3 +542,75 @@ def flatten_dims(
...
+
+def set_item(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    original_tensor: TRTTensor,
+    index: int,
+    new_value: Union[TRTTensor, int, float, torch.Tensor, np.ndarray],
+) -> TRTTensor:
```
gs-olive, line 577: This implementation could be very useful for other `setter`-style operators like `index_put` and the like, so consider adding documentation of the scope of this. For instance, does `index` refer to an index in the first dimension of the original tensor, or a rank in the Tensor, or something else? Additionally, what are the constraints on `new_value` here?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -263,22 +244,21 @@ def aten_ops_embedding(
...
+    indices = meta.get("tensor_meta")
+    if indices is None:
+        indices = meta.get("val")
     if indices is None:
         return False
-    return (
-        bool(node.args[2].op == "get_attr")
-        and (mode == 0 or mode == 1 or mode == 2)
-        and len(indices.shape) == 1
-    )
```
gs-olive, line 281: I believe this validator still needs to verify that none of the downstream users access the items at any indices other than index 1. Essentially, it needs the code from here:
https://github.com/pytorch/TensorRT/blob/7f14221ca532d8c42daa43b74fe600ebb27e1537/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py#L48-L54

gs-olive (commented):
Overall a great improvement to `embedding_bag` added a few small comments and suggestions

zewenli98:
Thanks for the review! I just refactored the embedding_bag with native TRT layers.

tests/py/dynamo/conversion/test_embedding_bag_aten.py
```diff
@@ -22,29 +75,181 @@ class TestEmbeddingBagConverter(DispatchTestCase):
...
                 mode=0,
                 sparse=False,
-                per_sample_weights=torch.randn((6,)),
+                per_sample_weights=None,
+                include_last_offset=True,
+                padding_idx=-1,
+            ),
+            param(
+                test_name="1d_indices_8",
+                weight=torch.randn((10, 4), dtype=torch.float32),
```
gs-olive, line 103: Could one more test case be added or an existing one modified to try `weight` having dtype `torch.half`? I noticed some of the constant tensors are instantiated as `np.float32` and want to verify this wouldn't be an issue

zewenli98: Do you mean some tensors are instantiated as `torch.float16`?

gs-olive: Yes, that's right - trying out with the weight tensors, specifically

zewenli98:
This is currently blocked by a Myelin bug. Filed a NVBug internally.

zewenli98:
With the latest `Pytorch`, `TRT-10 GA`, and `Torch-TRT main branch`. The `embedding_bag` converter works on DLRM now.

Issue #2665, opened 2024-02-28
❓ [Question] operator being decomposed rather than being converted when a corresponding converter exists?


HolyWu:
## ❓ Question

From the debug log below, it seems that the `aten.grid_sampler_2d` operator gets decomposed into several lower-level operators. But isn't there a corresponding [converter](https://github.com/pytorch/TensorRT/blob/9a100b6414bee175040bcaa275ecb71df54836e4/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py#L333-L358) which should be used?

## What you have already tried

```py
import torch
import torch.nn as nn
import torch.nn.functional as F
import torch_tensorrt


class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
        
    def forward(self, input, grid):
        return F.grid_sample(input, grid, mode="bilinear", padding_mode="border", align_corners=True)
    
model = MyModule().eval().cuda()

inputs = [
    torch.randn((1, 3, 8, 8), dtype=torch.float, device="cuda"),
    torch.randn((1, 16, 16, 2), dtype=torch.float, device="cuda")
]

optimized_model = torch_tensorrt.compile(
    model,
    ir="dynamo",
    inputs=inputs,
    enabled_precisions={torch.float},
    debug=True,
    min_block_size=1,
    truncate_long_and_double=True,
    output_format="fx",
)
```

```
DEBUG:torch_tensorrt.dynamo.conversion.aten_ops_converters:_to_copy converter rejected node _to_copy with dtype torch.int64
DEBUG:torch_tensorrt.dynamo.conversion.aten_ops_converters:_to_copy converter rejected node _to_copy with dtype torch.int64
DEBUG:torch_tensorrt.dynamo.conversion.aten_ops_converters:_to_copy converter rejected node _to_copy_1 with dtype torch.int64
DEBUG:torch_tensorrt.dynamo.conversion.aten_ops_converters:_to_copy converter rejected node _to_copy_1 with dtype torch.int64
DEBUG:torch_tensorrt.dynamo.conversion.aten_ops_converters:_to_copy converter rejected node _to_copy_2 with dtype torch.int64
DEBUG:torch_tensorrt.dynamo.conversion.aten_ops_converters:_to_copy converter rejected node _to_copy_2 with dtype torch.int64
DEBUG:torch_tensorrt.dynamo.conversion.aten_ops_converters:_to_copy converter rejected node _to_copy_3 with dtype torch.int64
[...]

PR #2686, opened 2024-03-11
Removing grid lowering

apbose:
This PR addresses the issue #2665 
1. Removing grid_sampler_2d from lowering ops
2. Adding grid_sampler.default and grid_sampler_2d.default to the CONVERTER key registry
3. Adding tests for grid_sampler.default and grid_sampler_2d.default tests
4. Restructing the test
5. Correcting the interpolation mode and padding mode in the tests

### DIFF
--- a/tests/py/dynamo/conversion/test_grid_aten.py
+++ b/tests/py/dynamo/conversion/test_grid_aten.py
@@ -6,112 +6,74 @@ from parameterized import parameterized
 from torch.testing._internal.common_utils import run_tests
 from torch_tensorrt import Input
 
+grid_sampler_aten_ops = {
+    "torch.ops.aten.grid_sampler": torch.ops.aten.grid_sampler,
+    "torch.ops.aten.grid_sampler_2d": torch.ops.aten.grid_sampler_2d,
+    "torch.ops.aten.grid_sampler.default": torch.ops.aten.grid_sampler.default,
+    "torch.ops.aten.grid_sampler_2d.default": torch.ops.aten.grid_sampler_2d.default,
+}
+
 grid_sampler_ops = [
     (
         "input_grid_interpolation_nearest_sample_fill",
-        (lambda x, grid: torch.ops.aten.grid_sampler(x, grid, 0, 0, True)),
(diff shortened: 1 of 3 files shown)


apbose:
PR #2682 seems to adddress partially the above.

tests/py/dynamo/conversion/test_grid_aten.py
```diff
@@ -126,19 +88,98 @@ class TestGridConverter(DispatchTestCase):
...
+    )
+    def test_grid(self, _, op_name, op, input_shape, dim_shape):
+        class TestModule(nn.Module):
+            def __init__(self, grid_sampler_op):
+                super().__init__()
+                self.grid_sampler_op = grid_sampler_op
+
+            def forward(self, x):
+                grid = torch.randint(-1, 1, dim_shape, dtype=torch.float32)
+                return self.grid_sampler_op(x, grid, grid_sampler_aten_ops[op_name])
```
gs-olive, line 104: Would this need to be `grid_sampler_aten_ops[op_name](x, grid)`?

apbose: @gs-olive yes it could be but then in that case it would be something like-
```
   def test_grid(self, _, op_name, input_shape, dim_shape, padding_mode, interpolation_mode, align_corners):
        class TestModule(nn.Module):
            def __init__(self):
                super().__init__()

            def forward(self, x):
                grid = torch.randint(-1, 1, dim_shape, dtype=torch.float32)
                return grid_sampler_aten_ops[op_name](x, grid, padding_mode, interpolation_mode, align_corners)

        inputs = [torch.randn(input_shape, dtype=torch.float32)]
        grid_model = TestModule()
        self.run_test(grid_model, inputs)
```
It would be code design choice to either  encapsulate it in the lambda function in the parameters, or declare it in the above way. In my opinion both should be good.

You could let me know if you think otherwise.

gs-olive: I see - this makes sense, thanks. In that case, it seems good to go - should just need a rebase to enable testing.

Issue #2586, opened 2024-01-09
aten.atan2


narendasan:
Implement a converter for aten.atan2 

https://pytorch.org/docs/stable/generated/torch.atan2.html

PR #2689, opened 2024-03-13
feat: support aten.atan2 converter

chohk88:
# Description

New feature to support aten.atan2 converter. I also add test case including edge case for inputs are zero. 

Fixes # ([issue](https://github.com/pytorch/TensorRT/issues/2586))

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1392,6 +1392,12 @@ def aten_ops_atanh(
 
 
 @dynamo_tensorrt_converter(torch.ops.aten.atan2.default)
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
+        1: (TRTTensor,),
+    }
+)
 def aten_ops_atan2(
     ctx: ConversionContext,
     target: Target,
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1391,6 +1391,24 @@ def aten_ops_atanh(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.atan2.default)
+def aten_ops_atan2(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.elementwise.atan2(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+    )
+
+
 @dynamo_tensorrt_converter(torch.ops.aten.ceil.default)
 def aten_ops_ceil(
     ctx: ConversionContext,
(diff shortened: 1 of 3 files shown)


zewenli98 (commented):
Overall looks good to me. It seems that `get_value_by_condition` is a helper function. Can you move it to `.../dynamo/conversion/converter_utils.py`?

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
```diff
@@ -213,6 +216,178 @@ def remainder(
     return fmod2_value
 
 
+def atan2(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    input: TRTTensor,
+    other: TRTTensor,
```
gs-olive, line 225: Does this function support `other` being a Python float or does it require `TRTTensor`?

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
```diff
@@ -556,3 +731,17 @@ def le(
...
+    condition: TRTTensor,
+) -> TRTTensor:
+    select_layer = ctx.net.add_select(condition, input, other)
+    set_layer_name(select_layer, target, name + "_select", source_ir)
+    return select_layer.get_output(0)
```
gs-olive, line 747: This can be removed - replace with usage of `impl.condition.select`, from here:
https://github.com/pytorch/TensorRT/blob/76c9ebdc7544f7f666b63e21d13abfc9e2d44c5d/py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py#L102-L113

gs-olive: Alternatively, you may need `where`:
https://github.com/pytorch/TensorRT/blob/76c9ebdc7544f7f666b63e21d13abfc9e2d44c5d/py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py#L18-L26

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1391,6 +1391,24 @@ def aten_ops_atanh(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.atan2.default)
```
gs-olive, line 1394: If you require one or both input tensors to be `TRTTensor`, consider using:
https://github.com/pytorch/TensorRT/blob/6b8ccbc69f05de3b6275cac4613735934fb559b0/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py#L62-L66

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
chohk88: I think it requires `TRTTensor` because both input and output are `Tensor`

image

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
chohk88: I add `enforce_tensor_types` for both input and other.

py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py
chohk88: I removed `get_value_by_condition` and replaced it with `impl.condition.select.`

chohk88:
@zewenli98 @gs-olive I've made the changes you suggested. I also noticed I missed the case when input or other is a single constant, so I fixed that using `broadcast` function. And, I added test cases to check it's all working right. Thanks for your helpful suggestions!

narendasan (approved):
LGTM

PR #2693, opened 2024-03-15
Selectively enable different frontends

narendasan:
# Description

Allows users to configure which frontends/features they want to include in the `torch-tensorrt` builds.

Features can be enabled as so: 
```py
NO_TORCHSCRIPT=1 pip install -e . ... # Includes the C++ runtime but no Torchscript frontend
PYTHON_ONLY=1 pip install -e . ... # No C++ dependencies at all, a pure python package 
```

A builds feature set can be accessed via the following struct 

```py
import torch_tensorrt

torch_tensorrt.ENABLED_FEATURES
```

where `ENABLED_FEATURES` is a `namedtuple` `FeatureSet`: 
```py
namedtuple('FeatureSet', [
    "torchscript_frontend",
    "torch_tensorrt_runtime",
    "dynamo_frontend",
    "fx_frontend"
])
```

In order to support optional features, a number of core types have been abstracted: 

- `torch_tensorrt.Device` has no direct dependencies on the torchscript core and can be translated to `torch_tensorrt.ts.Device` to access those features
- Similarly `torch_tensorrt.Input` behaves the same way
- enums for `dtype`,`DeviceType`, `memory_format` have been defined and can translate from `numpy`, `tensorrt`, `torch` and `torch_tensorrt._C` (assuming `torch_tensorrt.ENABLED_FEATURES.torchscript_frontend` is `True`)

Translating between different library enums now can take the form
```py
import torch
import tensorrt as trt
import torch_tensorrt

# Convert, throw error if no conversion exists
trt_dtype = torch_tensorrt.dtype._from(torch.float16).to(trt.DataType) # returns trt.DataType.HALF

#Try to convert, do not throw an error if there is no conversion, just return None
trt_dtype = torch_tensorrt.dtype.try_from(torch.float16).to(trt.DataType) # returns trt.DataType.HALF

# Alternatively, use a fallback type
trt_dtype = torch_tensorrt.dtype.unknown.to(trt.DataType, use_default=True) #returns trt.DataType.FLOAT
```

Fixes #1943
Fixes #2379 
 

## Type of change

Please delete options that are not relevant and/or add your own.

- New feature (non-breaking change which adds functionality)

# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas and hacks
- [x] I have made corresponding changes to the documentation
- [x] I have added tests to verify my fix or my feature
[...]

PR #2702, opened 2024-03-21
fix: Windows CI Experimentation

gs-olive:
Experimentation of windows CI job with upgraded C++ build and package runtime changes


HolyWu:
My previous commit can only build with TRT 8. To build with TRT 9 there are some necessary changes in https://github.com/HolyWu/TensorRT/commit/f644baa4f073afcb7e27ee655d62771607c4a7c2. Besides, if you prefer building from archives rather than using local sources, it's also doable in https://github.com/HolyWu/TensorRT/commit/b54a3bf9962b91a22f445596a53e24e7be740bfe.

HolyWu:
I don't think you can install/upgrade the display driver via conda because the nvidia packages in conda seems to not include the display driver. Only the official [CUDA Toolkit installer](https://developer.nvidia.com/cuda-12-1-1-download-archive) includes the display driver component. I'm not sure if you can successfully [execute the network installer in silent mode](https://docs.nvidia.com/cuda/archive/12.1.1/cuda-installation-guide-microsoft-windows/index.html#install-the-cuda-software) to only install the display driver on the windows runner, because it may require admin privilege or rebooting.

On the other hand, maybe poke the pytorch guys to upgrade the driver on their windows runner is a better solution.

gs-olive:
Thanks for this feedback - I am looking into getting the driver version for Windows updated [here](https://github.com/pytorch/builder/blob/89699bc54ceb04e687bb3b5c2d7d7238e25e214e/windows/internal/driver_update.bat#L1).

HolyWu:
The default shell on Windows runner is `powershell`. You need to change the shell to `bash` like https://github.com/gs-olive/TensorRT/blob/2b25c988970a8e4282cf5f5e58a2a116e74436b0/.github/workflows/windows-test.yml#L137 since most commands in the workflow are not supported by `powershell`.

NevermindNilas:
> @gs-olive I have managed to fix the CI errors in my [fork](https://github.com/HolyWu/TensorRT/tree/windows_CI_experimentation_2), including [successful C++ building](https://github.com/HolyWu/TensorRT/actions/runs/8602183974). Because I cannot use PyTorch-hosted runner in my fork, I have to run on GitHub-hosted runner. But GitHub-hosted runner doesn't have GPU support, so I cannot run pytest.
> 
> Some things you need to do besides reverting back to use pytorch's repo in [HolyWu@fa7ecc9](https://github.com/HolyWu/TensorRT/commit/fa7ecc973d58c960fd4d0f8ecb05467f772c583a):
> 
> * [build-test-windows.yml#L3](https://github.com/HolyWu/TensorRT/blob/41237cedd56809062ad8e29ed0a18fabd54dd3e1/.github/workflows/build-test-windows.yml#L3): change to `pull_request` only.
> * [build-test-windows.yml#L184](https://github.com/HolyWu/TensorRT/blob/41237cedd56809062ad8e29ed0a18fabd54dd3e1/.github/workflows/build-test-windows.yml#L184): uncomment
> * [windows-test.yml#L56](https://github.com/HolyWu/TensorRT/blob/41237cedd56809062ad8e29ed0a18fabd54dd3e1/.github/workflows/windows-test.yml#L56-L57): change to use PyTorch's runner.
> * [windows-test.yml#L92](https://github.com/HolyWu/TensorRT/blob/41237cedd56809062ad8e29ed0a18fabd54dd3e1/.github/workflows/windows-test.yml#L92-L93): change to `false`.
> * [pre_build_script_windows.sh#L13](https://github.com/HolyWu/TensorRT/blob/41237cedd56809062ad8e29ed0a18fabd54dd3e1/packaging/pre_build_script_windows.sh#L13-L14): change to only install driver since PyTorch's runner should already have CUDA Toolkit installed.
> * [smoke_test_windows.py#L33](https://github.com/HolyWu/TensorRT/blob/41237cedd56809062ad8e29ed0a18fabd54dd3e1/packaging/smoke_test_windows.py#L33): uncomment.
> * [WORKSPACE.win.release.tmpl#L46](https://github.com/HolyWu/TensorRT/blob/41237cedd56809062ad8e29ed0a18fabd54dd3e1/toolchains/ci_workspaces/WORKSPACE.win.release.tmpl#L46): change to `${CUDA_HOME}`? I'm not sure whether PyTorch's Windows runner have that variable exported.

With this we can assume windows wheels will be coming soon????

HolyWu:
Don't forget to set `setup-miniconda` to false in `windows-test.yml` when using pytorch's Windows runner.

HolyWu:
Hmm...maybe you need to set https://github.com/gs-olive/test-infra/blob/0c5cc5b410cfaa6953a15fdd564c2399a95b185e/.github/workflows/build_wheels_windows.yml#L80 back to `windows-2019`? It's the culprit of the `Called workflows cannot be queued onto self-hosted runners across organizations/enterprises.` error.

gs-olive:
That worked - thank you! Now looking into why this error is appearing during install:
```
ERROR: torch_tensorrt-2.3.0.dev0+6031b50-cp39-cp39-win_amd64.whl is not a supported wheel on this platform.
```
It seems that even though the builds specify `py38`, they are building for `py39`, which might be contributing

HolyWu:
> That worked - thank you! Now looking into why this error is appearing during install:
> 
> ```
> ERROR: torch_tensorrt-2.3.0.dev0+6031b50-cp39-cp39-win_amd64.whl is not a supported wheel on this platform.
> ```
> 
> It seems that even though the builds are named `py38`, they are building for `py39`, which might be contributing

Ah...the reason might be [this](https://github.com/gs-olive/test-infra/blob/30af0234835077d084d46b047fd52ff1f1f2df3a/.github/actions/setup-binary-builds/action.yml#L88-L89). I think you can revert to use pytorch's test-infra instead of self-forked one, since pytorch's runner is accessible to this repo, unlike mine.

Also [tests-py-torch-compile-be](https://github.com/gs-olive/TensorRT/blob/a0ec0e2c90e764568db23328f6e32067f11a1b01/.github/workflows/build-test-windows.yml#L159) needs to add pre-script just like [this](https://github.com/gs-olive/TensorRT/blob/a0ec0e2c90e764568db23328f6e32067f11a1b01/.github/workflows/build-test-windows.yml#L26-L37) so as to update display driver on pytorch's Windows runner.

HolyWu:
Weird. I thought `tensorrt-bindings` and `tensorrt-libs` would be automatically installed when installing `tensorrt` itself. Probably also adding `tensorrt-bindings==9.3.0.post12.dev1` and `tensorrt-libs==9.3.0.post12.dev1` to [here](https://github.com/gs-olive/TensorRT/blob/0e8ae1ee4188c319f0b64c3ec6b6980d33d3dd6a/.github/scripts/install-torch-tensorrt-windows.sh#L4) for guarantee.

gs-olive:
Thanks @HolyWu! First batch of Python-runtime-only E2E tests seem to be passing now in CI for Python 3.9 + cu121! If I revert to the mainline `pytorch/test-infra`, the `py3.8`/`py3.9` issue may come back, however. I will test it out

HolyWu:
Should `use_python_runtime=True` be set in `compile_spec` for the other enabled tests?

gs-olive:
Since this branch is based on `min_cpp_build` (#2693), it should be automatically detecting whether the C++ runtime is present and selecting the preferable runtime anyway. I can likely remove that flag now from the tests to which it was added

HolyWu:
It seems updating driver via CUDA Toolkit installer occasionally fails for unknown reason on CI. It's probably better to use [driver_update.bat](https://github.com/pytorch/builder/blob/main/windows/internal/driver_update.bat) instead. Note that don't simply copy the content from the bat to `pre_build_script_driver_upgrade.sh` but just use the bat itself, because it behaves differently.

HolyWu:
Try adding a line of `PYTHONUTF8: 1` inside [env](https://github.com/gs-olive/TensorRT/blob/a4fa67fcea4cd9808ec2cb3539ba366173c94dec/.github/workflows/windows-test.yml#L48) to eliminate the `UnicodeEncodeError`.

PR #2740, opened 2024-04-09
scatter_add_decomposition

apbose:
The issue in this PR is for cases where there is index collision. Example of such cases-
`scatter_add(input_tensor = torch.zeros(3,5), dim=1, index_tensor = torch.tensor([[0, 1, 2, 0]]).cuda(), src_tensor = torch.tensor([[1, 2, 3, 1]], dtype=torch.int32).cuda())`
Looks like straighforward decomposition by stacking the src_tensor would not work, since it would again lead to collision for some cases.
The better way to implement is to do converter implementation and allot the values using advanced indexing.
Working on this now.

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -243,6 +243,37 @@ def empty_strided_decomposition(*args, **kwargs) -> torch.Tensor:
     )
 
 
+@register_torch_trt_decomposition(
+    torch.ops.aten.scatter_add.default, registry=TORCH_TRT_DECOMPOSITIONS
+)
+def scatter_add_decomposition(
+    input_tensor: torch.Tensor,
+    src_tensor: torch.Tensor,
+    dim: int,
+    index: torch.Tensor,
+) -> torch.Tensor:
+    scatter_add_tensor = input_tensor
+    src_copy = src_tensor
+    src_shape = list(src_tensor.shape)
+    del src_shape[dim]
+    select_src_dim = src_copy.shape[dim]
+    to_stack_dummy_src = tuple(torch.empty(src_shape) for _ in range(select_src_dim))
+    for index_src_dim in range(0, select_src_dim, 1):
+        select_tensor_dim = torch.select(src_copy, dim, index_src_dim)
+        to_stack_src = to_stack_dummy_src
+        if(index_src_dim == 0):
+            to_stack_src = (select_tensor_dim.cpu(),) + to_stack_dummy_src[index_src_dim+1:] 
+        elif(index_src_dim == select_src_dim - 1 ):
+            to_stack_src = to_stack_dummy_src[:index_src_dim] + (select_tensor_dim.cpu(),)
+        else:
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -174,6 +174,19 @@ def empty_permuted_decomposition(*args, **kwargs) -> torch.Tensor:
...
+@register_torch_trt_decomposition(
+    torch.ops.aten.scatter_add.default, registry=TORCH_TRT_DECOMPOSITIONS
+)
+def scatter_add_decomposition(
+    input_tensor: torch.Tensor,
+    src_tensor: torch.Tensor,
+    dim: int,
+    index: torch.Tensor,
+) -> torch.Tensor:
+    input_tensor_to_add = torch.scatter(torch.empty_like(input_tensor), dim, index, src_tensor)
```
gs-olive, line 186: Could `torch.empty_like(input_tensor)` instead be just `input_tensor`, or is it required to be empty?

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -174,6 +174,19 @@ def empty_permuted_decomposition(*args, **kwargs) -> torch.Tensor:
...
+    scatter_add_tensor = torch.add(input_tensor, input_tensor_to_add.cuda())
```
gs-olive, line 187: Without `.cuda()` does the decomposition fail?

py/torch_tensorrt/dynamo/lowering/_decompositions.py
apbose: Yes it would be required. The logic is like we would apply scatter with the indices to an empty tensor and then add it to the input tensor.

py/torch_tensorrt/dynamo/lowering/_decompositions.py
apbose: Yes in this case it complains that the two tensors are on different devices.

apbose:
@gs-olive though the test cases pass, I don't think that the decomposition is taking place.
It fails in the backend at `aot_export_joint_simple`. Shows me something like

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -243,6 +243,44 @@ def empty_strided_decomposition(*args, **kwargs) -> torch.Tensor:
...
 
+@register_torch_trt_decomposition(
+    torch.ops.aten.scatter_add.default, registry=TORCH_TRT_DECOMPOSITIONS
+)
+def scatter_add_decomposition(
+    input_tensor: torch.Tensor,
+    src_tensor: torch.Tensor,
+    dim: int,
+    index: torch.Tensor,
+) -> torch.Tensor:
```
zewenli98, line 254: It seems the order of args doesn't match with tests, which made CI failed.

apbose: Thanks for pointing this out! Corrected.

zewenli98 (approved):
LGTM

PR #2745, opened 2024-04-12
empty_memory_format evaluator

apbose:
In this PR I am facing issue in the test case of 
```
    (
        "empty_four_dimension_memformat",
        [1, 2, 2, 1],
        torch.float32,
        "cuda",
        torch.channels_last,
    ),
    (
        "empty_five_dimension_memformat",
        [1, 2, 2, 2, 1],
        torch.float32,
        "cuda",
        torch.channels_last_3d,
    ),
```
In the `TRTInterpreter.run()`, the stride of the empty_tensor  
`torch.ops.aten.empty.memory_format([1,2,2,1], dtype = torch.int32, memory_format = torch.channels_last)` 
is (4,1,2,2) (the desired torch output. This is an evaluator so could check the output in this [line](https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py#L308) - 
However when it comes to [line](https://github.com/pytorch/TensorRT/blob/main/tests/py/dynamo/conversion/harness.py#L149) the stride is (4,2,1,1). This is the output from the TRT engine from the the above INetwork.

Would anyone know what is going wrong?

fixes: #2738

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
+++ b/py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
@@ -18,6 +18,7 @@ from torch_tensorrt.dynamo.conversion.converter_utils import (
 )
 from torch_tensorrt.dynamo.conversion.impl.elementwise import sub, trunc_div
 from torch_tensorrt.fx.types import TRTTensor
+from torch_tensorrt.fx.utils import Frameworks, unified_dtype_converter
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -118,3 +119,50 @@ def aten_ops_randperm(
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     return np.random.permutation(args[0])
+
+
+def empty_validator(empty_node: Node) -> bool:
+    layout = empty_node.kwargs.get("layout", None)
+    pin_memory = empty_node.kwargs.get("pin_memory", None)
+    memory_format = empty_node.kwargs.get("memory_format", None)
```
zewenli98, line 127: It seems `pin_memory` and `memory_format` are not used in the validator.

tests/py/dynamo/conversion/test_empty_aten.py
```diff
@@ -0,0 +1,126 @@
...
+        torch.float32,
+        "cuda",
+        torch.channels_last_3d,
+    ),
+]
+
+
+class TestRandConverter(DispatchTestCase):
+    @parameterized.expand(
+        [(empty_op[0], empty_op[1], empty_op[2], empty_op[3]) for empty_op in empty_ops]
```
zewenli98, line 78: Since each test has 5 arguments, do we need `empty_op[4]` here?

tests/py/dynamo/conversion/test_empty_aten.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    def test_empty(self, name, shape_or_input, data_type, device):
+        class TestModule(nn.Module):
+            def __init__(self):
+                super().__init__()
+
+            def forward(self, x):
+                shape_or_input[0] = x.shape[0]
+                return torch.empty(shape_or_input)
```
zewenli98, line 87: For consistence, I think it's better to use `torch.ops.aten.empty.memory_format` with necessary args.

tests/py/dynamo/conversion/test_empty_aten.py
```diff
@@ -0,0 +1,126 @@
...
+        torch.channels_last_3d,
+    ),
+]
+
+
+class TestRandConverter(DispatchTestCase):
+    @parameterized.expand(
+        [(empty_op[0], empty_op[1], empty_op[2], empty_op[3]) for empty_op in empty_ops]
+    )
+    def test_empty(self, name, shape_or_input, data_type, device):
```
zewenli98, line 80: It seems `data_type` and `device` are not tested.

tests/py/dynamo/conversion/test_empty_aten.py
```diff
@@ -0,0 +1,126 @@
+import numpy as np
+import torch
+import torch.nn as nn
+import torch_tensorrt
+from harness import DispatchTestCase
```
zewenli98, line 5: I think it should be `from .harness import DispatchTestCase`

zewenli98 (changes requested):

tests/py/dynamo/conversion/test_empty_aten.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class TestEmptyConverter(DispatchTestCase):
+    @parameterized.expand(
+        [(empty_op[0], empty_op[1], empty_op[2]) for empty_op in empty_ops]
+    )
+    def test_empty(self, name, shape_or_input, data_type):
+        class TestModule(nn.Module):
+            def __init__(self):
+                super().__init__()
```
zewenli98, line 56: The `init` seems not necessary.

zewenli98 (approved):
Just a minor comment. The other LGTM

Issue #2743, opened 2024-04-11
`aten._local_scalar_dense`


chohk88:
The `torch.ops.aten._local_scalar_dense` operation is a low-level operation in PyTorch used internally to extract a scalar value from a tensor containing a single element.

PR #2752, opened 2024-04-16
feat: support aten._local_scalar_dense converter

chohk88:
# Description

A converter for the torch.ops.aten._local_scalar_dense operation, a low-level operation in PyTorch used internally to extract a scalar value from a tensor containing a single element.

To reviewer, I did not incorporate type-specific considerations in the implementation of the converter and test code. However, all test cases have successfully passed. I would greatly appreciate any feedback or suggestions regarding this aspect—especially if there are potential implications or improvements I might have overlooked. 

image

Fixes # ([issue](https://github.com/pytorch/TensorRT/issues/2743))

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1714,6 +1714,23 @@ def aten_ops_isnan(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten._local_scalar_dense.default)
+def aten_ops_local_scalar_dense(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.unary.local_scalar_dense(
+        ctx,
+        target,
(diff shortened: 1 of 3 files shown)


chohk88:
I would appreciate your feedback on the type-related issues mentioned in the description.

zewenli98:
@chohk88 @narendasan I checked the aten schema where there are only two funcs return `Scalar` type, i.e.,
```
- func: item(Tensor self) -> Scalar
- func: _local_scalar_dense(Tensor self) -> Scalar
```
I found some people have similar issues like: https://github.com/pytorch/xla/issues/909 
Their cases are moving data from XLA to CPU, so I'm thinking if this op is only called when moving data to CPUs. 

For the functionality, my understanding to this function is to extract the first element in the input, whatever it is a single tensor or list of tensors, and then move it to CPUs. If true, I don't think it's meaningful to implement this converter. Instead, we can have a lowering pass to remove this op to avoid potential graph breaks.

Do you know why this converter is needed? or is this op used in any models?

narendasan:
Needs #2810 to be implemented to fully work

peri044 (commented):
This looks good to me. Seems like all this op does is to gather the first element. 
> To reviewer, I did not incorporate type-specific considerations in the implementation of the converter and test code.

The type should always be the input data type correct ?

PR #2763, opened 2024-04-18
feat: Implement FP8 functionality

peri044:
# Description

This PR adds FP8 & BF16 datatype support. It also implements converter for FP8 quantized ops. 

## Type of change

# Checklist:


zewenli98:
@peri044 I remember we've already removed `cudnn` dependency on the release/2.3, but it still stays here: https://github.com/pytorch/TensorRT/blob/3f6999d6ab2f9b62c63b78e8405cacf216370214/py/torch_tensorrt/__init__.py#L9

This causes an error when I import torch-trt:
```
>>> import torch_tensorrt
  File "/home/scratch.zewenl_sw/docker_workspace/TRT10/TensorRT/py/torch_tensorrt/__init__.py", line 7, in <module>
    from torch_tensorrt._version import (  # noqa: F401
ImportError: cannot import name '__cudnn_version__' from 'torch_tensorrt._version' (/home/scratch.zewenl_sw/docker_workspace/TRT10/TensorRT/py/torch_tensorrt/_version.py)
```

Can you take a look?

peri044:
> @peri044 I remember we've already removed `cudnn` dependency on the release/2.3, but it still stays here:
> 
> https://github.com/pytorch/TensorRT/blob/3f6999d6ab2f9b62c63b78e8405cacf216370214/py/torch_tensorrt/__init__.py#L9
> 
> This causes an error when I import torch-trt:
> 
> ```
> >>> import torch_tensorrt
>   File "<stdin>", line 1, in <module>
>   File "/home/scratch.zewenl_sw/docker_workspace/TRT10/TensorRT/py/torch_tensorrt/__init__.py", line 7, in <module>
>     from torch_tensorrt._version import (  # noqa: F401
> ImportError: cannot import name '__cudnn_version__' from 'torch_tensorrt._version' (/home/scratch.zewenl_sw/docker_workspace/TRT10/TensorRT/py/torch_tensorrt/_version.py)
> ```
> 
> Can you take a look?

fixed it

zewenli98:
Cool thanks! And did you implement the unit test for `torch.ops.trt.quantize_fp8.default`?

examples/int8/training/vgg16/main.py
```diff
@@ -167,6 +173,50 @@ def main():
                 ckpt_dir=args.ckpt_dir,
             )
 
+    if args.fp8_epochs > 0:
```
peri044, line 176: Let's move this into a separate file. 
1) Create a vgg16_fp8_ptq file in `examples/dynamo` directory ([reference](https://github.com/pytorch/TensorRT/blob/68a5892021d9d273c8cbb67777d664bd2b55f716/examples/dynamo/torch_compile_stable_diffusion.py))

2) This script should instantiate a model and load regular checkpoint weights. ([example](https://github.com/pytorch/TensorRT/blob/c21866bcb84b3cb3212ce4007a6dbad8196d2c31/examples/int8/training/vgg16/export_ckpt.py#L43-L58))
3) Then perform the `model mtq.quantize(model, quant_cfg, forward_loop=calibrate_loop)` with the calibrate loop loading examples from from training dataset.
4) Print out the accuracy at the end. 
The final demo should be rendered similar to https://pytorch.org/TensorRT/tutorials/_rendered_examples/dynamo/torch_compile_resnet_example.html
So the training code will stay in `examples/int8/training/vgg16` directory while the PTQ demo will be in `examples/dynamo/`

zewenli98:
@peri044 Thanks for the comments. I have refactored based on your suggestions.

examples/dynamo/vgg16_fp8_ptq.py
```diff
@@ -0,0 +1,261 @@
...
+        512,
+        "pool",
+    ]
+    return VGG(vgg16_cfg, num_classes, init_weights)
+
+
+PARSER = argparse.ArgumentParser(
+    description="Load pre-trained VGG model and then tune with FP8 and PTQ"
+)
+PARSER.add_argument("--ckpt", type=str, help="Path to the pre-trained checkpoint")
```
peri044, line 105: set required=True here

examples/dynamo/vgg16_fp8_ptq.py
```diff
@@ -0,0 +1,261 @@
...
+    default=128,
+    type=int,
+    help="Batch size for tuning the model with PTQ",
+)
+PARSER.add_argument(
+    "--fp8-epochs",
+    default=10,
+    type=int,
+    help="The number of epochs to quantize the model to FP8",
+)
```
peri044, line 117: We don't need this right ?

examples/dynamo/vgg16_fp8_ptq.py
```diff
@@ -0,0 +1,261 @@
...
+crit = nn.CrossEntropyLoss()
+
+# %%
+# Define Calibration Loop for quantization
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+def calibrate_loop(model):
+    # calibrate on a small number of batches
+    for fp8_ep in range(args.fp8_epochs):
```
peri044, line 177: I think this should be only 1 epoch, so we don't need the argument. Calibration only happens once on the dataset (or subset of it), we don't need to iterate it for all epochs

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
```diff
@@ -0,0 +1,46 @@
...
+    input_tensor: TRTTensor,
+    scale: np.ndarray,
+) -> TRTTensor:
+    """
+    Adds quantize and dequantize ops (QDQ) which quantize to INT8 or FP8 based
+    on the output_type set and dequantizes them back.
+    """
+    if (isinstance(input_tensor, TRTTensor)) and not (
+        input_tensor.dtype != trt.float32 or input_tensor.dtype != trt.float16
+    ):
```
gs-olive, line 27: Consider changing to `not (input_tensor.dtype == trt.float32 or input_tensor.dtype == trt.float16)`. Current logic seems to always throw an error.

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
```diff
@@ -0,0 +1,46 @@
...
+        raise ValueError(
+            f"quantize_fp8 converter received an input of {input_tensor.dtype} type. Supported types: float32 | float16"
+        )
+
+    if isinstance(scale, np.ndarray):
+        scale = get_trt_tensor(ctx, scale, name + "_scale")
```
gs-olive, line 33: Can `scale` also be a TRT tensor? If so, can the function header schema be changed to `Union[np.ndarray, TRTTensor]`

py/torch_tensorrt/dynamo/lowering/_remove_sym_nodes.py
```diff
@@ -10,17 +10,18 @@ def remove_sym_nodes(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
...
     # Extract SymInt placeholder Tensors
-    placeholders = [
+    placeholder_sym_ints = [
         node
         for node in gm.graph.nodes
         if (
             node.op == "placeholder"
             and isinstance(node.type, type)
             and issubclass(node.type, torch.SymInt)
+            and not node.users
```
gs-olive, line 20: Do `SymInt` inputs always have 0 users?

py/torch_tensorrt/dynamo/lowering/passes/remove_detach.py
```diff
@@ -0,0 +1,25 @@
...
+    for node in gm.graph.nodes:
+        # node.target = "detach" in torch.compile workflow
+        if node.target == torch.ops.aten.detach.default or node.target == "detach":
+            # Detach node has only one input
+            node_input = node.all_input_nodes[0]
+            node.replace_all_uses_with(node_input)
+            gm.graph.erase_node(node)
+            count += 1
+
+    logger.debug(f"Removed {count} detach nodes:\n{gm.graph}")
```
gs-olive, line 23: Consider replacing with the following, to clean up graph + avoid logging if no changes were made to the graph
```python
    if count > 0:
        gm = clean_up_graph_after_modifications(gm)
        logger.debug(f"Removed {count} detach nodes:\n{gm.graph}")

    return gm
```

gs-olive (commented):
Overall looks good! Added a few comments

py/torch_tensorrt/dynamo/lowering/_remove_sym_nodes.py
peri044: Nope they have users. But there's one no-user sym_int node added by torch.compile workflow when we do `torch._dynamo.mark_dynamic` and this is removing that specific node.

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
peri044: scale can only be constants.

py/torch_tensorrt/dynamo/lowering/passes/remove_detach.py
peri044: I've intentionally not used it since `clean_up_graph_after_modifications` will invoke `gm.recompile()` which I'm avoiding. The next step of this pass is getting ep.module() which calls recompile inside it.

py/torch_tensorrt/_enums.py
```diff
@@ -24,9 +23,9 @@ class dtype(Enum):
     f32 = auto()
     f64 = auto()
     b = auto()
-    # TODO: Enable FP8 and BF16
-    # f8 = auto()
-    # bf16 = auto()
+
+    f8 = auto()
+    bf16 = auto()
```
narendasan, line 28: Make sure this is aligned with main, otherwise the enum values would change version to version

py/torch_tensorrt/dynamo/_defaults.py
```diff
@@ -27,7 +27,7 @@
 REQUIRE_FULL_COMPILATION = False
 DRYRUN = False
 HARDWARE_COMPATIBLE = False
-SUPPORTED_KERNEL_PRECISIONS = {dtype.f32, dtype.f16, dtype.i8}
+SUPPORTED_KERNEL_PRECISIONS = {dtype.f32, dtype.f16, dtype.i8, dtype.f8}
```
narendasan, line 30: Missing bf16 here, can you just use a cherrypick of main?

narendasan: https://github.com/pytorch/TensorRT/pull/2845

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -579,6 +579,34 @@ def aten_ops_neg(
     )
 
 
+try:
+    import modelopt.torch.quantization as mtq
+
+    assert torch.ops.trt.quantize_fp8.default
+except Exception as e:
+    _LOGGER.warning(
+        "Unable to import torch.ops.trt.quantize_fp8 op. Please install modelopt library (https://github.com/NVIDIA/TensorRT-Model-Optimizer?tab=readme-ov-file#installation) to register torch.ops.trt.quantize_fp8 op"
```
narendasan, line 588: Make this like, install model opt to add support for compiling quantized models

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
gs-olive: Can the `if isinstance` be removed then in that case, or are there situations where the `scale` might not be a numpy array?

py/torch_tensorrt/dynamo/_defaults.py
peri044: This will be added when we merge this https://github.com/pytorch/TensorRT/pull/2845 right ?

py/torch_tensorrt/dynamo/lowering/passes/_aten_lowering_pass.py
```diff
@@ -72,23 +79,35 @@ def add_lowering_pass(
...
+    logging.debug(
+        f"Invoking DynamoPassManager and applying lowering passes: {ATEN_POST_LOWERING_PASSES}"
+    )
+    return ATEN_POST_LOWERING_PASSES(gm, sample_inputs)
+
+
+def pre_export_lowering(
+    ep: torch.export.ExportedProgram, sample_inputs: Sequence[torch.Tensor]
+) -> torch.fx.GraphModule:
+    """Applies the lowering passes to a graph module after torch.export/ torch.compile and their decompositions, returns the modified GraphModule"""
```
zewenli98, line 102: `after` -> `before`?
I'm wondering what belong to `pre_lowering` and what belong to `post_lowering`?

peri044: remove_detach belongs to pre_lowering (which happens before decompositions and other lowering passes)

PR #2699, opened 2024-03-19
chore: Upgrade TensorRT version to TRT 10 EA

peri044:
# Description

Upgrade TensorRT version to TRT 10 EA

## Type of change

Please delete options that are not relevant and/or add your own.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

# Checklist:

- [ ] My code follows the style guidelines of this project (You can use the linters)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #2726, opened 2024-04-04
feat: support aten._cdist_forward converter

chohk88:
# Description

New feature to support aten._cdist_forward converter converter. I have added test cases ensuring compatibility with both matching and broadcasting input shapes.

Fixes # ([issue](https://github.com/pytorch/TensorRT/issues/2725))

## Type of change

- New feature (non-breaking change which adds functionality)

# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests to verify my fix or my feature
- [x] New and existing unit tests pass locally with my changes
- [x] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #2774, opened 2024-04-23
chore: Upgrade TensorRT version to TRT 10 EA (#2699)

peri044:
# Description

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/impl/elementwise/base.py
```diff
@@ -139,27 +143,51 @@ def convert_binary_elementwise(
...
-
-    lhs_val, rhs_val = broadcast(
-        ctx.net, lhs_val, rhs_val, f"{name}_lhs", f"{name}_rhs"
-    )
+    if has_dynamic_shape(lhs_val.shape) or has_dynamic_shape(rhs_val.shape):
+        lhs_val, rhs_val = broadcast(
+            ctx.net, lhs_val, rhs_val, f"{name}_lhs", f"{name}_rhs"
+        )
+    else:
+        lhs_val_shape = lhs_val.shape
```
chohk88, line 158: @peri044 

I followed the code changes to address the error caused by broadcasting, and it seems to be related to this issue.

If we perform element-wise operations on two inputs with shapes lhs `(2, 2, 14, 1, 5)` and rhs `(2, 1, 3, 5)`, the output of the element-wise operation layer should be `(2, 2, 14, 3, 5)`. 
However, with the current code, the output of the element-wise operation layer follows the larger rank of lhs_val and becomes `(2, 2, 14, 1, 5)`. 

As a result, an error that did not occur on the local machine was found during CI/CD (https://github.com/pytorch/TensorRT/pull/2726)

Is it necessary to make adjustments to fix this code?

peri044: Yes, it is necessary to fix this code as any converter that uses this would likely have that failure. It sounds like the fix would be broadcasting across mismatched dimensions instead of using rank heuristic. @zewenli98 probably you have an idea about these specific changes while fixing test_binary_ops ?

PR #2784, opened 2024-04-26
"empty_like" decomposition test correction

apbose:
This excludes the test of the values in `empty_like` and instead checks for equal shapes from torch and torchTRT


gs-olive (approved):
Looks good to me, pending CI verification

PR #2796, opened 2024-04-29
feat: Add validators for dynamic shapes in converter registration

peri044:
# Description

This PR does the following

1) The default behavior is all converters are marked `dynamic=False`. 
2) We conservatively/explicitly mark `dynamic=True` for the nodes we know that support dynamic shapes. 
3) If a converter is marked dynamic=True, nothing happens. Both static and dynamic shapes are supported. 
4) If a converter is marked dynamic=False, it checks if the node received dynamic inputs. If it received dynamic inputs, we disable this converter and the node will fallback to PyTorch. If it received static inputs, compilation proceeds fine with no warnings.
5) For static inputs to the converter, this flag doesn't affect anything. 

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2080,7 +2080,9 @@ def conv_param_validator(conv_node: Node) -> bool:
 
 
 @dynamo_tensorrt_converter(
-    torch.ops.aten.convolution.default, capability_validator=conv_param_validator
+    torch.ops.aten.convolution.default,
+    capability_validator=conv_param_validator,
+    dynamic=True,
```
gs-olive, line 2085: Small detail - consider switching to `allows_dynamic_shapes=True`

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
```diff
@@ -323,6 +329,18 @@ def __getitem__(
...
                 if isinstance(converters, (list, tuple)):
                     for candidate in converters:
+                        # TODO: Importing this here avoids circular import issue. One potential fix is moving this function into _ConverterRegistry file.
+                        from torch_tensorrt.dynamo.conversion.converter_utils import (
+                            dynamic_unsupported,
+                        )
+
+                        has_static_inputs = dynamic_unsupported(node)
+                        # If there are dynamic inputs but the converter doesn't support it explicitly, throw a warning.
+                        if not has_static_inputs and not candidate.dynamic:
```
gs-olive, line 339: Switch to `if not candidate.dynamic and not dynamic_unsupported(node):`, to run the converter boolean check first and avoid the function call in most cases.

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
```diff
@@ -323,6 +329,18 @@ def __getitem__(
 
                 if isinstance(converters, (list, tuple)):
                     for candidate in converters:
+                        # TODO: Importing this here avoids circular import issue. One potential fix is moving this function into _ConverterRegistry file.
+                        from torch_tensorrt.dynamo.conversion.converter_utils import (
+                            dynamic_unsupported,
+                        )
+
+                        has_static_inputs = dynamic_unsupported(node)
```
gs-olive, line 337: Consider renaming this function to `has_dynamic_shapes`, because the function is no longer a validator

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -358,7 +358,7 @@ def aten_ops_grid(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.relu.default)
+@dynamo_tensorrt_converter(torch.ops.aten.relu.default, dynamic=True)
```
gs-olive, line 361: Are there cases where we want `dynamic=None` to mean _both_ dynamic and not dynamic, and reserve `dynamic=True` to imply _only_ dynamic?

gs-olive (commented):
Overall looks great, added a few suggestions

narendasan:
`WARNING:torch_tensorrt.dynamo.conversion._ConverterRegistry:The converter for node aten.convolution.default received dynamic shaped inputs but the static version of the converter is being used. Please report this issue at https://github.com/pytorch/TensorRT/issues`

This is a severe warning, it implies that it probably wont work. Is the intention for us to treat it as something that probably would work but might fail like (70/30) or like (20/80) pass/fail? If its the latter this should be an error, if its the former, we need to adjust the warning

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
```diff
@@ -79,6 +79,7 @@ class ConverterSupport:
 
     converter_implementation: ConverterImplSignature
     capability_validator: Callable[[Node], bool] = field(default=lambda node: True)
+    dynamic: bool = False
```
narendasan, line 82: call this `supports_dynamic_shapes` or something

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
```diff
@@ -323,6 +329,18 @@ def __getitem__(
 
                 if isinstance(converters, (list, tuple)):
                     for candidate in converters:
+                        # TODO: Importing this here avoids circular import issue. One potential fix is moving this function into _ConverterRegistry file.
```
narendasan, line 332: Move it into this file

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
narendasan: I think it should be a support vs not support since any dynamic shape converter should be able to support both right?

peri044:
> `WARNING:torch_tensorrt.dynamo.conversion._ConverterRegistry:The converter for node aten.convolution.default received dynamic shaped inputs but the static version of the converter is being used. Please report this issue at https://github.com/pytorch/TensorRT/issues`
> 
> This is a severe warning, it implies that it probably wont work. Is the intention for us to treat it as something that probably would work but might fail like (70/30) or like (20/80) pass/fail? If its the latter this should be an error, if its the former, we need to adjust the warning

Intention is "would work but might fail". How about the following message ? 
```
The converter for node aten.convolution.default received dynamic shaped inputs although it was designed for static inputs. This shouldn't likely cause issues unless there are some dimensions which are dynamic (excluding the batch). If you encounter any issues, please post at https://github.com/pytorch/TensorRT/issues`
```

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
peri044: Current assumption is `dynamic=True` implies it supports both static and dynamic. I'm not aware of any converters which only work for dynamic (but not static).

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -645,14 +644,11 @@ def aten_ops_softmax(
 
 
 @dynamo_tensorrt_converter(
-    torch.ops.aten.split.Tensor, capability_validator=dynamic_unsupported_with_args([1])
-)
-@dynamo_tensorrt_converter(
-    torch.ops.aten.split.sizes, capability_validator=dynamic_unsupported_with_args([1])
```
peri044, line 651: Revisit the use of capability_validator here

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
```diff
@@ -75,22 +78,102 @@ class ConverterSupport:
...
     capability_validator: Callable[[Node], bool] = field(default=lambda node: True)
+    supports_dynamic_shapes: bool = False
 
 
 # Dictionary representing Dynamo aten-only converters
 # Each converter maps to a sequence of at least one ConverterSupport object(s)
 DYNAMO_ATEN_CONVERTERS: Dict[Target, Sequence[ConverterSupport]] = {}
 
 
+def has_dynamic_shapes(node: torch.fx.Node) -> bool:
```
narendasan, line 94: Seems redundant now

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
```diff
@@ -323,6 +410,15 @@ def __getitem__(
 
                 if isinstance(converters, (list, tuple)):
                     for candidate in converters:
+                        # If there are dynamic inputs but the converter doesn't support it explicitly, throw a warning.
+                        if (
```
narendasan, line 414: You could have a flag similar to experimental_decompositions that lets all converters through but default behavior is non dynamic converters should not be used in a dynamic context.

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
peri044: We still need it here right ?

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
peri044: Implemented it now

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
```diff
@@ -324,13 +414,30 @@ def __getitem__(
...
+                        ):
+                            # If node has dynamic inputs and the converter supports dynamic shapes, it is enabled
+                            return (
+                                candidate.converter_implementation,
+                                calling_convention,
+                            )
+                        elif candidate.capability_validator(
+                            node
+                        ) and not has_dynamic_shapes(node):
+                            # For static shapes all converters are turned on based on capability_validator check
```
gs-olive, line 432: Consider condensing logic to:
```python
node_has_dynamic_shapes = has_dynamic_shapes(node)

if candidate.capability_validator(node) and (
    self.assume_dynamic_shape_support
    or not node_has_dynamic_shapes 
    or (
        node_has_dynamic_shapes 
        and candidate.supports_dynamic_shapes
    )
):
```
The above might assist with showing explicitly the three conditions that allow a converter to "pass", from my understanding:

- Validator Passes [Required]
- Need at least _one_ of the following:
  - `assume_dynamic_shape_support` [Optional]
  - Static Node [Optional]
  - Dynamic Node + Dynamic Support Converter [Optional]

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
gs-olive: This one seems to be more of an edge case (it seems most converters can either support dynamic shapes, or not), perhaps the capability validator could be allowed to assess dynamic shape support in certain arguments for some challenging edge cases like this.

py/torch_tensorrt/dynamo/_defaults.py
```diff
@@ -6,6 +6,7 @@
 DEBUG = False
 DEVICE = None
 DISABLE_TF32 = False
+ASSUME_DYNAMIC_SHAPE_SUPPORT = False
```
gs-olive, line 9: Are there cases where one might want this to be `True`?

gs-olive (commented):
Overall looks good to me - very useful feature for the Converter Registry. Added a few comments + suggestions

py/torch_tensorrt/dynamo/_defaults.py
peri044: - Some of the ops might support dynamic shapes supports_dynamic_shapes=True might not be enabled (until we have validation).  
- The existing models which compiled successfully assumed all ops supported dynamic shapes but we have limited that now. This flag lists that restriction.

gs-olive (approved):
Looks good to me

PR #2832, opened 2024-05-14
fix: bugs in TRT 10 upgrade

### LINKED ISSUE #2811: 🐛 [Bug] An assertion error when upgrading to enqueueV3 interface
##  Bug Description

https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py#L63-L65

This should be:
```
assert (
    self.engine.num_io_tensors == len(self.input_names) + len(self.output_names)
)
```
because `num_io_tensors` is not inflated by the number of optimization profiles, for enqueueV3 interface. This is different from the enqueueV2 interface: 
```
assert engine.num_bindings == num_io_tensors * engine.num_optimization_profiles
```

zewenli98:
# Description

Fix bugs in TRT 10 upgrade.

Fixes #2811 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
+++ b/py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
@@ -29,7 +29,7 @@ class PythonTorchTensorRTModule(Module):  # type: ignore[misc]
 
     def __init__(
         self,
-        engine: trt.tensorrt.IHostMemory,
+        engine: bytes,
         input_names: Optional[List[str]] = None,
         output_names: Optional[List[str]] = None,
         target_device: Device = Device._current_device(),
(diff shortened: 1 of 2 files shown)


peri044 (approved):
LGTM

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -29,7 +29,7 @@ class PythonTorchTensorRTModule(Module):  # type: ignore[misc]
 
     def __init__(
         self,
-        engine: trt.ICudaEngine,
+        engine: trt.tensorrt.IHostMemory,
```
narendasan, line 32: Lets use the bytes interface (may be just changing the type annotation), but we use the same thing for the C++ runtime

zewenli98: Thanks. Currently, the `engine` actually saves `serialized engine` since TRT 10 changes API that uses `builder.build_serialized_network`. Do you recommend saving deserialized engines or serialized engines?

peri044:
Can this be merged @zewenli98 ? If so, please raise a cherry pick to release/2.3 it it's needed

PR #2841, opened 2024-05-16
add aten.topk implementation

lanluo-nvidia:
# Description

Add aten.topk implementation

## Type of change

# Checklist:


tests/py/dynamo/conversion/test_topk_aten.py
```diff
@@ -0,0 +1,31 @@
...
+import torch.nn as nn
+from parameterized import parameterized
+from torch.testing._internal.common_utils import run_tests
+
+from .harness import DispatchTestCase
+
+
+class TestSortConverter(DispatchTestCase):
+    @parameterized.expand(
+        [
```
apbose, line 11: Can you add more test cases with 2D and 3D cases? Also cases on how it would handle duplicate values.
Also could you once print the decomposed graph to see if aten.topk.value is invoked or not.

py/torch_tensorrt/dynamo/conversion/impl/topk.py
```diff
@@ -134,3 +134,38 @@ def sort(
         return topk_layer.get_output(0), topk_layer.get_output(1)
     else:
         return topk_layer.get_output(0)
+
+
+def topk(
```
apbose, line 139: Hi can you consolidate the topk() and the sort() function, something like the sort() function calls the topk function with the k arg as the dim?

Also in the below the sorted argument should also be handled to return the elements in sorted order.

py/torch_tensorrt/dynamo/conversion/impl/topk.py
```diff
@@ -103,31 +103,38 @@ def argmin(
     )
 
 
-def sort(
+def topk(
```
narendasan, line 106: Make a sort function that just calls topk with the right parameters.

tests/py/dynamo/conversion/test_topk_aten.py
apbose, line 1: The function structure as discussed last time looks good to me.
In the above you mentioned that topk in TRT returns sorted by default. So was curious if there was a mismatch in the test cases where sorted=False.
Also could you add a validator to the topk function such that k values greater than 3840 will not be supported by TRT layer. Something like this - https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py#L264-L269

lanluo-nvidia: added the topk validator, was having difficulties try to get the input tensor shape will discuss this offline with @narendasan for help.

apbose (commented):
Looks good to me! Pending on CI.

PR #2853, opened 2024-05-20
fix: Handle dynamic shapes in where ops

keehyuna:
# Description

Dynamic shape cannot be used in where ops because of exception in torch.broadcast_shapes().
Proposed fix removes expand() for static shape input and only performs prepend ones to have same rank size for static/dynamic shape input. I think broadcast is not required as it is applied to ISelectLayer([addSelect](https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_network_definition.html#a8178e3d6fa27651feff2f7e0652e2101))

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/_Input.py
```diff
@@ -364,9 +364,13 @@ def example_tensor(
                     )
 
                 if isinstance(self.shape, dict):
-                    return torch.rand(self.shape[optimization_profile_field]).to(
-                        dtype=self.dtype.to(torch.dtype, use_default=True)
```
keehyuna, line 368: If dtype is torch.bool in input_spec, torch.rand() returns random numbers on [0,1) and casting it to bool value is True

py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
```diff
@@ -10,8 +9,7 @@
     broadcastable,
     get_trt_tensor,
 )
-from torch_tensorrt.dynamo.conversion.impl.slice import expand
-from torch_tensorrt.fx.converters.converter_utils import set_layer_name
+from torch_tensorrt.fx.converters.converter_utils import prepend_ones, set_layer_name
```
peri044, line 12: can we add prepend_ones and the set_layer_name functions in our dynamo codebase instead of using from fx ?

tests/py/dynamo/conversion/test_where_aten.py
```diff
@@ -76,6 +77,86 @@ def forward(self, condition):
...
+    def test_with_dynamic_shape(self, *args):
+        class Where(nn.Module):
+            def forward(self, condition, x, y):
+                return torch.ops.aten.where.self(condition, x, y)
+
+        input_specs = [
+            Input(
+                shape=args[1],
+                dtype=torch.bool,
+                shape_ranges=[(args[2], args[3], args[4])],
```
peri044, line 145: Instead of `shape_ranges`, can you use `torch_tensorrt.Input(min_shape=<>, opt_shape=<> ...)` directly ?

peri044 (changes requested):
Also please mark torch.where converter in aten_ops_converters.py with `supports_dynamic_shapes=True` flag. Example : https://github.com/pytorch/TensorRT/blob/release/2.3/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py#L62

tests/py/dynamo/conversion/test_where_aten.py
```diff
@@ -76,6 +77,86 @@ def forward(self, condition):
...
+                (
+                    -1,
+                    -1,
+                ),
+                (1, 1),
+                (2, 2),
+                (2, 4),
+            ),
+            (
+                "2d_condition_3d_xshape_2d_yshape",
```
chohk88, line 117: It would be great if you could also add a test case to verify broadcasting.

```
            (
                "name",
                (-1, -1),
                (4, 1),
                (4, 2),
                (5, 4),
                (-1, -1, -1),
                (1, 1, 1),
                (3, 1, 2),
                (3, 1, 4),
                (
                    -1,
                    -1,
                ),
                (1, 1),
                (1, 2),
                (1, 4),
            ),
```

py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
```diff
@@ -30,73 +28,30 @@ def where(
     x_shape = list(input.shape)
```
chohk88, line 28: Although this isn't related to the changes you made, isn't the check for broadcastability unnecessary since it is already verified by the torch operation?

py/torch_tensorrt/_Input.py
```diff
@@ -364,9 +364,13 @@ def example_tensor(
...
-                    )
+                    if self.dtype == dtype.b:
```
chohk88, line 367: Could you provide more details on why this part was changed? I'd like to understand it better.

chohk88 (commented):
I have left some comments. It looks like there aren't any major issues with the functionality.

tests/py/dynamo/conversion/test_where_aten.py
keehyuna: Thanks for suggestion. I replaced 2d_condition_3d_xshape_2d_yshape test with it

py/torch_tensorrt/_Input.py
keehyuna: Casting from random [0,1) value to bool type is not efficient as all output is True except 0.
I think random [0,1) < 0.5 should be helpful to create better random bool values

py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
keehyuna: Yes I saw error in tensorrt when invalid input that is not boradcastable is passed. But maybe early failure could be helpful

py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
keehyuna: I moved it to py/torch_tensorrt/dynamo/conversion/converter_utils.py

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -15,9 +15,11 @@
     ConverterRegistry,
     DynamoConverterImplSignature,
 )
-from torch_tensorrt.fx.converters.converter_utils import (
+from torch_tensorrt.fx.converters.converter_utils import (  # noqa: F401
     broadcast,
     get_axes_for_reduce_op,
+    prepend_ones,
+    set_layer_name,
```
peri044, line 22: Copy the source of these functions directly into this file here. 
1) We should not be using utilities from `torch_tensorrt.fx.converters.converter_utils` but rather have them in `torch_tensorrt.dynamo.converters.converter_utils`
2) Also, add these types to our dynamo types `from torch_tensorrt.fx.types import TRTDataType, TRTTensor` 
One of the side goals of adding DS support for converters is to move away from using legacy utilities (`torch_tensorrt.fx` in this case).

peri044 (changes requested):

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -686,3 +676,183 @@ def calculate_strides(shape: Sequence[int]) -> Sequence[int]:
...
+        b = prepend_ones(network, b, f"{b_name}_broadcast", diff)
+    elif diff < 0:
+        a = prepend_ones(network, a, f"{a_name}_broadcast", -diff)
+
+    return a, b
+
+
+def get_axes_for_reduce_op(
+    dim: Union[int, Sequence[int]],
+    has_implicit_batch_dimension: bool = False,
```
keehyuna, line 723: default parameter was used to merge below function
get_axes_for_reduce_op = functools.partial(
    get_axes_for_reduce_op, has_implicit_batch_dimension=False
)

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -686,3 +676,183 @@ def calculate_strides(shape: Sequence[int]) -> Sequence[int]:
...
+    Args:
+        shape (Shape): Shape of a tensor. Essentially is a sequence of integers.
+
+    Returns:
+        A boolean value indicates whether there's dynamic dim in the shape.
+    """
+    count = 0
+    for s in shape:
+        count += 1 if s == -1 else 0
+    return count > 0
```
keehyuna, line 766: return type was changed when after it's copied from fx.converters.has_dynamic_shape
- return count
+ return count > 0

chohk88:
Looks good to me!

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -684,3 +676,183 @@ def calculate_strides(shape: Sequence[int]) -> Sequence[int]:
...
+        b = prepend_ones(network, b, f"{b_name}_broadcast", diff)
+    elif diff < 0:
+        a = prepend_ones(network, a, f"{a_name}_broadcast", -diff)
+
+    return a, b
+
+
+def get_axes_for_reduce_op(
+    dim: Union[int, Sequence[int]],
+    has_implicit_batch_dimension: bool = False,
```
zewenli98, line 723: @peri044 do we still have `has_implicit_batch_dimension`? Is it possible to remove the arg in the dynamo's converter_utils?

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -684,3 +676,183 @@ def calculate_strides(shape: Sequence[int]) -> Sequence[int]:
...
+    Returns:
+        A boolean value indicates whether there's dynamic dim in the shape.
+    """
+    count = 0
+    for s in shape:
+        count += 1 if s == -1 else 0
+    return count > 0
+
+
+def type_cast(
```
zewenli98, line 769: It seems the function does the same functionality as `cast_trt_tensor`: https://github.com/pytorch/TensorRT/blob/68f08ea09744d4ecc84f9f2be88faff497a9283e/py/torch_tensorrt/dynamo/conversion/converter_utils.py#L61-L97

zewenli98 (commented):
@peri044 I think it's reasonable to deprecate FX converter_utils in Dynamo converter implementation. In my opinion, when copying these helper functions from FX utils, it would be better to change the first arg from `network: TRTNetwork` to `ctx: ConversionContext` for two reasons: 1) consistent with other helper functions 2) convenient to call them, just passing in `ctx`.

py/torch_tensorrt/dynamo/conversion/converter_utils.py
keehyuna: Thanks for review. I agree with it. replaced type_cast with cast_trt_tensor

py/torch_tensorrt/dynamo/conversion/impl/condition/ops.py
```diff
@@ -30,73 +29,28 @@ def where(
...
     y_shape = list(other.shape)
     condition_shape = list(condition.shape)
+    max_shape_len = max(len(x_shape), len(y_shape), len(condition_shape))
 
-    output_shape = list(torch.broadcast_shapes(condition_shape, x_shape, y_shape))
-
-    # expand shape
     if not isinstance(condition, TRTTensor):
         assert condition.dtype in (torch.bool, np.bool_), "condition dtype is not bool"
```
zewenli98, line 35: If `condition.dtype` in (torch.bool, np.bool_), is it possible to convert to a supported type for `condition` so that it will not error out? because our goal is not to throw out errors if pytorch supports it. If not, I think a capability validator is needed.

keehyuna: I saw uint8 type of condition was available for torch. Added type conversion.

zewenli98: looks good to me. It would be better if add a suffix for the name here: https://github.com/pytorch/TensorRT/pull/2853/files#diff-84205b80e84f9b477f7b0b6dfaab7ea72f8426cc2673e240d2221401fe74c14cR41

py/torch_tensorrt/dynamo/conversion/converter_utils.py
peri044: 1) We should remove `has_implicit_batch_dimension`.
2) > It would be better to change the first arg from network: TRTNetwork to ctx: ConversionContext. 

Yes we should make these changes and not use FX code/ data structures as much as possible. If we use them, we should be consistent with dynamo APIs

keehyuna: thanks. Removed has_implicit_batch_dimension flag in get_axes_for_reduce_op()
first arg of boardcast()/prepend_ones() is ctx: ConversionContext, instead of network: TRTNetwork

keehyuna:
Moved "chore: Better random bool values for example_tensor" change into [PR2878](https://github.com/pytorch/TensorRT/pull/2878) to combine random input changes to one PR.
https://github.com/pytorch/TensorRT/pull/2878/commits/436e99b9c1c341dd0eba9c47a0fbad68119a66dc

py/torch_tensorrt/dynamo/conversion/converter_utils.py
peri044: We already have a function `has_dynamic_shapes` here : https://github.com/pytorch/TensorRT/blob/5e7ca649e14742504e318c84464b20498652a598/py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py#L99
Can you merge these two functionalities in here ? The function can accept either `torch.fx.Node` or `ITensor` in your case and should return boolean accordinly

py/torch_tensorrt/dynamo/types.py
```diff
@@ -0,0 +1,24 @@
...
+    TRTNetwork = "trt.INetworkDefinition"
+    TRTTensor = "trt.tensorrt.ITensor"
+    TRTLayer = "trt.ILayer"
+    TRTPluginFieldCollection = "trt.PluginFieldCollection"
+    TRTPlugin = "trt.IPluginV2"
+    TRTDataType = "trt.DataType"
+    TRTElementWiseOp = "trt.ElementWiseOperation"
+
+Shape = Sequence[int]
+ShapeRange = Tuple[Shape, Shape, Shape]
```
peri044, line 24: If these are being used by dynamo converters, keep it or else we can remove them.

peri044 (changes requested):
Thanks. added a few more minor changes.

py/torch_tensorrt/dynamo/types.py
keehyuna: Grepped and checked it's not used. Removed in new commit

peri044 (approved):
LGTM

PR #2856, opened 2024-05-22
feat: support aten.diagonal converter

chohk88:
# Description

A converter for the torch.ops.aten.diagonal operation. There are very good reference for aten.diagonal operation ( https://github.com/pytorch/pytorch/blob/082251e76b93b277ff2791d0e2b64934add34644/torch/_refs/__init__.py#L4255 )

Fixes # ([issue](https://github.com/pytorch/TensorRT/issues/2846))

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -16,6 +16,7 @@ from torch_tensorrt.dynamo.conversion._ConverterRegistry import (
 from torch_tensorrt.dynamo.conversion.converter_utils import (
     dynamic_unsupported_with_args,
     enforce_tensor_types,
+    get_positive_dim,
     is_only_operator_on_placeholder,
 )
 from torch_tensorrt.fx.types import TRTTensor
@@ -2935,12 +2936,26 @@ def aten_ops_flip(
 
 
 def zero_diag_size_validator(node: Node) -> bool:
-    input_shape, offset, dim1, dim2 = (
-        node.args[0].meta["tensor_meta"][0],
+    meta = node.args[0].meta.get("tensor_meta")
+    if meta:
+        input_shape = meta.shape
+    else:
+        _LOGGER.warning(
+            "Meta information of input is missing, unable to validate diagonal size."
+        )
+        return True
+
+    offset, dim1, dim2 = (
         node.args[1],
         node.args[2],
         node.args[3],
     )
+
+    num_dims = len(input_shape)
+
+    # Adjust dimensions to be positive and canonicalize
+    dim1 = get_positive_dim(dim1, num_dims)
+    dim2 = get_positive_dim(dim2, num_dims)
 
     if offset >= 0:
         diag_size = max(min(input_shape[dim1], input_shape[dim2] - offset), 0)
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2941,9 +2941,9 @@ def zero_diag_size_validator(node: Node) -> bool:
         input_shape = meta.shape
     else:
         _LOGGER.warning(
-            "Meta information of input is missing, unable to validate diagonal size."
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2934,6 +2934,50 @@ def aten_ops_flip(
     )
 
 
+def zero_diag_size_validator(node: Node) -> bool:
+    input_shape, offset, dim1, dim2 = (
+        node.args[0].meta["tensor_meta"][0],
```
zewenli98, line 2939: Can you change this line to something like: 
```
meta = node.args[0].meta.get("tensor_meta")
if meta:
    input_shape = meta.shape
```
Because I observed if `enable_passes=False` in the unit test, the `meta` will be empty. You can play with the arg as well, but don't need to change to False at the end.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2934,6 +2934,50 @@ def aten_ops_flip(
...
+        node.args[1],
+        node.args[2],
+        node.args[3],
+    )
+
```
zewenli98, line 2944: Do we need to call `get_positive_dim` like you did in the ops.py?

py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
```diff
@@ -262,6 +262,75 @@ def flip(
     return layer.get_output(0)
 
 
+def calculate_strides(shape: Sequence[int]) -> Sequence[int]:
+    strides = [1] * len(shape)
+    for i in range(len(shape) - 2, -1, -1):
+        strides[i] = strides[i + 1] * shape[i + 1]
+    return strides
```
zewenli98, line 269: Is this function a general util or just specific for this case? If general, I think it would be perfect to move it into the `converter_utils.py` with detailed explanations on limitations and how to use it. Otherwise, it could be moved into the function of `diagonal`

zewenli98 (commented):
Overall LGTM, just added some comments for your reference.

py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
chohk88: Yes, this function can be a general utility function. Therefore, I have moved the `calculate_strides` function to `converter_utils.py`. Thank you for the useful comment.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
chohk88: You are definitely right. The `get_positive_dim` function is also needed here. Thank you!

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
chohk88: Thank you so much.

I also encountered the issue where meta was empty, so I revised enable_passes=True in the tested code. From your comment, I understand that enable_passes could be False in CI tests or other cases. You suggested revising the code as `meta = node.args[0].meta.get("tensor_meta")`.

I agree with your comments and have revised the code accordingly. However, if the validator returns False whenever the meta of the node is empty, it could cause problems because having a diagonal size of zero is a very rare case. Therefore, I included a warning message for empty meta and made the validator return True.

If I have misunderstood your comments, please let me know what I missed.

zewenli98: I meant if your implementation depends on `input_shape`, considering in some cases like `enable_passes=False` where you cannot get `input_shape`, the validator will report an error. However, if you change to what I said, I think it will not get errors, instead, just fall back to pytorch operation.

> Therefore, I included a warning message for empty meta and made the validator return True.

If the validator returns True when meta is empty, it will be converted whatever the real `input_shape` is. Is this what you expect? If not, I think it should be `return False`. Correct me if my understanding is wrong.

zewenli98 (approved):
LGTM

Issue #2872, opened 2024-05-31
`aten.resize_`


chohk88:
support https://pytorch.org/docs/stable/generated/torch.Tensor.resize_.html#torch.Tensor.resize_ converter

PR #2874, opened 2024-05-31
feat: support aten.resize_ converter

chohk88:
# Description
Support converter for `aten.resize_` operation: https://pytorch.org/docs/stable/generated/torch.Tensor.resize_.html#torch.Tensor.resize_

One critical aspect of this operation is handling cases where the target size (output size) is larger than the input tensor size. When `the target size` is larger than `the input tensor size`, the values of the additional elements are unpredictable. In PyTorch, these additional elements are not initialized, which can result in values that are close to `zero` but are not guaranteed to be `zero`.

image

In the converter developed for this PR, the additional elements are initialized to `zero` using `numpy.zeros`. While this approach ensures a predictable output, it does not exactly replicate the behavior of PyTorch, where the values of the additional elements are not initialized and can be unpredictable.

Fixes # ([issue](https://github.com/pytorch/TensorRT/issues/2872))

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2695,6 +2695,30 @@ def aten_ops_pixel_unshuffle(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.resize.default)
+@dynamo_tensorrt_converter(torch.ops.aten.resize_.default)
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
+    }
+)
+def aten_ops_resize(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.shuffle.resize_(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        input=args[0],
+        sizes=args[1],
+    )
+
+
 @enforce_tensor_types({0: (TRTTensor,)})
 @dynamo_tensorrt_converter(torch.ops.aten.argmax.default)
 def aten_ops_argmax(
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2695,7 +2695,6 @@ def aten_ops_pixel_unshuffle(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.resize.default)
 @dynamo_tensorrt_converter(torch.ops.aten.resize_.default)
 @enforce_tensor_types(
     {
@@ -2709,7 +2708,7 @@ def aten_ops_resize(
     kwargs: Dict[str, Argument],
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
-    return impl.shuffle.resize_(
+    return impl.shuffle.resize(
         ctx,
         target,
         SourceIR.ATEN,
--- a/py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
@@ -1,5 +1,7 @@
 from typing import Optional, Sequence, Union
 
+import numpy as np
+import tensorrt as trt
 import torch_tensorrt.dynamo.conversion.impl as impl
 from torch.fx.node import Target
 from torch_tensorrt import _enums
@@ -7,9 +9,14 @@ from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContex
 from torch_tensorrt.dynamo.conversion.converter_utils import (
     SourceIR,
     cast_trt_tensor,
+    flatten_dims,
     get_trt_tensor,
 )
-from torch_tensorrt.fx.converters.converter_utils import set_layer_name
+from torch_tensorrt.fx.converters.converter_utils import (
+    Frameworks,
+    set_layer_name,
+    unified_dtype_converter,
+)
 from torch_tensorrt.fx.types import TRTTensor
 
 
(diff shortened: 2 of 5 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2610,6 +2610,30 @@ def aten_ops_pixel_unshuffle(
     )
 
 
+@dynamo_tensorrt_converter(torch.ops.aten.resize.default)
```
zewenli98, line 2613: Is there `torch.ops.aten.resize.default` in the [schema](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml)? I didn't find it. Can you double check?

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
```diff
@@ -120,3 +130,60 @@ def pixel_unshuffle(
         permuted_tensor,
         shape[:-3] + (out_channels, out_height, out_width),
     )
+
+
+def resize_(
```
zewenli98, line 135: I think the function name `resize`, instead of `resize_`, is good enough, for consistency.

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
```diff
@@ -120,3 +130,60 @@ def pixel_unshuffle(
...
+    if new_num_elements > current_num_elements:
+        # Create a padding tensor with the required size and initialize new elements with zeros
+        padding_size = new_num_elements - current_num_elements
+        padding_tensor = ctx.net.add_constant(
+            (padding_size,), trt.Weights(np.zeros(padding_size, dtype=input_np_dtype))
+        ).get_output(0)
+
+        # Flatten input tensor to 1D for concatenation
+        flatten_shape = flatten_dims(input_val, 0, -1)
+        flattened_input = impl.shuffle.reshape(
```
zewenli98, line 161: `impl.shuffle.reshape` -> `reshape`

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
```diff
@@ -120,3 +130,60 @@ def pixel_unshuffle(
...
+            ctx, target, source_ir, f"{name}_flatten_input", input_val, flatten_shape
+        )
+
+        # Concatenate the flattened input tensor and padding tensor
+        concat_layer = ctx.net.add_concatenation([flattened_input, padding_tensor])
+        concat_layer.axis = 0
+        reshaped_tensor = concat_layer.get_output(0)
```
zewenli98, line 168: Can you use `impl.cat.cat` function?

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
```diff
@@ -120,3 +130,60 @@ def pixel_unshuffle(
...
+
+    elif new_num_elements < current_num_elements:
+        # Flatten input tensor to 1D for slicing
+        flatten_shape = flatten_dims(input_val, 0, -1)
+        flattened_input = impl.shuffle.reshape(
```
zewenli98, line 173: `impl.shuffle.reshape` -> `reshape`

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
```diff
@@ -120,3 +130,60 @@ def pixel_unshuffle(
...
+
+        # Slice the flattened input tensor to the desired number of elements
+        slice_layer = ctx.net.add_slice(flattened_input, [0], [new_num_elements], [1])
+        reshaped_tensor = slice_layer.get_output(0)
+
+    else:
+        reshaped_tensor = input_val
+
+    # Reshape the final output tensor to the target sizes
+    resized_output = impl.shuffle.reshape(
```
zewenli98, line 185: `impl.shuffle.reshape` -> `reshape`

tests/py/dynamo/conversion/test_resize_aten.py
```diff
@@ -0,0 +1,117 @@
...
+            ((8, 3),),
+            ((7, 7),),
+            ((5, 5, 5),),
+            ((3, 3, 5),),
+        ]
+    )
+    def test_resize_1d_input_float(self, target_shape):
+        class Resize(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
```
zewenli98, line 24: It seems `__init__` is not necessary, same for below

zewenli98 (commented):
Overall looks good, just left some comments for your reference.

Although the additional elements are unpredictable but they are close enough to zero, as you said, I think it is able to pass our checkpoints since the tolerance is acceptable

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
chohk88: Thank you for your review. I have addressed your comment.

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
chohk88: Thank you for your review. I've made the changes as per your review.

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
chohk88: Thank you for your review. I've made the changes as per your review.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: Typically, we only register the functions starting with `- func: ...` in the schema.
Can you try only registering `resize_`, whether the unit test with `resize` is able to pass?
If pass, I think only `resize_` is enough.

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
```diff
@@ -1,10 +1,20 @@
...
-from torch_tensorrt.fx.converters.converter_utils import set_layer_name
+from torch_tensorrt.dynamo.conversion.converter_utils import (
+    SourceIR,
+    flatten_dims,
+    get_trt_tensor,
+)
+from torch_tensorrt.fx.converters.converter_utils import (
+    Frameworks,
+    set_layer_name,
+    unified_dtype_converter,
```
keehyuna, line 16: You might need to copy Framework/unified_dtype_converter from torch_tensorrt.fx.converters.converter_utils.
set_layer_name is copied like below.
https://github.com/pytorch/TensorRT/pull/2853/files#diff-f6ddcec33365053481a53cde22972164c5c051197dcd0ec83855b422fb272975R817

chohk88:
@apbose @zewenli98 The converter for `resize_` has been fully implemented functionally. 

However, while investigating the cause of the CI/CD failure, I found that tensors in uninitialized locations sometimes contain extremely large values, whereas the description always showed values very close to zero (in the range of `10^-40`) or `zero`. Do you have any suggestions on how to handle tests in such cases?

image

narendasan:
Use run_tests custom compare and check the original elements and the shape to verify correctness

apbose:
@chohk88 I was talking about the function  `run_test_custom_compare_results()`. Ypu can implement the attributes of the TRTTensor and the ref Tensor which you would want to compare in the comparators arg.

tests/py/dynamo/conversion/test_resize_aten.py
```diff
@@ -0,0 +1,202 @@
...
+    )
+    def test_resize_2d_input_float(self, target_shape):
+        class Resize(torch.nn.Module):
+            def forward(self, x):
+                return torch.ops.aten.resize_.default(x, target_shape)
+
+        input_shape = (4, 4)
+        inputs = [torch.randint(1, 10, input_shape)]
+
+        def compare_resized_tensors(tensor1, tensor2, input_shape, target_shape):
```
apbose, line 127: The comparators look good, just that I think they are the same for all the cases. So can we have a single comparator for all?

tests/py/dynamo/conversion/harness.py
```diff
@@ -157,12 +157,20 @@ def run_test_custom_compare_results(
...
+                    comp_func = comparator[0]
+                    args = comparator[1]
+                    self.assertTrue(comp_func(res_trt, res_cpu, *args))
+            else:
+                for output_trt, output_cpu, comparator in zip(
+                    res_trt, res_cpu, comparators
+                ):
+                    comp_func = comparator[0]
+                    args = comparator[1]
+                    self.assertTrue(comp_func(output_trt, output_cpu, *args))
```
apbose, line 174: Is there a specific case where the len(cuda_inputs) == 1 is required? In general I assume the `len(cuda_inputs)` would be 1 in most cases. And since it conditioned on `res_trt ` could you highlight the cases where it would be required?

tests/py/dynamo/conversion/test_resize_aten.py
chohk88: Thank you for useful comment! 
I have moved the `compare_resized_tensors` function to a class method to avoid redundancy.

tests/py/dynamo/conversion/harness.py
chohk88: Thank you for your comment!

As you mentioned, in most cases, the length of `cuda_inputs` is 1. Below are the outputs when I set a breakpoint in the original code for two cases where `torch.ops.aten.resize_.default(x, target_shape)` returns tensors with shapes `(3,)` and `(10, 15, 10)`. 

image
image

In both cases, the length of `cuda_inputs` is also `1`. When `cuda_inputs` has a length of `1`, `res_trt` and `res_cpu` are not lists of length `1` but are `torch.Tensors` with shapes `(3,)` and `(10, 15, 10)` for each case. Therefore, when we use `zip` in `for output_trt, output_cpu, comparator in zip(res_trt, res_cpu, comparators)`, and the comparators list has a length of `1`, it results in comparing fewer elements because one dimension is lost from `res_trt` and `res_cpu`.

py/torch_tensorrt/dynamo/conversion/impl/shuffle.py
chohk88: I've resolve this issue (remove dependency on fx util, move to dynamo)

PR #2879, opened 2024-06-03
replace itemset due to numpy version 2.0 removed itemset api

lanluo-nvidia:
# Description

Numpy has removed itemset api since 2.0, so need to replace the itemset call

Fixes # (issue)

>           offsets.itemset(-1, len_embed)
E           AttributeError: `itemset` was removed from the ndarray class in NumPy 2.0. Use `arr[index] = value` instead.

Collecting numpy (from torchvision==0.18.0->-r /__w/TensorRT/TensorRT/pytorch/tensorrt/tests/py/requirements.txt (line 8))
398
  Downloading numpy-2.0.0rc2-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (19.3 MB)
  
## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/embedding.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/embedding.py
@@ -94,7 +94,10 @@ def embedding_bag_with_traversable_offsets(
         # however, pytorch doc says if `include_last_offset` is True, the size of offsets
         # is equal to the number of bags + 1. The last element is the size of the input,
         # or the ending index position of the last bag (sequence).
-        offsets.itemset(-1, len_embed)
+        offsets_shape = offsets.shape
+        offsets = offsets.flatten()
+        offsets[-1] = len_embed
+        offsets.reshape(offsets_shape)
     else:
         # add the end index to offsets
         offsets = np.append(offsets, len_embed)
--- a/py/torch_tensorrt/dynamo/conversion/impl/embedding.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/embedding.py
@@ -99,7 +99,7 @@ def embedding_bag_with_traversable_offsets(
         if len(offsets.shape) != 1:
             raise TypeError(
                 f"The offsets should be in 1d array, here offset shape is {offsets.shape}."
-        )
+            )
         offsets[-1] = len_embed
     else:
         # add the end index to offsets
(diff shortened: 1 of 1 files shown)


py/torch_tensorrt/dynamo/conversion/impl/embedding.py
```diff
@@ -94,7 +94,10 @@ def embedding_bag_with_traversable_offsets(
         # however, pytorch doc says if `include_last_offset` is True, the size of offsets
         # is equal to the number of bags + 1. The last element is the size of the input,
         # or the ending index position of the last bag (sequence).
-        offsets.itemset(-1, len_embed)
+        offsets_shape = offsets.shape
+        offsets = offsets.flatten()
+        offsets[-1] = len_embed
+        offsets.reshape(offsets_shape)
```
zewenli98, line 100: According to the error: 
```
AttributeError: `itemset` was removed from the ndarray class in NumPy 2.0. Use `arr[index] = value` instead.
```
maybe you can just use `offsets[-1] = len_embed`?

lanluo-nvidia: @zewenli98  for the multi dimensional arry, offset[-1] is actually updating the last dimension as the value
what we want is only to update the last element as the value
```
>>> l = torch.rand(2,3)
>>> l
tensor([[0.0770, 0.6687, 0.1859],
        [0.5044, 0.9199, 0.2451]])
>>> l[-1]=9.99
>>> l
tensor([[0.0770, 0.6687, 0.1859],
        [9.9900, 9.9900, 9.9900]])
```

zewenli98: This usage is correct for a general case, but I think in this case the `offsets` is 1d, per https://pytorch.org/docs/stable/generated/torch.nn.EmbeddingBag.html

zewenli98:
```
A few notes about input and offsets:

input and offsets have to be of the same type, either int or long

If input is 2D of shape (B, N), it will be treated as B bags (sequences) each of fixed length N, and this will return B values aggregated in a way depending on the mode. offsets is ignored and required to be None in this case.

If input is 1D of shape (N), it will be treated as a concatenation of multiple bags (sequences). offsets is required to be a 1D tensor containing the starting index positions of each bag in input. Therefore, for offsets of shape (B), input will be viewed as having B bags. Empty bags (i.e., having 0-length) will have returned vectors filled by zeros.
```

zewenli98 (approved):
The modification LGTM. Pending on CI

Issue #2544, opened 2023-12-15
`aten.index_put`


gs-olive:

gs-olive:
Required for HF model `google/recurrentgemma-2b`

PR #2880, opened 2024-06-03
feat: support aten index_put converter for accumulate=False

chohk88:
# Description

I have implemented the `aten::index_put` operation using the `add_scatter` layer with `trt.ScatterMode.ND`. However, I was unable to implement the `accumulate=True` case, which is currently handled by the validator.

Fixes # ([issue](https://github.com/pytorch/TensorRT/issues/2544))

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/select.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/select.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Optional, Sequence, Tuple, Union, cast
+from typing import Optional, Sequence, Union, cast
 
 import numpy as np
 import tensorrt as trt
@@ -455,7 +455,7 @@ def index_put_converter(
     source_ir: Optional[SourceIR],
     name: str,
     input_tensor: TRTTensor,
-    indices: Union[TRTTensor, Tuple[TRTTensor, ...]],
+    indices: Sequence[Union[TRTTensor, np.ndarray, torch.Tensor]],
     values: TRTTensor,
     accumulate: bool = False,
 ) -> TRTTensor:
--- a/py/torch_tensorrt/dynamo/conversion/impl/select.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/select.py
@@ -6,6 +6,7 @@ import tensorrt as trt
 import torch
 from torch.fx.node import Target
 from torch_tensorrt.dynamo._SourceIR import SourceIR
+from torch_tensorrt.dynamo.conversion import impl
 from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
 from torch_tensorrt.dynamo.conversion.converter_utils import (
     broadcastable,
@@ -410,7 +411,7 @@ def scatter(
     dim = get_positive_dim(dim, len(input_shape))
     src_tensor = src
     # scatter.value
-    if isinstance(src, int) or isinstance(src, float):
+    if isinstance(src, (int, float)):
         src_tensor = get_trt_tensor(
             ctx, src * np.ones(index_shape_list), name + "_value_tensor"
         )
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -428,3 +429,41 @@ def scatter(
...
+    source_ir: Optional[SourceIR],
+    name: str,
+    input_tensor: TRTTensor,
+    indices: Union[TRTTensor, Tuple[TRTTensor, ...]],
+    values: TRTTensor,
+    accumulate: bool = False,
+) -> TRTTensor:
+    # Reshape indices to add an extra dimension if necessary (indices is a Tuple of ITensors)
+    reshaped_indices = []
+    for i, each_input in enumerate(indices):
```
zewenli98, line 446: Since `indices` is possible to be `ITensor` per the schema, you may not be able to iterate an ITensor.
In the test case, you can try to change the line 173 to `inputs=[source_tensor, indices_tensor, value_tensor],`.
It's kind of similar to the `offsets` in the annoying `embedding_bag`. You can think about how to use native TRT Layers to do this, like ILoop.

zewenli98: Besides, what blocks you when `accumulate=True`?

chohk88: When `accumulate=True`, if there are duplicate pairs of index in `indices`, the corresponding `values` should be summed and then removed from the elements. Therefore, I aimed to obtain `indices` without duplicated pairs and corresponding modified `values`, and then use these to input into the scatter layer. However, I encountered difficulties in implementing the for loop to check for duplicate pairs of index in `indices`.

zewenli98: Thanks for the detailed explanations! Yes you are right, the `indices` should be list or tuple, and thus it could be iterated over. Then your current implementation LGTM.
> One more question I have is about the type definition of indices when it is a tuple of tensors. Is it correct to define indices as Union[TRTTensor, Tuple[TRTTensor, ...]]?

I think it could be `Sequence[Union[TRTTensor, np.ndarray, torch.Tensor]]` since a single TRTTensor cannot be iterated, right?

tests/py/dynamo/conversion/test_index_put_aten.py
apbose, line 1: The implementation looks good to me. Can you add a test case like this -
tensor = torch.zeros([4, 4, 4, 4], dtype = torch.int32)
indices = (torch.tensor([0,1,2,3]), torch.tensor([2,3, 1, 0]))
values = torch.tensor([10, 20, 30, 40],  dtype = torch.int32)
out = torch.index_put_(tensor, indices, values)

py/torch_tensorrt/dynamo/conversion/impl/select.py
apbose: When you say `accumulate=True` is causing issue, I believe the duplicate indices causes issues. I faced the same in `scatter_reduce` and I believe advanced indexing would be the way to deal with it (lengthy code that would be I believe :( ). Do you have any other ideas?

tests/py/dynamo/conversion/test_index_put_aten.py
narendasan: Lets write a validator for this case and resolve in a new PR

chohk88: I have write validator for broadcasting.

py/torch_tensorrt/dynamo/conversion/impl/select.py
chohk88: I have write a validator to handle the `accumulate=True` case. And I have created a separate [issue](https://github.com/pytorch/TensorRT/issues/2939) for implementing the converter for `accumulate=True`. It would be great to share ideas and work together on this.

PR #2881, opened 2024-06-03
feat: Cudagraphs integration for Torch-TRT + Non-default Stream Utilization

gs-olive:
# Description

- Add option to enable Cudagraphs in the runtime for additional acceleration of TRT engines
- Add C++ and Python toggles and well as full integration for C++ and Python runtimes
- Add support for dynamic shape cases via shape keys with cache invalidation
- Add test cases for cudagraphs support
- Add support for non-default CUDA Stream usage by default

Addresses #2736 

## Type of change

# Checklist:


core/runtime/execute_engine.cpp
```diff
@@ -168,27 +210,66 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
       auto dims = core::util::toVec(out_shape);
       auto type = util::TRTDataTypeToScalarType(compiled_engine->exec_ctx->getEngine().getTensorDataType(name.c_str()));
       outputs[pyt_idx] = std::move(at::empty(dims, {at::kCUDA}).to(type).contiguous());
+
+      // In cudagraphs mode, the allocated output buffers are stored for reuse
```
narendasan, line 214: Is there a downside to always using the buffers?

narendasan: Also what is the required lifetime of these buffers, can they be deallocated at any point or do they need to be reserved as long as the cuda graph is active?

core/runtime/execute_engine.cpp
```diff
@@ -168,27 +210,66 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
...
-          std::make_unique<torch::autograd::profiler::RecordProfile>(compiled_engine->enqueue_profile_path);
-    }
+  std::unique_ptr<torch::autograd::profiler::RecordProfile> enqueue_profiler_guard;
+
+  // nvinfer1::IExecutionContext::enqueue is not thread safe and we need a mutex for it.
+  std::unique_lock<std::mutex> lock(compiled_engine->mu);
+
+  if (!CUDAGRAPHS_MODE) {
+    // If not in cudagraphs mode, proceed with enqueueV3 as normal
     c10::cuda::CUDAStream stream = c10::cuda::getCurrentCUDAStream(inputs[0].device().index());
```
narendasan, line 229: Can we take this oppurtunity to fix the default stream issue? I think we need to create a new stream for the enqueue operation

core/runtime/execute_engine.cpp
```diff
@@ -168,27 +210,66 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
...
-    std::unique_lock<std::mutex> lock(compiled_engine->mu);
     compiled_engine->exec_ctx->enqueueV3(stream);
-    if (compiled_engine->profile_execution) {
-      LOG_INFO(std::endl << *compiled_engine->trt_engine_profiler);
-      dump_trace(compiled_engine->trt_engine_profile_path, *compiled_engine->trt_engine_profiler);
-      compiled_engine->dump_engine_layer_info();
+  } else if (need_cudagraphs_record) {
+    // If cudagraphs needs to record a graph, capture the enqueueV3 call in a graph
+
+    // Cudagraphs cannot record on the default stream, so use an alternate
```
narendasan, line 234: Lets reuse this for both modes

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -270,3 +325,23 @@ def get_layer_info(self) -> str:
...
+
+    def cudagraphs_validate_shapes(self, inputs: Sequence[torch.Tensor]) -> bool:
+        """
+        Validates the input shapes of the forward function
+        versus the version currently active for the
+        """
+        # Representation of input shapes to a given model
+        # Shapes are concatenated as so:
+        # x: (3, 4), y: (4, 5) --> Key: (3, 4)(4, 5)
+        new_shape_key = "".join(str(tuple(t.shape)) for t in inputs)
```
narendasan, line 337: Can we have no whitespace in the key, ensures that there is only one valid representation of a key

py/torch_tensorrt/runtime/cudagraphs.py
```diff
@@ -0,0 +1,55 @@
+import logging
+from importlib.util import find_spec
+from typing import Any
+
+import torch
+
+if find_spec("torch_tensorrt._C") is not None:
```
narendasan, line 7: You can use the ENABLED_FEATURES structure for this

py/torch_tensorrt/runtime/cudagraphs.py
```diff
@@ -0,0 +1,55 @@
...
+logger = logging.getLogger(__name__)
+
+
+def set_cudagraphs_mode(mode: bool) -> None:
+    # Set new cudagraphs mode for Python
+    global _PY_RT_CUDAGRAPHS
+    _PY_RT_CUDAGRAPHS = mode
+
+    # Set new mode for C++
+    if find_spec("torch_tensorrt._C") is not None:
```
narendasan, line 22: same here

tests/py/dynamo/runtime/test_cudagraphs.py
```diff
@@ -0,0 +1,112 @@
+import unittest
+
+import torch
+from torch.testing._internal.common_utils import TestCase, run_tests
+
+import torch_tensorrt
+
+from ..testing_utilities import DECIMALS_OF_AGREEMENT
+
+
```
narendasan, line 10: We should only skip the CPP tests specifically right? What about python runtime tests?

core/runtime/execute_engine.cpp
```diff
@@ -58,6 +59,29 @@ RTDevice select_rt_device(const RTDevice& engine_device, const RTDevice& curr_de
...
 }
 
+bool _cudagraphs_validate_shapes(std::vector<at::Tensor> inputs, c10::intrusive_ptr<TRTEngine> compiled_engine) {
+  // Validate whether the current input shapes to the engine
+  // invalidate the existing cudagraphs object
+
+  // Populate the shape key for the inputs
+  std::stringstream new_shape_key_ss;
+  for (auto input : inputs) {
+    new_shape_key_ss << input.sizes();
```
narendasan, line 69: Does this create the same shape key as python?

core/runtime/execute_engine.cpp
gs-olive: If not in cudagraphs mode, the main downside is the requirement that buffers need to change size/be reallocated in Dynamic shape usecases.

My understanding is that these I/O buffers need to be reserved for the duration of the cudagraph being active

core/runtime/execute_engine.cpp
gs-olive: Yes, I think in both the Python and C++ cases, we can use a non-default stream. I am not sure if we want to grab a new stream from the pool on every `forward` pass, or if we should store the stream after first inference call.

core/runtime/execute_engine.cpp
gs-olive: I believe so - I will check the bracketing to see if `.sizes()` uses `[]` or `()`, but the scheme is the same. I will remove the whitespace in this case as well.

core/runtime/execute_engine.cpp
gs-olive: `self.stream`
--> Synchronize at I/O boundaries
```python
stream = torch.cuda.Stream()
torch.cuda.set_stream(stream)
```

narendasan:
```py
torch.cuda.synchronize()
with torch.cuda.stream(s1):
    C = torch.mm(A, A)
```

peri044 (commented):
LGTM

peri044 (approved):
LGTM

zewenli98 (approved):
LGTM

PR #2898, opened 2024-06-07
feat: Use a global timing cache and add a save option

peri044:
# Description

1) Adds an option to specify timing cache path. This is now a global cache used across multiple TRT engines. 
2) Serializes the cache to ensure this can be reused in the next runs.

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/_defaults.py
```diff
@@ -28,6 +28,7 @@
 DRYRUN = False
 HARDWARE_COMPATIBLE = False
 SUPPORTED_KERNEL_PRECISIONS = {dtype.f32, dtype.f16, dtype.bf16, dtype.i8, dtype.f8}
+TIMING_CACHE_PATH = "/tmp/timing_cache.bin"
```
HolyWu, line 31: Could you use a more cross-platform default path because `/tmp` is not a valid path on Windows

peri044: I modified it to use `tempfile.gettempdir()` now which gives `C:\\Users\\GSTEFA~1\\AppData\\Local\\Temp` on windows and `/tmp` on linux

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -276,30 +277,48 @@ def _populate_trt_builder_config(
...
-        else:
-            cache = builder_config.create_timing_cache(b"")
+        timing_cache_path: str = "",
+    ) -> None:
+        """
+        Create a timing cache to enable faster build time for TRT engines.
+        By default the timing_cache_path="/tmp/timing_cache.bin"
+        """
+        buffer = b""
+        if os.path.exists(timing_cache_path):
```
HolyWu, line 287: Consider using `os.path.isfile` because `os.path.exists` also returns True for directory.

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -276,30 +277,48 @@ def _populate_trt_builder_config(
...
+    ) -> None:
+        """
+        This is called after a TensorRT engine is built. Save the timing cache
+        """
+        timing_cache = builder_config.get_timing_cache()
+        if timing_cache_path:
+            with open(timing_cache_path, "wb") as timing_cache_file:
+                timing_cache_file.write(memoryview(timing_cache.serialize()))
+        else:
+            tmp_dir_path = os.path.join(tempfile.gettempdir(), "timing_cache.bin")
```
HolyWu, line 307: Is it not preferred that simply assign `TIMING_CACHE_PATH` in `_defaults.py` to `os.path.join(tempfile.gettempdir(), "timing_cache.bin")`? Otherwise `_create_timing_cache` will always receive an empty string by default and hence a new cache will be created every time.

peri044: I don't have a strong preference here. With the first run, there's a possibility that a torch-trt session might be using a timing cache from another model, but there's no harm in that I guess.

narendasan (approved):
LGTM

PR #2915, opened 2024-06-13
Better random input for dynamic shape

keehyuna:
# Description

Random bool/integer type input for dynamic shape is casted from random float in [0,1)
All zero input might cause false positive test result.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/_Input.py
+++ b/py/torch_tensorrt/_Input.py
@@ -352,9 +352,20 @@ class Input(object):
                 )
             else:
                 if isinstance(self.shape, tuple):
-                    return torch.rand(self.shape).to(
-                        dtype=self.dtype.to(torch.dtype, use_default=True)
-                    )
+                    if self.dtype in [dtype.u8, dtype.i8, dtype.i32, dtype.i64]:
+                        type = self.dtype.to(torch.dtype, use_default=True)
+                        return torch.randint(
+                            torch.iinfo(type).min,
+                            torch.iinfo(type).max,
+                            self.shape,
+                            dtype=type,
+                        )
+                    elif self.dtype == dtype.b:
+                        return torch.rand(self.shape) < 0.5
+                    else:
+                        return torch.rand(self.shape).to(
+                            dtype=self.dtype.to(torch.dtype, use_default=True)
+                        )
                 else:
                     RuntimeError(
                         f"Input shape is dynamic but shapes are not provided as sequence (found: {self.shape})"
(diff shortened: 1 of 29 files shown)


py/torch_tensorrt/_Input.py
```diff
@@ -352,9 +353,20 @@ def example_tensor(
                 )
             else:
                 if isinstance(self.shape, tuple):
-                    return torch.rand(self.shape).to(
-                        dtype=self.dtype.to(torch.dtype, use_default=True)
-                    )
+                    if self.dtype in [dtype.u8, dtype.i8, dtype.i32, dtype.i64]:
```
chohk88, line 356: It appears that the changes ensure values are within the range of -128 to 127 for `dtype.i8`, `dtype.i32`, and `dtype.i64`, and within the range of 0 to 127 for `dtype.u8`. 

Previously, it seems there was an issue where casting random values between 0 and 1 to integers resulted in most values being 0.

This fix seems necessary and addresses the problem effectively.

tests/py/dynamo/conversion/harness.py
```diff
@@ -353,6 +353,15 @@ def run_test_compare_tensor_attributes_only(
             debug=True,
         )
 
+        output_dtypes = None
+        if check_dtype:
```
chohk88, line 357: I believe your modification is correct, but I'm curious about the history. Why was it originally necessary to specify `output_dtypes` as an input in the `dynamic shape test` function? It seems that specifying `output_dtypes` is no longer needed, and your modification looks like the right approach.

chohk88 (approved):
LGTM, but I didn't know the history of `output_dtypes` so I left a comment about it.

tests/py/dynamo/conversion/harness.py
keehyuna: Thanks, I agree. it was included from other issue.

py/torch_tensorrt/_Input.py
keehyuna: Yes, but It caused problem in model test as random int value was used in fake tensor as index.
Added option to initialize it with zero.

py/torch_tensorrt/_Input.py
```diff
@@ -183,9 +183,9 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
                     kwargs["opt_shape"], dtype=kwargs["dtype"]
                 )
             elif self.shape_mode == Input._ShapeMode.DYNAMIC:
-                self.torch_tensor = self.example_tensor("opt_shape")
+                self.torch_tensor = self.example_tensor("opt_shape", zero_filled=True)
             else:
-                self.torch_tensor = self.example_tensor()
+                self.torch_tensor = self.example_tensor(zero_filled=True)
```
keehyuna, line 188: If input is used as fake tensor and used as index, it caused cuda error. Input buffer is initialized as zero.

py/torch_tensorrt/_Input.py
```diff
@@ -384,4 +384,21 @@ def example_tensor(
                 raise ValueError(
                     "Requested an example tensor from a dynamic shaped input but did not specific which profile field to use."
                 )
+        type = self.dtype.to(torch.dtype, use_default=True)
+        if zero_filled:
+            return torch.zeros(shape, dtype=type)
+        else:
+            if self.dtype in [dtype.u8, dtype.i8, dtype.i32, dtype.i64]:
+                return torch.randint(
+                    max(torch.iinfo(torch.int8).min, torch.iinfo(type).min),
```
keehyuna, line 393: unary like ACOS does not support int32 in the TRT, but supports int8. Using 8bit range random integer to cover all int type.

PR #2917, opened 2024-06-13
full_like evaluator

apbose:


py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -18,6 +18,7 @@
 )
 from torch_tensorrt.dynamo.conversion.impl.elementwise import sub, trunc_div
 from torch_tensorrt.fx.types import TRTTensor
+from torch_tensorrt.fx.utils import Frameworks, unified_dtype_converter
```
peri044, line 21: We have our own types and switch from `unified_dtype_converter` to using `_enums.dtype.to` and `_enums.dtype.from` for type conversions. We shouldn't be using FX utils anymore.

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -165,3 +166,49 @@ def aten_ops_randperm(
...
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    full_like_np_tensor = None
+    tensor_shape = args[0].shape
+    if kwargs.get("dtype") is not None:
+        full_like_np_tensor = np.full(
```
peri044, line 202: Can you also add dynamic shape support ? what if the input shape has -1 in it ? Please refer to fill layer which can help you out here.

peri044 (changes requested):

peri044:
@apbose  what is the status of full_like ?

apbose:
I did not work on this since long. I should be able to take a look at this next week when I am done with dynamic shape of aten::chunk and aten::tile and merge the remaining PRs.

peri044:
Closing this as we now have a lowering pass that handles this.

PR #2918, opened 2024-06-13
chore: fix ValueRanges computation in symbolic nodes

peri044:
# Description

This PR fixes the following: 

1) Sometimes there are subgraphs which receive a SymInt  node (`so*s1`) as input ( that's the product of two other SymInt  `s0` and `s1`). However, this node (`so*s1`) does not have ValueRanges computed automatically but it just has the data for `s0` and `s1`

2) It also adds support for `operator.mul`. This operator gets added in the added test_linear test case where it computes product of first two dimensions (which are `SymInt` nodes). Since `SymInt` nodes are handled as shape tensors in TensorRT, our existing converter for `mul` operators can be used for this operator.

## Type of change

# Checklist:


peri044:
Context: https://github.com/pytorch/pytorch/issues/128640

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1779,6 +1779,7 @@ def aten_ops_add(
     )
 
 
+@dynamo_tensorrt_converter(operator.mul, supports_dynamic_shapes=True)
```
gs-olive, line 1782: Is `operator.mul` elicited outside of dynamic Linear by any chance? As in, is there a smaller model which can result in this operator appearing

py/torch_tensorrt/dynamo/partitioning/common.py
```diff
@@ -35,8 +35,12 @@ def construct_dynamic_input(
...
             expr = node.expr
             shape_env = node.shape_env
-            var_range = shape_env.var_to_range.get(expr, None)
-            var_val = shape_env.var_to_val.get(expr, None)
+            var_range = shape_env.var_to_range.get(expr, None) or shape_env.bound_sympy(
+                expr
+            )
+            var_val = shape_env.var_to_val.get(expr, None) or expr.xreplace(
+                shape_env.var_to_val
+            )
```
gs-olive, line 43: Does this line prompt evaluation of the symbolic expression, like `s0*s1`, which has been mentioned in the PR summary? Could a comment be added to the above two lines to describe their effect on the symbolic expressions

gs-olive (commented):
Overall looks good - added two clarifying comments

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
peri044: Maybe, I couldn't reproduce a model with just this operator. The smallest one is the linear example I added. Often times, if you just trace `x*2` or `operator.mul(x, 2` it records `torch.ops.aten.mul.Tensor ` in the graph.

py/torch_tensorrt/dynamo/partitioning/common.py
peri044: added it now.

gs-olive (approved):
Looks good to me, pending CI

PR #2925, opened 2024-06-15
scatter CI failures

apbose:
]


py/torch_tensorrt/dynamo/lowering/_decompositions.py
apbose, line 1: @gs-olive will there be fake tensor issue in this? The torch.compile test seems to be passing here.

gs-olive: Since it is within a decomposition, it should be fine since the `torch.ones` Tensor itself is never accessed or used

gs-olive (approved):
Looks good to me - thanks for the quick fix! Just pending 1 CI test case

PR #2929, opened 2024-06-17
chore: dynamic shape support for rsqrt/erf ops

keehyuna:
# Description

dynamic shape support for rsqrt/erf ops

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/tests/py/dynamo/conversion/test_rsqrt_aten.py
+++ b/tests/py/dynamo/conversion/test_rsqrt_aten.py
@@ -10,16 +10,21 @@ from .harness import DispatchTestCase
 class TestRSqrtConverter(DispatchTestCase):
     @parameterized.expand(
         [
-            ("2d_dim_alpha", (2, 1), 2),
-            ("3d_dim_alpha", (2, 1, 2), 2),
+            ("2d_dim_float", (2, 1), torch.float),
+            ("3d_dim_float", (2, 1, 2), torch.float),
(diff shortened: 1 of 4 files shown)


tests/py/dynamo/conversion/test_rsqrt_aten.py
```diff
@@ -25,6 +25,46 @@ def forward(self, input):
...
+                torch.half,
+            ),
+            (
+                "3d_dim_dtype_float",
+                (1, 1, 1),
+                (1, 2, 3),
+                (3, 3, 3),
+                torch.float,
+                torch.float,
+            ),
```
peri044, line 45: Can you add a testcase with int32 as input and it returns float correct ?

peri044 (commented):
LGTM. minor change

tests/py/dynamo/conversion/test_rsqrt_aten.py
keehyuna: Thanks for pointing this out. Integer input has issue with both static/dynamic input.
below error from TRT when int input is passed to SQRT. I added cast to float
ERROR:torch_tensorrt [TensorRT Conversion Context]:3: [UNARY]-[aten_ops.rsqrt.default]-[rsqrt_default_sqrt]: operation SQRT not allowed on type Int32
https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/conversion/impl/elementwise/ops.py#L124-L131

PR #2931, opened 2024-06-17
feat: support `native_dropout` dynamo converter

### LINKED ISSUE #2494: `aten.native_dropout`

zewenli98:
# Description

Support `native_dropout` dynamo converter.

Per [the schema](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L283), this function returns two tensors. The first is `output` and the other is `mask`. For inference, i.e., `train` is not `True` or `None`, whatever the `p` is, it always returns `input` and all `True`s.

Fixes #2494 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -3247,7 +3247,13 @@ def aten_ops_index_select(
 
 def dropout_inference_validator(node: Node) -> bool:
     train_mode = args_bounds_check(node.args, 2, None)
-    return train_mode is False
+    if train_mode is False:
+        return True
+    else:  # train_mode is True or None
+        _LOGGER.debug(
+            "Currently only inference mode is supported for dropout operation."
+        )
+        return False
 
 
 @dynamo_tensorrt_converter(
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -3243,3 +3243,30 @@ def aten_ops_index_select(
         args[1],
         args[2],
     )
+
+
+def dropout_inference_validator(node: Node) -> bool:
+    train_mode = args_bounds_check(node.args, 2, None)
+    return train_mode is False
+
+
+@dynamo_tensorrt_converter(
+    torch.ops.aten.native_dropout.default,
+    capability_validator=dropout_inference_validator,
+)
+def aten_ops_native_dropout(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    return impl.unary.native_dropout(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        args[0],
+        args[1],
+        args_bounds_check(args, 2, None),
+    )
--- a/tests/py/dynamo/conversion/test_dropout_aten.py
+++ b/tests/py/dynamo/conversion/test_dropout_aten.py
@@ -1,0 +1,56 @@
+import torch
+import torch.nn as nn
+from parameterized import parameterized
+from torch.testing._internal.common_utils import run_tests
+
+from .harness import DispatchTestCase
+
+
+class TestDropOutConverter(DispatchTestCase):
+    @parameterized.expand(
+        [
+            ((10,), 0, False),
+            ((1, 3), 0.3, False),
+            ((2, 2, 2), 0.5),
+            ((2, 2, 2, 2), 1),
+        ]
+    )
+    def test_native_dropout(self, input_shape, p, train=False):
+        class NativeDropout(nn.Module):
+            def forward(self, input):
+                return torch.ops.aten.native_dropout.default(input, p, train)
+
+        inputs = [torch.randn(input_shape)]
+        self.run_test(
+            NativeDropout(),
+            inputs,
+        )
+
+    @parameterized.expand(
+        [
+            (
+                torch.randn(
+                    10,
+                ),
+                0,
+                False,
+            ),
+            (torch.randn(1, 3), 0.3, False),
+            (torch.randn(2, 2, 2), 0.5),
+            (torch.randn(2, 2, 2, 2), 1),
+        ]
+    )
+    def test_native_dropout_pytorch(self, input, p, train=False):
+        class NativeDropout(nn.Module):
+            def forward(self):
+                return torch.ops.aten.native_dropout.default(input, p, train)
+
+        inputs = []
+        self.run_test(
+            NativeDropout(),
+            inputs,
+        )
+
+
+if __name__ == "__main__":
+    run_tests()
--- a/py/torch_tensorrt/dynamo/conversion/impl/unary/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/unary/ops.py
@@ -571,3 +571,20 @@ def isnan(
     )
 
     return nan_values_mask
+
+
+def native_dropout(
+    ctx: ConversionContext,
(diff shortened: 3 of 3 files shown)


tests/py/dynamo/conversion/test_dropout_aten.py
```diff
@@ -0,0 +1,56 @@
...
+            (torch.randn(2, 2, 2), 0.5),
+            (torch.randn(2, 2, 2, 2), 1),
+        ]
+    )
+    def test_native_dropout_pytorch(self, input, p, train=False):
+        class NativeDropout(nn.Module):
+            def forward(self):
+                return torch.ops.aten.native_dropout.default(input, p, train)
+
+        inputs = []
```
chohk88, line 48: Could you please explain the reason for dividing the test cases into `inputs = [torch.randn(input_shape)]` and `inputs = []`?

tests/py/dynamo/conversion/test_dropout_aten.py
```diff
@@ -0,0 +1,56 @@
...
+            ((2, 2, 2), 0.5),
+            ((2, 2, 2, 2), 1),
+        ]
+    )
+    def test_native_dropout(self, input_shape, p, train=False):
+        class NativeDropout(nn.Module):
+            def forward(self, input):
+                return torch.ops.aten.native_dropout.default(input, p, train)
+
+        inputs = [torch.randn(input_shape)]
```
chohk88, line 23: The test cases with `inputs = [torch.randn(input_shape)] `fail both in my local tests and in the CI.

```
======================================================================
ERROR: test_native_dropout_0 (__main__.TestDropOutConverter)
----------------------------------------------------------------------
  File "/home/hoonkyungc/miniconda3/envs/torch-trt-10/lib/python3.9/site-packages/torch/testing/_internal/common_utils.py", line 2741, in wrapper
    method(*args, **kwargs)
  File "/home/hoonkyungc/miniconda3/envs/torch-trt-10/lib/python3.9/site-packages/parameterized/parameterized.py", line 620, in standalone_func
    return func(*(a + p.args), **p.kwargs, **kw)
  File "/home/hoonkyungc/workspace/TorchTRT/TensorRT/tests/py/dynamo/conversion/test_dropout_aten.py", line 24, in test_native_dropout
    self.run_test(
  File "/home/hoonkyungc/workspace/TorchTRT/TensorRT/tests/py/dynamo/conversion/harness.py", line 311, in run_test
    super().run_test(
  File "/home/hoonkyungc/workspace/TorchTRT/TensorRT/tests/py/dynamo/conversion/harness.py", line 72, in run_test
  File "/home/hoonkyungc/workspace/TorchTRT/TensorRT/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py", line 323, in run
    super().run()
  File "/home/hoonkyungc/miniconda3/envs/torch-trt-10/lib/python3.9/site-packages/torch/fx/interpreter.py", line 145, in run
    self.env[node] = self.run_node(node)
  File "/home/hoonkyungc/workspace/TorchTRT/TensorRT/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py", line 364, in run_node
    trt_node: torch.fx.Node = super().run_node(n)
  File "/home/hoonkyungc/miniconda3/envs/torch-trt-10/lib/python3.9/site-packages/torch/fx/interpreter.py", line 202, in run_node
    return getattr(self, n.op)(n.target, args, kwargs)
  File "/home/hoonkyungc/workspace/TorchTRT/TensorRT/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py", line 472, in call_function
    return converter(self.ctx, target, args, kwargs, self._cur_node_name)
  File "/home/hoonkyungc/workspace/TorchTRT/TensorRT/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py", line 3264, in aten_ops_native_dropout
    return impl.unary.native_dropout(
  File "/home/hoonkyungc/workspace/TorchTRT/TensorRT/py/torch_tensorrt/dynamo/conversion/impl/unary/ops.py", line 588, in native_dropout
    mask = np.ones(input_val.shape, dtype=np.bool)
  File "/home/hoonkyungc/miniconda3/envs/torch-trt-10/lib/python3.9/site-packages/numpy/__init__.py", line 324, in __getattr__
    raise AttributeError(__former_attrs__[attr])
AttributeError: module 'numpy' has no attribute 'bool'.
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

```

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3243,3 +3243,30 @@ def aten_ops_index_select(
         args[1],
         args[2],
     )
+
+
+def dropout_inference_validator(node: Node) -> bool:
```
chohk88, line 3248: I think it would be good to add a warning message like 'The current implementation of native_dropout only supports inference mode (train=False)'.

chohk88 (changes requested):

tests/py/dynamo/conversion/test_dropout_aten.py
zewenli98: If passing in `inputs = [torch.randn(input_shape)]`, the input will be an ITensor. If passing the input directly into the pytorch function, the input will be a normal tensor.

tests/py/dynamo/conversion/test_dropout_aten.py
zewenli98: Thanks for the details!

tests/py/dynamo/conversion/test_dropout_aten.py
chohk88: LGTM. The code looks good to merge.

I have one more question about input types for my understanding. Usually, we handle inputs as `input: TRTTensor` and only consider ITensor input. However, sometimes we use `input_val: Union[TRTTensor, torch.Tensor, np.ndarray]` to handle both ITensor and normal tensor types. Even though the schema doesn't show a difference, how do we distinguish between them?

chohk88: LGTM. The code looks good to merge.

I have one more question about input types for my understanding. Usually, we handle inputs as input: TRTTensor and only consider ITensor input. However, sometimes we use input_val: Union[TRTTensor, torch.Tensor, np.ndarray] to handle both ITensor and normal tensor types. Even though the schema doesn't show a difference, how do we distinguish between them?

chohk88 (approved):

tests/py/dynamo/conversion/test_dropout_aten.py
zewenli98: Yeah very good questions.

> 1. Even though the schema doesn't show a difference, how do we distinguish between them?

From my understanding, `ITensor` is a concept in TRT so PyTorch schema isn't able to distinguish them. (TRT is just one backend among many backends)

> 2. Usually, we handle inputs as input: TRTTensor and only consider ITensor input. However, sometimes we use input_val: Union[TRTTensor, torch.Tensor, np.ndarray] or input: Sequence[Union[TRTTensor, torch.Tensor, np.ndarray]] to handle both ITensor and normal tensor types.

I think in the entry point the inputs are not promised to be `ITensor`. Instead, they could be `torch.Tensor` or `np.ndarray`. So there are two ways to handle it. 1) use `enforce_tensor_types` to make it as `ITensor` 2) manually cast it to `ITensor` inside of the function. I think in most cases both are fine, but if this function is called by another function that passes in non-`ITensor`, the first solution will error out.

chohk88: I understand better now. Thank you so much for the explanation!

PR #2855, opened 2024-05-22
[WIP] upgrade to public TRT 10.1.0.27

zewenli98:
# Description

Upgrade to TRT 10.1.0.27

Fixes # (issue)

## Type of change

Please delete options that are not relevant and/or add your own.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

# Checklist:

- [ ] My code follows the style guidelines of this project (You can use the linters)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #2937, opened 2024-06-18
support native int64 as valid datatype

apbose:


HolyWu:
It seems that https://github.com/pytorch/TensorRT/pull/2855/files#diff-97c67f0cea25ec7aa88ca61594d9ad62e9fb575cab3647b079cb718f81308454 is already planned to do relevant changes.

apbose:
Thanks @HolyWu for pointing out the above. @narendasan would the changes in PRhttps://github.com/pytorch/TensorRT/pull/2855/files#diff-97c67f0cea25ec7aa88ca61594d9ad62e9fb575cab3647b079cb718f81308454 work or would we have to create a type map for kINT32 to kINT64 in the torchscript path?

narendasan:
Yeah neither of these PRs implements the feature fully since we need to either amend truncation in TorchScript or have a TorchScript specific type map

narendasan:
@apbose going to close this. reopen a new pr with the comprehensive changes if its still needed

PR #2979, opened 2024-07-02
feat: add dynamic support for eq/ne/lt/le

chohk88:
# Description

Add dynamic shape for eq/ne/lt/le for Tensor and Scalar operations.

## Type of change

Please delete options that are not relevant and/or add your own.

- New feature (non-breaking change which adds functionality)

# Checklist:

- [ ] My code follows the style guidelines of this project (You can use the linters)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #2982, opened 2024-07-03
feat: dynamic shape support for pow/mod/eq operator

keehyuna:
# Description

dynamic shape support for pow/mod/eq operator
Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2002,6 +2002,7 @@ def aten_ops_div(
 @dynamo_tensorrt_converter(
     torch.ops.aten.pow.Tensor_Scalar, supports_dynamic_shapes=True
 )
+@dynamo_tensorrt_converter(operator.pow, supports_dynamic_shapes=True)
 def aten_ops_pow(
     ctx: ConversionContext,
     target: Target,
@@ -2278,6 +2279,7 @@ def aten_ops_bitwise_not(
 
 @dynamo_tensorrt_converter(torch.ops.aten.eq.Tensor, supports_dynamic_shapes=True)
 @dynamo_tensorrt_converter(torch.ops.aten.eq.Scalar, supports_dynamic_shapes=True)
+@dynamo_tensorrt_converter(operator.eq, supports_dynamic_shapes=True)
 @enforce_tensor_types(
     {
         0: (TRTTensor,),
@@ -3149,8 +3151,13 @@ def aten_ops_copy(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.remainder.Scalar)
-@dynamo_tensorrt_converter(torch.ops.aten.remainder.Tensor)
+@dynamo_tensorrt_converter(
+    torch.ops.aten.remainder.Scalar, supports_dynamic_shapes=True
+)
+@dynamo_tensorrt_converter(
+    torch.ops.aten.remainder.Tensor, supports_dynamic_shapes=True
+)
+@dynamo_tensorrt_converter(operator.mod, supports_dynamic_shapes=True)
 @enforce_tensor_types(
     {
         0: (TRTTensor,),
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2278,6 +2279,7 @@ def aten_ops_bitwise_not(
 
 @dynamo_tensorrt_converter(torch.ops.aten.eq.Tensor)
 @dynamo_tensorrt_converter(torch.ops.aten.eq.Scalar)
+@dynamo_tensorrt_converter(operator.eq, supports_dynamic_shapes=True)
```
keehyuna, line 2282: I will rebase after https://github.com/pytorch/TensorRT/pull/2979/files#diff-9cbf43cfb62cdf682eef77f6fd5cabc488bb5a91ff09bfee5d593827542b2476R2256-R2257 is merged

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2002,6 +2002,7 @@ def aten_ops_div(
 @dynamo_tensorrt_converter(
     torch.ops.aten.pow.Tensor_Scalar, supports_dynamic_shapes=True
 )
+@dynamo_tensorrt_converter(operator.pow, supports_dynamic_shapes=True)
```
chohk88, line 2005: From the PR (https://github.com/pytorch/TensorRT/pull/2918), it looks like dynamic testing and registration for `operator.mul` were added because PyTorch internally uses `operator.mul` when using `torch.nn.Linear` layers. This might be why `operator.mul` was registered with the converter. 

If I'm misunderstanding, could you please explain? (cc. @peri044 )

Then, in this PR, why do we need to register `operator.pow`, `operator.eq`, and `operator.mod` with the converter?

zewenli98: It seems `operator.*` overrides original Python ops, which results in running any Python op will create a TRT Layer to handle it. However, they are not listed in [the schema](https://github.com/pytorch/pytorch/blob/a408920817ac3852d27fc96ba85109ff4c981d52/aten/src/ATen/native/native_functions.yaml). I'm also curious if these ops' converters are needed

keehyuna: Thanks @chohk88 / @zewenli98
If operator.pow is missing and ** is used in module, there is exception from TRTInterpreter.

E torch_tensorrt.dynamo.conversion._TRTInterpreter.UnsupportedOperatorException: Conversion of function _operator.pow not currently supported!

I saw such usage of operator in openchat model. I think we need to register these operators to support such usage.
https://github.com/imoneoi/openchat/blob/master/ochat/models/unpadded_llama.py#L105

zewenli98: Thanks for the example. If then, it seems to be necessary I think. 
cc: @narendasan @dheerajperi

tests/py/dynamo/conversion/test_pow_aten.py
```diff
@@ -59,6 +59,65 @@ def forward(self, lhs_val):
...
+                torch.float,
+            ),
+        ]
+    )
+    def test_pow_dynamic_shape(
+        self, _, min_shape, opt_shape, max_shape, type, output_type
+    ):
+        class pow(nn.Module):
+            def forward(self, lhs_val, rhs_val):
+                return torch.ops.aten.floor_divide.default(lhs_val, rhs_val)
```
peri044, line 87: Why does pow test case use floor_divide here ?

peri044 (commented):
Overall LGTM. Added a question

tests/py/dynamo/conversion/test_pow_aten.py
keehyuna: Thanks for pointing this. I fixed.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
chohk88: Thanks for the example. LGTM!

PR #2990, opened 2024-07-09
feat: dynamic shape support for aten.select.int

chohk88:
# Description

Support dynamic shapes for aten.select.int. As shown below, since the arguments dim and index for select.int are int and not list, no reshaping is required.

```
- func: select.int(Tensor(a) self, int dim, SymInt index) -> Tensor(a)
  variants: function, method
  device_check: NoCheck
  device_guard: False
  dispatch:
    CompositeExplicitAutograd: select_symint
    SparseCsrCPU, SparseCsrCUDA: select_sparse_csr
    NestedTensorCPU, NestedTensorCUDA: select_nested
  tags: core
```

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/select.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/select.py
@@ -47,26 +47,16 @@ def select(
     if dynamic_shape:
         # Check whether slice target dim is dynamic shape dim
         assert input.shape[dim] != -1, "Can't select on negative shape dimension!"
-    index = index
 
     if index >= input.shape[dim]:
         raise RuntimeError(
             f"cannot have index greater than the dimension length! {input.shape[dim]}"
         )
-    output_shape = list(input.shape)
-    output_shape[dim] = 1
-    if dynamic_shape > 0:
-        output_shape = get_shape_with_dynamic_shape(
-            ctx, target, source_ir, name, output_shape, input
-        )
+
     index_value = np.array(index, dtype=np.int32)
-    indices_tensor = ctx.net.add_constant(
-        index_value.shape, to_numpy(index_value)
-    ).get_output(0)
+    indices_tensor = ctx.net.add_constant(index_value.shape, index_value).get_output(0)
     layer = ctx.net.add_gather(input, indices_tensor, dim)
-    out = layer.get_output(0)
-    if len(out.shape) != 1:
-        layer = ctx.net.add_shuffle(out)
+
     return layer.get_output(0)
 
 
--- a/py/torch_tensorrt/dynamo/conversion/impl/select.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/select.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Optional, Sequence, Union, cast
+from typing import Optional, Sequence, Union
 
 import numpy as np
 import tensorrt as trt
@@ -21,7 +21,7 @@ from torch_tensorrt.fx.converters.converter_utils import (
     has_dynamic_shape,
     set_layer_name,
 )
-from torch_tensorrt.fx.types import Shape, TRTTensor
+from torch_tensorrt.fx.types import TRTTensor
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
@@ -32,8 +32,8 @@ def select(
     source_ir: Optional[SourceIR],
     name: str,
     input: TRTTensor,
-    dim: Shape,
-    index: Shape,
+    dim: int,
+    index: int,
 ) -> TRTTensor:
     if not isinstance(input, TRTTensor):
         raise RuntimeError(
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -47,26 +47,16 @@ def select(
...
             f"cannot have index greater than the dimension length! {input.shape[dim]}"
         )
-    output_shape = list(input.shape)
-    output_shape[dim] = 1
-    if dynamic_shape > 0:
-        output_shape = get_shape_with_dynamic_shape(
-            ctx, target, source_ir, name, output_shape, input
-        )
+
     index_value = np.array(index, dtype=np.int32)
```
peri044, line 56: Can index itself be dynamic (ITensor)?

peri044 (commented):
Mark `torch.ops.aten.select.int` with supports_dynamic_shapes=True

py/torch_tensorrt/dynamo/conversion/impl/select.py
chohk88: Based on the [PyTorch docs](https://pytorch.org/docs/stable/generated/torch.select.html) and [the schema](https://github.com/pytorch/pytorch/blob/a408920817ac3852d27fc96ba85109ff4c981d52/aten/src/ATen/native/native_functions.yaml#L4911), the `index` is always an integer and cannot be a list or tuple of integers. The size of the `indices_tensor` created with `index` and `index_value` will always be a scalar. Therefore, it cannot be dynamic.

chohk88:
> Mark `torch.ops.aten.select.int` with supports_dynamic_shapes=True

I have marked `torch.ops.aten.select.int` with supports_dynamic_shapes=True.

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -47,26 +47,16 @@ def select(
...
         raise RuntimeError(
             f"cannot have index greater than the dimension length! {input.shape[dim]}"
         )
-    output_shape = list(input.shape)
-    output_shape[dim] = 1
-    if dynamic_shape > 0:
-        output_shape = get_shape_with_dynamic_shape(
-            ctx, target, source_ir, name, output_shape, input
-        )
+
```
keehyuna, line 55: If above asserts() are removed and full dynamic shape was used(e.g.(-1,-1,-1)), test worked.
I'm wondering if select on dynamic dim can be supported.

chohk88: I have tested the test you mentioned and confirmed that using -1, -1, -1 also passes the test cases successfully.

However, I didn't remove the asserts() because the test cases fail when the `index` is out of range, meaning the `index` is larger than the corresponding dimension in the dynamic input shape. 

```
            (
                "success case",
                (1, 1, 1),
                (2, 2, 2),
                (3, 3, 3),
                torch.float,
                0,
                1,
            ),
            (
                "fail case",
                (1, 1, 1),
                (2, 2, 2),
                (3, 3, 3),
                torch.float,
                0,
                3,
            ),
```

It seems that modifying the `index` like dimension (`dim = get_positive_dim(cast(int, dim), ranks)`) when the `index` is greater than the `size` would solve the issue. Do you have an example that handles it this way?

keehyuna: I think we cannot check invalid index for dynamic dim. Error will happen in runtime.
Maybe we can check index for only static dim.
if DYNAMIC_DIM != input.shape[dim] and index >= input.shape[dim]:

chohk88: I also tested by removing the `assert()` and `raise RuntimeError()` statements and tested with (-1, -1, -1). 

The test case mentioned as a `success case` above passes, but the test case mentioned as a `fail case` fails. The reason for this is that in the `fail case`, the size of the 0-th dimension is 3, but the `index` selects positions larger than that (`index`=3, 4th position). The current converter does not handle this case where the `index` is larger than the size of the dimension specified by `dim`, as in the 'fail case'. (Note: It might be possible to handle this using the slice layer, lt (less than) and div functions, but currently, it is handled with an assert statement.) 

Therefore, I left the `assert()` and `raise RuntimeError()` statement without removing it.

chohk88: Oh! I misunderstood. If the `index` is larger than the size of the dimension specified by `dim`, it won't work in PyTorch either, so we don't need to handle this case. Therefore, we don't need to consider test cases like the 'fail case' mentioned above. This means that a dynamic shape is supported for all dimensions. 

<Example of correct usage of select.int in normal PyTorch>
image
<Example of IndexError occurring in PyTorch when using select.int>
image

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -47,26 +47,16 @@ def select(
...
     index_value = np.array(index, dtype=np.int32)
-    indices_tensor = ctx.net.add_constant(
-        index_value.shape, to_numpy(index_value)
-    ).get_output(0)
+    indices_tensor = ctx.net.add_constant(index_value.shape, index_value).get_output(0)
```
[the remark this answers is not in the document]
chohk88: Thanks for the suggestion!

`dim` and `index` are now `int` (not `Shape`), and I've changed to using `get_trt_tensor` for `indices_tensor`.

py/torch_tensorrt/dynamo/conversion/impl/select.py
chohk88: @keehyuna Thanks for the suggestion!

Since PyTorch already raises an error when the index exceeds the input size, there’s no need for us to check this here. I removed those checks. Additionally, following your suggestion, I’ve added test cases to fully support dynamic shapes.

peri044 (approved):
LGTM

PR #2995, opened 2024-07-10
feat: engine caching

zewenli98:
# Description

Engine caching feature. More details see: https://github.com/pytorch/TensorRT/discussions/2957

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -236,6 +236,9 @@ def compile(
     logger.debug("Lowered Input graph: " + str(gm.graph))
 
     if cache_built_engines or reuse_cached_engines:
+        assert (
+            make_refitable
+        ), "Engine caching requires make_refitable to be set to True"
         if custom_engine_cache is None:
             custom_engine_cache = DiskEngineCache(engine_cache_dir, engine_cache_size)
 
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -18,6 +18,7 @@ from torch_tensorrt.dynamo._DryRunTracker import (
     dryrun_stats_display,
     parse_non_trt_nodes,
 )
+from torch_tensorrt.dynamo._engine_caching import BaseEngineCache, EngineCache
 from torch_tensorrt.dynamo.conversion import (
     CompilationSettings,
     UnsupportedOperatorException,
@@ -82,6 +83,11 @@ def compile(
     hardware_compatible: bool = _defaults.HARDWARE_COMPATIBLE,
     timing_cache_path: str = _defaults.TIMING_CACHE_PATH,
     lazy_engine_init: bool = _defaults.LAZY_ENGINE_INIT,
+    save_engine_cache: bool = _defaults.SAVE_ENGINE_CACHE,
+    load_engine_cache: bool = _defaults.LOAD_ENGINE_CACHE,
+    engine_cache_dir: str = _defaults.ENGINE_CACHE_DIR,
+    engine_cache_size: int = _defaults.ENGINE_CACHE_SIZE,
+    engine_cache_instance: Optional[BaseEngineCache] = None,
     **kwargs: Any,
 ) -> torch.fx.GraphModule:
     """Compile an ExportedProgram module for NVIDIA GPUs using TensorRT
@@ -147,6 +153,11 @@ def compile(
         hardware_compatible (bool): Build the TensorRT engines compatible with GPU architectures other than that of the GPU on which the engine was built (currently works for NVIDIA Ampere and newer)
         timing_cache_path (str): Path to the timing cache if it exists (or) where it will be saved after compilation
         lazy_engine_init (bool): Defer setting up engines until the compilation of all engines is complete. Can allow larger models with multiple graph breaks to compile but can lead to oversubscription of GPU memory at runtime.
+        save_engine_cache (bool): Whether to save the compiled TRT engines to hard disk
+        load_engine_cache (bool): Whether to load the compiled TRT engines from hard disk
+        engine_cache_dir (str): Directory to store the cached TRT engines
+        engine_cache_size (int): Maximum hard-disk space to use for the engine cache
+        engine_cache_instance (Optional[BaseEngineCache]): Engine cache instance to use for saving and loading engines. Users can provide their own engine cache by inheriting from BaseEngineCache
         **kwargs: Any,
     Returns:
         torch.fx.GraphModule: Compiled FX Module, when run it will execute via TensorRT
@@ -224,6 +235,11 @@ def compile(
     gm = post_lowering(gm)
     logger.debug("Lowered Input graph: " + str(gm.graph))
 
+    if engine_cache_instance is None:
+        engine_cache_instance = EngineCacheInstanceCreator.get_creator(
+            engine_cache_size, engine_cache_dir
+        ).engine_cache_instance
+
     compilation_options = {
         "enabled_precisions": (
             enabled_precisions if enabled_precisions else _defaults.ENABLED_PRECISIONS
@@ -257,6 +273,11 @@ def compile(
         "hardware_compatible": hardware_compatible,
         "timing_cache_path": timing_cache_path,
         "lazy_engine_init": lazy_engine_init,
+        "save_engine_cache": save_engine_cache,
+        "load_engine_cache": load_engine_cache,
+        "engine_cache_dir": engine_cache_dir,
+        "engine_cache_size": engine_cache_size,
+        "engine_cache_instance": engine_cache_instance,
     }
 
     settings = CompilationSettings(**compilation_options)
@@ -665,3 +686,21 @@ def convert_exported_program_to_serialized_trt_engine(
 
     serialized_engine: bytes = interpreter_result.serialized_engine
     return serialized_engine
+
+
+class EngineCacheInstanceCreator:
+    engine_cache_creator = None
+
+    def __init__(self, engine_cache_size: int, engine_cache_dir: str) -> None:
+        self.engine_cache_instance = EngineCache(
+            engine_cache_size=engine_cache_size,
+            engine_cache_dir=engine_cache_dir,
+        )
+
+    @classmethod
+    def get_creator(
+        cls, engine_cache_size: int, engine_cache_dir: str
+    ) -> EngineCacheInstanceCreator:
+        if cls.engine_cache_creator is None:
+            cls.engine_cache_creator = cls(engine_cache_size, engine_cache_dir)
+        return cls.engine_cache_creator
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -18,7 +18,7 @@ from torch_tensorrt.dynamo._DryRunTracker import (
     dryrun_stats_display,
     parse_non_trt_nodes,
 )
-from torch_tensorrt.dynamo._engine_caching import BaseEngineCache, EngineCache
+from torch_tensorrt.dynamo._engine_caching import BaseEngineCache, DiskEngineCache
 from torch_tensorrt.dynamo.conversion import (
     CompilationSettings,
     UnsupportedOperatorException,
@@ -83,11 +83,11 @@ def compile(
     hardware_compatible: bool = _defaults.HARDWARE_COMPATIBLE,
     timing_cache_path: str = _defaults.TIMING_CACHE_PATH,
     lazy_engine_init: bool = _defaults.LAZY_ENGINE_INIT,
-    save_engine_cache: bool = _defaults.SAVE_ENGINE_CACHE,
-    load_engine_cache: bool = _defaults.LOAD_ENGINE_CACHE,
+    cache_built_engines: bool = _defaults.CACHE_BUILT_ENGINES,
+    reuse_cached_engines: bool = _defaults.REUSE_CACHED_ENGINES,
     engine_cache_dir: str = _defaults.ENGINE_CACHE_DIR,
     engine_cache_size: int = _defaults.ENGINE_CACHE_SIZE,
-    engine_cache_instance: Optional[BaseEngineCache] = None,
+    custom_engine_cache: Optional[BaseEngineCache] = _defaults.CUSTOM_ENGINE_CACHE,
     **kwargs: Any,
 ) -> torch.fx.GraphModule:
     """Compile an ExportedProgram module for NVIDIA GPUs using TensorRT
@@ -153,11 +153,11 @@ def compile(
         hardware_compatible (bool): Build the TensorRT engines compatible with GPU architectures other than that of the GPU on which the engine was built (currently works for NVIDIA Ampere and newer)
         timing_cache_path (str): Path to the timing cache if it exists (or) where it will be saved after compilation
         lazy_engine_init (bool): Defer setting up engines until the compilation of all engines is complete. Can allow larger models with multiple graph breaks to compile but can lead to oversubscription of GPU memory at runtime.
-        save_engine_cache (bool): Whether to save the compiled TRT engines to hard disk
-        load_engine_cache (bool): Whether to load the compiled TRT engines from hard disk
+        cache_built_engines (bool): Whether to save the compiled TRT engines to storage
+        reuse_cached_engines (bool): Whether to load the compiled TRT engines from storage
         engine_cache_dir (str): Directory to store the cached TRT engines
         engine_cache_size (int): Maximum hard-disk space to use for the engine cache
-        engine_cache_instance (Optional[BaseEngineCache]): Engine cache instance to use for saving and loading engines. Users can provide their own engine cache by inheriting from BaseEngineCache
+        custom_engine_cache (Optional[BaseEngineCache]): Engine cache instance to use for saving and loading engines. Users can provide their own engine cache by inheriting from BaseEngineCache. If used, engine_cache_dir and engine_cache_size will be ignored.
         **kwargs: Any,
     Returns:
         torch.fx.GraphModule: Compiled FX Module, when run it will execute via TensorRT
@@ -235,10 +235,9 @@ def compile(
     gm = post_lowering(gm)
     logger.debug("Lowered Input graph: " + str(gm.graph))
 
-    if engine_cache_instance is None:
-        engine_cache_instance = EngineCacheInstanceCreator.get_creator(
-            engine_cache_size, engine_cache_dir
-        ).engine_cache_instance
+    if cache_built_engines or reuse_cached_engines:
+        if custom_engine_cache is None:
+            custom_engine_cache = DiskEngineCache(engine_cache_dir, engine_cache_size)
 
     compilation_options = {
         "enabled_precisions": (
@@ -273,11 +272,9 @@ def compile(
         "hardware_compatible": hardware_compatible,
         "timing_cache_path": timing_cache_path,
         "lazy_engine_init": lazy_engine_init,
-        "save_engine_cache": save_engine_cache,
-        "load_engine_cache": load_engine_cache,
-        "engine_cache_dir": engine_cache_dir,
-        "engine_cache_size": engine_cache_size,
-        "engine_cache_instance": engine_cache_instance,
+        "cache_built_engines": cache_built_engines,
+        "reuse_cached_engines": reuse_cached_engines,
+        "custom_engine_cache": custom_engine_cache,
     }
 
     settings = CompilationSettings(**compilation_options)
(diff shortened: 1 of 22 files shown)


py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -83,6 +84,11 @@ def compile(
     hardware_compatible: bool = _defaults.HARDWARE_COMPATIBLE,
     timing_cache_path: str = _defaults.TIMING_CACHE_PATH,
     lazy_engine_init: bool = _defaults.LAZY_ENGINE_INIT,
+    save_engine_cache: bool = _defaults.SAVE_ENGINE_CACHE,
```
narendasan, line 87: Lets invert this like: `disable_caching_engines` since it should default to engines being cached

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -83,6 +84,11 @@ def compile(
...
+    load_engine_cache: bool = _defaults.LOAD_ENGINE_CACHE,
+    engine_cache_dir: str = _defaults.ENGINE_CACHE_DIR,
```
narendasan, line 89: Optional in the case of a user cache instance?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -83,6 +84,11 @@ def compile(
...
+    engine_cache_size: int = _defaults.ENGINE_CACHE_SIZE,
```
narendasan, line 90: Same here?

py/torch_tensorrt/dynamo/_compiler.py
narendasan: Maybe we should specialize these settings like `disk_engine_cache_dir` or we should document that if you give engine_cache_instance, then these do nothing

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -83,6 +84,11 @@ def compile(
...
+    engine_cache_instance: Optional[BaseEngineCache] = None,
```
narendasan, line 91: Lets call this `custom_engine_cache`

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -703,3 +724,21 @@ def convert_exported_program_to_serialized_trt_engine(
 
     serialized_engine: bytes = interpreter_result.serialized_engine
     return serialized_engine
+
+
+class EngineCacheInstanceCreator:
```
narendasan, line 729: Why do we need this indirection?

py/torch_tensorrt/dynamo/_defaults.py
```diff
@@ -31,8 +32,17 @@
...
-TIMING_CACHE_PATH = os.path.join(tempfile.gettempdir(), "timing_cache.bin")
+TIMING_CACHE_PATH = os.path.join(
+    tempfile.gettempdir(), "torch_tensorrt_engine_cache", "timing_cache.bin"
+)
 LAZY_ENGINE_INIT = False
+SAVE_ENGINE_CACHE = True
+LOAD_ENGINE_CACHE = True
+ENGINE_CACHE_DIR = os.path.join(tempfile.gettempdir(), "torch_tensorrt_engine_cache")
+ENGINE_CACHE_SIZE = 1073741824
+ENGINE_CACHE_INSTANCE = EngineCache(
```
narendasan, line 43: I think we can leave this as `None` as default and just create the cache object in compile

py/torch_tensorrt/dynamo/_engine_caching.py
```diff
@@ -0,0 +1,268 @@
...
+        Returns:
+            bool: whether the serialized engine is saved successfully
+        """
+        pass
+
+    @abstractmethod
+    def load(
+        self, hash: str
+    ) -> Tuple[Optional[bytes], List[str], List[str], Optional[Dict[str, Any]]]:
+        """Load the serialized engine from hard disk
```
narendasan, line 74: Load the serialized engine from ~~hard disk~~ storage

py/torch_tensorrt/dynamo/_engine_caching.py
```diff
@@ -0,0 +1,268 @@
...
+    @abstractmethod
+    def save(
+        self,
+        hash: str,
+        serialized_engine: bytes,
+        input_names: List[str],
+        output_names: List[str],
+        weight_name_map: Optional[Dict[str, Any]] = None,
+    ) -> bool:
+        """Save the serialized engine to hard disk
```
narendasan, line 56: ~~Save the serialized engine to hard disk~~ Store serialized engine in cache

py/torch_tensorrt/dynamo/_engine_caching.py
```diff
@@ -0,0 +1,268 @@
...
+        Args:
+            hash (str): hash value of the GraphModule
+
+        Returns:
+            Sequence[Optional[bytes], List[str], List[str], Optional[Dict[str, Any]]]: serialized engine, input names, output names, weight name map
+        """
+        pass
+
+
+class EngineCache(BaseEngineCache):
```
narendasan, line 85: Call this `DiskEngineCache`

py/torch_tensorrt/dynamo/_engine_caching.py
```diff
@@ -0,0 +1,268 @@
...
+    def load(
+        self, hash: str
+    ) -> Tuple[Optional[bytes], List[str], List[str], Optional[Dict[str, Any]]]:
+        """Load the serialized engine from hard disk
+
+        Args:
+            hash (str): hash value of the GraphModule
+
+        Returns:
+            Sequence[Optional[bytes], List[str], List[str], Optional[Dict[str, Any]]]: serialized engine, input names, output names, weight name map
```
narendasan, line 80: You need to define exactly what a cache hit looks like and how a cache would signal that there isn't a hit.

py/torch_tensorrt/dynamo/_engine_caching.py
```diff
@@ -0,0 +1,268 @@
...
+
+            hash_val = cast(str, FxGraphCachePickler.get_hash(new_gm))
+
+        return hash_val
+
+    @abstractmethod
+    def save(
+        self,
+        hash: str,
+        serialized_engine: bytes,
```
narendasan, line 51: Do you think we should give people a single blob to save with all the info or have them pack it themselves? This blob could be a pkl or something that is easy to store in a bin file or db

py/torch_tensorrt/dynamo/_engine_caching.py
narendasan: If we just give the cache a blob then its easy its just this function returns `bytes | None`

py/torch_tensorrt/dynamo/_engine_caching.py
narendasan: We can define the packing as a static method of the base class where `(serialized_engine, input_names, output_names, weight_map) -> bytes` then this function can be `def save(self, hash, blob) -> bool`

py/torch_tensorrt/dynamo/_engine_caching.py
```diff
@@ -0,0 +1,268 @@
...
+                        f"The weight_name_map was saved to {weight_name_map_path}"
+                    )
+                except Exception as e:
+                    del self.hash2size_map[hash]
+                    self.available_engine_cache_size += serialized_engine_size
+                    shutil.rmtree(directory)
+                    _LOGGER.warning(
+                        f"Failed to save the weight_name_map to {weight_name_map_path}: {e}"
+                    )
+                    return False
```
narendasan, line 229: What happens if False is returned?

narendasan (commented):
Do you have tests for this?

py/torch_tensorrt/dynamo/_defaults.py
zewenli98: This is because if users run `compile` multiple times, the EngineCache class will be instantiated multiple times. e.g.
```
for i in range(3):
    torch.compile(...)
```

Please let me know if I'm wrong or there's better way to bypass the issue

narendasan: But a cache object is just an interface to a system or process outside of compile right? like we can have as many EngineCache objects as we want since they all ultimately hit the disk

py/torch_tensorrt/dynamo/_engine_caching.py
zewenli98: Currently there's no effect return either True or False. Is it necessary to differentiate them?

py/torch_tensorrt/dynamo/_defaults.py
zewenli98: yes but if that's created too many times. Do the instances use too much memory?

zewenli98: Do you want to create multiple EngineCache objects or just single one?

py/torch_tensorrt/dynamo/_engine_caching.py
```diff
@@ -0,0 +1,268 @@
...
+
+class EngineCache(BaseEngineCache):
+
+    def __init__(
+        self,
+        engine_cache_size: int,
+        engine_cache_dir: str,
+    ) -> None:
+        self.total_engine_cache_size = engine_cache_size
+        self.available_engine_cache_size = engine_cache_size
```
narendasan, line 93: This should be the current total_cache_size - directories size

py/torch_tensorrt/dynamo/_engine_caching.py
```diff
@@ -0,0 +1,268 @@
...
+        self.engine_cache_dir = engine_cache_dir
+        self.hash2size_map: Dict[str, int] = {}
```
narendasan, line 95: Create directory if doesnt exist?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -148,6 +154,11 @@ def compile(
         hardware_compatible (bool): Build the TensorRT engines compatible with GPU architectures other than that of the GPU on which the engine was built (currently works for NVIDIA Ampere and newer)
         timing_cache_path (str): Path to the timing cache if it exists (or) where it will be saved after compilation
         lazy_engine_init (bool): Defer setting up engines until the compilation of all engines is complete. Can allow larger models with multiple graph breaks to compile but can lead to oversubscription of GPU memory at runtime.
+        cache_built_engines (bool): Whether to save the compiled TRT engines to storage
```
peri044, line 157: Minor: Consider renaming this to `save_engine_cache`

py/torch_tensorrt/dynamo/_engine_caching.py
```diff
@@ -0,0 +1,251 @@
...
+        """Get the hash value of the GraphModule
+
+        Args:
+            gm (torch.fx.GraphModule): GraphModule to hash
+
+        Returns:
+            str: hash value of the GraphModule
+        """
+        # parameters are set to 0
+        with maybe_disable_fake_tensor_mode():
```
peri044, line 37: This changed to `unset_fake_temporarlly`

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -468,15 +469,48 @@ def run(
...
+                blob = engine_cache.load(hash_val)
+                if blob is not None:  # hit the cache
+                    serialized_engine, input_names, output_names, weight_name_map = (
+                        engine_cache.unpack(blob)
+                    )
+                    self._input_names = input_names
+                    self._output_names = output_names
+                    self.weight_name_map = weight_name_map
+                    _LOGGER.info(
+                        "Hit the cached TRT engine. It is loaded and skip recompilation."
```
peri044, line 497: Consider changing this to - Found engines that correspond to this graph, cached at {engine_dir_path}. Please clear the cache incase you want to recompile.

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -257,6 +272,9 @@ def compile(
         "hardware_compatible": hardware_compatible,
         "timing_cache_path": timing_cache_path,
         "lazy_engine_init": lazy_engine_init,
+        "cache_built_engines": cache_built_engines,
+        "reuse_cached_engines": reuse_cached_engines,
+        "custom_engine_cache": custom_engine_cache,
```
peri044, line 277: is engine_cache_dir not being here intentional ?

peri044 (commented):
Added some comments. Functionality looks good. One question - Does the hash change if the module has the same graph but different weights ?

zewenli98:
> Added some comments. Functionality looks good. One question - Does the hash change if the module has the same graph but different weights ?

Nope, all weights are set to 0 before hashing, so if the architectures are the same they will be considered isomorphic

py/torch_tensorrt/dynamo/_compiler.py
zewenli98: It was `save_engine_cache`. after discussion with Naren, we deem `cache_built_engines` might be better.

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
zewenli98: Thanks for the suggestion. Since users are able to customize EngineCache class that might not have `engine_cache_dir` attribute, for example, engines are cached on a database instead of hard disk, I keep the front part of the sentence.

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -473,15 +474,47 @@ def run(
...
+                    serialized_engine, input_names, output_names, weight_name_map = (
+                        engine_cache.unpack(blob)
+                    )
+                    self._input_names = input_names
+                    self._output_names = output_names
+                    self.weight_name_map = weight_name_map
+                    _LOGGER.info(
+                        "Found the cached engine that corresponds to this graph. It is directly loaded."
+                    )
+                    return TRTInterpreterResult(
```
narendasan, line 504: Where does refit happen?

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -107,6 +107,36 @@ def interpret_module_to_result(
         compilation_settings=settings,
     )
     interpreter_result = interpreter.run()
+
+    if settings.make_refitable:
```
narendasan, line 111: Wouldnt engine caching require refit to be on?

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -107,6 +107,36 @@ def interpret_module_to_result(
...
+        )
+        try:
+            _refit_single_trt_engine_with_gm(
+                new_gm=module,
+                old_engine=refit_test_engine,
+                input_list=inputs,
+                settings=settings,
+                weight_name_map=interpreter_result.weight_name_map,
+            )
+        except AssertionError:
```
narendasan, line 130: Why do we expect an assertion failure here?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
narendasan: Are we always going to make weight stripped engines now or something?

narendasan:
> > Added some comments. Functionality looks good. One question - Does the hash change if the module has the same graph but different weights ?
> 
> Nope, all weights are set to 0 before hashing, so if the architectures are the same they will be considered isomorphic

I dont really get where we refit for functionality in the code now if we store weight stripped (or zero'd) graphs. Also if we set all weights to 0 what happens in data-dependent cases? Can we detect these cases?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
zewenli98: Adrian said the **fast refit** would be called even without calling engine caching, so that's why it's after `interpreter.run()`: https://github.com/pytorch/TensorRT/pull/2995/files#diff-5d0cff736de42c72a7c404440a15e371fac0984794a5de7e651cc2a32566856dR111-R129

I just realized that it's incorrect to put refitting code after `interpreter.run()` because `interpreter.run()` could be called in other places. In these scenarios, the engine weights will be 0 if use engine caching. I plan to move it into `interpreter.run()`.

py/torch_tensorrt/dynamo/conversion/_conversion.py
zewenli98: > Wouldnt engine caching require refit to be on?

yep. it requires make_refitable == True.

py/torch_tensorrt/dynamo/conversion/_conversion.py
zewenli98: Based on what Adrian said, here's just a fast refit. 
`_refit_single_trt_engine_with_gm` will run very fast to check if `interpreter_result.weight_name_map` is valid. If the map doesn't match, `_refit_single_trt_engine_with_gm` will throw out an exception, which means fast refit failed and `weight_name_map` will be cleared. The future refit would use the slow version.

zewenli98:
> I dont really get where we refit for functionality in the code now if we store weight stripped (or zero'd) graphs. 

The current code calls refit after `interpreter.run()` but I plan to move it into `interpreter.run()`. 

> Also if we set all weights to 0 what happens in data-dependent cases? Can we detect these cases?

The current engine caching happens after partition.
If an input is data dependent, does it generate different TRT engines given different actual inputs?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -473,15 +474,65 @@ def run(
...
+                    )
+
+                    serialized_engine = bytes(engine.serialize())
+
+                    return TRTInterpreterResult(
+                        serialized_engine,
+                        self._input_names,
+                        self._output_names,
+                        self.weight_name_map,
+                    )
```
zewenli98, line 527: Here should I directly use `serialized_engine` to build `TRTInterpreterResult` or use `engine_str`:
```
with io.BytesIO() as engine_bytes:
return TRTInterpreterResult(
    engine_str, self._input_names, self._output_names, self.weight_name_map
)
```
like line 570 below

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -148,6 +154,11 @@ def compile(
         hardware_compatible (bool): Build the TensorRT engines compatible with GPU architectures other than that of the GPU on which the engine was built (currently works for NVIDIA Ampere and newer)
         timing_cache_path (str): Path to the timing cache if it exists (or) where it will be saved after compilation
         lazy_engine_init (bool): Defer setting up engines until the compilation of all engines is complete. Can allow larger models with multiple graph breaks to compile but can lead to oversubscription of GPU memory at runtime.
+        cache_built_engines (bool): Whether to save the compiled TRT engines to storage
+        reuse_cached_engines (bool): Whether to load the compiled TRT engines from storage
+        engine_cache_dir (str): Directory to store the cached TRT engines
+        engine_cache_size (int): Maximum hard-disk space to use for the engine cache
```
narendasan, line 160: Add the units

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -148,6 +154,11 @@ def compile(
         hardware_compatible (bool): Build the TensorRT engines compatible with GPU architectures other than that of the GPU on which the engine was built (currently works for NVIDIA Ampere and newer)
         timing_cache_path (str): Path to the timing cache if it exists (or) where it will be saved after compilation
         lazy_engine_init (bool): Defer setting up engines until the compilation of all engines is complete. Can allow larger models with multiple graph breaks to compile but can lead to oversubscription of GPU memory at runtime.
+        cache_built_engines (bool): Whether to save the compiled TRT engines to storage
+        reuse_cached_engines (bool): Whether to load the compiled TRT engines from storage
+        engine_cache_dir (str): Directory to store the cached TRT engines
```
narendasan, line 159: Optional?

py/torch_tensorrt/dynamo/_compiler.py
narendasan, line 160: Optional

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -83,6 +84,11 @@ def compile(
     hardware_compatible: bool = _defaults.HARDWARE_COMPATIBLE,
     timing_cache_path: str = _defaults.TIMING_CACHE_PATH,
     lazy_engine_init: bool = _defaults.LAZY_ENGINE_INIT,
+    cache_built_engines: bool = _defaults.CACHE_BUILT_ENGINES,
+    reuse_cached_engines: bool = _defaults.REUSE_CACHED_ENGINES,
+    engine_cache_dir: str = _defaults.ENGINE_CACHE_DIR,
```
narendasan, line 89: Optional?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -83,6 +84,11 @@ def compile(
...
+    engine_cache_size: int = _defaults.ENGINE_CACHE_SIZE,
```
narendasan, line 90: Optional?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -83,6 +84,11 @@ def compile(
...
+    custom_engine_cache: Optional[BaseEngineCache] = _defaults.CUSTOM_ENGINE_CACHE,
```
narendasan, line 91: custom engine cache should not be in the Compilation settings, should be a standalone arg

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -547,6 +608,17 @@ def run(
         self._save_timing_cache(
             builder_config, self.compilation_settings.timing_cache_path
         )
+        if (
```
narendasan, line 611: Do we need to re-save everytime if theres a cache hit?

narendasan (approved):
LGTM!

PR #3008, opened 2024-07-15
scatter reduce decomposition

apbose:
#2740 should be using this. Will change it once this PR is finalized

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -333,6 +333,7 @@ def scatter_reduce_decomposition(
     index: torch.Tensor,
     src_tensor: torch.Tensor,
     reduce: str,
+    include_self: bool = True,
 ) -> torch.Tensor:
     scatter_loop_tensor = input_tensor
     device_input_tensor = input_tensor.device
@@ -340,7 +341,8 @@ def scatter_reduce_decomposition(
     scatter_count_tensor = torch.zeros_like(input_tensor)
     src_shape = list(src_tensor.shape)
     src_dim = src_shape[dim]
-
+    if include_self == False:
+        raise AssertionError("include_self False for scatter reduce not yet supported")
     for i in range(0, src_dim):
         src_slice = torch.select(src_tensor, dim, i)
         index_slice = torch.select(index, dim, i)
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -318,7 +318,7 @@ class ReduceOperation(Enum):
             print("Invalid Operation for Reduce op!!")
 
         operation_rhs = torch.scatter(scatter_tensor, dim, index_tensor, src_tensor)
-        device = to_torch_device(default_device())
+        device = to_torch_device(scatter_tensor.device)
         operation_lhs = operation_lhs.to(device)
         operation_rhs = operation_rhs.to(device)
         return self.func(operation_lhs, operation_rhs)
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -1,4 +1,5 @@
 import logging
+from enum import Enum, auto
 from typing import Any, Callable, Dict, List, Optional
 
 import torch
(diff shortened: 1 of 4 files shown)


tests/py/dynamo/lowering/test_decompositions.py
```diff
@@ -962,6 +962,414 @@ def forward(self, input):
...
+        input = torch.zeros(3, 5, dtype=torch.int32).cuda()
+        inputs = [input_reduce_op]
+
+        fx_graph = torch.fx.symbolic_trace(TestModule())
+        unexpected_ops_seen, expected_ops_unseen = lower_graph_testing(
+            fx_graph,
+            inputs,
+            expected_ops=expected_ops,
+            unexpected_ops=unexpected_ops,
+            min_block_size=2,
```
peri044, line 1334: Why is the min_block size = 2 here ? What  does lower_graph_testing do ? Does it use our partitioning ?

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -243,6 +244,99 @@ def empty_strided_decomposition(*args, **kwargs) -> torch.Tensor:
     )
 
 
+# enum class for reduce operation of scatter_reduce
+class reduceOperation(Enum):
```
peri044, line 248: minor - consider renaming it to `ReduceOperation`

tests/py/dynamo/lowering/test_decompositions.py
apbose: Yes lower_graph_testing uses our partition. It is used for returning the expected ops and the unexpected ops. The default I think is 3, but sometimes the graph in the test case is too small, and the block size is lesser than 3. Thats why I set it to 1 or 2, since that is not something which we are testing explicitly in the test.
Let me try with the default 3. If it passes, I will remove the `min_block_size =2` then.

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -281,6 +282,99 @@ def scatter_add_decomposition(
...
+        elif self == ReduceOperation.PROD:
+            scatter_tensor = torch.ones_like(initial_tensor)
+        elif self == ReduceOperation.AMIN or self == ReduceOperation.AMAX:
+            scatter_tensor = initial_tensor
+        else:
+            # This case would not be encountered from torch itself
+            print("Invalid Operation for Reduce op!!")
+
+        operation_rhs = torch.scatter(scatter_tensor, dim, index_tensor, src_tensor)
+        device = to_torch_device(default_device())
```
peri044, line 315: let's use the device where the initial_tensor exists.

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -281,6 +282,99 @@ def scatter_add_decomposition(
...
+    src_shape = list(src_tensor.shape)
+    src_dim = src_shape[dim]
+
+    for i in range(0, src_dim):
+        src_slice = torch.select(src_tensor, dim, i)
+        index_slice = torch.select(index, dim, i)
+        # unsqueeze src and index in dim
+        src_slice = torch.unsqueeze(src_slice, dim)
+        index_slice = torch.unsqueeze(index_slice, dim)
+        device = to_torch_device(default_device())
```
peri044, line 343: let's use the device where the input_tensor exists

peri044:
@apbose  CI is failing on scatter tests

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -285,6 +286,99 @@ def scatter_add_decomposition(
...
+        elif self == ReduceOperation.PROD:
+            scatter_tensor = torch.ones_like(initial_tensor)
+        elif self == ReduceOperation.AMIN or self == ReduceOperation.AMAX:
+            scatter_tensor = initial_tensor
+        else:
+            # This case would not be encountered from torch itself
+            print("Invalid Operation for Reduce op!!")
+
+        operation_rhs = torch.scatter(scatter_tensor, dim, index_tensor, src_tensor)
+        device = to_torch_device(default_device())
```
peri044, line 319: use the device of initial_tensor here instead of default

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -287,6 +288,99 @@ def scatter_add_decomposition(
...
+@register_torch_trt_decomposition(
+    torch.ops.aten.scatter_reduce.two, registry=TORCH_TRT_DECOMPOSITIONS
+)
+def scatter_reduce_decomposition(
+    input_tensor: torch.Tensor,
+    dim: int,
+    index: torch.Tensor,
+    src_tensor: torch.Tensor,
+    reduce: str,
+) -> torch.Tensor:
```
HolyWu, line 336: There is a kwarg `include_self` in https://github.com/pytorch/pytorch/blob/bc1b8f094d24de27432f4c29f0729e85a6b5ba63/aten/src/ATen/native/native_functions.yaml#L8237. Is it intentionally not handled in our decomposition?

apbose: Most of the cases which I have seen is with `include_self = True`. No particular reason, I could add cases with `include_self = False`

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -186,11 +187,14 @@ def get_model_device(module: torch.fx.GraphModule) -> torch.device:
     device = None
     for parameter in list(module.parameters()):
         if isinstance(parameter, (torch.nn.parameter.Parameter, torch.Tensor)):
-            device = parameter.device
-            break
+            return parameter.device
+
+    for buffer in list(module.buffers()):
```
peri044, line 192: The buffer device overrides the parameter device here which shouldn't be the case. Check device of parameters first, if not found, use buffers.
Also consider adding break once the device is found.

py/torch_tensorrt/dynamo/lowering/_decompositions.py
peri044: Add `include_self=True` in the function arguments. And raise an error saying we don't support the case when user sets it `False`

py/torch_tensorrt/dynamo/utils.py
peri044: nvm

PR #3014, opened 2024-07-17
Group norm bug fix

cehongwang:
# Description

Fixes # (issue)

## Type of change

# Checklist:

- [ x] My code follows the style guidelines of this project (You can use the linters)
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas and hacks
- [ x] I have added tests to verify my fix or my feature
- [ x] New and existing unit tests pass locally with my changes
- [ x] I have added the relevant labels to my PR in so that relevant reviewers are notified

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
@@ -189,6 +189,12 @@ def native_group_norm(
         shape,
     )
 
+    if weight is None:
+        weight = to_numpy(1.0)
+
+    if bias is None:
+        bias = to_numpy(0.0)
+
     weight = get_trt_tensor(ctx, weight, f"{name}_weight")
     bias = get_trt_tensor(ctx, bias, f"{name}_bias")
     weight_bias_shape = (1, C) + (1,) * (len(input.shape) - 2)
--- a/py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
@@ -134,6 +134,100 @@ def layer_norm(
     return layer_norm.get_output(0)
 
 
+# def native_group_norm(
+#     ctx: ConversionContext,
+#     target: Target,
+#     source_ir: Optional[SourceIR],
+#     name: str,
+#     input: TRTTensor,
+#     weight: Optional[Union[torch.Tensor, np.ndarray]],
+#     bias: Optional[Union[torch.Tensor, np.ndarray]],
+#     N: int,
+#     C: int,
+#     HxW: int,
+#     group: int,
+#     eps: float,
+#     return_mean_rstd: bool = True,
+# ) -> Union[TRTTensor, Sequence[TRTTensor]]:
+#     assert (
+#         len(input.shape) >= 3
+#     ), f"The input dimension should not be less than 3, got {len(input.shape)}!"
+
+#     B = input.shape[0]
+#     # if C is provided, it must be as same as the channel from the input shape,
+#     # else if C is zero, we should get the channel from the input shape
+#     if C == 0:
+#         C = input.shape[1]
+#     assert (
+#         C == input.shape[1]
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -222,6 +295,40 @@ def native_group_norm(
...
+        weight_bias_shape,
+    )
+    reshaped_output = impl.elementwise.add(
+        ctx,
+        target,
+        source_ir,
+        f"{name}_add_beta",
+        reshaped_output,
+        reshaped_bias,
+    )
```
apbose, line 331: How is the TensorRT INormalization layer different from this? I am assuming that the implementation has been changed since the INormalization layer would not work for Group norm case. Where was it going wrong?

cehongwang: I don't know how TRT implementation is different, but in reality INormalization layer does not produce correct results under FP16 and stable diffusion fails on that. The break-down implementation produces correct results

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -189,26 +191,97 @@ def native_group_norm(
...
+    var_trt = impl.slice.expand(
+        ctx,
+        target,
+        source_ir,
+        f"{name}_expand_var_trt",
+        var_trt,
+        reshaped_input.shape,
+    )
+
+    eps_trt = get_trt_tensor(ctx, eps, f"{name}_eps", np.float32)
```
apbose, line 258: Why is this converted to np.float32?

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -189,26 +191,97 @@ def native_group_norm(
...
+
+    # y = (X - E[X]) / sqrt((var + eps))
+    output = impl.elementwise.div(
+        ctx,
+        target,
+        source_ir,
+        f"{name}_div",
+        sub_trt,
+        sqrt_trt,
+    )
```
apbose, line 285: Wanted to clarify if this div would require any mode, eg: trunc? Are the data types always compatible with the output types?

cehongwang: I don't think so. The previous implementation from Evan did not include any mode.

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
cehongwang: It is 1e-6 in most cases.

cehongwang: float16 cannot handle that close to 0

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
zewenli98: just FYI: we have `impl.elementwise.floor_divide` and `impl.elementwise.trunc_div` helper functions, if needed.

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -189,26 +191,97 @@ def native_group_norm(
 
     weight = get_trt_tensor(ctx, weight, f"{name}_weight")
     bias = get_trt_tensor(ctx, bias, f"{name}_bias")
```
zewenli98, line 193: Per [the schema](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L2965), does this work for `weight==None` or `bias==None`?

cehongwang: Nice catch. Added that case

zewenli98 (approved):
LGTM

PR #3026, opened 2024-07-21
chore: dynamic shape support for any/sort/trunc ops

keehyuna:
# Description

dynamic shape support for any/sort/trunc ops

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -19,7 +19,6 @@ from torch_tensorrt.dynamo.conversion.converter_utils import (
     get_positive_dim,
     is_only_operator_on_placeholder,
 )
-from torch_tensorrt.dynamo.utils import TRT_TOPK_MAX_ELEMENT
 from torch_tensorrt.fx.types import TRTTensor
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
@@ -2649,21 +2648,22 @@ def topk_validator(node: Node) -> bool:
 
 
 def sort_validator(node: Node) -> bool:
-    # if meta data is not available(e.g. dynamic shape), validate k value during runtime.
-    if not node.args[0].meta:
-        return True
-
-    shape = node.args[0].meta.get("tensor_meta").shape
+    meta_data = node.args[0].meta.get("tensor_meta")
+    if meta_data is None:
+        return False
+    shape = meta_data.shape
     dim = node.args[1]
     dim = get_positive_dim(dim, len(shape))
     k = shape[dim]
+    if not isinstance(k, int):
+        return False
     return topk_sort_validator(k)
 
 
 def topk_sort_validator(k: int) -> bool:
-    if k > TRT_TOPK_MAX_ELEMENT:
+    if k > 3840:
         _LOGGER.debug(
-            f"Currently only topk values up to {TRT_TOPK_MAX_ELEMENT} are supported, got k={k}."
+            f"Currently only topk values up to 3840 are supported, got k={k}."
         )
         return False
     return True
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -19,6 +19,7 @@ from torch_tensorrt.dynamo.conversion.converter_utils import (
     get_positive_dim,
     is_only_operator_on_placeholder,
 )
+from torch_tensorrt.dynamo.utils import TRT_TOPK_MAX_ELEMENT
 from torch_tensorrt.fx.types import TRTTensor
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


py/torch_tensorrt/dynamo/conversion/impl/topk.py
```diff
@@ -155,6 +158,37 @@ def topk(
...
+
+        cond = le(
+            ctx,
+            target,
+            source_ir,
+            f"{name}_k_cond",
+            shuffle_layer.get_output(0),
+            TRT_TOPK_MAX_ELEMENT,
+        )
+        ctx.net.add_assertion(
```
keehyuna, line 186: when dynamic k > 3840, there is segmentation fault rather than runtime error. Error is like below

Python error: Segmentation fault

Thread 0x00007fda56bfe640 (most recent call first):
  File "/root/.pyenv/versions/3.10.14/lib/python3.10/threading.py", line 324 in wait
  File "/root/.pyenv/versions/3.10.14/lib/python3.10/threading.py", line 607 in wait
  File "/root/.pyenv/versions/3.10.14/lib/python3.10/site-packages/tqdm/_monitor.py", line 60 in run
  File "/root/.pyenv/versions/3.10.14/lib/python3.10/threading.py", line 1016 in _bootstrap_inner
  File "/root/.pyenv/versions/3.10.14/lib/python3.10/threading.py", line 973 in _bootstrap

Current thread 0x00007fdc02f65740 (most recent call first):
  File "/root/trt/TensorRT/py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py", line 287 in forward
...
Extension modules: numpy.core._multiarray_umath, ...
Segmentation fault

I tried catch it from add_assertion() but it doesn't work while simple test with add_assertion() was ok.
Any idea to use it? Or we can with this behavior to enable dynamic k value for sort?
There is no problem if dynamic k value is in range as in test code

zewenli98: I think the issue is due to [TRT's limitation](https://docs.nvidia.com/deeplearning/tensorrt/archives/tensorrt-1000-ea/api/python_api/infer/Graph/Layers.html#tensorrt.ITopKLayer). One possible solution is to fall back to pytorch if k > 3840. We don't expect to throw any errors due to TRT

py/torch_tensorrt/dynamo/conversion/impl/topk.py
```diff
@@ -155,6 +158,37 @@ def topk(
...
+            cond,
+            message=f"Currently only topk values up to {TRT_TOPK_MAX_ELEMENT} are supported",
+        )
+
+        topk_layer.set_input(1, shuffle_layer.get_output(0))
     # TensorRT ITopKLayer does not have a sorted flag, it is always returning the sorted topk elements
     # so here no matter sorted is True or False the returned the topk Tensor object is always sorted
     set_layer_name(topk_layer, target, name, source_ir)
```
zewenli98, line 194: same here

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2648,10 +2648,15 @@ def topk_validator(node: Node) -> bool:
...
 def sort_validator(node: Node) -> bool:
-    shape = node.args[0].meta.get("tensor_meta").shape
+    meta_data = node.args[0].meta.get("tensor_meta")
+    if meta_data is None:
+        return False
+    shape = meta_data.shape
     dim = node.args[1]
     dim = get_positive_dim(dim, len(shape))
     k = shape[dim]
+    if not isinstance(k, int):
```
keehyuna, line 2658: If dim k is static and other dims are dynamic, we can support sort ops.
This is to validate if k is static or not.

tests/py/dynamo/conversion/test_sort_aten.py
```diff
@@ -32,5 +33,55 @@ def forward(self, x):
...
+                opt_shape=opt_shape,
+                max_shape=max_shape,
+                dtype=torch.float,
+            ),
+        ]
+        self.run_test_with_dynamic_shape(
+            Sort(),
+            input_specs,
+            output_dtypes=[torch.float, torch.int64],
+            use_dynamo_tracer=True,
```
keehyuna, line 82: tensor meta data is available when use_dynamo_tracer=True

py/torch_tensorrt/dynamo/conversion/impl/topk.py
keehyuna: Thanks for review. As we discuss, I changed to allow this ops when k value is static because we cannot identify dynamic k value at compile time.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: Thanks for the comment. Is it possible that k is -1?

keehyuna: Thanks for bringing this to my attention.
I checked only dynamic dims are passed to export() in torch_tensorrt.dynamo.trace()
https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/_tracer.py#L81
So I think static dim value in meta_data will be valid dim value(>0). If you think static shape dim value from meta data could be -1, I will update to validate this case as well.

+elif isinstance(k, int) and k < 0:
+    return False

peri044 (approved):
LGTM

PR #3039, opened 2024-07-25
Dynamic shape index

apbose:


peri044:
@apbose  is this ready ?

apbose:
@peri044 Right now the dynamic input cases are supported, I will add some test cases for dynamic index. Although the implementation should support it. Also, is there any example script to test it with mistral-7b and SD? I did not test it with the models yet

peri044:
> @peri044 Right now the dynamic input cases are supported, I will add some test cases for dynamic index. Although the implementation should support it. Also, is there any example script to test it with mistral-7b and SD? I did not test it end to end with the models yet

For mistral model, you can use the same example https://github.com/pytorch/TensorRT/blob/llm_examples_main/examples/dynamo/torch_export_llama2.py and swap the model name to `mistralai/Mistral-7B-v0.3`

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -169,12 +172,51 @@ def index(
...
+                        ctx,
+                        target,
+                        source_ir,
+                        name + f"_transpose_tensor_shape_mult_d0_{i}",
+                        transpose_tensor,
+                        i,
+                    )
+                    if transpose_tensor_shape[i] == DYNAMIC_DIM
+                    else transpose_tensor_shape[i]
+                ),
```
peri044, line 196: consider storing this in a variable for better readability

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -169,12 +172,51 @@ def index(
...
+                        ctx,
+                        target,
+                        source_ir,
+                        name + f"_transpose_tensor_shape_mult_d1_{i}",
+                        transpose_tensor,
+                        i,
+                    )
+                    if transpose_tensor_shape[i] == DYNAMIC_DIM
+                    else transpose_tensor_shape[i]
+                ),
```
peri044, line 218: same as above

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -204,11 +246,7 @@ def index(
                 ctx, cum_adv_index, name + "_index_sum_intermediate"
             )
         else:
-            multiplier = get_trt_tensor(
-                ctx,
-                dim_tensor_list[adv_indx_indices[adv_indx_count - 1]],
-                name + "_dim_last",
-            )
+            multiplier = dim_tensor_list[adv_indx_indices[adv_indx_count - 1]]
```
peri044, line 249: Does this change imply multiplier is always an ITensor since you were using `get_trt_tensor` before ?

apbose: Yes we do not need `get_trt_tensor` in this, even in the previous non dynamic cases it was not required since dim_tensor_list already stores a list of dimension tensors

apbose:
I ran the Mistral-7B-v0.3, it shows me this in the end-
```
INFO:torch_tensorrt.dynamo.conversion._TRTInterpreter:TRT INetwork construction elapsed time: 0:08:12.032557
INFO:torch_tensorrt [TensorRT Conversion Context]:Global timing cache in use. Profiling results in this builder pass will be stored.
ERROR:torch_tensorrt [TensorRT Conversion Context]:1: [defaultAllocator.cpp::allocate::19] Error Code 1: Cuda Runtime (out of memory)
WARNING:torch_tensorrt [TensorRT Conversion Context]:Requested amount of GPU memory (28354543872 bytes) could not be allocated. There may not be enough free memory for allocation to succeed.
double free or corruption (!prev)
```
Looks like it got past the dynamic index case

peri044:
`TestIndexDynamicConstantConverter` tests are failing in CI

PR #3061, opened 2024-08-05
fix: Relax thresholds of dynamo converters' tests

### LINKED ISSUE #3053: ✨[Feature] Relax thresholds in tests for TRT 10 upgrade
After bumping to TRT 10, some Dynamo converters' tests in Torch-TRT failed due to threshold/tolerance issue. This is because TRT 10 uses different kernels for some ops, which were not enabled on previous architectures. For fixing the issue, TRT team suggests that **`5e-3`** (we currently use `1e-3`) can be used as the threshold and we can assume it would work for the following TRT releases. Thus, we need to update all the tests to use the suggested threshold. Since input data is torch.randn (normal distribution), we can simply set relative tolerance = absolute tolerance.

zewenli98:
# Description

After bumping to TRT 10, some Dynamo converters' tests in Torch-TRT failed due to threshold/tolerance issue. TRT team suggests using 5e-3 as the threshold.

Fixes #3053 

## Type of change

# Checklist:


narendasan:
Change the TS tests as well

PR #3072, opened 2024-08-08
Test Only: Windows Build failure issue

lanluo-nvidia:
# Description

For the windows build issue, I am able to identify the issue is caused by the torchtrt.dll:
It is complaining the getinstance method in the UndefinedTensorImpl class
```
The procedure entry point
?getinstance@UndefinedTensorImpl@c10@@CAAEAU12@XZ
could not be located in the dynamic link library
C:\Users\lanl\git\venv_py310\Lib\site-packages\torch_tensorrt\
lib\torchtrt.dll.
```
I believe this is related to the libtorch version:
currently we use the one from:
```
https://download.pytorch.org/libtorch/${CHANNEL}/${CU_VERSION}/libtorch-win-shared-with-deps-latest.zip
```
which is causing the issue.

Fixes # (issue)
need to find out a good libtorch mapping.

## Type of change

# Checklist:


HolyWu:
The CI failure issue in Windows workflow is caused by version mismatch between the torch library being used for building the wheel and the torch library being used at runtime. In `Build torch-tensorrt whl package` phase, it's always using the latest nightly version of libtorch. But in the subsequent Test phases, it's installing the version specified in `py/requirements.txt`. However, torchvision specified `>=0.19.0.dev,<0.20.0`, and the last torch version matching torchvision 0.19.0 is `2.5.0.dev20240617`. The dll from 20240617 doesn't contain the symbol appeared in the newer library, hence the issue. I believe the issue should be gone if you upgrade torch and torchvision by `pip3 install --pre -U torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu124` on your local machine. Albeit now there is an upstream breaking change causing `maybe_disable_fake_tensor_mode` import error after upgrading, it's not related to our issue here.

lanluo-nvidia:
> The CI failure issue in Windows workflow is caused by version mismatch between the torch library being used for building the wheel and the torch library being used at runtime. In `Build torch-tensorrt whl package` phase, it's always using the latest nightly version of libtorch. But in the subsequent Test phases, it's installing the version specified in `py/requirements.txt`. However, torchvision specified `>=0.19.0.dev,<0.20.0`, and the last torch version matching torchvision 0.19.0 is `2.5.0.dev20240617`. The dll from 20240617 doesn't contain the symbol appeared in the newer library, hence the issue. I believe the issue should be gone if you upgrade torch and torchvision by `pip3 install --pre -U torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu124` on your local machine. Albeit now there is an upstream breaking change causing `maybe_disable_fake_tensor_mode` import error after upgrading, it's not related to our issue here.

Yes, ultimately if we can always keep update with pytorch latest version, then nightly latest is alright for us.
However it is possible there is some reason we cannot keep up with the latest pytorch version, so I changed it to local_repository instead of getting from nightly latest.
this way libtorch is always synced with torch.
Here is the new PR:
https://github.com/pytorch/TensorRT/pull/3086
Closing this one, since it is a test only.

PR #3085, opened 2024-08-14
tile dynamic dim

apbose:


py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
```diff
@@ -467,10 +469,40 @@ def tile(
...
 
-    shapes = [i * j for i, j in zip(input.shape, dims)]
     starts = [0] * len(dims)
     strides = [1] * len(dims)
-    layer = ctx.net.add_slice(input, tuple(starts), tuple(shapes), tuple(strides))
+    if not (has_dynamic_shape_input) and not (has_dynamic_shape_dims):
+        shapes = [i * j for i, j in zip(input.shape, dims)]
+        layer = ctx.net.add_slice(input, tuple(starts), tuple(shapes), tuple(strides))
+    else:
+        shapes = None
```
peri044, line 478: should shapes be a list here ?

py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
```diff
@@ -467,10 +469,40 @@ def tile(
...
+                trt.ElementWiseOperation.PROD,
+                i,
+                j,
+            )
+            shapes.append(prod_shape)
+            layer = ctx.net.add_slice(
+                input, start=trt.Dims(), shape=trt.Dims(), stride=trt.Dims()
+            )
+            layer.set_input(1, tuple(starts))
+            layer.set_input(2, tuple(shapes))
```
peri044, line 503: Looks like we are setting a list of ITensors as a second input. But shouldn't the second input be an ITensor (which is the concatenated output of these shapes ) ? or both of them work ?

tests/py/dynamo/conversion/test_tile_aten.py
```diff
@@ -71,5 +72,94 @@ def forward(self, x):
...
+class TestTileConverterDynamicShape(DispatchTestCase):
+    @parameterized.expand(
+        [
+            ((3,), (3,), (6,), (1,)),
+            ((3,), (3,), (6,), (0,)),
+            ((3,), (3,), (6,), (2,)),
+            ((2,), (3,), (6,), (2, 2)),
+            ((2,), (3,), (6,), (0, 2)),
+            # 2d cases
+            ((3, 1), (3, 1), (6, 1), (0,)),
```
peri044, line 84: is tiling with dims = (0,) same as tiling with (1,) ?

apbose: Tiling with dims = (0,) results in that particular dimension as 0 in the tiled tensor. In the above case though it prepends -1 resulting in dims being (-1, 0) post which the dynamic input tensor is tiled

PR #3086, opened 2024-08-14
fix the libtorch version mismatch issue

lanluo-nvidia:
# Description

For the windows build issue, I am able to identify the issue is caused by the torchtrt.dll:
It is complaining the getinstance method in the UndefinedTensorImpl class

The procedure entry point
?getinstance@UndefinedTensorImpl@c10@@CAAEAU12@XZ
could not be located in the dynamic link library
C:\Users\lanl\git\venv_py310\Lib\site-packages\torch_tensorrt\
lib\torchtrt.dll.
I believe this is related to the libtorch version:
currently we use the one from:

https://download.pytorch.org/libtorch/${CHANNEL}/${CU_VERSION}/libtorch-win-shared-with-deps-latest.zip
which is causing the issue.

## Type of change

# Checklist:


toolchains/ci_workspaces/MODULE.bazel.tmpl
```diff
@@ -110,17 +89,23 @@ http_archive(
...
+    path = "${TORCH_INSTALL_PATH}",
+    build_file = "third_party/libtorch/BUILD"
+)
 
-#new_local_repository(
-#    name = "libtorch_pre_cxx11_abi",
-#    path = "${TORCH_INSTALL_PATH}",
-#    build_file = "third_party/libtorch/BUILD"
-#
+new_local_repository(
```
narendasan, line 98: Keep in mind if we do this it becomes hard to build the cxx11 abi builds in CI. Consider just replacing the problematic one (libtorch_win) for now. Alternatively, we could replace both `libtorch_win`, and `libtorch_pre_cxx11_abi` but keep `libtorch` as `http_archive`, thats probably best of both worlds for CI

lanluo-nvidia: Yes, this is a very good point, I have modified as suggested and also enabled release build to see ensure cxx11 abi build is successful in ci.

HolyWu:
I don't think simply changing to local_repository will solve the issue, since the locally installed torch in building wheel phase is still the latest nightly version.

lanluo-nvidia:
> I don't think simply changing to local_repository will solve the issue, since the locally installed torch in building wheel phase is still the latest nightly version.

I have uninstall ed latest nightly version, always use the version we have dependencies on py/requirements.txt
this does no longer show us the error in windows.

However there is a new error in windows for cu118:
 torchvision 0.19.0.dev20240617+cu118 depends on torch==2.5.0.dev20240616+cu118 in windows, 
 however https://download.pytorch.org/whl/nightly/torch/ does not have a 2.5.0.dev20240616+cu118

narendasan (approved):
LGTM

.github/scripts/install-torch-tensorrt.sh
```diff
@@ -8,6 +8,11 @@ PLATFORM=$(python -c "import sys; print(sys.platform)")
 pip install --pre ${TORCH_TORCHVISION} --index-url ${INDEX_URL}
 pip install --pre -r ${PWD}/tests/py/requirements.txt --use-deprecated legacy-resolver
 
+export CUDA_HOME="$(echo ${CUDA_PATH} | sed -e 's#\\#\/#g')"
+export TORCH_INSTALL_PATH="$(python -c "import torch, os; print(os.path.dirname(torch.__file__))" | sed -e 's#\\#\/#g')"
+
+cat ${PWD}/toolchains/ci_workspaces/MODULE.bazel.tmpl | envsubst > ${PWD}/MODULE.bazel
```
HolyWu, line 14: Was the addition of these three lines necessary? `MODULE.bazel` is only used in build job to create the wheel. But `install-torch-tensorrt.sh` is only used in tests job so `MODULE.bazel` is having no use here.

lanluo-nvidia: @HolyWu  thanks for pointing out this, it is indeed not required, removed as suggested.

Issue #3074, opened 2024-08-09
🐛 [Bug] Support ITensor type arguments for Batch Norm


zewenli98:
##  Bug Description

While testing tests/py/dynamo/models/test_models.py, there's an error:
```
>       scale = to_numpy(weight) / np.sqrt(to_numpy(running_var) + eps)
E       TypeError: unsupported operand type(s) for +: 'NoneType' and 'float'
E       
E       While executing %_native_batch_norm_legit_no_training : [num_users=1] = call_function[target=torch.ops.aten._native_batch_norm_legit_no_training.default](args = (%convolution, %arg4_1, %arg5_1, %arg2_1, %arg3_1, 0.1, 1e-05), kwargs = {_itensor_to_tensor_meta: {<tensorrt.tensorrt.ITensor object at 0x73f131e20ff0>: ((1, 3, 224, 224), torch.float16, False, (150528, 50176, 224, 1), torch.contiguous_format, False, {}), <tensorrt.tensorrt.ITensor object at 0x73f131e22bb0>: ((64, 3, 7, 7), torch.float16, True, (147, 49, 7, 1), torch.contiguous_format, False, {}), <tensorrt.tensorrt.ITensor object at 0x73f131d92ab0>: ((1, 64, 112, 112), torch.float16, False, (802816, 12544, 112, 1), torch.contiguous_format, False, {}), <tensorrt.tensorrt.ITensor object at 0x73f131e69d30>: ((64,), torch.float16, True, (1,), torch.contiguous_format, False, {}), <tensorrt.tensorrt.ITensor object at 0x73f132272ef0>: ((64,), torch.float16, True, (1,), torch.contiguous_format, False, {}), <tensorrt.tensorrt.ITensor object at 0x73f132270b30>: ((64,), torch.float16, False, (1,), torch.contiguous_format, False, {}), <tensorrt.tensorrt.ITensor object at 0x73f131bbf7b0>: ((64,), torch.float16, False, (1,), torch.contiguous_format, False, {})}})
E       Original traceback:
E       None
```

This is because current implementation of `batch_norm` doesn't support ITensor type arguments, but it should based on the schema:
```
- func: batch_norm(Tensor input, Tensor? weight, Tensor? bias, Tensor? running_mean, Tensor? running_var, bool training, float momentum, float eps, bool cudnn_enabled) -> Tensor
[...]

PR #3099, opened 2024-08-19
feat: Add handling for ITensor mean and var in batch_norm

chohk88:
# Description

Support ITensor type running_mean and running_var arguments for Batch Norm converter.

Fixes # ([issue](https://github.com/pytorch/TensorRT/issues/3074))

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
@@ -36,59 +36,136 @@ def batch_norm(
     input: TRTTensor,
     weight: Optional[Union[torch.Tensor, np.ndarray]],
     bias: Optional[Union[torch.Tensor, np.ndarray]],
-    running_mean: Optional[Union[torch.Tensor, np.ndarray]],
-    running_var: Optional[Union[torch.Tensor, np.ndarray]],
+    running_mean: Union[TRTTensor, Optional[Union[torch.Tensor, np.ndarray]]],
+    running_var: Union[TRTTensor, Optional[Union[torch.Tensor, np.ndarray]]],
     training: bool,
     momentum: float,
     eps: float,
     cudnn_enabled: bool,
     return_mean_rstd: bool,
 ) -> Union[TRTTensor, Tuple[TRTTensor, torch.Tensor, torch.Tensor]]:
+
     if has_dynamic_shape(input.shape):
         assert input.shape[1] != -1, "Channel dim can't be dynamic for batch norm."
 
-    if weight is None:
-        weight = 1.0
-
-    if bias is None:
-        bias = 0.0
-
-    if running_mean is None:
-        running_mean = 0.0
-
-    if running_var is None:
-        running_var = 1.0
-
-    scale = to_numpy(weight) / np.sqrt(to_numpy(running_var) + eps)
-    bias = to_numpy(bias) - to_numpy(running_mean) * scale
-    power = np.ones_like(scale)
-
-    # For BatchNorm1d, reshape 1d to 2d
+    # Save the original output shape for later use
     output_shape = input.shape
-    if len(input.shape) < 4:
-        assert (
-            len(get_dynamic_dims(input.shape)) <= 1
-        ), "BatchNorm1D with more than one dynamic dims is not currently supported."
-        new_shape = (
-            (input.shape[0], input.shape[1], 1, 1)
-            if len(input.shape) == 2
-            else (input.shape[0], input.shape[1], input.shape[2], 1)
-        )
-        input = impl.shuffle.reshape(
-            ctx, target, source_ir, f"{name}_reshape_2d", input, new_shape
-        )
-    layer = ctx.net.add_scale(input, trt.ScaleMode.CHANNEL, bias, scale, power)
-    set_layer_name(layer, target, name, source_ir)
-    output = layer.get_output(0)
 
-    # For BatchNorm1d, reshape output back to 1d
+    # Handle case when running_mean or running_var is TRTTensor
+    if isinstance(running_mean, TRTTensor) or isinstance(running_var, TRTTensor):
+        # Default values if weight, bias, running_mean, running_var are None
+        if weight is None:
+            weight = get_trt_tensor(ctx, 1.0, f"{name}_weight", input.dtype)
+        if bias is None:
+            bias = get_trt_tensor(ctx, 0.0, f"{name}_bias", input.dtype)
+        if running_mean is None:
+            running_mean = get_trt_tensor(ctx, 0.0, f"{name}_running_mean", input.dtype)
+        if running_var is None:
+            running_var = get_trt_tensor(ctx, 1.0, f"{name}_running_var", input.dtype)
+
+        # eps_tensor for numerical stability
+        eps_tensor = get_trt_tensor(ctx, eps, f"{name}_eps")
+
+        # adjusted_var = running_var + eps
+        adjusted_var = impl.elementwise.add(
+            ctx, target, source_ir, f"{name}_adjusted_var", running_var, eps_tensor
+        )
+
+        # sqrt_adjusted_var = sqrt(adjusted_var)
+        sqrt_adjusted_var = impl.unary.sqrt(
+            ctx, target, source_ir, f"{name}_sqrt", adjusted_var
+        )
+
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -36,59 +36,136 @@ def batch_norm(
...
+    if isinstance(running_mean, TRTTensor) or isinstance(running_var, TRTTensor):
+        # Default values if weight, bias, running_mean, running_var are None
+        if weight is None:
+            weight = get_trt_tensor(ctx, 1.0, f"{name}_weight", input.dtype)
+        if bias is None:
+            bias = get_trt_tensor(ctx, 0.0, f"{name}_bias", input.dtype)
+        if running_mean is None:
+            running_mean = get_trt_tensor(ctx, 0.0, f"{name}_running_mean", input.dtype)
+        if running_var is None:
+            running_var = get_trt_tensor(ctx, 1.0, f"{name}_running_var", input.dtype)
```
zewenli98, line 64: 1) Do we need to cast these parameters to the type of `input`? I think there's probably a case that `input` is int type while `weight`, `bias`, `running_mean`, and/or `running_var` are float type. It seems problematic to force cast float to int. The `dtype` is optional so you can just leave it blank.

2) `weight` and `bias` could be ITensor as well right?

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -36,59 +36,136 @@ def batch_norm(
     input: TRTTensor,
     weight: Optional[Union[torch.Tensor, np.ndarray]],
     bias: Optional[Union[torch.Tensor, np.ndarray]],
-    running_mean: Optional[Union[torch.Tensor, np.ndarray]],
-    running_var: Optional[Union[torch.Tensor, np.ndarray]],
+    running_mean: Union[TRTTensor, Optional[Union[torch.Tensor, np.ndarray]]],
+    running_var: Union[TRTTensor, Optional[Union[torch.Tensor, np.ndarray]]],
```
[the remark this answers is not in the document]
zewenli98: I just noticed that `torch.ops.aten.native_batch_norm.default` and `torch.ops.aten.batch_norm.default` reuse the function. Since they require `running_mean` and `running_var` to be optional, you can put all these types to `Optional[Union[TRTensor, torch.Tensor, np.ndarray]]`

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
keehyuna: >>Do we need to cast these parameters to the type of input
I thinks this is required for [strongly-typed-networks](https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#strongly-typed-networks), different type can be allowed for weak typed network.
I noticed it by enabling strongly typed networks with some model. Here is [some changes](https://github.com/keehyuna/TensorRT/commit/2e32bde7370bf26d8f7b7112b8aa53b52ab39644#diff-47e9eb66cba97ff59bf6c1355e0506921be5780f8796d725f9dc3abcdb22cda5R127) to keep same type in ops. I only saw float32 and half float, I'm now sure if float and int tensor in same layer is possible.

peri044 (approved):
LGTM

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
chohk88: ->>Do we need to cast these parameters to the type of input?
If `input_val` is a `TRTTensor`, the input is returned unchanged ([code](https://github.com/pytorch/TensorRT/blob/db15ecaa6f18cfdae99e14b5ff497a60abcf231b/py/torch_tensorrt/dynamo/conversion/converter_utils.py#L386)), so the type stays the same. If `input_val` is something else, `create_constant` converts the type using `to_numpy` ([code](https://github.com/pytorch/TensorRT/blob/db15ecaa6f18cfdae99e14b5ff497a60abcf231b/py/torch_tensorrt/dynamo/conversion/converter_utils.py#L344)). However, if the `value` is an `np.ndarray` or `torch.Tensor`, it keeps the original type ([code](https://github.com/pytorch/TensorRT/blob/db15ecaa6f18cfdae99e14b5ff497a60abcf231b/py/torch_tensorrt/dynamo/conversion/converter_utils.py#L550) and [code](https://github.com/pytorch/TensorRT/blob/db15ecaa6f18cfdae99e14b5ff497a60abcf231b/py/torch_tensorrt/dynamo/conversion/converter_utils.py#L559)). This means setting `input.dtype` has no effect.

I confirmed this behavior when I removed the `input.dtype` argument.

As for the issue @keehyuna  mentioned, it's new to me, so I'll investigate.

->> `weight` and `bias` could be ITensor as well right?
Additionally, according to the schema, `weight` and `bias` can't be `ITensor`, but the converter works fine. I've tested it, and it works successfully.

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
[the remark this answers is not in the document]
chohk88: Thank you for your comments! I've resolved the issue based on your feedback.

chohk88:
> LGTM

If the CI results show no more batch_norm errors, it's good to merge.

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
zewenli98: > ->> weight and bias could be ITensor as well right?
Additionally, according to the schema, weight and bias can't be ITensor, but the converter works fine. I've tested it, and it works successfully.

@chohk88 Did I miss something? The schema is: 
```
- func: batch_norm(Tensor input, Tensor? weight, Tensor? bias, Tensor? running_mean, Tensor? running_var, bool training, float momentum, float eps, bool cudnn_enabled) -> Tensor
- func: _native_batch_norm_legit_no_training(Tensor input, Tensor? weight, Tensor? bias, Tensor running_mean, Tensor running_var, float momentum, float eps) -> (Tensor, Tensor, Tensor)
```

My understanding to `Tensor? weight, Tensor? bias` is that they could be None, ITensor, torch.Tensor, or np.ndarray. Please correct me if I'm wrong.

chohk88: Oh, I misunderstood that. You're right—`weight`, `bias`, `running_mean`, and `running_var` can all be `TRTTensor`. I’ve combined the separate converter and added some test cases for this.

chohk88:
@zewenli98  Although it's beyond the scope of this PR, I noticed something regarding the converters for `layer_norm`, `group_norm`, and `native_group_norm`.

Currently, `weight` and `bias` are defined as:
```python
weight: Optional[Union[torch.Tensor, np.ndarray]],
bias: Optional[Union[torch.Tensor, np.ndarray]],
```

Is this incorrect? Fortunately, unlike batch norm, there's no issue with adding `eps` or applying `to_numpy` on the `weight` or `bias` of `TRTTensor` (or `ITensor`), so it doesn't seem to cause any problems with the converter. However, it seems like the case where the default value for `weight` in `layer_norm` is `None` might not be handled properly.

Here are schema:
- [`group_norm`](https://github.com/pytorch/pytorch/blob/d990dada86a8ad94882b5c23e859b88c0c255bda/aten/src/ATen/native/native_functions.yaml#L2961)
- [`layer_norm`](https://github.com/pytorch/pytorch/blob/d990dada86a8ad94882b5c23e859b88c0c255bda/aten/src/ATen/native/native_functions.yaml#L3250)
- [`native_group_norm`](https://github.com/pytorch/pytorch/blob/d990dada86a8ad94882b5c23e859b88c0c255bda/aten/src/ATen/native/native_functions.yaml#L2963)

peri044:
Merging this as the other failures are unrelated.

zewenli98:
@chohk88 Thanks for pointing out the issue. I think your understanding is correct. Although there's no errors out now, they should be `Optional[Union[TRTTensor, torch.Tensor, np.ndarray]]`. Since the Issue #3114 has been opened above to track it, could you modify them to `Optional[Union[TRTTensor, torch.Tensor, np.ndarray]]` like what you did for the batch_norm?

PR #3120, opened 2024-08-27
chunk converter validator

apbose:
This validator is for the purpose of dynamic shape support

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -692,7 +692,9 @@ def aten_ops_softmax(
 
 @dynamo_tensorrt_converter(
     torch.ops.aten.split.Tensor,
-    capability_validator=has_static_shapes_in_args([1]),
+    capability_validator=(
+        has_static_shapes_in_args([0]) and has_static_shapes_in_args([1])
+    ),
     supports_dynamic_shapes=True,
 )
 @dynamo_tensorrt_converter(
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -692,9 +692,7 @@ def aten_ops_softmax(
 
 @dynamo_tensorrt_converter(
     torch.ops.aten.split.Tensor,
-    capability_validator=(
-        has_static_shapes_in_args([0]) and has_static_shapes_in_args([1])
-    ),
+    capability_validator=has_static_shapes_in_args([1]),
     supports_dynamic_shapes=True,
 )
 @dynamo_tensorrt_converter(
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -903,7 +904,17 @@ def aten_ops_slice(
     )
 
 
-@dynamo_tensorrt_converter(torch.ops.aten.chunk.default)
+def chunk_validator(node: Node) -> bool:
+    meta_data = node.args[0].meta.get("tensor_meta")
```
peri044, line 908: Also check for the key `val` for the shape first, else check `tensor_meta` and if not return False

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -903,7 +904,17 @@ def aten_ops_slice(
...
+    if meta_data is None:
+        return False
+    shape = meta_data.shape
+    dynamic_shape = has_dynamic_shape(shape)
+    if (dynamic_shape):
```
peri044, line 913: consider using if has_dynamic_shape(shape) directly

tests/py/dynamo/conversion/test_chunk_aten.py
```diff
@@ -51,6 +52,7 @@ def forward(self, input):
         self.run_test(
             TestChunk(),
             input,
+            enable_passes=True,
```
peri044, line 55: Any reason we need the passes for this test ?

tests/py/dynamo/conversion/test_chunk_aten.py
```diff
@@ -75,8 +78,102 @@ def forward(self, input):
...
+    #         ((1,), (1,), (3,), 3, 0),
+    #         ((3,), (3,), (4,), 3, 0),
+    #         ((4,), (4,), (6,), 3, 0),
+    #         ((6,), (6,), (9,), 3, 0),
+    #         ((3,), (3,), (4,), 1, -1),
+    #         ((3,), (3,), (4,), 3, -1),
+    #         ((3,), (3,), (4,), 4, -1),
+    #     ]
+    # )
+    # def test_chunk_1D(self, min_shape, opt_shape, max_shape, chunks, dim):
```
peri044, line 97: Probably it's better to add a unittest.skip similar to https://github.com/pytorch/TensorRT/blob/95cc532aa6a11c8bb55bc43bc2b04a76d264577e/tests/py/dynamo/conversion/test_evaluators.py#L13 and add a link to the issue you created.

peri044 (changes requested):
Since this supports only static shapes now, is it better if we remove supports_dynamic_shapes=True ?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -692,7 +692,9 @@ def aten_ops_softmax(
 
 @dynamo_tensorrt_converter(
     torch.ops.aten.split.Tensor,
-    capability_validator=has_static_shapes_in_args([1]),
+    capability_validator=(
+        has_static_shapes_in_args([0]) and has_static_shapes_in_args([1])
```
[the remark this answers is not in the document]
apbose: args[0] is `input` and args[1] is `split_size_or_sections`. Initially I thought we should add this since dynamic input tensor split is not supported, but yes you are right, the error comes in torch.export before the interpretor stage, so having this validator does not make sense. We are anyways skipping the test so that CI does not fail. So will remove this.

tests/py/dynamo/conversion/test_chunk_aten.py
```diff
@@ -75,6 +80,104 @@ def forward(self, input):
         self.run_test(
             TestChunk(),
             input,
+            use_dynamo_tracer=True,
+        )
+
+
+#######################Dynamic cases#######################
+#The tests are skipped for now. Will be addressed once https://github.com/pytorch/pytorch/issues/134663 is addressed
+@unittest.skip("Pending aten.split converter. Currently tested by E2E")
```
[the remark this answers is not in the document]
apbose: Oh I think it was a misinterpretation, as in in an earlier comment I thought you wanted me to add this message. I will change this to @unittest.skip("Pending aten.split dynamic input torch.export guard bug. Issue- https://github.com/pytorch/pytorch/issues/134663")

PR #3128, opened 2024-08-29
Fix typo in _compile.py

juliusgh:
# Description

There is a typo in the `convert_method_to_trt_engine` function in `_compile.py` causing a Python syntax error.

Adding a comma fixes the issue.

## Type of change


peri044 (approved):
Thanks for the patch.

Issue #3145, opened 2024-09-04
🐛 [Bug] Log level changed in client code


seymurkafkas:
[This logging call](https://github.com/pytorch/TensorRT/blob/49771b520be13bb2be7d275aae30c03f7f5de9eb/py/torch_tensorrt/fx/__init__.py#L16)sets up the root logger at import time. This is problematic, because:

- Client projects want to configure logging on their own. 
- A second handler may be added to the root logger (e.g. when the client code doesn't use `logging.basicConfig`). This means duplicated logs, which leads to double the storage costs, should someone use `torch_tensorrt` in production.

How to Reproduce:
```python
import logging
import torch_tensorrt  # This import triggers logging.basicConfig 


_LOGGER = logging.getLogger()
_LOGGER.addHandler(logging.StreamHandler())

print(f"Number of handlers on root logger: {len(_LOGGER.handlers)}")
# THIS MEANS DUPLICATE LOGS
```

Result: `Number of handlers on root logger: 2`

narendasan:
Seems reasonable to me to remove this, would you be able to submit a patch?

seymurkafkas:
> Seems reasonable to me to remove this, would you be able to submit a patch?

For sure, will do it 👍

seymurkafkas:
Submitted a fix in this PR #3147

narendasan:
Fixed in #3147

PR #3147, opened 2024-09-04
fix: remove global logging setup

seymurkafkas:
Removes the global logging setup (which is initialized at import time). Solves #3145


peri044 (approved):
LGTM

PR #3153, opened 2024-09-11
scatter reduce lowering with include_self=False

apbose:
This is for scatter_reduce decomposition where include_self=False


narendasan:
Has been 9mo with no progress so closing

PR #3156, opened 2024-09-11
MODULE.bazel file for NGC docker container

apbose:


narendasan (commented):
Do you need the torch-tensorrt-testing dependency?

docker/MODULE.bazel.ngc
```diff
@@ -0,0 +1,73 @@
...
+
+bazel_dep(name = "googletest", version = "1.14.0")
+bazel_dep(name = "platforms", version = "0.0.10")
+bazel_dep(name = "rules_cc", version = "0.0.9")
+bazel_dep(name = "rules_python", version = "0.34.0")
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(
+    ignore_root_user_error = True,
+    python_version = "3.11",
```
narendasan, line 15: This should probably be 3.12 right?

apbose: In the recent release I kept it as 3.11 and it went fine. Should the toolchain be 3.12? I will change to it. For 3.10, we used 3.11, so how should we as such determine the version?

PR #3158, opened 2024-09-11
remove torchvision as dependency

lanluo-nvidia:
# Description

Fixes # (issue)

## Type of change

# Checklist:


tests/py/requirements.txt
```diff
@@ -12,3 +12,6 @@ transformers==4.40.2
 # TODO @lanlao-nvidia Renable when modelopt can be install properly to run the tests
 # "nvidia-modelopt[all]">=0.16.1,<0.17.0
 --extra-index-url https://pypi.nvidia.com
+torchvision>=0.20.0.dev,<0.21.0
+--extra-index-url https://download.pytorch.org/whl/nightly/cu124
```
HolyWu, line 16: I think this `extra-index-url` is redundant. We already have `INDEX_URL` defined in `install-torch-tensorrt.sh` which has templated channel and cu_version.

lanluo-nvidia: yes, It is redundant.
in the install-torch-tensorrt.sh it is templated with channel and cu_version.
I still want to keep this --extra-index-url here just in case that developer run:
python -m pip install --pre -r tests/py/requirements.txt in their local environment to test

narendasan:
How many tests are getting altered from this changed? Are there tests where we could just use a simple network that we host and get similar impact for testing and not have to disable it?

I think we really only care about torchvision models in cases where we specifically want to test that model architecture (so we should have at least 1 resnet, 1 bert, 1 vit test) or we need real weights for some reason

lanluo-nvidia:
> Note that the latest nightly torchvision has depended on torch 2.6.0.dev, so installing torchtrt wheel will fail because it still depends on torch 2.5.0.dev.

we have specified the torchvison version so that make sure it is compatible with torch version in the test.


> How many tests are getting altered from this changed? Are there tests where we could just use a simple network that we host and get similar impact for testing and not have to disable it?
> 
> I think we really only care about torchvision models in cases where we specifically want to test that model architecture (so we should have at least 1 resnet, 1 bert, 1 vit test) or we need real weights for some reason

@narendasan  
There are total 30 testcases which uses torchvision.resnet18(28 testcases) torchvision.mobilenet_v2(2 testcases) models 
for the Bertmodel we used it from transformers, did not use torchvision.

I can see most times we just use torchvision.resnet18 to test our features, it can be replaced with other simple network.
For example: 
[test_engine_cache.py ](https://github.com/pytorch/TensorRT/blob/main/tests/py/dynamo/models/test_engine_cache.py#L62) we have 5 testcases uses resnet18 model to test engine cache related features
[test_model_refit.py](https://github.com/pytorch/TensorRT/blob/main/tests/py/dynamo/models/test_model_refit.py#L95)
we have 9 testcases use resnet18 model to test our refit related feature

I think for test_models.py is where we should actually use these torchvision.resnet18 and movilenet_v2 models for testing.

HolyWu:
Be aware that all tests were failed to run even though they all showed green. https://github.com/pytorch/TensorRT/actions/runs/10887326997/job/30210220619?pr=3158#step:12:522

PR #3159, opened 2024-09-12
feat: exclude refit sensitive ops from TRT compilation

peri044:
# Description

1) Fix refit errors on certain ops by forcing them to fallback on pytorch. The global default for `make_refittable` is set to `False` even after this addition.
2) Adds a capability validator for cumsum and embedding bag to let them fallback to pytorch incase user builds refittable engines
3) Expose user compilation settings to ConverterRegistry
4) Added a testcase to verify fallback
5) The converter testcases for cumsum and embedding bag are intentionally marked as `make_refittable=False`.  This is because in the future, we want `make_refitable=True` by default and hence these tests fail. So, explicitly marking these now will avoid those errors.

Fixes https://github.com/pytorch/TensorRT/actions/runs/10637055400/job/29490744653?pr=3131

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -317,6 +318,10 @@ def compile_module(
     # Assume converters support dynamic shapes and disable validation
     CONVERTERS.set_dynamic_shape_support(settings.assume_dynamic_shape_support)
 
+    # Set non-refitable ops as disallowed targets.
+    if settings.make_refitable:
+        settings.torch_executed_ops.update(REFIT_SENSITIVE_OPS)
```
narendasan, line 323: I think its better to do this through validators than maintaining another list

peri044: 1) Currently, the capability validator doesn't have access to user settings. So, I think making the compilation settings in _compiler.py (compile call) a global variable and passing it here would work. https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py#L439 

2) The other alternative if we don't want to modify user settings is 
```py
CONVERTERS.set_disallowed_targets(settings. REFIT_SENSITIVE_OPS)
```
What would be preferred ?

narendasan: Dont see a drawback to let validators see user settings, could be relevant later like if a certain converter only works for one data type

narendasan: So it would involve updating the converter registry to take the settings as an arg in addition to the node, and some adjustment in the decorator. But overall I think its the best design since we shouldnt be spreading this info in a bunch of global lists, a converter should be able to tell you what it can and cant do on its own 

Dont think we need to make settings global we could make a reference a member of the registry like we do the dynamic shape setting https://github.com/pytorch/TensorRT/blob/4d2a04a7ccd4917f3f15666840d5175f1b9ed9c6/py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py#L300 and then it gets injected into the call to the validator

narendasan: Just an additional arg here: https://github.com/pytorch/TensorRT/blob/4d2a04a7ccd4917f3f15666840d5175f1b9ed9c6/py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py#L439

peri044: Got it. modified it now.

narendasan (approved):
LGTM, make sure to mark needs cherrypick and ping @lanluo-nvidia

PR #3167, opened 2024-09-19
feat: Support weight-stripped engine and REFIT_IDENTICAL flag

### LINKED ISSUE #3146: ✨[Feature] Weight specific engine caching

Caching right now is weight agnostic, but at the cost of creating lower performance engines. 

If we know that weights would be identical, then we can cache engines that are higher performance. The caching system would need to be able to distinguish these two caches and based on user settings select the right one 

TensorRT has a flag called kREFIT_IDENTICAL for this workflow 

**Additional context**

zewenli98:
# Description

1) Supported weight-stripped engine
2) Added REFIT_IDENTICAL flag

Fixes #3146 

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -79,6 +79,7 @@ def __init__(
         name: str = "",
         settings: CompilationSettings = CompilationSettings(),  # Assumes engine was built with default compilation settings if object not passed
         weight_name_map: Optional[dict[Any, Any]] = None,
+        graph_module: torch.fx.GraphModule = None,
```
zewenli98, line 82: @narendasan I tried to do refitting like in Python runtime but didn't work. Any suggestions? should I do in C++ or Python?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -619,27 +609,32 @@ def run(
             builder_config, self.compilation_settings.timing_cache_path
         )
 
-        serialized_engine = self.builder.build_serialized_network(
+        # if strip_engine_weights is true, the serialized engine need to be refitted before using
+        maybe_unrefitted_serialized_engine = self.builder.build_serialized_network(
```
cehongwang, line 613: Why is this maybe unrefitted engine?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
narendasan: Doesnt refit already work on both apis?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -121,6 +124,52 @@ def setup_engine(self) -> None:
         self.engine = runtime.deserialize_cuda_engine(self.serialized_engine)
         self.context = self.engine.create_execution_context()
 
+        if self.settings.strip_engine_weights:
+            assert (
+                self.settings.make_refittable
+            ), "weight-stripped engines must be refittable, please set make_refittable=True"
+
+            # Refit the weights
+            refitter = trt.Refitter(self.engine, TRT_LOGGER)
```
cehongwang, line 133: Can you use this function?
https://github.com/pytorch/TensorRT/blob/fa02fd3e3a85a6042d11a00cd386f6b69c1d6c4b/py/torch_tensorrt/dynamo/_refit.py#L138

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
narendasan: Also why do we need the graph module in this module?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -121,6 +124,52 @@ def setup_engine(self) -> None:
         self.engine = runtime.deserialize_cuda_engine(self.serialized_engine)
         self.context = self.engine.create_execution_context()
 
+        if self.settings.strip_engine_weights:
```
narendasan, line 127: We likely shouldnt be doing the refit **in** these modules

I think for weight stripping there are 2 workflows.

1. a user just wants a weight stripped engine. They should use `convert_exported_program_to_trt_engine` with settings `strip_weights`. The choice of `make_refittable` can be used to decide between `kREFIT` and `kREFIT_IDENTICAL` (though it might not be entirely clear so we might want to think about that setting). 
2. We want to utilize weight stripping to have a lighter weight cache. Here this choice is opaque to the user. The user choice of `make_refittable` controls if we use `kREFIT` or `kREFIT_IDENTICAL`. But once the engine is loaded or we pull from cache we immediately refit (prior to passing the engine to the TRTModule).  
3. The user wants a stripped weights compiled program (im not sure why or if this is a real usecase). Here, this is basically the same as lazy engine loading. We would require that users need to run through `refit_engine_weights` before executing.

narendasan (commented):
@zewenli98 do you have a design for this feature?

zewenli98:
@narendasan Ok, at first the overall design was like:

In TRTInterpreter.run():
```python
if compilation_settings.strip_engine_weights is True:
    if engine_cache not hit:
        1. build a weight-stripped engine
        2. save the weight-stripped engine if engine_cache is set
        3. return the weight-stripped engine (not yet refit)
    else:
        load and return the weight-stripped engine (not yet refit)
else:
    if engine_cache not hit:
        1. build a weight-included engine
        2. save the weight-included engine if engine_cache is set
        3. return the weight-included engine (don't need to refit)
    else:
        load and return the weight-included engine (not yet refit)
```
Then, in TRTModule, refit if necessary before inference.
The reason that I didn't put the refitting part into TRTInterpreter.run() is that I want to avoid repeated de/serializations of TRT engines: (1) deserialize in TRTInterpreter.run() for refitting and then serialize (2) deserialize in TRTModule again.

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
zewenli98: 1. In this PR I moved the refitting part into TRTModule, so only works for Python runtime.

2. graph module is used for refitting

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
zewenli98: please see the design in the comment below. If compilation_settings.strip_engine_weights is true, it needs to be refitted, else it doesn't. so it's _maybe_

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
zewenli98: The function requires `input_list` which is not provided in the caller.

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
zewenli98: Got it. The very beginning idea/design is commented below. I'll move the refitting part back to TRTInterpreter.run()

> The choice of `make_refittable` can be used to decide between `kREFIT` and `kREFIT_IDENTICAL`

Do you mean we use `make_refittable` to control both `kREFIT` and `kREFIT_IDENTICAL`?

zewenli98:
@narendasan The design was updated. 

From the users' perspective, they are able to set `make_refittable` and `refit_identical_engine_weights`. 
`make_refittable` for general refitting and `refit_identical_engine_weights` only for refitting with identical weights
```
if self.compilation_settings.make_refittable:
    if version.parse(trt.__version__) >= version.parse("10.0"):
        if self.compilation_settings.refit_identical_engine_weights:
            builder_config.set_flag(trt.BuilderFlag.REFIT_IDENTICAL)
        else:
            builder_config.set_flag(trt.BuilderFlag.REFIT)
    else:
        builder_config.set_flag(trt.BuilderFlag.REFIT)
```

Besides, users can specify `strip_engine_weights`. If `strip_engine_weights` is True, TRTInterpreter.run() will return weight-stripped engine. Otherwise, return general engine (with weights). 

For the 3 workflows mentioned above,
1. controlling the args above, users can call `convert_exported_program_to_trt_engine` specifying `strip_engine_weights=True` to get weight-stripped engine.

2. For engine caching, the implementation of weight-stripped engine is opaque to users, which means engine caching mechanism will (1) save weight-stripped engine no matter what settings users specify (`make_refittable` is required to be true) and then (2) load and refit the weight-stripped engine while reusing cached engines.
If `strip_engine_weights` is True, the engine will not be refitted. Instead, just returns weight-stripped engine.

3. If users specify `strip_engine_weights=True`, calling `torch.compile()` or `torch_trt.dynamo.compile()` will return weight-stripped compiled program. If running the compiled program with inputs, all the results will be zeros. Then, calling `refit_module_weights` will make weights back, e.g.:
```
from torch_tensorrt.dynamo._refit import refit_module_weights
refitted_trt_gm = refit_module_weights(trt_gm, exp_program)
refitted_output = refitted_trt_gm(*inputs)
```
Please see more details in the tests.

narendasan:
> The reason that I didn't put the refitting part into TRTInterpreter.run() is that I want to avoid repeated de/serializations of TRT engines: (1) deserialize in TRTInterpreter.run() for refitting and then serialize (2) deserialize in TRTModule again.

I think that we need to separate the runtime and the compiler so im willing to spend the time serializing and deserializing. 

I think we should frame PR this around moving TRTInterpreter to default to building weight stripped engines. There will be 3 kinds of engines now. 

1. weight strip + refittable (`strip_weights` + `kREFIT`) - should move towards this being the default
2. weight strip + refittable with original weights (`strip_weights` + `kREFIT_INDIVIDUAL`)
3. non_refittable 

The first 2 need separate cache entries. So we need to be able to hash on the weights in the case that the model is being built with `kREFIT_INDIVIDUAL` 

We should look to prefer case 1 in the long term as it allows us to reuse the most work, case 2 would be the next preference. Case 2 should produce faster engines than Case 1 so there remains a need to support `kREFIT_IDENTICAL` 

> Do you mean we use make_refittable to control both kREFIT and kREFIT_IDENTICAL?

The case for type 3 engines now is only valid if building a non refittable engine is faster than building a refit_identical engine then refitting the weights. If it is not by a significant enough margin I propose we remove that workflow and just have `refit` or `refit_identical` engines. 

So assuming that we can remove type 3 engines, `make_refittable` really means "allows the weights to be changed" (we can change the name if needed here), since now both engines are refittable they just have different weight constraints.

narendasan:
Some of the open questions are: 
- how we determine if the weights have been refit prior to running the engine. Can TRT tell us without an error? 
- How can we tell if a user is trying to refit an engine with different weights to an engine built with REFIT_IDENTICAL? 
- If building strip weights refit identical  + refit is slower than just building?

zewenli98:
> * how we determine if the weights have been refit prior to running the engine. Can TRT tell us without an error?

My current design is: If users specify `strip_engine_weights=True` in compile, the weights will not be refitted. They will get a weight-stripped engine. 
However, if they get an engine somewhere, they can call [get_missing_weights()](https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Core/Refitter.html#tensorrt.Refitter.get_missing_weights) to see if there's any weight not gets refitted.

> * How can we tell if a user is trying to refit an engine with different weights to an engine built with REFIT_IDENTICAL?

I also thought about it earlier. The [TRT doc](https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/namespacenvinfer1.html#abdc74c40fe7a0c3d05d2caeccfbc29c1ac38aae7d53507fc823b7d223ca3a0a8d) says "if the refit weights are not identical to the build-time weights, behavior is undefined... This enables use of a single set of weights with different inference backends, or with TensorRT plans for multiple GPU architectures." 
My understanding is that we cannot tell if weights are identical in build time and refitting, from the perspective of engine itself, because weight-stripped engine doesn't compare weights in build time and refitting phase, or give any prompts. So users need to be clear what they are refitting.

> * If building strip weights refit identical  + refit is slower than just building?

will investigate on it. But if users want a weight-included engine, they can specify `strip_engine_weights=False`

zewenli98:
> The case for type 3 engines now is only valid if building a non refittable engine is faster than building a refit_identical engine then refitting the weights. If it is not by a significant enough margin I propose we remove that workflow and just have refit or refit_identical engines.

@narendasan I tested on building Resnet18 and vgg16 via the two paths: (1) `strip weights + refit_identical + refit` (2) `non-refittable`, build time of the two ways are almost same (diff < 1%). I'm not sure if there are other benefits from non-refittable engines even though the build time are the same, like in deployment weights are not allowed to be changed?

zewenli98:
@narendasan I just confirmed with TRT team, the conclusion is `engine built with STRIP_PLAN + REFIT_IDENTICAL + refit` is almost same as `non-refittable engine`. Do you prefer to remove non-refittable engine path?
If yes, the paths would be:
1. weight strip + refittable (strip_weights + kREFIT) - default
2. weight strip + refittable with original weights (strip_weights + kREFIT_IDENTICAL)

> So assuming that we can remove type 3 engines, make_refittable really means "allows the weights to be changed" (we can change the name if needed here), since now both engines are refittable they just have different weight constraints.

I think we can rename `make_refittable` to `refit_mode: str: Union["general", "identical"]` (may be easier to extend in the future?) or `refit_identical_weights: bool`. Then, we can remove `refit_identical_engine_weights` arg which has been committed in this PR.

On top of this, `STRIP_PLAN` will be always on while building engines. we have `strip_engine_weights` arg to allow users to control if they want to get weight-stripped engines.

In summary, the 3 workflows mentioned above would be:
1. Users just want a weight stripped engine. They can call `convert_exported_program_to_trt_engine` specifying `strip_engine_weights=True` to get weight-stripped engine. It is also supported if the engine is loaded from engine cache.

2. We want to utilize weight stripping to have a lighter weight engine cache. The implementation of weight-stripped engine is opaque to users. However, if users specify `kREFIT` or `kREFIT_IDENTICAL`, they would be considered as different engine and cached twice.

3. Users want a stripped weights compiled program. They just need to call `torch.compile()` or `torch_trt.dynamo.compile()` with `strip_engine_weights=True`. If running the compiled program with inputs immediately, all the results will always be zeros. Calling `refit_module_weights()` will make weights back

narendasan:
I think we should remove non-refittable then and we can add it back as a non default workflow later if theres some reason to. 


> Users want a stripped weights compiled program. They just need to call torch.compile() or torch_trt.dynamo.compile() with strip_engine_weights=True. If running the compiled program with inputs immediately, all the results will always be zeros. Calling refit_module_weights() will make weights back

I still dont know what the usecase for this is

narendasan:
>> How can we tell if a user is trying to refit an engine with different weights to an engine built with REFIT_IDENTICAL?

> I also thought about it earlier. The [TRT doc](https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/namespacenvinfer1.html#abdc74c40fe7a0c3d05d2caeccfbc29c1ac38aae7d53507fc823b7d223ca3a0a8d) says "if the refit weights are not identical to the build-time weights, behavior is undefined... This enables use of a single set of weights with different inference backends, or with TensorRT plans for multiple GPU architectures."
My understanding is that we cannot tell if weights are identical in build time and refitting, from the perspective of engine itself, because weight-stripped engine doesn't compare weights in build time and refitting phase, or give any prompts. So users need to be clear what they are refitting.

We should think about a solution for this since behavior is undefined

zewenli98:
> I think we should remove non-refittable then and we can add it back as a non default workflow later if theres some reason to.

I was aware that some ops/nodes are not refittable, like `cumsum` or `embedding_bag`. That was handled in the previous PR: https://github.com/pytorch/TensorRT/pull/3159. However, if `make_refittable` is removed, we may need another flag to tell TRT to build a non-refittable engine?

narendasan:
> > I think we should remove non-refittable then and we can add it back as a non default workflow later if theres some reason to.
> > > Users want a stripped weights compiled program. They just need to call torch.compile() or torch_trt.dynamo.compile() with strip_engine_weights=True. If running the compiled program with inputs immediately, all the results will always be zeros. Calling refit_module_weights() will make weights back
> > 
> > 
> > I still dont know what the usecase for this is
> 
> I think this allows users to have a weight-stripped compiled program first, and then they can refit with different weights. For example:

Yeah but why would they care that there are no weights inside? The thing I'm worried about is now you get these non live programs returned to users. So either we need to be able to check if the weights have been refitted before running or we shouldn't let exported programs have stripped weights

zewenli98:
> Yeah but why would they care that there are no weights inside? 

I was thinking if this is a possible scenario that users call compile function to get a weight-stripped compiled program and then distribute the program to different nodes/GPUs, and each node refits with different weights?

> The thing I'm worried about is now you get these non live programs returned to users. So either we need to be able to check if the weights have been refitted before running or we shouldn't let exported programs have stripped weights

Natively TRT doesn't restrict the running of weight-stripped engines. It just returns results with all zeros. But yeah if the scenario above is not a possible case, we can just return weight-included exported programs.

narendasan:
> Natively TRT doesn't restrict the running of weight-stripped engines. It just returns results with all zeros. But yeah if the scenario above is not a possible case, we can just return weight-included exported programs.

I guess thats better than nothing, ideally still we have a way to detect if weights have been refitted or not

py/torch_tensorrt/dynamo/_refit.py
```diff
@@ -414,6 +410,10 @@ def refit_module_weights(
                 "The type of graph module is not supported for refitting or two compiled modules do not match."
             )
 
+        assert (
```
narendasan, line 413: Weight stripped engines can only be refit once?

py/torch_tensorrt/dynamo/_refit.py
```diff
@@ -451,17 +451,24 @@ def refit_module_weights(
                     weight_name_map=None,
                 )
 
+        # clear EXCLUDE_WEIGHTS flag
+        serialization_config = engine.create_serialization_config()
+        serialization_config.clear_flag(trt.SerializationFlag.EXCLUDE_WEIGHTS)
+        serialized_engine = engine.serialize_with_config(serialization_config)
```
narendasan, line 457: Why do we serialize then immediately deserialize here?

py/torch_tensorrt/dynamo/_refit.py
```diff
@@ -451,17 +451,24 @@ def refit_module_weights(
...
+        engine = runtime.deserialize_cuda_engine(serialized_engine)
+
+        if isinstance(compiled_submodule, PythonTorchTensorRTModule):
+            compiled_submodule.engine = engine
+
         if isinstance(compiled_submodule, TorchTensorRTModule):
-            serialized_engine = bytes(engine.serialize())
             new_engine_info = list(engine_info)
-            new_engine_info[ENGINE_IDX] = serialized_engine
+            new_engine_info[ENGINE_IDX] = bytes(serialized_engine)
```
narendasan, line 465: Seems like we only need to deserialize in a PythonTorchTensorRTModule and we should probably use setup_engine instead. The standard interface should be like provide the serialized engine, then setup engine will set the module up properly

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -532,7 +548,10 @@ def run(
         # self.engine_cache could be None if:
```
narendasan, line 548: We are kind of avoiding this but I feel like we might want to restructure completely `run` to assume refit and check if the user wants immutable weights. Also we might want to pull the cache pulling and inserting code into helpers just to make easier to understand

narendasan: Like lines from 551-598 and 671-711 should probably be helpers that pull and insert the weight stripped engine

narendasan: Something like 

```py
if self.reuse_cached_engines:
    def _pull_cached_engine(hash, settings, inputs, 
```

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -629,35 +657,68 @@ def run(
         assert serialized_engine
```
narendasan, line 657: cant just be weight_stripped_serialized_engine now?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -629,35 +657,68 @@ def run(
...
 
         _LOGGER.info(
-            f"Build TRT engine elapsed time: {datetime.now() - build_engine_start_time}"
+            f"Build weight-stripped TRT engine elapsed time: {datetime.now() - build_engine_start_time}"
```
narendasan, line 660: Probably just leave this as is, if the user requests immutable_weights, it wouldnt apply

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -629,35 +657,68 @@ def run(
...
+            if (
+                self.engine_cache is not None
+                and self.compilation_settings.cache_built_engines
+            ):
+                # Cache the weight-stripped engine
+                if self.compilation_settings.strip_engine_weights:
+                    weight_stripped_serialized_engine = serialized_engine
+                else:
+                    # Serialize the refitted engine where the EXCLUDE_WEIGHTS flag must be cleared
+                    runtime = trt.Runtime(TRT_LOGGER)
```
narendasan, line 680: Why do we refit then strip the weights again? If refit is enabled shouldnt the builder always us a weight stripped engine?

narendasan (commented):
Make sure to add the refit settings to the `_SETTINGS_TO_BE_ENGINE_INVARIANT`

py/torch_tensorrt/dynamo/_refit.py
zewenli98: because we want the engine to clear the EXCLUDE_WEIGHTS flag. Is there a way to clear the flag without doing serialization?

py/torch_tensorrt/dynamo/_refit.py
zewenli98: > Seems like we only need to deserialize in a PythonTorchTensorRTModule

@narendasan Did you mean we only deserialize if `compiled_submodule` is a `PythonTorchTensorRTModule`? why not deserialize for `TorchTensorRTModule`?

zewenli98: modified, please take a look again. thanks

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1049,7 +1058,7 @@ def aten_ops_permute(
 
 
 def to_copy_dtype_validator(
-    placeholder_only: bool, settings: CompilationSettings = None
+    placeholder_only: bool, settings: Optional[CompilationSettings] = None
```
peri044, line 1061: why the change to optional here ?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -180,14 +182,17 @@ def compile(
...
         warnings.warn(
-            "Refit is deprecated. Please use make_refittable=True if you want to enable refitting of the engine.",
+            "`refit` is deprecated. All engines are refittable now. If you want to disable refitting, please open an issue on the Github repo with reasons.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    if "make_refittable" in kwargs.keys():
+        warnings.warn(
+            "`make_refittable` is deprecated. All engines are refittable now. If you want to disable refitting, please open an issue on the Github repo with reasons.",
```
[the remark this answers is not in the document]
zewenli98: We do provide the option but rename to `immutable_weights`.

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -282,9 +282,16 @@ def _populate_trt_builder_config(
...
-        if self.compilation_settings.make_refittable:
+        if version.parse(trt.__version__) >= version.parse("10.0"):
+            if self.compilation_settings.refit_identical_engine_weights:
+                builder_config.set_flag(trt.BuilderFlag.REFIT_IDENTICAL)
+            else:
+                builder_config.set_flag(trt.BuilderFlag.REFIT)
+        else:
             builder_config.set_flag(trt.BuilderFlag.REFIT)
 
+        builder_config.set_flag(trt.BuilderFlag.STRIP_PLAN)
```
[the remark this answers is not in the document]
zewenli98: It looks like you were reviewing an outdated commit (not submit after that review lol). To clarify the intention here, previously we planned to embed weight stripping into the compilation workflow whatever `strip_engine_weights` is because it would benefit engine caching. Afterwards, we will add weights back if users want a weighted engine. However, due to the lack of TRT support, as we discussed, it was deprecated. So for now the latest commit added the if condition.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: Lint pointed it out. I think it's because the default is None

keehyuna:
@zewenli98 / @peri044 
I could reproduce "misaligned address cuda error" with weight streaming test.
This problem started when refit build flag is add as default. When I remove trt.BuilderFlag.REFIT, test passed.
I will check further.

    def _populate_trt_builder_config(
        self,
        strict_type_constraints: bool = False,
        algorithm_selector: Optional[trt.IAlgorithmSelector] = None,
        tactic_sources: Optional[int] = None,
    ) -> trt.IBuilderConfig:
...
        if self.compilation_settings.immutable_weights:
            # non-refittable engine
            if self.compilation_settings.strip_engine_weights:
                _LOGGER.warning("strip_engine_weights will be ignored.")
        else:
            # refittable engine
            if self.compilation_settings.refit_identical_engine_weights:
                builder_config.set_flag(trt.BuilderFlag.REFIT_IDENTICAL)
            else:
                builder_config.set_flag(trt.BuilderFlag.REFIT)          <----- refit flag is set as default

zewenli98:
Thanks @keehyuna! Does this error happen with trt.BuilderFlag.REFIT_IDENTICAL flag?

keehyuna:
> Thanks @keehyuna! Does this error happen with trt.BuilderFlag.REFIT_IDENTICAL flag?

This is no problem when trt.BuilderFlag.REFIT_IDENTICAL is used. It was tested on main branch.

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -193,14 +193,17 @@ def cross_compile_for_windows(
...
         warnings.warn(
-            "Refit is deprecated. Please use make_refittable=True if you want to enable refitting of the engine.",
+            "`refit` is deprecated. Engines are refittable by default. Please set immutable_weights=True to build a non-refittable engine whose weights will be fixed.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    if "make_refittable" in kwargs.keys():
+        warnings.warn(
+            "`make_refittable` is deprecated. Engines are refittable by default. Please set immutable_weights=True to build a non-refittable engine whose weights will be fixed.",
```
narendasan, line 203: where do we translate the old apis to immutable_weights?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -468,14 +475,17 @@ def compile(
 
     if "refit" in kwargs.keys():
         warnings.warn(
-            "Refit is deprecated. Please use make_refittable=True if you want to enable refitting of the engine.",
+            "`refit` is deprecated. Engines are refittable by default. Please set immutable_weights=True to build a non-refittable engine whose weights will be fixed.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    if "make_refittable" in kwargs.keys():
```
narendasan, line 483: same here

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -287,8 +287,19 @@ def _populate_trt_builder_config(
         if self.compilation_settings.disable_tf32:
             builder_config.clear_flag(trt.BuilderFlag.TF32)
 
-        if self.compilation_settings.make_refittable:
-            builder_config.set_flag(trt.BuilderFlag.REFIT)
+        if self.compilation_settings.immutable_weights:
+            # non-refittable engine
+            if self.compilation_settings.strip_engine_weights:
```
narendasan, line 292: what about refit_identical

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3634,3 +3651,27 @@ def aten_ops_full(
         fill_value=args[1],
         dtype=kwargs.get("dtype", None),
     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.chunk.default)
```
peri044, line 3656: support for dynamic shapes ? Is chunk a core aten op ?

tests/py/dynamo/models/test_weight_stripped_engine.py
```diff
@@ -0,0 +1,573 @@
...
+            exp_program,
+            example_inputs,
+            **settings,
+        )
+        gm1_output = gm1(*example_inputs)
+
+        # 2. Compile with torch_trt.compile using dynamo backend
+        gm2 = torch_trt.compile(
+            pyt_model, ir="dynamo", inputs=example_inputs, **settings
+        )
```
peri044, line 44: both of these are same

tests/py/dynamo/models/test_weight_stripped_engine.py
```diff
@@ -0,0 +1,573 @@
...
```
peri044, line 94: same as above comment

tests/py/dynamo/models/test_weight_stripped_engine.py
```diff
@@ -0,0 +1,573 @@
...
+
+        trt_gm = torch_trt.dynamo.compile(
+            exp_program,
+            tuple(example_inputs),
+            use_python_runtime=True,
+            enabled_precisions={torch.float},
+            debug=False,
+            min_block_size=1,
+            strip_engine_weights=False,  # engine cache will save the stripped engine even if this is False
```
peri044, line 229: This comment conflicts with the 
```
@unittest.skip(
        "For now, torch-trt will save weighted engine if strip_engine_weights is False. In the near future, we plan to save weight-stripped engine regardless of strip_engine_weights, which is pending on TRT's feature development: NVBug #4914602"
    )
```

py/torch_tensorrt/_enums.py
```diff
@@ -220,7 +220,7 @@ def _from(
                 return dtype.f32
             elif t == np.float64:
                 return dtype.f64
-            elif t == np.bool:
+            elif t == np.bool_:
```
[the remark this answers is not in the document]
zewenli98: np.bool has been deprecated since NumPy 1.20.0 in favor of np.bool_, so maybe np.bool_ is enough?

tests/py/dynamo/models/test_weight_stripped_engine.py
zewenli98: Do you mean we only need to keep one of these two?

peri044: yes

peri044 (approved):
LGTM

py/ci/Dockerfile.ci
```diff
@@ -1,4 +1,4 @@
-FROM pytorch/manylinux-builder:cuda12.4
+FROM pytorch/manylinux2_28-builder:cuda12.4
```
narendasan, line 1: Bump this to 12.6

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -193,14 +197,44 @@ def cross_compile_for_windows(
...
+            raise ValueError(
+                "Use flag `immutable_weights` only. Flag `refit` is deprecated."
+            )
         else:
-            make_refittable = kwargs["refit"]
+            immutable_weights = not kwargs["refit"]
+
+    if "make_refittable" in kwargs.keys():
+        warnings.warn(
+            "`make_refittable` is deprecated. Please set `immutable_weights=True` to build a non-refittable engine whose weights will be fixed.",
```
narendasan, line 213: Shouldnt it be `immutable_weights=False` since this should only exist if a user does `make_refittable=True`?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -952,12 +1025,48 @@ def convert_exported_program_to_serialized_trt_engine(
...
+        if immutable_weights:
+            raise ValueError(
+                "Use flag `immutable_weights` only. Flag `refit` is deprecated."
+            )
+        else:
+            immutable_weights = not kwargs["refit"]
+
+    if "make_refittable" in kwargs.keys():
+        warnings.warn(
+            "`make_refittable` is deprecated. Please set `immutable_weights=True` to build a non-refittable engine whose weights will be fixed.",
```
narendasan, line 1044: Same here

packaging/env_vars.txt
```diff
@@ -1,2 +1,3 @@
 export CI_BUILD="1"
-export RELEASE="1"
\ No newline at end of file
+export RELEASE="1"
+export USE_CXX11_ABI="1"
```
narendasan, line 3: This needs to be conditioned on the cuda version right?

narendasan (approved):
@zewenli98 can you just pull the CI changes into their own branch? Those seem ready now

narendasan (changes requested):
sorry did not mean to approve

setup.py
```diff
@@ -141,7 +141,9 @@ def load_dep_info():
     CXX11_ABI = True
 
 if (cxx11_abi_env_var := os.environ.get("USE_CXX11_ABI")) is not None:
-    if cxx11_abi_env_var == "1":
+    if (
```
narendasan, line 144: not sure if this is the right way to do this.

narendasan: Likely this should be handled outside the setup.py so that behavior isnt being masked

narendasan (approved):
LGTM

PR #3181, opened 2024-09-26
Adjust cpp torch trt logging level with compiler option

keehyuna:
# Description

Log level of python and native TRT logger is set differently. Same level from compiler option should be helpful to work with development version of torch-TRT module.
- python trt logger: when compiler option provide debug=True
- c++ trt logger: compiled at build time

> TorchTRTLogger.cpp
#ifndef NDEBUG
  static TorchTRTLogger global_logger("[Torch-TensorRT - Debug Build] - ", LogLevel::kDEBUG, true);
#else
  static TorchTRTLogger global_logger("[Torch-TensorRT] - ", LogLevel::kWARNING, false);
#endif

And profile option for c++ trt runtime module is enabled by debug level logging. This is helpful to measure/evaluate
performance with development version of torch-TRT module

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/runtime/TRTEngine.cpp
+++ b/core/runtime/TRTEngine.cpp
@@ -201,10 +201,9 @@ TRTEngine::TRTEngine(
     }
     num_io = std::make_pair(inputs_size, outputs);
   }
-
-#ifndef NDEBUG
-  this->enable_profiling();
-#endif
+  if (util::logging::get_logger().get_reportable_log_level() == util::logging::LogLevel::kDEBUG) {
+    this->enable_profiling();
+  }
   LOG_DEBUG(*this);
 }
 
--- a/core/runtime/TRTEngine.cpp
+++ b/core/runtime/TRTEngine.cpp
@@ -201,10 +201,9 @@ TRTEngine::TRTEngine(
     }
     num_io = std::make_pair(inputs_size, outputs);
   }
+
 #ifndef NDEBUG
-  if (util::logging::get_logger().get_reportable_log_level() == util::logging::LogLevel::kDEBUG) {
-    this->enable_profiling();
-  }
+  this->enable_profiling();
 #endif
   LOG_DEBUG(*this);
 }
--- a/core/runtime/TRTEngine.cpp
+++ b/core/runtime/TRTEngine.cpp
@@ -201,9 +201,11 @@ TRTEngine::TRTEngine(
     }
     num_io = std::make_pair(inputs_size, outputs);
   }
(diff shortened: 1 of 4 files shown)


core/util/logging/TorchTRTLogger.cpp
```diff
@@ -125,7 +125,7 @@ namespace {
 
 TorchTRTLogger& get_global_logger() {
 #ifndef NDEBUG
-  static TorchTRTLogger global_logger("[Torch-TensorRT - Debug Build] - ", LogLevel::kDEBUG, true);
+  static TorchTRTLogger global_logger("[Torch-TensorRT - Debug Build] - ", LogLevel::kINFO, true);
```
narendasan, line 128: I think this should still be debug

core/runtime/TRTEngine.cpp
```diff
@@ -201,9 +201,10 @@ TRTEngine::TRTEngine(
     }
     num_io = std::make_pair(inputs_size, outputs);
   }
-
 #ifndef NDEBUG
-  this->enable_profiling();
+  if (util::logging::get_logger().get_reportable_log_level() == util::logging::LogLevel::kDEBUG) {
```
narendasan, line 205: Could just remove this if you feel its not necessary

keehyuna: Ok, I removed

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -204,10 +204,26 @@ def set_log_level(parent_logger: Any, level: Any) -> None:
...
+    elif level == logging.WARNING:
+        log_level = trt.ILogger.Severity.WARNING
+    elif level == logging.ERROR:
+        log_level = trt.ILogger.Severity.ERROR
+    elif level == logging.CRITICAL:
+        log_level = trt.ILogger.Severity.INTERNAL_ERROR
+    else:
+        raise AssertionError(f"{level} is not valid log level")
+
+    torch.ops.tensorrt.set_logging_level(int(log_level))
```
narendasan, line 225: What happens in the case that it is a python only build? You might want to use enabled features to gate access to this code to only when the C++ runtime is available.

keehyuna: You're right, I see the problem. Updated the change.
```
AttributeError: '_OpNamespace' 'tensorrt' object has no attribute 'set_logging_level'
```

narendasan (approved):
LGTM

PR #3191, opened 2024-09-27
Fix for warning as default stream was used in enqueueV3

### LINKED ISSUE #3190: 🐛 [Bug] Warning as default stream was used in enqueueV3()
##  Bug Description

Some time below warning is seen while running the model.

WARNING: [Torch-TensorRT - Debug Build] - Using default stream in enqueueV3() may lead to performance issues due to additional calls to cudaStreamSynchronize() by TensorRT to ensure correct synchronization. Please use non-default stream instead.

## To Reproduce
It was reproduced with resnet model with multiple interference call. Both use_python_runtime=False/True have issue

```
model = models.resnet18(pretrained=True).eval().to("cuda")
input = torch.randn((1, 3, 224, 224)).to("cuda")
compile_spec = {
    "inputs": [
        torchtrt.Input(
            input.shape, dtype=torch.float, format=torch.contiguous_format
        )
    ],
    "device": torchtrt.Device("cuda:0"),
    "ir": "dynamo",
    "cache_built_engines": False,
    "reuse_cached_engines": False,
    "use_python_runtime": True,
}

trt_mod = torchtrt.compile(model, **compile_spec)
for i in range(5):
    trt_mod(input)
# Clean up model env
torch._dynamo.reset()
```

## Expected behavior

No warning message

## Environment

 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - Python version:
 - CUDA version:

keehyuna:
# Description

torch.cuda.current_stream()/c10::cuda::getCurrentCUDAStream() always returns default stream and it leads running enqueueV3() with default stream.
torch.cuda.set_stream/c10::cuda::setCurrentCUDAStream is required to set current stream when new stream is acquired from pool

Fixes #3190

## Type of change

# Checklist:

### DIFF
--- a/core/runtime/execute_engine.cpp
+++ b/core/runtime/execute_engine.cpp
@@ -305,9 +305,6 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
   if (compiled_engine->engine_stream == c10::cuda::getDefaultCUDAStream(current_device_id)) {
     // Create a new stream if the engine stream is the default stream
     compiled_engine->engine_stream = c10::cuda::getStreamFromPool(false, current_device_id);
-    c10::cuda::setCurrentCUDAStream(compiled_engine->engine_stream);
-  } else {
-    compiled_engine->engine_stream = compiled_engine->caller_stream;
   }
 
   // nvinfer1::IExecutionContext::enqueue is not thread safe and we need a mutex for it.
@@ -334,7 +331,7 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
       if (need_cudagraphs_record) {
         // If cudagraphs needs to record a graph, capture the enqueueV3 call in a graph
         c10::cuda::CUDAStream recording_stream = compiled_engine->engine_stream;
-        compiled_engine->cudagraph.capture_begin(compiled_engine->cudagraph_mempool_id);
+        compiled_engine->cudagraph.capture_begin();
         compiled_engine->exec_ctx->enqueueV3(recording_stream);
         compiled_engine->cudagraph.capture_end();
 
--- a/core/runtime/execute_engine.cpp
+++ b/core/runtime/execute_engine.cpp
@@ -305,6 +305,7 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
   if (compiled_engine->engine_stream == c10::cuda::getDefaultCUDAStream(current_device_id)) {
     // Create a new stream if the engine stream is the default stream
     compiled_engine->engine_stream = c10::cuda::getStreamFromPool(false, current_device_id);
+    c10::cuda::setCurrentCUDAStream(compiled_engine->engine_stream);
   } else {
     compiled_engine->engine_stream = compiled_engine->caller_stream;
   }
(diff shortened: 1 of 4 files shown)


narendasan:
@keehyuna there is some code related to cudagraphs, can you check how it handles streams and perhaps write some docs on how the runtime is suppose to do streams in all cases?

peri044:
- Add a doc (runtime) which explains how streams work in our runtime. (2nd part of this task)

core/runtime/execute_engine.cpp
```diff
@@ -333,7 +331,7 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
       if (need_cudagraphs_record) {
         // If cudagraphs needs to record a graph, capture the enqueueV3 call in a graph
         c10::cuda::CUDAStream recording_stream = compiled_engine->engine_stream;
-        compiled_engine->cudagraph.capture_begin(compiled_engine->cudagraph_mempool_id);
```
keehyuna, line 336: This was reverted to fix below assert() from torch. We don't share memory across captures, I think we can use internally created pool.

https://pytorch.org/docs/stable/notes/cuda.html#graph-memory-management

```
  File "/root/trt/TensorRT/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py", line 274, in forward
    outputs: List[torch.Tensor] = torch.ops.tensorrt.execute_engine(
  File "/root/.pyenv/versions/3.10.14/lib/python3.10/site-packages/torch/_ops.py", line 1113, in __call__
    return self._op(*args, **(kwargs or {}))
RuntimeError: it->second->use_count > 0 INTERNAL ASSERT FAILED at "../c10/cuda/CUDACachingAllocator.cpp":2056, please report a bug to PyTorch
```

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -464,7 +461,8 @@ def cudagraphs_validate_shapes(self, inputs: Sequence[torch.Tensor]) -> bool:
         if new_shape_key != self.shape_key:
             logger.debug(f"Resetting Cudagraph on new shape key {new_shape_key}")
             self.shape_key = new_shape_key
-            self.cudagraph.reset()  # type: ignore
+            if self.cudagraph:
+                self.cudagraph.reset()
```
keehyuna, line 465: self.cudagraph can be None when torch_compile backend is used.

self.cudagraph is initialized when cudagraphs mode is enabled. But this init was called at compile()
https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py#L144-L145

keehyuna (commented):
This is when problem happend. capture stream is changed to default and non default stream.
image

This is when torch.cuda.set_stream() is used. Non default stream is used for cuda graph/enqueueV3(). But stream is not restored after Forward()
image.

This is proposed fix to keep side stream to cuda graph or enqueueV3()
image

narendasan (approved):
Good catch LGTM

narendasan:
@lanluo-nvidia please cherry-pick

PR #3193, opened 2024-09-27
Require full compilation arg

apbose:


py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -107,6 +107,8 @@ def _pretraced_backend(
             torchtrt_inputs = prepare_inputs(
                 torch_inputs, disable_memory_format_check=True
             )
+            if settings.require_full_compilation:
+                raise AssertionError("this argument is only applicable for ir=dynamo")
```
narendasan, line 111: I think this isnt very clear since the backend will get called from torch.compile. You could instead throw a warning that `requires_full_compilation` does nothing for the tensorrt backend for torch.compile

apbose: I changed this to a warning. But should there be some additional logic to make this argument invalid for the backend = "torch_tensorrt" since it would still be passed to the same partitioning code, and error out in case of torch.compile if there is a graph break

py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -107,6 +107,10 @@ def _pretraced_backend(
             torchtrt_inputs = prepare_inputs(
                 torch_inputs, disable_memory_format_check=True
             )
+            if settings.require_full_compilation:
+                logger.warning(
+                    "req_full_compilation arg is not applicable for torch.compile with backend='torch_tensorrt"
```
narendasan, line 112: Use the actual name of the argument `require_full_compilation`

PR #3212, opened 2024-10-06
skip run_shape_analysis

lanluo-nvidia:
# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_exporter.py
+++ b/py/torch_tensorrt/dynamo/_exporter.py
@@ -16,7 +16,6 @@ from torch.export.exported_program import (
     OutputSpec,
     TensorArgument,
 )
-from torch_tensorrt.dynamo import partitioning
 
 
 def export(
@@ -58,11 +57,9 @@ def transform(
     if kwarg_inputs is None:
         kwarg_inputs = {}
     gm = copy.deepcopy(gm)
-    # Run shape analysis
-    _, outputs_map = partitioning.run_shape_analysis(gm, inputs, kwarg_inputs)
 
     # Inline TensorRT submodules
-    inline_trt_modules(gm, outputs_map)
+    inline_trt_modules(gm)
 
     # Inline pytorch submodules
     inline_torch_modules(gm)
@@ -361,9 +358,7 @@ def create_trt_exp_program(
     return trt_exp_program
 
 
-def inline_trt_modules(
-    gm: torch.fx.GraphModule, outputs_map: Dict[Any, Sequence[Any]]
-) -> torch.fx.GraphModule:
+def inline_trt_modules(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
     """
     Replace TRT submodules with trt engine nodes.
     """
@@ -379,7 +374,7 @@ def inline_trt_modules(
         trt_module_node = trt_module_node[0]
         assert trt_module_node.args
 
-        num_outputs = len(outputs_map[trt_module_node.name])
+        num_outputs = len(trt_module.output_shapes)
         # Insert a call_function node to perform inference on TRT engine
         with gm.graph.inserting_before(trt_module_node):
             engine_name = f"{name}_engine"
@@ -398,8 +393,8 @@ def inline_trt_modules(
                     cast(
                         FakeTensor,
                         torch.empty_strided(
-                            tuple(outputs_map[trt_module_node.name][idx]),
-                            tuple([1] * len(outputs_map[trt_module_node.name][idx])),
+                            tuple(trt_module.output_shapes[idx]),
+                            tuple([1] * len(trt_module.output_shapes[idx])),
                         ),
                     )
                 )
--- a/py/torch_tensorrt/dynamo/conversion/_conversion.py
+++ b/py/torch_tensorrt/dynamo/conversion/_conversion.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
 import logging
-from typing import Any, List, Optional, Sequence
+from typing import Any, List, Optional, Sequence, Tuple
 
+import tensorrt as trt
 import torch
 from torch.fx.experimental.proxy_tensor import unset_fake_temporarily
 from torch_tensorrt._Device import Device
@@ -16,11 +17,85 @@ from torch_tensorrt.dynamo.conversion._TRTInterpreter import (
     TRTInterpreterResult,
 )
 from torch_tensorrt.dynamo.runtime import PythonTorchTensorRTModule, TorchTensorRTModule
-from torch_tensorrt.dynamo.utils import get_model_device, get_torch_inputs
-
-import tensorrt as trt
+from torch_tensorrt.dynamo.utils import (
+    get_model_device,
+    get_torch_inputs,
+    unwrap_tensor_shape,
+)
 
 logger = logging.getLogger(__name__)
+
+
+def get_interpret_result(
+    module: torch.fx.GraphModule,
+    inputs: Sequence[Input],
+    output_dtypes: Sequence[dtype],
+    settings: CompilationSettings = CompilationSettings(),
+    engine_cache: Optional[BaseEngineCache] = None,
+) -> TRTInterpreterResult:
+    interpreter = TRTInterpreter(
+        module,
+        inputs,
+        logger_level=(trt.Logger.VERBOSE if settings.debug else trt.Logger.WARNING),
+        output_dtypes=output_dtypes,
+        compilation_settings=settings,
+        engine_cache=engine_cache,
+    )
+
+    interpreter_result = interpreter.run()
+    return interpreter_result
+
+
+def infer_module_output_shapes_dtypes(
+    module: torch.fx.GraphModule,
+    inputs: Sequence[Input],
+    device: Device,
+    kwarg_inputs: Optional[dict[str, Any]] = None,
+    truncate_double: bool = False,
+) -> Tuple[List[Tuple[int]], List[dtype]]:
+    """
+    This function performs model inference to determine the output dtypes
+    and truncates them accordingly. inputs can be either arg_inputs or flattened input list.
+    If it is flattened list, kwarg_inputs should be None, as it is already included in the flattened input.
+    """
+    # TODO: We can also determine output dtypes from the module.graph based on node metadata.
+    # However, our converter tests use fx.symbolic_trace which sometimes does not provide metadata,
+    # so we stick to the model inference approach currently.
+    with unset_fake_temporarily():
+        # Get the device on which the model exists
+        # For large models, this can be done on CPU to save GPU memory allocation for TRT.
+        device = get_model_device(module)
+        torch_inputs = get_torch_inputs(inputs, device)
+        if kwarg_inputs is None:
+            kwarg_inputs = {}
+        torch_kwarg_inputs = get_torch_inputs(kwarg_inputs, device)
+        module_outputs = module(*torch_inputs, **torch_kwarg_inputs)
+        if not isinstance(module_outputs, (list, tuple)):
+            module_outputs = [module_outputs]
+
+    # Int64 outputs can sometimes be generated from within other operators
+    # such as aten.sum - such outputs can be truncated
+    output_dtypes = []
+    output_shapes = []
+    for output in module_outputs:
+        output_ = output
+        # We don't need to check if output is nested here because the input module will be flattened
+        if not isinstance(output, torch.Tensor):
+            if isinstance(output, str):
+                raise ValueError(
+                    f"Received an output type {type(output)} that's not in the acceptable datatypes (https://pytorch.org/docs/stable/tensor_attributes.html#torch.dtype)"
+                )
+            else:
+                output_ = torch.tensor(output)
+
+        output_shapes.append(unwrap_tensor_shape(output_))
+        # output_shapes.append(output_.shape)
+        if truncate_double and output_.dtype == dtype.float64:
+            output_dtypes.append(dtype.float32)
+        else:
+            output_dtypes.append(dtype._from(output_.dtype))
+
+    return output_shapes, output_dtypes
 
 
 def infer_module_output_dtypes(
@@ -139,8 +214,16 @@ def convert_module(
     Returns:
         PythonTorchTensorRTModule or TorchTensorRTModule
     """
-    interpreter_result = interpret_module_to_result(
-        module, inputs, settings, engine_cache=engine_cache
+
+    output_shapes, output_dtypes = infer_module_output_shapes_dtypes(
+        module,
+        inputs,
+        settings.device,
+        truncate_double=settings.truncate_double,
+    )
+
+    interpreter_result = get_interpret_result(
+        module, inputs, output_dtypes, settings, engine_cache=engine_cache
     )
 
     rt_cls = PythonTorchTensorRTModule
@@ -163,6 +246,7 @@ def convert_module(
         serialized_engine=interpreter_result.serialized_engine,
         input_binding_names=list(interpreter_result.input_names),
         output_binding_names=list(interpreter_result.output_names),
+        output_shapes=output_shapes,
         name=name,
         settings=settings,
         weight_name_map=interpreter_result.weight_name_map,
--- a/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
+++ b/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
@@ -75,6 +75,7 @@ class TorchTensorRTModule(torch.nn.Module):  # type: ignore[misc]
         serialized_engine: Optional[bytes] = None,
         input_binding_names: Optional[List[str]] = None,
         output_binding_names: Optional[List[str]] = None,
+        output_shapes: Optional[List[Tuple[int]]] = None,
         *,
         name: str = "",
         settings: CompilationSettings = CompilationSettings(),  # Assumes engine was built with default compilation settings if object not passed
@@ -125,6 +126,7 @@ class TorchTensorRTModule(torch.nn.Module):  # type: ignore[misc]
         self.output_binding_names = (
             output_binding_names if output_binding_names is not None else []
         )
+        self.output_shapes = output_shapes
         self.name = name
         self.hardware_compatible = settings.hardware_compatible
         self.settings = copy.deepcopy(settings)


py/torch_tensorrt/_compile.py
```diff
@@ -502,19 +502,24 @@ def save(
                 "Provided model is a torch.jit.ScriptModule but the output_format specified is exported_program. Please verify the output_format"
             )
         else:
+            if arg_inputs is not None:
+                raise ValueError(
+                    "Provided model is a torch.jit.ScriptModule, do not allow user to provide inputs or arg_inputs."
```
narendasan, line 507: This needs to be rephased to be more user facing

e.g "Provided model is a torch.jit.ScriptModule, arg_inputs is not necessary"

Also is this a real error or can it just be a warning?

py/torch_tensorrt/_compile.py
```diff
@@ -502,19 +502,24 @@ def save(
...
+                )
             torch.jit.save(module, file_path)
     elif module_type == _ModuleType.ep:
         if output_format == "torchscript":
             raise ValueError(
                 "Provided model is a torch.export.ExportedProgram but the output_format specified is torchscript. Please verify the output_format"
             )
         else:
+            if arg_inputs is not None:
```
narendasan, line 516: Same here

py/torch_tensorrt/_compile.py
```diff
@@ -525,11 +530,19 @@ def save(
             if not retrace:
                 from torch_tensorrt.dynamo._exporter import export
 
-                exp_program = export(module, arg_inputs, kwarg_inputs)
+                if arg_inputs is not None:
```
narendasan, line 533: same here

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -17,20 +17,24 @@
...
 
-def infer_module_output_dtypes(
+def infer_module_output_shapes_dtypes(
     module: torch.fx.GraphModule,
     inputs: Sequence[Input],
     device: Device,
     kwarg_inputs: Optional[dict[str, Any]] = None,
     truncate_double: bool = False,
-) -> List[dtype]:
+) -> Tuple[List[Tuple[int]], List[dtype]]:
```
narendasan, line 35: This probably should be some sort of struct very similar to Input, maybe we can use a fake tensor or something?

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -78,8 +84,8 @@ def interpret_module_to_result(
     arg_inputs: Optional[Sequence[Input]] = None,
     kwarg_inputs: Optional[dict[str, Any]] = None,
     engine_cache: Optional[BaseEngineCache] = None,
-) -> TRTInterpreterResult:
-    """Interpret an FX module to a TRTInterpreterResult
+) -> Tuple[List[Tuple[int]], TRTInterpreterResult]:
```
narendasan, line 87: TRTInterpreterResult first

py/torch_tensorrt/dynamo/conversion/_conversion.py
narendasan: @peri044 can we replace this with graph inspection?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -36,6 +35,7 @@ def __init__(
         serialized_engine: Optional[bytes] = None,
         input_binding_names: Optional[List[str]] = None,
         output_binding_names: Optional[List[str]] = None,
+        output_shapes: Optional[List[Tuple[int]]] = None,
```
narendasan, line 38: Why is this needed? shouldnt the engine have the shapes?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -125,6 +126,7 @@ def __init__(
         self.output_binding_names = (
             output_binding_names if output_binding_names is not None else []
         )
+        self.output_shapes = output_shapes
```
narendasan, line 129: Same thing here?

narendasan: This can be a function on the engine that pulls the data from TRT

py/torch_tensorrt/_compile.py
lanluo-nvidia: changed to warning.
add the warning is to let user know: 
if user provide input1 for torch_tensorrt.dynamo.compile(*****)
and then user call torch_tensorrt.save(****) with input2
since the torch_tensorrt model is compiled with input1, when save it is not going to use the input2

py/torch_tensorrt/_compile.py
lanluo-nvidia: changed to warning instead of raising error

py/torch_tensorrt/_compile.py
lanluo-nvidia: changed to warning instead of raising error

py/torch_tensorrt/dynamo/conversion/_conversion.py
lanluo-nvidia: modified as suggested

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
lanluo-nvidia: during compile stage, we have the output shapes information, however this information is not saved to the TorchTensorRTModule, when user call torch_tensorrt.save() we run_shape_analysis() again to get the output shapes, which is not only unnecessary, also run_shape_analysis requires the Engine which won't work for the cross runtime compilation feature.
Hence @dheerajperi  suggested that I should skip run_shape_analysis in the torch_tensorrt.save()
instead save the output shaped in compile stage.

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
lanluo-nvidia: do you mean I should add:
```
    def get_output_shapes(self) -> List[Tuple[int]]:
        return self.output_shapes
    
    def set_output_shapes(self, output_shapes: List[Tuple[int]]) -> None:
        self.output_shapes = output_shapes
```

Instead of using trt_module.output_shapes 
I should use:  trt_module.get_output_shapes()

peri044:
Here's what I think could be a simpler way of doing this 

1) We probably don't have to store output_shapes in `TorchTensorRTModule` class. Once the compilation is finished, verify if the nodes of the TRT graph modules have metadata in them (if not we can update it by node.meta["val"] = original metadata) 
Reference: https://github.com/pytorch/TensorRT/blob/3eb48d786d403b12bd3700004c60e08c5c002f7b/py/torch_tensorrt/dynamo/_compiler.py#L496-L499

Here the node corresponding to `_run_on_acc0` can be queried as 
```py
trt_module_node = [node for node in gm.graph.nodes if node.name == "_run_on_acc0"]
trt_module_node.meta["val"] - This should already have fake tensors which need to be used in the exporter. 
```

2) exporter
We have the TRT module node here : https://github.com/pytorch/TensorRT/blob/3eb48d786d403b12bd3700004c60e08c5c002f7b/py/torch_tensorrt/dynamo/_exporter.py#L364
We could directly set (ensuring trt_module_node.meta["val"] always exists)
```py
trt_node.meta["val"] = trt_module_node.meta["val"]
```

3) infer_module_types
We can replace the dummy inference with graph inspection by reading output metadata. The output of this function could be a list of FakeTensors and we can extract the dtypes from this to pass it to TRTInterpreter. 

Replacing the dummy inference will also need changes to our converter test suite.

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
lanluo-nvidia: no longer required. now all the output shape and dtype info is in the graph node.meta['val']

py/torch_tensorrt/_compile.py
```diff
@@ -502,19 +502,24 @@ def save(
                 "Provided model is a torch.jit.ScriptModule but the output_format specified is exported_program. Please verify the output_format"
             )
         else:
+            if arg_inputs is not None:
```
peri044, line 505: do you also need to check if inputs is not None ?

py/torch_tensorrt/_compile.py
```diff
@@ -502,19 +502,24 @@ def save(
...
+                    "Provided model is a torch.jit.ScriptModule, inputs or arg_inputs is not necessary during save."
+                )
             torch.jit.save(module, file_path)
     elif module_type == _ModuleType.ep:
         if output_format == "torchscript":
             raise ValueError(
                 "Provided model is a torch.export.ExportedProgram but the output_format specified is torchscript. Please verify the output_format"
             )
         else:
+            if arg_inputs is not None:
```
peri044, line 516: same as above

py/torch_tensorrt/_compile.py
peri044, line 533: same as above

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -379,7 +366,11 @@ def inline_trt_modules(
         trt_module_node = trt_module_node[0]
         assert trt_module_node.args
 
-        num_outputs = len(outputs_map[trt_module_node.name])
+        if "val" not in trt_module_node.meta:
+            raise ValueError(
+                f"trt_module_node: {trt_module_node.name} does not have the meta['val'] info, it should be set during dynamo compile_module step."
```
peri044, line 371: Consider using "trt_module_node: {trt_module_node.name} does not have the metadata which should be set during compile_module step"

py/torch_tensorrt/dynamo/_tracer.py
```diff
@@ -115,6 +115,9 @@ def get_dynamic_shapes_args(mod: torch.nn.Module, inputs: Any) -> dict[str, Any]
     args = list(signature(mod.forward).parameters.keys())
     dynamic_shapes = {}
     for input, input_name in zip(inputs, args[: len(inputs)]):
+        # if input.name is not None, also not empty str, use the input.name
+        if input.name is not None and len(input.name) > 0 and input.name != input_name:
```
peri044, line 119: what is happening here ?

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -17,12 +17,32 @@
...
+    return get_output_dtypes(outputs, truncate_double)
+
+
+# this method is only used in our converter test to infer the module output dtypes via dummy inference
+# which is due to fx.symbolic_trace does not have the meta['val'] info in the node
+# TODO: lan to remove this once our converter test is moved from fx.symbolic_trace to dynamo trace
+def infer_module_output_dtypes_for_test(
     module: torch.fx.GraphModule,
     inputs: Sequence[Input],
     device: Device,
```
peri044, line 48: move this to test (testing_utilities.py or harness.py)

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -661,3 +661,57 @@ def get_flat_args_with_check(
...
+    output_dtypes = []
+    if isinstance(output, torch.fx.node.Node):
+        if "val" in output.meta:
+            output_meta = output.meta["val"]
+            if isinstance(output_meta, (FakeTensor, torch.Tensor)):
+                if truncate_doulbe and output_meta.dtype == torch.float64:
+                    output_dtypes.append(dtype.float32)
+                else:
+                    output_dtypes.append(dtype._from(output_meta.dtype))
+        else:
```
peri044, line 706: also check "tensor_meta" field in output.meta

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -661,3 +661,57 @@ def get_flat_args_with_check(
...
+        if "val" not in output.meta:
+            output.meta["val"] = outputs_meta_val[0]
+        outputs_meta_val.pop(0)
+    elif isinstance(output, (tuple, list)):
+        for node in output:
+            set_output_meta_val(node, outputs_meta_val)
+    else:
+        raise ValueError(
+            f"expect torch.fx.node.Node or a tuple/list of torch.fx.node.Node type, got unexpected types: {type(output)=}"
+        )
```
peri044, line 693: We already have set_metadata and get_metadata functions in pass_utils.py. Consider merging them and using them across lowering pass for consistency

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -445,6 +451,19 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
...
+                f"node_name: {name} does not exist in the submodule node dictionary"
+            )
+
+        # set the submodule meta val back to the parent trt_module_node
+        if "val" not in submodule_node_dict[name].meta:
+            outputs = [node for node in submodule.graph.nodes if node.op == "output"]
+            outputs = outputs[0].args
+            outputs_meta_val = get_output_meta_val(outputs)
+            assert len(outputs_meta_val) > 0
+            submodule_node_dict[name].meta["val"] = outputs_meta_val
```
peri044, line 465: consider wrapping this into a helper function and  moving this into utils.py

peri044 (commented):
Besides checking `val` field in node.meta, also check `tensor_meta` if `val` is not present. Apparently, torch.compile stores metadata in `tensor_meta` while export uses `val` fields in node.meta

py/torch_tensorrt/_compile.py
lanluo-nvidia: It has the logic below, only need to check arg_inputs:
```
if arg_inputs is not None and not all(
        isinstance(input, torch.Tensor) for input in arg_inputs
    ):
        raise ValueError(
            "Not all inputs provided are torch.tensors. Please provide torch.tensors as inputs"
        )
    if arg_inputs and inputs:
        raise AssertionError(
        )

    arg_inputs = inputs or arg_inputs
```

py/torch_tensorrt/_compile.py
lanluo-nvidia: It has the logic below, so only need to check arg_inputs:
```
if arg_inputs is not None and not all(
        isinstance(input, torch.Tensor) for input in arg_inputs
    ):
        raise ValueError(
            "Not all inputs provided are torch.tensors. Please provide torch.tensors as inputs"
        )
    if arg_inputs and inputs:
        raise AssertionError(
        )

    arg_inputs = inputs or arg_inputs
```

py/torch_tensorrt/_compile.py
lanluo-nvidia: It has the logic below, so only need to check arg_inputs:
```
if arg_inputs is not None and not all(
        isinstance(input, torch.Tensor) for input in arg_inputs
    ):
        raise ValueError(
            "Not all inputs provided are torch.tensors. Please provide torch.tensors as inputs"
        )
    if arg_inputs and inputs:
        raise AssertionError(
        )

    arg_inputs = inputs or arg_inputs
```

py/torch_tensorrt/dynamo/_exporter.py
lanluo-nvidia: modified as suggested

py/torch_tensorrt/dynamo/_tracer.py
lanluo-nvidia: This is to handle the scenario, in case the forward signure paramater sequence is x, y,
And our InputSpec sequence is y, x, we honor the sequence user defined in the InputSpec:
```
class logical_or(nn.Module):
    def forward(self, x, y):
        return torch.ops.aten.logical_or.default(x, y)
input_specs = [
    Input(
        min_shape=(3,),
        opt_shape=(4,),
        max_shape=(6,),
        name="y",
    ),
    Input(
        min_shape=(2, 1),
        opt_shape=(3, 1),
        max_shape=(5, 1),
        name="x",
    ),
]

```

py/torch_tensorrt/dynamo/utils.py
lanluo-nvidia: added as suggested

py/torch_tensorrt/dynamo/_tracer.py
narendasan: I feel like we need to do spec matching via the pytree flattening scheme not names, it is not ergonomic if every input spec needs to be named

narendasan: So the idea is if I have a function input `Tuple[Tuple[Any], Dict[str, Any]]`, the spec collection `Tuple[Tuple[Input], Dict[str, Input]]` made from `arg_inputs` and `kwargs_inputs` and the forward function should flatten to the same ordering using pytorch's pytree system. See refit for an example

peri044: I think in the above example, we should let Pytorch handle wrong ordering and we shouldn't be re-ordering for users (since we don't expect them to provide names). We could throw a warning instead if we detect incorrect ordering.
In PyTorch, there are two ways to provide dynamic shapes for the above example
```py
ep = torch.export.export(gm, (x, y), dynamic_shapes = {"x": {0: Dim(min=2, max=5)}, "y": {0: Dim(min=3, max=6)} }

# (or)
ep = torch.export.export(gm, (x, y), dynamic_shapes = ({0: Dim(min=2, max=5)},  {0: Dim(min=3, max=6)},)
```
In the second option, the names are not considered but the order is assumed to be correct (x, y) and PyTorch will throw an error (or pass) if users provide wrong order (y, x) based on the model. I believe our behavior should be the same as well.

peri044 (commented):
Added minor comments. Mostly LGTM

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -452,6 +458,33 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
...
+        if "tensor_meta" not in submodule_node_dict[name].meta:
+            tensor_meta_list = [
+                metadata["tensor_meta"]
+                for metadata in metadata_list
+                if "tensor_meta" in metadata
+            ]
+            submodule_node_dict[name].meta["tensor_meta"] = tensor_meta_list
+            logger.debug(
+                f"Updated tensor_meta metadata for node: {name} with its corresponding submodule outputs"
+            )
```
peri044, line 486: Consider merging these in a single update to avoid code duplication. Eg: 
```py
metadata_keys = ["val", "tensor_meta"]
for key in metadata_keys:
     if key not in submodule_node_dict[name].meta:
          .......
          logger.debug(f"Updated {key} metadata for node ........")
          break # if val is found no need to check tensor_meta
```

peri044 (changes requested):

peri044 (approved):
LGTM

PR #3250, opened 2024-10-19
using nccl ops from TRT-LLM namespace

apbose:
This PR illustrates the use of nccl ops from TRT-LLM for the example `examples/distributed_inference/tensor_parallel_simple_example.py`


examples/distributed_inference/tensor_parallel_nccl_ops.py
```diff
@@ -0,0 +1,196 @@
...
+    tensorrt_llm_lib_path = tensorrt_llm.__file__
+    plugin_lib_path = tensorrt_llm_lib_path.replace(
+        "__init__.py", "/libs/libnvinfer_plugin_tensorrt_llm.so"
+    )
+    try:
+        ctypes.CDLL(plugin_lib_path)
+        logger.info(f"plugin loaded successfully")
+    except OSError as e:
+        logger.info(f"unsuccessful load : {e}")
+    trt.init_libnvinfer_plugins(None, "")
```
narendasan, line 110: Do you need these lines as well?

examples/distributed_inference/tensor_parallel_nccl_ops.py
```diff
@@ -0,0 +1,196 @@
...
+    # Iterate over all registered plugin creators
+    plugin_registry = trt.get_plugin_registry()
```
narendasan, line 112: Is this just for debugging purposes?

examples/distributed_inference/tensor_parallel_nccl_ops.py
```diff
@@ -0,0 +1,196 @@
...
+        args: Tuple[Argument, ...],
+        kwargs: Dict[str, Argument],
+        name: str,
+    ) -> Union[TRTTensor, Sequence[TRTTensor]]:
+        plug_inputs = [args[0]]
+        allgather_plg_creator = trt.get_plugin_registry().get_plugin_creator(
+            "AllGather", "1", "tensorrt_llm"
+        )
+        assert allgather_plg_creator is not None
+        world_size = dist.get_world_size()
```
narendasan, line 131: How might the converter get this info if it was in library?

narendasan (commented):
Did you verify that numerical results are correct here?

examples/distributed_inference/tensor_parallel_nccl_ops.py
apbose: I am not clear what is meant by library here? You mean the aten_ops_converters.py? Generally the converter should get this info when the distributed environment is initialized. It is implicitly done when using torhrun but we explicitly initialize this in the initialize_distributed_env()

examples/distributed_inference/tensor_parallel_nccl_ops.py
apbose: I don't think these lines are required actually. Just tested the code without these lines and having "import tensorrt_llm" should be fine to have the plugins with namespace as tensorrt_llm to be loaded.

examples/distributed_inference/tensor_parallel_nccl_ops.py
apbose: Yes to see if the the plugins with "tensorrt_llm" namespace have been loaded properly or not

apbose:
Yes @narendasan , the numerical results come out to be correct for this example and the llama3 within 0.01 error threshold

examples/distributed_inference/tensor_parallel_nccl_ops.py
narendasan: Ok so we dont need a dist object? can we use that version here?

apbose: Yes we could if we maintain a global variable for it and use that in the file. But the dist object would be required for initialization

examples/distributed_inference/tensor_parallel_nccl_ops.py
```diff
@@ -0,0 +1,186 @@
...
+        allgather_plg_creator = trt.get_plugin_registry().get_plugin_creator(
+            "AllGather", "1", "tensorrt_llm"
+        )
+        assert allgather_plg_creator is not None
+        _world_size = int(os.environ["WORLD_SIZE"])
+        group = list(range(_world_size))
+        group = trt.PluginField(
+            "group", np.array(group, dtype=np.int32), trt.PluginFieldType.INT32
+        )
+        p_dtype = trt.float16
```
narendasan, line 126: Do these kernels only support FP16?

py/torch_tensorrt/dynamo/lowering/passes/fuse_distributed_ops.py
```diff
@@ -0,0 +1,72 @@
...
+
+# dead-code elimination, linting, and recompilation for graph, in-place
+from torch_tensorrt.dynamo.lowering.passes.pass_utils import (
+    clean_up_graph_after_modifications,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def custom_fused_all_gather_op(args0, args1, args2):
```
narendasan, line 15: Lets call this something like `tensorrt_fused_nccl_all_gather` or something

examples/distributed_inference/tensor_parallel_nccl_ops.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # TensorRT NCCL plugins
+    # Iterate over all registered plugin creators
+    plugin_registry = trt.get_plugin_registry()
+    for plugin_creator in plugin_registry.plugin_creator_list:
+        logger.info(
+            f"Plugin Name: {plugin_creator.name}, Namespace: {plugin_creator.plugin_namespace}, Version: {plugin_creator.plugin_version}"
+        )
+
+    @dynamo_tensorrt_converter(custom_fused_all_gather_op)
```
narendasan, line 108: We want to start thinking about how these might get added as actual converters like how we support quantization. I think the global variable dependency is a issue. How might we work around that?

examples/distributed_inference/tensor_parallel_nccl_ops.py
```diff
@@ -0,0 +1,186 @@
...
+    torch.manual_seed(1111)
+
+    return local_rank, world_size
+
+
+def register_nccl_ops(logger_file_name):
+    # Initialization
+    initialize_distributed_env()
+    # create a device mesh based on the given world_size.
+    _world_size = int(os.environ["WORLD_SIZE"])
```
narendasan, line 90: Things like this I am ok pulling in "globally"

examples/distributed_inference/tensor_parallel_nccl_ops.py
apbose: Yes pulling in the global variable assuming that the environment variable is set and initialized in the initialization part can be done, instead of using the dist package

examples/distributed_inference/tensor_parallel_nccl_ops.py
apbose: No they can support FP32 too

examples/distributed_inference/tensor_parallel_nccl_ops.py
```diff
@@ -0,0 +1,185 @@
...
+        allreduce_plg_creator = trt.get_plugin_registry().get_plugin_creator(
+            "ReduceScatter", "1", "tensorrt_llm"
+        )
+
+        assert allreduce_plg_creator is not None
+
+        counter = 0
+        strategy = AllReduceStrategy.NCCL
+        config = AllReduceConfig(0)
+        _world_size = int(os.environ["WORLD_SIZE"])
```
narendasan, line 154: Can you add some error handling for if this is not set?

examples/distributed_inference/tensor_parallel_nccl_ops.py
```diff
@@ -0,0 +1,185 @@
...
+        args: Tuple[Argument, ...],
+        kwargs: Dict[str, Argument],
+        name: str,
+    ) -> Union[TRTTensor, Sequence[TRTTensor]]:
+        plug_inputs = [args[0]]
+        allgather_plg_creator = trt.get_plugin_registry().get_plugin_creator(
+            "AllGather", "1", "tensorrt_llm"
+        )
+        assert allgather_plg_creator is not None
+        _world_size = int(os.environ["WORLD_SIZE"])
```
narendasan, line 121: Same here

narendasan:
Overall I think this is close to be ready to be merged. The next step after merge is to move the converters into the converter registry and conditionally enable them similar to how we do model opt. There we need to start thinking about the install workflow since I think TRT-LLM only supports 3.10.

apbose:
Ok @narendasan , I will add the error check. Currently in the Llama3 example, in the `apply_rotary_emb` the complex tensors are distributed and is an input to the torchTRT backend. Wrapping it in `aot_autograd` gets over the trace input error, but TorchTRT does not currently support torch.complex input tensors. I am working on filtering out the above.

Since this PR shows the use of simple example (which is working) and LLama3, I will push in the above change. Also need to try with strict=False
```
ep = torch.export._trace._export(
                model,
                (inputs,),
                dynamic_shapes=({1: seq_len},),
                strict=False,
                allow_complex_guards_as_runtime_asserts=True,
            )
```

examples/distributed_inference/tensor_parallel_nccl_ops.py
```diff
@@ -0,0 +1,185 @@
...
+import torch_tensorrt
+from torch.distributed._tensor.device_mesh import init_device_mesh
+from torch.fx import GraphModule, Node
+from torch.fx.node import Argument, Target
+from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
+from torch_tensorrt.dynamo.conversion._ConverterRegistry import (
+    dynamo_tensorrt_converter,
+)
+from torch_tensorrt.dynamo.lowering.passes.fuse_distributed_ops import (
+    custom_fused_all_gather_op,
```
peri044, line 22: I can't seem to find this name defined. Is this defined in some other way ?

examples/distributed_inference/requirement.txt
```diff
@@ -1,3 +1,5 @@
 accelerate
 transformers
-diffusers
\ No newline at end of file
+diffusers
+site
```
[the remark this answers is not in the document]
apbose: Removed site. diffusers is for the data parallel examples

py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -89,22 +106,37 @@ def _pretraced_backend(
...
+                    ),
+                )
 
             logger.debug("Post-AOT Autograd graph:\n" + str(gm.graph))
 
             gm = post_lowering(gm, settings)
 
             logger.debug("Lowered Input graph:\n " + str(gm.graph))
 
+            complex_nodes = find_complex_nodes(gm)
```
narendasan, line 136: Why isnt this part of lowering?

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -0,0 +1,120 @@
...
+    allreduce_plg_creator = trt.get_plugin_registry().get_plugin_creator(
+        "ReduceScatter", "1", "tensorrt_llm"
+    )
+
+    assert allreduce_plg_creator is not None
+
+    counter = 0
+    strategy = AllReduceStrategy.NCCL
+    config = AllReduceConfig(0)
+    _world_size = os.environ.get("WORLD_SIZE")
```
narendasan, line 85: Does WORLD_SIZE get baked into the engine? if i load from serialized do i need the env variable?

py/torch_tensorrt/dynamo/backend/backends.py
apbose: Need to get the complex_nodes before I do the lowering pass of `replace_complex_placeholder_to_tuple()`. Can put this in` replace_complex_placeholder_to_tuple` lowering pass, but since its a util function and pertaining more to `modify_complex_nodes` I put this in converter_utils

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
apbose: NoI dont think so. This relies on the initialize_distributed_env() function in tensor_parallel_dist_env.py to do that. In torchrun command it would implicitly do it, but since we do mpirun for the nccl commands TRT-LLM support we need to initialize the variables

narendasan:
@apbose run the linter

py/torch_tensorrt/dynamo/backend/backends.py
narendasan: I mean seems to me these 3-4 lines are a pass that can be added to lowering

py/torch_tensorrt/dynamo/lowering/passes/_replace_complex_placeholder_to_tuple.py
```diff
@@ -0,0 +1,112 @@
...
+    visited_nodes = set()
+    stack = [start_node]
+    while stack:
+        node = stack.pop()
+        if node in visited_nodes:
+            continue
+        visited_nodes.add(node)
+        update_node_meta(node, fake_mode)
+        for user in node.users:
+            if (
```
narendasan, line 92: Why do we have this special case?

apbose: It is more like a terminating case for the Llama model complex placeholder nodes.  The model is like `complex placeholder->reshape->slice->complex mul` , we need the meta data for reshape and slice to be amended, stopping at mul node (we are removing the complex mul to a custom torchTRT mul later in `modify_reshape_complex_nodes`)

narendasan: It seems like it is too broad, either we should denote that this function should only be used as a helper for the complex lowering or we should make sure that it only will run against `complex placeholder->reshape->slice->complex mul`. Right now `propogate_shape_change` makes it seem like a generic util. We could call it something like `_propogate_complex_num_shape_change`

apbose: Modified it

PR #3258, opened 2024-10-22
fix: cumsum add_constant bug fix (add dtype for np zeros)

chohk88:
# Description

When compiling the `roberta-base` model from Hugging Face (https://huggingface.co/FacebookAI/roberta-base), a `TypeError` occurs in the `cumsum` operation. For static shape input, the default datatype of `np.zeros(new_dims)` function is `np.float64` which is not handled properly by the `create_constant` utility function.

Fixes # ([issue](https://github.com/pytorch/TensorRT/issues/3187))

```
  File "/usr/local/lib/python3.10/dist-packages/torch_tensorrt/dynamo/conversion/aten_ops_converters.py", line 934, in aten_ops_cumsum
    return impl.slice.cumsum(
  File "/usr/local/lib/python3.10/dist-packages/torch_tensorrt/dynamo/conversion/impl/slice/ops.py", line 387, in cumsum
    zero_trttensor = get_trt_tensor(ctx, zeros, f"{name}_initial_value")
  File "/usr/local/lib/python3.10/dist-packages/torch_tensorrt/dynamo/conversion/converter_utils.py", line 388, in get_trt_tensor
    return create_constant(ctx, input_val, name, dtype, min_rank)
  File "/usr/local/lib/python3.10/dist-packages/torch_tensorrt/dynamo/conversion/converter_utils.py", line 349, in create_constant
    constant = ctx.net.add_constant(
torch._dynamo.exc.BackendCompilerFailed: backend='torch_tensorrt' raised:
TypeError: add_constant(): incompatible function arguments. The following argument types are supported:
    1. (self: tensorrt.tensorrt.INetworkDefinition, shape: tensorrt.tensorrt.Dims, weights: tensorrt.tensorrt.Weights) -> tensorrt.tensorrt.IConstantLayer
Invoked with: <tensorrt.tensorrt.INetworkDefinition object at 0x7fee84ebd770>, (1,), array([0.])
```

### Reproduction Code:

```python
# https://huggingface.co/FacebookAI/roberta-base
import torch
from transformers import RobertaTokenizer, RobertaModel
import torch_tensorrt

backend = "torch_tensorrt"
device = "cuda:0"

# Load tokenizer and model
tokenizer = RobertaTokenizer.from_pretrained('roberta-base')
model = RobertaModel.from_pretrained('roberta-base')
model = model.to(device)

# Tokenize input text
text = "Replace me by any text you'd like."
encoded_input = tokenizer(text, return_tensors='pt')
encoded_input = {k: v.to(device) for k, v in encoded_input.items()} 

# Compile model with Torch-TensorRT
model = torch.compile(
    model,
    backend=backend,
    options={
        "enabled_precisions": {torch.float16},
    },
    dynamic=False,
)

# Run inference
output = model(**encoded_input)
print(output)
```

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
@@ -370,7 +370,7 @@ def cumsum(
         )
     else:
         new_dims = tuple(data.shape)
-        zeros = np.zeros(new_dims)
+        zeros = np.zeros(new_dims, dtype=np.float32)
         zero_trttensor = get_trt_tensor(ctx, zeros, f"{name}_initial_value")
 
     running_sum = loop.add_recurrence(zero_trttensor)
--- a/py/torch_tensorrt/dynamo/utils.py
+++ b/py/torch_tensorrt/dynamo/utils.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import warnings
 from dataclasses import fields, replace
 from enum import Enum
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
@@ -10,6 +11,8 @@ import sympy
 import tensorrt as trt
 import torch
 from torch._subclasses.fake_tensor import FakeTensor
+
+from packaging import version
 from torch_tensorrt._Device import Device
 from torch_tensorrt._enums import dtype
 from torch_tensorrt._features import ENABLED_FEATURES
@@ -18,8 +21,6 @@ from torch_tensorrt.dynamo import _defaults
 from torch_tensorrt.dynamo._defaults import default_device
 from torch_tensorrt.dynamo._engine_cache import BaseEngineCache
 from torch_tensorrt.dynamo._settings import CompilationSettings
-
-from packaging import version
 
 from .types import TRTDataType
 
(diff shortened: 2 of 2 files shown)


py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
```diff
@@ -370,7 +370,7 @@ def cumsum(
         )
     else:
         new_dims = tuple(data.shape)
-        zeros = np.zeros(new_dims)
+        zeros = np.zeros(new_dims, dtype=np.float32)
```
peri044, line 373: should this dtype be dependent on input dtype or always float32 ?

chohk88: Using `np.float32` for the input works fine regardless of the input type. 

However, if we trace the root cause of the error, we can see that in [this line](https://github.com/pytorch/TensorRT/blob/6d40ff1442572b8808961689c5ecb4ee5c975456/py/torch_tensorrt/dynamo/_settings.py#L99), the name `truncate_double` is used, but we are passing `truncate_long_and_double` as an argument to `torch.compile`, as shown [here](https://github.com/pytorch/TensorRT/blob/6d40ff1442572b8808961689c5ecb4ee5c975456/examples/dynamo/torch_compile_stable_diffusion.py#L41). 

Because of this, at [this point](https://github.com/pytorch/TensorRT/blob/6d40ff1442572b8808961689c5ecb4ee5c975456/py/torch_tensorrt/dynamo/utils.py#L483), the `truncate_long_and_double` argument is not handled and then removed, which leads to an error when trying to process the default type `float64` of `np.zeros`.

According to [this section](https://github.com/pytorch/TensorRT/blob/6d40ff1442572b8808961689c5ecb4ee5c975456/py/torch_tensorrt/dynamo/_compiler.py#L174), `torch_tensorrt.dynamo.compile` prefers `truncate_double` as the input but can also handle `truncate_long_and_double`.

What would be the best way to fix this issue?

peri044: Change the `truncate_long_and_double` to `truncate_double` in this example https://github.com/pytorch/TensorRT/blob/6d40ff1442572b8808961689c5ecb4ee5c975456/examples/dynamo/torch_compile_stable_diffusion.py#L40 and that should work right ?

peri044: `truncate_long_and_double` is deprecated. But it looks like it is not correctly handled if user provides this argument in `torch.compile` workflow. Can you add this check in https://github.com/pytorch/TensorRT/blob/6d40ff1442572b8808961689c5ecb4ee5c975456/py/torch_tensorrt/dynamo/utils.py#L483 ? similar to https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/_compiler.py#L180-L185

chohk88: @peri044  Using `float32` ensures compatibility, even with varying inputs, as it is the most commonly used data type. Additionally, I added exception handling for cases where the deprecated `truncated_long_and_double` argument might still be used.

peri044 (approved):
LGTM

PR #3261, opened 2024-10-23
switch from fx.symbolic_trace to dynamo_trace for converter test part-1

lanluo-nvidia:
# Description

Switch from fx.symbolic_trace to dynamo trace for converter test part-1

Since we ~2000 test cases I splited into batches to migrate them off fx.symbolic_trace

current batch:  tests/py/dynamo/conversion/test_a*.py

There mainly two types of testcases require change:
1. for ops require custom dynamic shape support, will be changed to use dynamic_shapes instead of Input
2. for ops will be decomposed into multiple ops, these op test will be added in the test_composite_aten_op.py

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/utils.py
```diff
@@ -711,7 +711,13 @@ def get_output_metadata(
 
 def get_output_dtypes(output: Any, truncate_doulbe: bool = False) -> List[dtype]:
     output_dtypes = []
-    if isinstance(output, torch.fx.node.Node):
+    if isinstance(output, int):
```
narendasan, line 714: you can just use type conversion here instead of a switch statement. 

```py
output_dtypes.append(dtype._from(output))
```

If dtype doesnt support python types we should probably add it

tests/py/dynamo/conversion/harness.py
```diff
@@ -29,6 +32,25 @@
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
 
+# this is to enable dynamo tracer as True in the converter test files batch by batch
+def get_use_dynamo_tracer(use_dynamo_tracer: Any) -> bool:
```
narendasan, line 36: why do we need this? any test using our harness should be on dynamo right? the fx tests are in the py/ tree

tests/py/dynamo/conversion/harness.py
```diff
@@ -277,10 +299,21 @@ def generate_graph(
...
         mod = mod.eval()
         if use_dynamo_tracer:
-            exported_program = torch_tensorrt.dynamo.trace(mod, tuple(original_inputs))
+            if torch_export_dynamic_shapes is None:
+                torch_export_dynamic_shapes = get_dynamic_shapes_args(
+                    mod, original_inputs
+                )
+            device = default_device()
+            torch_export_inputs = get_torch_inputs(original_inputs, device)
+            exported_program = torch.export.export(
```
narendasan, line 312: Is there some lower level representation we can use to minimized deps in the test?

tests/py/dynamo/conversion/harness.py
lanluo-nvidia: @narendasan  since there are 74 testfiles(~500 testcases) are impacted by switching from fx.symbolic_tracer to dynamo tracer, I use this to method to split the changes batch by batch
so the first batch is to automatically set use_dynamo_tracer to True for the below operators starts with a:
tests/py/dynamo/conversion/test_a*.py
I will eventually migrate all the symbolic tracer to dynamo tracer and remove this temporary helper method.

tests/py/dynamo/conversion/test_attention.py
```diff
@@ -48,15 +56,16 @@ def forward(self, query, key, value):
...
+                (4, 2, 12, 4),
+                (6, 3, 16, 8),
+                (32, 8, 18, 16),
                 (4, 2, 4, 16),
                 (6, 3, 8, 32),
                 (32, 8, 12, 64),
             ),
         ]
     )
+    @unittest.skip("need to change to custom dynamic shapes")
```
peri044, line 68: Are these SDPA tests skipped because they are added in test_composite_aten_op.py ? If so, can we remove them here ?

peri044 (approved):
minor question. LGTM

PR #3270, opened 2024-10-31
Wrapper module around TRT + pytorch subgraphs

keehyuna:
# Description

Wrapper module around TRT + pytorch subgraphs and record/replay cuda graphs

Fixes # (issue)

## Type of change

# Checklist:


core/runtime/TRTEngine.h
```diff
@@ -102,7 +103,8 @@ struct TRTEngine : torch::CustomClassHolder {
   std::vector<at::Tensor> input_buffers = {};
   std::vector<at::Tensor> output_buffers = {};
   std::string shape_key;
-
+  bool cudagraphs_enabled = false;
```
narendasan, line 106: What if we made this like an enum that describes the runtime mode `STANDARD, SUBGRAPH_CUDAGRAPHS, WHOLE_GRAPH_CUDAGRAPHS` so we avoid the checks in execute engine?

narendasan: Actually why does the engine need to know if its in whole graph cudagraph mode? seems like it doesnt actually change the behavior from "STANDARD" mode

py/torch_tensorrt/dynamo/runtime/_WrapperTorchTensorRTModule.py
```diff
@@ -0,0 +1,271 @@
...
+    """This Wrapper runtime module to record/replay cuda graph in sub modules"""
+
+    def __init__(
+        self,
+        original_module: torch.nn.Module,
+        output_shapes: List[torch.Size],
+        output_dtypes: List[torch.dtype],
+    ):
+        super(WrapperTorchTensorRTModule, self).__init__()
+        self.original_module = original_module
```
narendasan, line 29: Would this module be the original PyTorch one or the compiled one?

narendasan (commented):
@keehyuna can you add an example on how this would be used by a user?

py/torch_tensorrt/dynamo/runtime/_WrapperTorchTensorRTModule.py
keehyuna: It is compiled module. I renamed it to be descriptive.

py/torch_tensorrt/dynamo/runtime/_WrapperTorchTensorRTModule.py
```diff
@@ -0,0 +1,285 @@
...
+
+        # If the new shape key differs from the existing one, infer new output shape
+        if new_shape_key != self.shape_key:
+            logger.debug(f"Input shape changed {self.shape_key} -> {new_shape_key}")
+            self.shape_key = new_shape_key
+
+            if self.input_is_dynamic:
+                with FakeTensorMode() as mode:
+                    fake_inputs = [mode.from_tensor(input) for input in inputs]
+                    tmp_outputs = self.original_module(*fake_inputs)
```
keehyuna, line 88: Maybe I can use complied_module() instead. I'm now sure which one is better.

py/torch_tensorrt/dynamo/runtime/_WrapperTorchTensorRTModule.py
```diff
@@ -0,0 +1,285 @@
...
+        self._engine_stream: Optional[torch.cuda.Stream] = None
+        self.input_is_dynamic = input_is_dynamic(self.inputs)
+
+        # Disable cudagrphs in submodules as it will be enabled in wrapper
+        for name, rt_mod in self.compiled_module.named_children():
+            if "_run_on_acc" in name:
+                rt_mod.set_whole_cudagraphs(True)
+
+        # Warm up is necessary to ensure that memory allocations and initializations are not recorded in cuda graphs
+        with unset_fake_temporarily():
```
keehyuna, line 62: If this warm up is not used I can see error with below test. It seems cudnn ops is added by pytorch and warmup is required.
https://github.com/pytorch/TensorRT/pull/3270/files#diff-0355c793a29056c378c96cae4e041427e64ca145a9199310c25b43ac1091bbecR144-R148

`RuntimeError: cuDNN error: CUDNN_STATUS_INTERNAL_ERROR_DEVICE_ALLOCATION_FAILED`

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -832,6 +835,15 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
 
     dryrun_stats_display(dryrun_tracker, settings.dryrun)
 
+    if len(dryrun_tracker.to_run_in_torch) > 0:
```
keehyuna, line 838: I enabled wrapper module when there is graph break.
@narendasan Do you think we need another runtime api to enable wrapped module?

py/torch_tensorrt/dynamo/runtime/_WrapperTorchTensorRTModule.py
narendasan: compiled module is probably better

py/torch_tensorrt/dynamo/_compiler.py
narendasan: I think we could have a compile setting that is something like `run_with_cudagraphs` or something that returns the wrapped module. But I think its confusing that we have both, I think at some point we might want to just pick one

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -131,6 +131,7 @@ def __init__(
         self.weight_name_map = weight_name_map
         self.serialized_engine = serialized_engine
         self.engine = None
+        self.cudagraphs_enabled_parent_module = False
```
peri044, line 134: is this used ?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -195,6 +196,9 @@ def set_device_memory_budget(self, budget_bytes: int) -> int:
 
         return budget_bytes
 
+    def set_whole_cudagraphs(self, enable: bool) -> None:
+        self.engine.set_whole_cudagraphs(enable)
```
peri044, line 200: add a docstring

py/torch_tensorrt/dynamo/runtime/_WrapperTorchTensorRTModule.py
```diff
@@ -0,0 +1,285 @@
...
+            inputs_tensor = [spec.torch_tensor.cuda() for spec in self.inputs]
+            s = torch.cuda.Stream()
+            s.wait_stream(torch.cuda.current_stream())
+            with torch.cuda.stream(s):
+                for _ in range(3):
+                    self.compiled_module(*inputs_tensor)
```
peri044, line 68: should we wrap only the warm up runs in a error log level ? The scenario I'm thinking is let's say user ran his model in debug mode using 
```py
      trt_model = torch_tensorrt.compile(model, inputs, debug=True)
```
This would print the warm up logs as well which can be confusing.

py/torch_tensorrt/dynamo/runtime/_WrapperTorchTensorRTModule.py
```diff
@@ -0,0 +1,285 @@
...
+
+        return False
+
+    def __del__(self) -> None:
+        if self.cudagraph:
+            self.cudagraph.reset()
+
+    def forward(self, *inputs: torch.Tensor) -> torch.Tensor | Tuple[torch.Tensor, ...]:
+        # Ensure inputs are available in all scopes and cast symbolic integers to Tensors
+        contiguous_inputs: List[torch.Tensor] = [
```
peri044, line 103: Seems like some of the code here is similar to `PythonTorchTensorRTModule`. Can we functionalize them and reuse across these modules ?

py/torch_tensorrt/dynamo/runtime/_WrapperTorchTensorRTModule.py
```diff
@@ -0,0 +1,285 @@
...
+        self.profiling_enabled = False
+        self.cudagraphs_enabled = False
+        self._caller_stream: Optional[torch.cuda.Stream] = None
+        self._engine_stream: Optional[torch.cuda.Stream] = None
+        self.input_is_dynamic = input_is_dynamic(self.inputs)
+
+        # Disable cudagrphs in submodules as it will be enabled in wrapper
+        for name, rt_mod in self.compiled_module.named_children():
+            if "_run_on_acc" in name:
+                rt_mod.set_whole_cudagraphs(True)
```
peri044, line 59: how are we disabling them in submodules ? Also, do we now have two modes like intra cuda graphs (only within TRT modules) and inter cuda graphs ? I believe we should just have one concept of use_cuda_graphs exposed to users. What do you think ?

keehyuna (commented):
@narendasan / @peri044 
I would appreciate your feedback on three candidate API drafts.
Please let me know if you have any alternative ideas or suggestions for the API
1. Compiler option: enable_wrapper_module=True
```
wrapped_module = torchtrt.compile(
    ...
    enable_wrapper_module=True,
)
wrapped_module(*input)
```
simple usage. but it is difficult to handle torch.export(wrapped_module) when torch module is in input side.

2. Optional parameter in existing cuda runtime api
```
with torchtrt.runtime.enable_cudagraphs(optimized_model) as wrapped_module:
    wrapped_module(*input)
```
Reuse existing api. better usability to turn on both cuda graph and wrapped module

3. Import wrapper module
```
wrapped_module = WrapperTorchTensorRTModule(optimized_model)
with torchtrt.runtime.enable_cudagraphs():
    wrapped_module(*input)
```
Feels more intuitive.

https://github.com/pytorch/TensorRT/pull/3270/files#diff-0355c793a29056c378c96cae4e041427e64ca145a9199310c25b43ac1091bbecR235-R282

examples/dynamo/cudagraphs_wrapper_example.py
```diff
@@ -0,0 +1,98 @@
...
+#
+# Using a wrapped runtime module with CUDA Graphs allows you to encapsulate sequences of operations into graphs
+# that can be executed efficiently, even in the presence of graph breaks. When a CUDA Graph context manager is
+# used with the TensorRT module as a positional argument, it returns a wrapped_module. This module captures the
+# execution graph, enabling efficient replay during subsequent inferences by reducing kernel launch overheads
+# and improving performance. Note that initializing with the wrapper module involves a warm-up phase where the
+# module is executed several times. This warm-up ensures that memory allocations and initializations are not
+# recorded in CUDA Graphs, which helps maintain consistent execution paths and optimize performance.
+with torch_tensorrt.runtime.enable_cudagraphs(trt_model) as wrapped_module:
+    wrapped_module(input)
```
peri044, line 98: What do you think of the idea where we only allow users to do this and not `enable_cudagraphs()` ?

py/torch_tensorrt/_compile.py
```diff
@@ -12,6 +12,9 @@
 from torch_tensorrt._features import ENABLED_FEATURES
 from torch_tensorrt._Input import Input
 from torch_tensorrt.dynamo import _defaults
+from torch_tensorrt.dynamo.runtime._WrapperTorchTensorRTModule import (
+    WrapperTorchTensorRTModule,
```
peri044, line 16: Consider renaming this to `CudaGraphsTorchTensorRTModule`

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -373,6 +376,7 @@ def compile(
     use_explicit_typing: bool = _defaults.USE_EXPLICIT_TYPING,
     use_fp32_acc: bool = _defaults.USE_FP32_ACC,
     enable_weight_streaming: bool = _defaults.ENABLE_WEIGHT_STREAMING,
+    enable_wrapper_module: bool = _defaults.ENABLE_WRAPPER_MODULE,
```
peri044, line 379: Would this be needed if we decide to go with the following option ?
```py
with torch_tensorrt.runtime.enable_cuda_graphs(module) as torchtrt_module:
```

peri044 (commented):
Should we just have cudagraphs to be on or off and return the cudagraph module if its on ? I'm not sure if we have a value add with the `SUBGRAPH_CUDAGRAPHS` mode. Let me know if you think otherwise.

examples/dynamo/cudagraphs_wrapper_example.py
keehyuna: As we discussed, I changed `enable_cudagraphs()` to get module as mandatory.
Context manager checks and return original or wrapper module.

py/torch_tensorrt/runtime/_cudagraphs.py
```diff
@@ -37,19 +64,52 @@ class _CudagraphsContextManager(object):
...
+            logger.debug(
+                f"{num_torch_module} torch modules are in subgraphs. Using wrapper module for cuda graphs"
+            )
+            return CudaGraphsTorchTensorRTModule(self.compiled_module)
+        else:
+            if num_trt_module > 0:
+                logger.debug(
+                    "There is no graph breaks. Using original module for cuda graphs"
+                )
+            else:
```
keehyuna, line 99: If `torch_compile` backend is used, graph information is not available when context manager is generated.
Context manager will return original module in this case. Documented that `dynamo` backend is required for cuda graphs+ graphs breaks.
https://github.com/pytorch/TensorRT/pull/3270/files#diff-7f89360445589890395e2fa77a673994a449f464d82482f51f1df3c448700a04R96

narendasan: Also, so is this saying that if the user is using torch.compile that we will use subgraph_cudagraphs? I think this is probably fine not sure if it needs to be a warning but we should print in the debug logs which one we choose.

py/torch_tensorrt/runtime/_cudagraphs.py
```diff
@@ -37,19 +64,52 @@ class _CudagraphsContextManager(object):
...
 
     def __exit__(self, *args: Any) -> None:
         # Set cudagraphs back to old mode
         set_cudagraphs_mode(self.old_mode)
 
 
-def enable_cudagraphs() -> _CudagraphsContextManager:
-    return _CudagraphsContextManager()
+def enable_cudagraphs(
+    compiled_module: torch.nn.Module,
```
narendasan, line 113: Should be `Union[torch.fx.GraphModule, torch.nn.Module]`

py/torch_tensorrt/runtime/_cudagraphs.py
```diff
@@ -37,19 +64,52 @@ class _CudagraphsContextManager(object):
...
+                torch.ops.tensorrt.set_cudagraphs_mode(_PY_RT_CUDAGRAPHS)
+
+            logger.debug(
+                f"{num_torch_module} torch modules are in subgraphs. Using wrapper module for cuda graphs"
+            )
+            return CudaGraphsTorchTensorRTModule(self.compiled_module)
+        else:
+            if num_trt_module > 0:
+                logger.debug(
+                    "There is no graph breaks. Using original module for cuda graphs"
```
narendasan, line 97: Think this could be a better message. like `No graph breaks detected, using runtime cudagraphs mode`

py/torch_tensorrt/runtime/_cudagraphs.py
```diff
@@ -37,19 +64,52 @@ class _CudagraphsContextManager(object):
...
+                num_torch_module += 1
+
+        if num_torch_module > 0:
+            # Set whole cudagraphs mode and returns wrapped module
+            _PY_RT_CUDAGRAPHS = CudaGraphsMode.WHOLE_GRAPH_CUDAGRAPHS
+            # Set new mode for C++
+            if torch_tensorrt.ENABLED_FEATURES.torch_tensorrt_runtime:
+                torch.ops.tensorrt.set_cudagraphs_mode(_PY_RT_CUDAGRAPHS)
+
+            logger.debug(
```
narendasan, line 90: Maybe reword `Found pytorch subgraphs in module, wrapping module in CudaGraphsTorchTensorRTModule`

narendasan (approved):
LGTM, thanks

PR #3282, opened 2024-11-06
allow null arg_inputs if kwarg_inputs is set

technillogue:
# Description

Currently, a module cannot be compiled if it is called with only kwarg inputs, because arg_inputs must not be none. This prevents some usages of MutableTorchTensorRTModule. arg_inputs should be allowed to be None if kwarg_inputs is not none.

Fixes # (issue)

## Type of change

# Checklist:


peri044 (commented):
Can you also change it here ? https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/_compiler.py#L671-L672

PR #3300, opened 2024-11-20
fix: attribute error caused by `trt.BuilderFlag.STRICT_TYPES`

jiwoong-choi:
# Description

When `strict_type_constraints` is `True`, accessing `trt.BuilderFlag.STRICT_TYPES` causes `AttributeError: type object 'tensorrt.tensorrt.BuilderFlag' has no attribute 'STRICT_TYPES'`. Indeed, there is no such builder flag in TensorRT, and it should be replaced by `trt.BuilderFlag.OBEY_PRECISION_CONSTRAINTS`.

## Type of change

Bug fix (non-breaking change which fixes an issue)

# Checklist:


peri044 (commented):
I think we should remove this if block because we don't expose strict_type_constraints flag to the user. We advocate for strong typing and here is the usage for it https://pytorch.org/TensorRT/user_guide/mixed_precision.html

peri044:
Closing this PR. Please re-open incase you think this feature is necessary.

PR #3311, opened 2024-12-04
fix: Remove pytorch overhead while finding fusions for fully convertible models

peri044:
# Description

By default, we use utilities such as `FxNetAccFusionsFinder` to find fusions which are used in partitioning. For big models like flux, this takes a lot of time. We can avoid this for fully convertible models. 

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -685,6 +685,8 @@ def compile_module(
         )
 
     breakpoint()
+    import time 
+    start = time.time()
     # Skip partitioning if the whole graph is supported to reduce partitioning overhead
     if num_supported_ops == total_ops:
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -786,7 +788,7 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
             require_full_compilation=settings.require_full_compilation,
         )
 
-    dryrun_tracker.unsupported_ops = supported_ops.unsupported_operators
+    # dryrun_tracker.unsupported_ops = supported_ops.unsupported_operators
```
narendasan, line 791: Should we should condition this on debug=True?

peri044: actually this wasn't supposed to land.

PR #3356, opened 2025-01-15
DLFW changes 25.01 onwards

apbose:
Some minor changes for the NGC container release

### DIFF
--- a/core/util/Exception.h
+++ b/core/util/Exception.h
@@ -1,3 +1,12 @@
+#if defined(__GNUC__) && !defined(__clang__)
+#if __GNUC__ >= 13
+#include <cstdint>
+#endif
+#elif defined(__clang__)
+#if __clang_major__ >= 13
+#include <cstdint>
+#endif
+#endif
 #pragma once
 
 #include <exception>
--- a/core/util/Exception.h
+++ b/core/util/Exception.h
@@ -1,11 +1,6 @@
-#if defined(__GNUC__) && !defined(__clang__)
-#if __GNUC__ >= 13
+// Include cstdint for GCC 13+ or Clang 13+
+#if (defined(__GNUC__) && !defined(__clang__) && (__GNUC__ >= 13)) || (defined(__clang__) && (__clang_major__ >= 13))
 #include <cstdint>
-#endif
-#elif defined(__clang__)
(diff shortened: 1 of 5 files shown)


core/util/Exception.h
```diff
@@ -1,3 +1,12 @@
+#if defined(__GNUC__) && !defined(__clang__)
+#if __GNUC__ >= 13
+#include <cstdint>
+#endif
+#elif defined(__clang__)
+#if __clang_major__ >= 13
+#include <cstdint>
+#endif
+#endif
```
peri044, line 9: Consider changing to  
```
// Include cstdint for GCC 13+ or Clang 13+
#if (defined(__GNUC__) && !defined(__clang__) && (__GNUC__ >= 13)) || \
    (defined(__clang__) && (__clang_major__ >= 13))
    #include <cstdint>
#endif
```

peri044 (approved):
LGTM. posted a minor suggestion

noxfile.py
```diff
@@ -34,7 +34,7 @@
 # Set epochs to train VGG model for accuracy tests
 EPOCHS = 25
 
-SUPPORTED_PYTHON_VERSIONS = ["3.9", "3.10", "3.11", "3.12"]
+SUPPORTED_PYTHON_VERSIONS = ["3.7", "3.8", "3.9", "3.10", "3.12"]
```
narendasan, line 37: IIRC we dont support py 3.7, 3.8 anymore

tests/py/ts/ptq/test_ptq_dataloader_calibrator.py
```diff
@@ -4,9 +4,9 @@
 import torch
 import torch.nn as nn
 import torch_tensorrt as torchtrt
+import torch_tensorrt.ts.ptq as PTQ
```
narendasan, line 7: This is fine to have but we should drop these tests from DLFW as this API is being deprecated

apbose: Yes they are dropped in DLFW

noxfile.py
apbose: Ok I will remove them then

PR #3276, opened 2024-11-04
feat: Runtime output buffer optimization

keehyuna:
# Description

Latency hiding by creating the output tensor for next output buffer

Fixes #3275 

## Type of change

Please delete options that are not relevant and/or add your own.

- New feature (non-breaking change which adds functionality)

# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [x] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #3364, opened 2025-01-23
Support Data Dependent Shape (DDS) and NonZero op

### LINKED ISSUE #2516: `aten.nonzero`

zewenli98:
# Description

Revisit NonZero. Support Data Dependent Shape (DDS) and NonZero op in this PR.

Torch-TRT needs static output shapes to create output tensors in the runtime. However, for some ops with DDS, like NonZero, its output shape depends on the input data content, not just the input shape, so we need to propagate the shapes using `ShapeProp` like `ShapeProp(fx_module).propagate(*torch_inputs)`, and then pass in the output shape to the Torch-TRT runtime.

Fixes #2516 

## Type of change

# Checklist:


tests/py/dynamo/conversion/test_nonzero_aten.py
```diff
@@ -0,0 +1,33 @@
+import torch
+import torch.nn as nn
+from parameterized import parameterized
+from torch.testing._internal.common_utils import run_tests
+
+from .harness import DispatchTestCase
+
+
+class TestAtanConverter(DispatchTestCase):
```
peri044, line 9: Rename to nonzero converter

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -731,11 +734,16 @@ def run(
             engine_bytes.write(serialized_engine)
             engine_str = engine_bytes.getvalue()
 
+        for node in self.module.graph.nodes:
+            if node.op == "output":
+                self.output_shapes.append(tuple(node.meta["tensor_meta"].shape))
```
peri044, line 739: Check `node.meta["val"]` as well.

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -455,10 +460,11 @@ def forward(self, *inputs: torch.Tensor) -> torch.Tensor | Tuple[torch.Tensor, .
                 if can_use_pre_allocated_outputs:
                     outputs = self.pre_allocated_outputs
                 else:
-                    self.output_shapes = [
-                        tuple(self.context.get_tensor_shape(output_name))
-                        for output_name in self.output_names
-                    ]
+                    if self.output_shapes is None:
```
peri044, line 463: what happens here `self.output_shapes` has symbolic values ?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -100,6 +101,7 @@ def __init__(
             serialized_engine (bytes): Serialized TensorRT engine in the form of a bytearray
             input_binding_names (List[str]): List of input TensorRT engine binding names in the order they would be passed to the TRT modules
             output_binding_names (List[str]): List of output TensorRT engine binding names in the order they should be returned
+            output_shapes (Sequence[Tuple]): List of output shapes for the engine. For some cases, output shapes are dynamic and depends on input data, like NonZero op, so we need to explicitly provide output shapes
```
narendasan, line 104: So someone would need to precalculate the output shape for TensorRT?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -455,10 +460,11 @@ def forward(self, *inputs: torch.Tensor) -> torch.Tensor | Tuple[torch.Tensor, .
                 if can_use_pre_allocated_outputs:
                     outputs = self.pre_allocated_outputs
                 else:
-                    self.output_shapes = [
-                        tuple(self.context.get_tensor_shape(output_name))
-                        for output_name in self.output_names
-                    ]
+                    if self.output_shapes is None:
+                        self.output_shapes = [
```
narendasan, line 464: Where does this get used?

tests/py/dynamo/conversion/test_nonzero_aten.py
```diff
@@ -0,0 +1,33 @@
...
+    def test_atan_float(self, input_shape, dtype):
+        class atan(nn.Module):
+            def forward(self, input):
+                return torch.ops.aten.nonzero.default(input)
+
+        inputs = [torch.randint(low=0, high=3, size=input_shape, dtype=dtype)]
+        self.run_test(
+            atan(),
+            inputs,
+            propagate_shapes=True,  # it requires propagate_shapes=True to get output shapes
```
peri044, line 28: what happens if the input changes for the already compiled TRT engine ? We should have a test for that for nonzero right ?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
zewenli98: That's optional. We need to provide output_shapes only for DDS ops

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
zewenli98: If output_shapes is provided, like by `ShapeProp(fx_module).propagate(*torch_inputs)` we are going to use it directly. Otherwise to get output_shapes as before

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
zewenli98: Did we support using torch.empty to create a symbolic tensors? https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py#L350

narendasan (approved):
LGTM

tests/py/dynamo/conversion/test_nonzero_aten.py
```diff
@@ -0,0 +1,32 @@
...
+    @parameterized.expand(
+        [
+            ((10,), torch.int),
+            ((1, 20), torch.int32),
+            ((2, 3), torch.int64),
+            ((2, 3, 4), torch.float),
+            ((2, 3, 4, 5), torch.float),
+        ]
+    )
+    def test_non_zero_float(self, input_shape, dtype):
```
peri044, line 19: How about adding a test case with dynamic shapes ?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -23,6 +22,42 @@
...
+            if self.buffers[tensor_name].shape != shape:
+                self.buffers[tensor_name] = torch.empty(
+                    shape,
+                    dtype=self.dtypes[tensor_name],
+                    device=torch.cuda.current_device(),
+                )
+                self.shapes[tensor_name] = tuple(shape)
+        return self.buffers[tensor_name].data_ptr()
+
+    def notify_shape(self, tensor_name: str, shape: Tuple[int, ...]) -> None:
```
peri044, line 57: What is the order of the execution ? notify shape is called first for every output and then rellocate_output_sync gets called ?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -524,12 +537,26 @@ def forward(self, *inputs: torch.Tensor) -> torch.Tensor | Tuple[torch.Tensor, .
 
                 self._caller_stream.wait_stream(self._engine_stream)
 
-            if self.use_pre_allocated_outputs:
```
peri044, line 527: You might want to check with @keehyuna  on how this change would affect the use_pre_allocated_outputs optimization that he implemented ?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
zewenli98: Yes I think so. It will call `notify_shape` to get correct shape first and then allocate memory

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
zewenli98: yeah I think we can directly remove `pre_allocated_outputs` related arguments because we don't have to manually allocate memory to outputs. Instead, `reallocate_output_async` fuunction will take care of everything and automatically reuse pre_allocated_outputs if possible. @keehyuna Can you take a look?

keehyuna: Yes, `pre_allocated_outputs` is not required if allocation is properly reused. I tried to check the behavior but there was "Segmentation fault" during build_serialized_network()

keehyuna: I added nvtx event in PythonTorchTensorRTModule:TensorRTRuntime and reallocate_output_async.
They are running concurrently when I test resnet model. It seems pre_allocated_outputs is not required and async output will have benefit in most cases. I will check perf in more cases. I think this pr can go ahead as use_pre_allocated_outputs is disabled by default. I will revert pre-allocated output feature after some more test.

BTW, will cpp runtime have async output?

zewenli98: yeah I'm working on cpp runtime but it may take a couple of days

zewenli98: Hi @keehyuna I have updated the cpp runtime. Can you confirm if `pre_allocated_outputs` is not needed anymore?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -238,6 +274,8 @@ def setup_engine(self) -> None:
             for output_name in self.output_names
         ]
 
+        self.setup_output_allocator()
```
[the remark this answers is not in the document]
zewenli98: Thanks for pointing out. It should be set up during forward only so I removed here

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
zewenli98: @keehyuna Thanks for confirming. I have cleared up some `pre_allocated_outputs` related codes by the way. Can you take a look if that's complete? If anything else should be removed please let me know!

core/runtime/execute_engine.cpp
```diff
@@ -163,22 +163,23 @@ void setup_input_tensors(
...
+      output_dtypes_dict[name] =
+          util::TRTDataTypeToScalarType(compiled_engine->exec_ctx->getEngine().getTensorDataType(name.c_str()));
+    }
+    compiled_engine->output_allocator = std::make_shared<DynamicOutputAllocator>(output_dtypes_dict);
   }
 
-  return outputs;
+  for (const auto& output_name : compiled_engine->out_binding_names) {
+    if (!compiled_engine->exec_ctx->setOutputAllocator(output_name.c_str(), compiled_engine->output_allocator.get())) {
+      throw std::runtime_error("Failed to set output allocator for " + output_name);
```
narendasan, line 180: Use the error utilities we already have like `TORCHTRT_THROW_ERROR`

core/runtime/execute_engine.cpp
```diff
@@ -287,44 +287,20 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
...
-            compiled_engine->exec_ctx->setTensorAddress(
-                name.c_str(), compiled_engine->output_buffers[pyt_idx].data_ptr()),
-            "Error while setting the output tensor address");
-      } else {
-        TORCHTRT_CHECK(
-            compiled_engine->exec_ctx->setTensorAddress(name.c_str(), outputs[pyt_idx].data_ptr()),
-            "Error while setting the output tensor address");
-      }
-    }
+    setup_output_allocator(compiled_engine);
```
narendasan, line 296: We need to do this everytime the engine is going to run?

core/runtime/execute_engine.cpp
```diff
@@ -287,44 +287,20 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
...
   }
 
   auto current_device_id = -1;
   if (inputs.size() > 0) {
     current_device_id = inputs[0].device().index(); // Done this way to avoid a call to cudart
-  } else if (outputs.size() > 0) {
-    current_device_id = outputs[0].device().index(); // Done this way to avoid a call to cudart
+  } else {
+    current_device_id = c10::cuda::current_device();
```
narendasan, line 303: This is going to call cudart right? That will incur a significant cost

core/runtime/execute_engine.cpp
```diff
@@ -368,21 +344,32 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
...
+  if (compiled_engine->profile_execution) {
+    output_profiler_guard =
+        std::make_unique<torch::autograd::profiler::RecordProfile>(compiled_engine->output_profile_path);
+  }
+  std::vector<at::Tensor> outputs;
+  for (size_t i = 0; i < compiled_engine->out_binding_names.size(); i++) {
+    auto name = compiled_engine->out_binding_names[i];
+    auto dims = compiled_engine->output_allocator->getShapes().at(name);
+    auto dtype = util::TRTDataTypeToScalarType(compiled_engine->exec_ctx->getEngine().getTensorDataType(name.c_str()));
+    at::Tensor output = compiled_engine->output_allocator->getBuffers().at(name).clone().detach();
```
narendasan, line 362: Can we not directly copy to the result buffers like before instead of allocating new memory here? or does it not make a difference.

core/runtime/execute_engine.cpp
```diff
@@ -368,21 +344,32 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
...
+    int64_t prod = 1;
+    for (int i = 0; i < dims.nbDims; ++i) {
+      prod *= dims.d[i];
+    }
+    std::vector<int64_t> dims_vec(dims.nbDims);
+    for (int i = 0; i < dims.nbDims; ++i) {
+      dims_vec[i] = dims.d[i];
     }
+    output = output.reshape(-1).view(dtype).slice(0, 0, prod).reshape(dims_vec);
```
narendasan, line 371: why do we need this?

narendasan (commented):
My feeling is that the Output allocator needs to trigger rerecording of the cuda graph when the buffers change for each runtime

core/runtime/TRTEngine.cpp
```diff
@@ -30,6 +30,27 @@ std::vector<std::string> split(const std::string& str, char delim) {
...
+    void* currentMemory,
+    uint64_t size,
+    uint64_t alignment,
+    cudaStream_t stream) {
+  std::vector<int64_t> shape = {static_cast<int64_t>(size)};
+  auto it = buffers.find(tensorName);
+  if (it == buffers.end() || it->second.sizes() != shape) {
+    buffers[tensorName] = at::empty(shape, at::TensorOptions().dtype(dtypes.at(tensorName)).device(c10::kCUDA));
+  }
+  return buffers[tensorName].data_ptr();
```
narendasan, line 47: You are paying the index cost twice here? just return it.second in the else and return a handle to the new empty tensor after you insert it in the first

core/runtime/TRTEngine.cpp
```diff
@@ -30,6 +30,27 @@ std::vector<std::string> split(const std::string& str, char delim) {
...
+void* DynamicOutputAllocator::reallocateOutputAsync(
+    char const* tensorName,
+    void* currentMemory,
+    uint64_t size,
+    uint64_t alignment,
+    cudaStream_t stream) {
+  std::vector<int64_t> shape = {static_cast<int64_t>(size)};
+  auto it = buffers.find(tensorName);
+  if (it == buffers.end() || it->second.sizes() != shape) {
+    buffers[tensorName] = at::empty(shape, at::TensorOptions().dtype(dtypes.at(tensorName)).device(c10::kCUDA));
```
narendasan, line 45: How do you know what device to use?

narendasan, line 45: As soon as this happens the recapture flag needs to be set

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -23,55 +22,76 @@
...
 
     def set_runtime_states(
         self,
         new_cudagraphs: bool,
-        new_pre_allocated_output: bool,
-        shape_changed: bool,
-    ) -> Tuple[bool, bool, bool]:
-        # Evaluates whether certain conditions are met to enable CUDA Graph recording or to use pre-allocated outputs
-        # based on the current and previous states, as well as input shape has changed
+        input_shape_changed: bool,
```
keehyuna, line 70: same clean up is required in TRTEngine.h

keehyuna (commented):
Some more change should be removed.
core/runtime/TRTEngine.cpp
https://github.com/pytorch/TensorRT/pull/3276/files#diff-70fc185381a0909dbf7f1e62e5a8fb9704a105f2ffcdc8b35822ab28f758ce02R103

core/runtime/TRTEngine.h
pre_allocated states in struct TorchTRTRuntimeStates
https://github.com/pytorch/TensorRT/pull/3276/files#diff-3d6304b0f21f64bdf7867ceac346866d505619401e9c7b6a696cfe3bb567254eR37

docsrc/index.rst
https://github.com/pytorch/TensorRT/pull/3276/files#diff-8b1e4af58f12f81caed8b4870394f5f79f926a3a731eef74cedafda2f2ff2be3R89

zewenli98:
closing in favor of https://github.com/pytorch/TensorRT/pull/3388

PR #3387, opened 2025-02-10
Nccl ops correction changes

apbose:


py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -59,17 +59,25 @@ def aot_torch_tensorrt_aten_backend(
...
-        del settings_aot_autograd["decompositions"][aten.detach]
+    # transpose key deleted since not desirable to lower it to permute
+    to_delete = {
+        key
+        for key in settings_aot_autograd["decompositions"]
+        if "detach" in key._name or "transpose" in key._name
+    }
+
+    for key in to_delete:
+        del settings_aot_autograd["decompositions"][key]
```
peri044, line 76: we have a `remove_detach` lowering pass. Can that help here ?

py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -59,17 +59,26 @@ def aot_torch_tensorrt_aten_backend(
...
+    settings_aot_autograd["decompositions"] = get_decompositions(
         settings.enable_experimental_decompositions
     )
     # This is added since detach lowering leads to alias nodes
     # Error - View operation returned a tensor that is the same as the input base tensor
     # torch nop_decompositions in torch/_decomp/decompositions.py
-    if aten.detach in settings_aot_autograd["decompositions"]:
-        del settings_aot_autograd["decompositions"][aten.detach]
+    # transpose key deleted since not desirable to lower it to permute
+    to_delete = {
```
narendasan, line 69: Will this apply to all cases not just NCCL?

apbose: You mean in the non distributed example? I am not sure about that answer, I added this for the llama3 example since I was issues in the model lowering and it was generating graph breaks at the wrong part, leading to complex input error. It can be added to all cases in case if we want to not lower transpose to permute.

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -364,6 +364,15 @@ def forward(self, *inputs: torch.Tensor) -> torch.Tensor | Tuple[torch.Tensor, .
             (i.contiguous() if isinstance(i, torch.Tensor) else torch.tensor(i).cuda())
             for i in inputs
         ]
+
+        for i, contiguous_input in enumerate(contiguous_inputs):
```
narendasan, line 368: Does the C++ API not need these changes?

tests/py/dynamo/distributed/distributed_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+    world_size = int(os.environ.get("OMPI_COMM_WORLD_SIZE", world_size))
+
+    # Set up environment variable to run with mpirun
+    os.environ["RANK"] = str(local_rank)
+    os.environ["WORLD_SIZE"] = str(world_size)
+    os.environ["MASTER_ADDR"] = "127.0.0.1"
+    os.environ["MASTER_PORT"] = str(port)
+    os.environ["TRTLLM_PLUGINS_PATH"] = (
+        find_repo_root() + "/lib/libnvinfer_plugin_tensorrt_llm.so"
```
narendasan, line 53: Assume a tmp path instead of some arbitrary location in the source tree

tests/py/dynamo/conversion/harness.py
```diff
@@ -351,6 +351,7 @@ def generate_graph(
         enable_passes: bool,
         propagate_shapes: bool = False,
         settings: CompilationSettings = CompilationSettings(),
+        fuse_distributed_ops: bool = False,
```
narendasan, line 354: Why do we need this setting in the harness? Since it only applies to distributed, you should 1. have a test for the pass specifically, 2. have a test for converting the custom op without a pass and 3. have a test that applies the pass then converts the custom op

tests/py/dynamo/distributed/test_nccl_ops.sh
```diff
@@ -0,0 +1,137 @@
...
+        echo "pip is not installed. Please install pip first."
+        exit 1
+    fi
+}
+
+echo "Setting up the environment"
+
+OS="$(uname -s)"
+ARCH="$(uname -m)"
+PYTHON_VERSION="$(python3 -c 'import sys; print(f"cp{sys.version_info.major}{sys.version_info.minor}")')"
```
narendasan, line 75: Why does it matter what python version the system is using? Can we just use the same .so for all python versions?

py/torch_tensorrt/dynamo/backend/backends.py
apbose: That is not helping here, because the graph explicitly does not have detach ops to remove the nodes. Instead it encounters this in https://github.com/pytorch/pytorch/blob/main/torch/_decomp/decompositions.py#L2153. This might be due to the  `%hook_result_3 = call_function[target=torch._dynamo.variables.tensor.prim_to_local](args = (%outputs_3,), kwargs = {})` where it gets the DTensor to local tensor and might need to detach the distributed operation.

tests/py/dynamo/conversion/harness.py
apbose: I had the same doubt actually.
1. I can look for the test for pass
2. Without a pass, I do not think it is possible to encounter the custom op. 
```
gathered_tensor = torch.ops._c10d_functional.all_gather_into_tensor(x, world_size, group_name)
gathered_tensor = torch.ops._c10d_functional.wait_tensor(gathered_tensor)
```
together form the nccl_ops code, and for that the pass is required

3. That is what is done right now. Maybe applying the pass in harness is not the right place. Where would you suggest it to be?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
apbose: I am not clear on this aspect. Could you please let me know what would be required as part of this. I am running distributed python example and did not encounter the requirement for this.

py/torch_tensorrt/dynamo/backend/backends.py
apbose: Regarding the discussion

1. detach: `remove_detach` does not help since the graph explicitly does not have detach ops to remove the nodes. Instead it encounters this in https://github.com/pytorch/pytorch/blob/main/torch/_decomp/decompositions.py#L2153. This might be due to the `%hook_result_3 = call_function[target=torch._dynamo.variables.tensor.prim_to_local](args = (%outputs_3,), kwargs = {}) ` where it gets the DTensor to local tensor and needs to detach the distributed operation. This is in `tensor_parallel_simple_example.py`

2. transpose: `transpose` is more for tackling the `tensor_parallel_llama3.py`. The broad modification I did to handle the complex nos, are:
a. Modifying the placeholder node shape and type
b. Modifying the inputs to the reshape and slice ops with complex inputs
c. Replace the complex tensorrt mul
I see that if I don't decompose transpose to permute, the graph in gpu_0, has output of complex tensor mul as complex64 or complex 128 which goes as input to acc_* graph causing complex input error. Transpose being in the graph helps in it handling the complex input.

Regarding the discussion, would removal of transpose from decomposition affect the result-  I would think no, since this is not removal of op like detach, but instead it is just that we do not lower it to permute. But you could provide me more insights if not

3. Also if we would want it to be model specific and not apply to all models, I think it can be the next step to include it either in the UI or something like we do in the non distributed models, including in the `torch_disabled_decomposition` dictionary which applies to all model. Specifying UI since we are talking about model specific disabled decomposition. As of now since this part of code applies to only distributed model, it should be good to go.

tests/py/dynamo/conversion/harness.py
apbose: The  `mpirun -n 1 --allow-run-as-root python test_distributed_simple_example.py` ideally covers all of it

apbose:
I am not sure why lint is showing this 
```
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running ``pip install "black[jupyter]"``
would reformat /home/runner/work/TensorRT/TensorRT/tests/py/dynamo/distributed/test_nccl_ops.py

Oh no! 💥 💔 💥
1 file would be reformatted, 595 files would be left unchanged.
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running ``pip install "black[jupyter]"``
would reformat /home/runner/work/TensorRT/TensorRT/tests/py/dynamo/distributed/test_nccl_ops.py

All done! ✨ 🍰 ✨
1 file would be reformatted, 595 files would be left unchanged.
```
I see no error in local pre-commit

tests/py/dynamo/conversion/harness.py
apbose: Actually just putting `enable_passes=True`, works in the converter test. Earlier the constant folding was giving an error, but that is resolved.

py/torch_tensorrt/dynamo/backend/backends.py
narendasan: Ok, so the code needs to be restructured to make it clear that is is not the main codepath. 

```py
settings, engine_cache = parse_dynamo_kwargs(kwargs)
    if settings.use_aot_joint_export:
        return _pretraced_backend(gm, sample_inputs, settings, engine_cache)
    logger.debug("Wrapping the backend with aot_autograd\n")
    _pretraced_backend_autograd = functools.partial(
        _pretraced_backend, settings=settings, engine_cache=engine_cache
    )
    settings_aot_autograd = {}
    settings_aot_autograd["decompositions"] = get_decompositions(
        settings.enable_experimental_decompositions
    )
    # This is added since detach lowering leads to alias nodes
    # Error - View operation returned a tensor that is the same as the input base tensor
    # torch nop_decompositions in torch/_decomp/decompositions.py
    # transpose key deleted since not desirable to lower it to permute
    to_delete = {...
```

Its not immediately obvious that most models will run through 
```py
if settings.use_aot_joint_export:
        return _pretraced_backend(gm, sample_inputs, settings, engine_cache)
```

And I am still not sure if such a broad change should be made even in the case of MGMN. How would a user/we know know that this change is needed?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
narendasan: I mean try running the example using the C++ runtime, Id expect that it doesnt handle complex numerics correctly

apbose: It is showing this
```
[rank0]:  File "/root/.pyenv/versions/3.11.11/lib/python3.11/site-packages/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py", line 304, in forward
[rank0]:    outputs: List[torch.Tensor] = torch.ops.tensorrt.execute_engine(
[rank0]:                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:  File "/root/.pyenv/versions/3.11.11/lib/python3.11/site-packages/torch/_ops.py", line 1158, in __call__
[rank0]:    return self._op(*args, **(kwargs or {}))
[rank0]:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: RuntimeError: [Error thrown at core/runtime/execute_engine.cpp:111] Expected inputs[i].dtype() == expected_type to be true but got false
[rank0]: Expected input tensors to have type Float, found type c10::complex<float>
```
for C++ runtime. 
Can we please  record these as future To-dos for this feature, and merge this PR so that the main feature is merged to the main?

py/torch_tensorrt/dynamo/backend/backends.py
apbose: I restructured the code. Also could you please clarify what is meant by -And I am still not sure if such a broad change should be made even in the case of MGMN.

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
narendasan: Uh, I would not consider this feature shipable without supporting the C++ runtime, but it can be in a different PR

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -92,7 +92,7 @@ class CompilationSettings:
         enable_weight_streaming (bool): Enable weight streaming.
         enable_cross_compile_for_windows (bool): By default this is False means TensorRT engines can only be executed on the same platform where they were built.
             True will enable cross-platform compatibility which allows the engine to be built on Linux and run on Windows
-        use_aot_joint_export (bool): Use aot_export_joint_simple, else wrap backend with AOT_autograd, required for distributed tensors
+        USE_DISTRIBUTED_MODE_TRACE (bool): Use aot_export_joint_simple, else wrap backend with AOT_autograd, required for distributed tensors
```
narendasan, line 95: should not be all caps?

py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -52,25 +52,35 @@ def aot_torch_tensorrt_aten_backend(
...
-        del settings_aot_autograd["decompositions"][aten.detach]
-    return aot_autograd(
-        fw_compiler=_pretraced_backend_autograd,
-        decompositions=get_decompositions(settings.enable_experimental_decompositions),
-    )(gm, sample_inputs)
+
+    if settings.use_distributed_mode_trace:
+        logger.debug(
+            "Wrapping the backend with aot_autograd for Distributed examples\n"
+        )
```
peri044, line 59: Consider changing the message to - "Using aot_autograd to trace the graph. Enable this only if the model includes distributed operations"

py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -121,7 +131,7 @@ def _pretraced_backend(
                     )
 
             # Invoke AOTAutograd to translate operators to aten
-            if settings.use_aot_joint_export:
+            if not settings.use_distributed_mode_trace:
```
peri044, line 134: do we need this check now ? since we return the aot_autograd block before ?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
apbose: Yes I agree I can do it in the next PR. As such this change in code is to enable running llama3 distributed end to end on torchTRT for supporting complex inputs. At present it runs on python runtime but then additional changes are needed for C++ runtime. 
But the implementation of NCCL ops plugin which is the main feature and its basic functioning should be working.

py/torch_tensorrt/dynamo/backend/backends.py
apbose: Yes we would be needing this since while wrapping it in `aot_autograd` , we need to pass the pretraced_backend still to the fw_compiler arg in aot_autograd. 
`aot_print_fn = aot_function(fn, fw_compiler=compiler_fn, bw_compiler=compiler_fn)` But then we won't want it to do aot_joint_export tracing there

py/torch_tensorrt/dynamo/backend/backends.py
peri044: - Now the new bool flag (`use_distributed_mode_trace`) is better than previous approach for code restructuring since previous one was not conveying general path (pretraced_backend) clearly.
- These changes should not be model specific ideally. Infact, I was wondering if we really need the above flag. Because it might be not straightforward to users when to actually use it (as you mentioned in the comments). If we only support `all_gather_into_tensor`  and `reduce_scatter_tensor`, we should detect if these ops exist in the graph pre-tracing and then redirect it to `aot_autograd`  accordingly.
- There maybe other distributed variants of these ops which are not supported but users might expect them to be supported once `use_distributed_mode_trace`  is enabled. It's broad.

apbose: Thanks for the detailed analysis and bringing these points up!
1. Yes the new bool flag` (use_distributed_mode_trace)` is better
2. When you say the changes should not be model specific, are you referring to the specific decompositions to be excluded or change of the tracing from `aot_export_joint_simple`.  I would assume the latter- change of the tracing from `aot_export_joint_simple`. Right now for distributed case using the torch.distributed libraries I have encountered these two ops `all_gather_into_tensor` and `reduce_scatter_tensor` which lead to DTensors, requiring this tracing mode as directed by Pytorch folks.  I am yet to encounter the other ops, need to get into the task of running more models with this code. From user perspective, if the user is using distributed libs leading to these ops and consequently these distributed tensors, then this flag would be required. Thats what my idea is right now.
That being said we can discuss further on it if you see the point in changing the trace if these ops are present. That would be a design change and we could work on it in the PR with the C++ changes
3. Yes I do get this point. Other distributed variants of these ops are maybe untested. But I would say this flag tells the user that distributed mode trace should be enabled or not, rather than saying if these ops will work right out of box. We can discuss more on this.
As I mentioned earlier, we could maybe merge this PR for nccl ops support and discuss these points for further improvement.

py/torch_tensorrt/dynamo/backend/backends.py
narendasan: Not sure about only, 

1. have a warning for when users should use this trace mode but arent
2. have an Info level message when this mode is being used `Using AOTAutograd tracer for model lowering` or something to that extent

examples/distributed_inference/tensor_parallel_simple_example.py
```diff
@@ -21,6 +21,35 @@
...
+    compile_options = {
+        "truncate_long_and_double": True,
+        "enabled_precisions": {torch.float32, torch.float16},
+        "use_python_runtime": True,
+        "min_block_size": 1,
+    }
+
+    try:
+        return torch.compile(
+            tp_model, backend=backend, options=compile_options, dynamic=False
```
peri044, line 34: Do you need dynamic=False ? can it be the default None ?

tests/py/dynamo/distributed/test_distributed_simple_example.py
```diff
@@ -16,6 +16,36 @@
...
+    compile_options = {
+        "truncate_long_and_double": True,
+        "enabled_precisions": {torch.float32, torch.float16},
+        "use_python_runtime": True,
+        "min_block_size": 1,
+    }
+
+    try:
+        return torch.compile(
+            tp_model, backend=backend, options=compile_options, dynamic=False
```
peri044, line 30: same question as above

tests/py/dynamo/distributed/test_nccl_ops.sh
```diff
@@ -88,11 +88,17 @@ fi
 URL="https://pypi.nvidia.com/tensorrt-llm/$FILE"
 echo "Downloading $FILE from $URL..."
 
-echo "Downloading ...."
+echo "Downloading here...."
```
peri044, line 91: instead of `here`, consider adding the url name instead ?

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -92,7 +92,7 @@ class CompilationSettings:
         enable_weight_streaming (bool): Enable weight streaming.
         enable_cross_compile_for_windows (bool): By default this is False means TensorRT engines can only be executed on the same platform where they were built.
             True will enable cross-platform compatibility which allows the engine to be built on Linux and run on Windows
-        use_distributed_mode_trace (bool): Use aot_export_joint_simple for models, else wrap backend with AOT_autograd if this is True as required for distributed tensors
+        use_distributed_mode_trace (bool): Using aot_autograd to trace the graph. Enable this only if the model includes distributed operations
```
peri044, line 95: We could skip the word `only` as per this discussion https://github.com/pytorch/TensorRT/pull/3387#discussion_r1982204693

examples/distributed_inference/tensor_parallel_llama3.py
```diff
@@ -51,7 +51,7 @@
             "use_python_runtime": True,
             "workspace_size": 1 << 33,
             "debug": False,
-            "use_aot_joint_export": False,
+            "use_distributed_mode_trace": True,
```
peri044, line 54: would llama3 be supported now after the removal of complex tensor handling ? If so, is this file required?

py/torch_tensorrt/dynamo/lowering/passes/_replace_complex_placeholder_to_tuple.py
```diff
@@ -106,7 +106,12 @@ def update_node_meta(node: torch.fx.Node, fake_mode: FakeTensorMode) -> None:
 
     if op_target in shape_inference_funcs:
         new_shape = shape_inference_funcs[op_target](node)
-        real_tensor = torch.empty(new_shape, dtype=node.meta["val"].dtype)
+        new_node_dtype = None
+        if node.meta["val"].dtype == torch.complex64:
```
narendasan, line 110: Are we still doing complex changes in this PR?

examples/distributed_inference/tensor_parallel_simple_example.py
```diff
@@ -15,13 +15,41 @@
...
+        "min_block_size": 1,
+    }
+
+    try:
+        return torch.compile(
+            tp_model, backend=backend, options=compile_options, dynamic=None
+        )
+    except RuntimeError as e:
+        if (
+            "aot_export is not currently supported with traceable tensor subclass"
```
narendasan, line 38: What does traceable tensor subclass? and is every runtime error from torch compile going to be because of this issue?

examples/distributed_inference/tensor_parallel_simple_example.py
```diff
@@ -15,13 +15,41 @@
...
 
+def compile_tp_model(tp_model, backend):
+    compile_options = {
+        "truncate_long_and_double": True,
+        "enabled_precisions": {torch.float32, torch.float16},
+        "use_python_runtime": True,
+        "min_block_size": 1,
+    }
+
+    try:
```
narendasan, line 32: If this is an example why do you do all of this error handing stuff? just give users a set of settings that work and explain why they need to use those settings?

apbose: I thought this is what we discussed
`Detect these cases, throw a warning recommending the users to turn on this flag (use_aot_autograd_tracer)`

The example shows that 
1. If this case is detected then throw warning
2. Then turn on the flag
3. If the error is different then just throw the runtime error.

  use_distributed_mode_trace (use_aot_autograd_tracer in the discussion above) is directly not provided to the users as a setting, since an external user would not have much idea on when to use the flag.

Or do you want this to be handled in the backend? Then in that case why would we even require the flag then?

examples/distributed_inference/tensor_parallel_simple_example.py
apbose: This is the error which comes when DTensors are the inputs to aot_export_joint_simple which is the case for distributed examples. Yes this would be a runtime time error in every DTensor usecase.

examples/distributed_inference/tensor_parallel_simple_example.py
narendasan: The recommendation should be handled in compiler (both torch.compile and torch.export), it should directly catch this failure case only and make a suggestion that the flag might help. The user should turn on the flag themselves

examples/distributed_inference/tensor_parallel_simple_example.py
narendasan: So you are string matching the error from torch.compile to see if it matches? Is there not a better way to tell? How do they detect this case?

examples/distributed_inference/tensor_parallel_simple_example.py
```diff
@@ -15,13 +15,41 @@
...
+        ):
+            logger.warning(
+                "It is recommended to run the model with use_distributed_mode_trace=True. Running with that option"
+            )
+            compile_options["use_distributed_mode_trace"] = True
+            return torch.compile(
+                tp_model, backend=backend, options=compile_options, dynamic=None
+            )
+        else:
+            logger.debug("The distributed model fails with the following error")
```
narendasan, line 49: Also you never print the actual error

PR #3398, opened 2025-02-14
change in TRT-LLM loading mechanism and exposing aot_joint_export in _compiler.py

apbose:
…ing in dynamo.compile

# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -96,6 +96,7 @@ def cross_compile_for_windows(
     strip_engine_weights: bool = _defaults.STRIP_ENGINE_WEIGHTS,
     immutable_weights: bool = _defaults.IMMUTABLE_WEIGHTS,
     enable_weight_streaming: bool = _defaults.ENABLE_WEIGHT_STREAMING,
+    use_aot_joint_export: bool = _defaults.USE_AOT_JOINT_EXPORT,
     **kwargs: Any,
 ) -> torch.fx.GraphModule:
     """Compile an ExportedProgram module using TensorRT in Linux for Inference in Windows
@@ -169,6 +170,7 @@ def cross_compile_for_windows(
         strip_engine_weights (bool): Strip engine weights from the serialized engine. This is useful when the engine is to be deployed in an environment where the weights are not required.
         immutable_weights (bool): Build non-refittable engines. This is useful for some layers that are not refittable. If this argument is set to true, `strip_engine_weights` and `refit_identical_engine_weights` will be ignored.
         enable_weight_streaming (bool): Enable weight streaming.
+        use_aot_joint_export (bool): Use aot_export_joint_simple, else wrap backend with AOT_autograd, required for distributed tensors
         **kwargs: Any,
     Returns:
         torch.fx.GraphModule: Compiled FX Module, when run it will execute via TensorRT
@@ -326,6 +328,7 @@ def cross_compile_for_windows(
         "immutable_weights": immutable_weights,
         "enable_cross_compile_for_windows": True,
         "enable_weight_streaming": enable_weight_streaming,
+        "use_aot_joint_export": use_aot_joint_export,
     }
 
     # disable the following settings is not supported for cross compilation for windows feature
@@ -413,6 +416,7 @@ def compile(
     strip_engine_weights: bool = _defaults.STRIP_ENGINE_WEIGHTS,
     immutable_weights: bool = _defaults.IMMUTABLE_WEIGHTS,
     enable_weight_streaming: bool = _defaults.ENABLE_WEIGHT_STREAMING,
+    use_aot_joint_export: bool = _defaults.USE_AOT_JOINT_EXPORT,
     **kwargs: Any,
 ) -> torch.fx.GraphModule:
     """Compile an ExportedProgram module for NVIDIA GPUs using TensorRT
@@ -488,6 +492,7 @@ def compile(
         strip_engine_weights (bool): Strip engine weights from the serialized engine. This is useful when the engine is to be deployed in an environment where the weights are not required.
         immutable_weights (bool): Build non-refittable engines. This is useful for some layers that are not refittable. If this argument is set to true, `strip_engine_weights` and `refit_identical_engine_weights` will be ignored.
         enable_weight_streaming (bool): Enable weight streaming.
+
         **kwargs: Any,
     Returns:
         torch.fx.GraphModule: Compiled FX Module, when run it will execute via TensorRT
@@ -662,6 +667,7 @@ def compile(
         "immutable_weights": immutable_weights,
         "enable_cross_compile_for_windows": False,
         "enable_weight_streaming": enable_weight_streaming,
+        "use_aot_joint_export": use_aot_joint_export,
     }
 
     settings = CompilationSettings(**compilation_options)
@@ -950,6 +956,7 @@ def convert_exported_program_to_serialized_trt_engine(
     strip_engine_weights: bool = _defaults.STRIP_ENGINE_WEIGHTS,
     immutable_weights: bool = _defaults.IMMUTABLE_WEIGHTS,
     enable_weight_streaming: bool = _defaults.ENABLE_WEIGHT_STREAMING,
+    use_aot_joint_export: bool = _defaults.USE_AOT_JOINT_EXPORT,
     **kwargs: Any,
 ) -> bytes:
     """Convert an ExportedProgram to a serialized TensorRT engine
@@ -1013,6 +1020,7 @@ def convert_exported_program_to_serialized_trt_engine(
         strip_engine_weights (bool): Strip engine weights from the serialized engine. This is useful when the engine is to be deployed in an environment where the weights are not required.
         immutable_weights (bool): Build non-refittable engines. This is useful for some layers that are not refittable. If this argument is set to true, `strip_engine_weights` and `refit_identical_engine_weights` will be ignored.
         enable_weight_streaming (bool): Enable weight streaming.
+        use_aot_joint_export (bool): Use aot_export_joint_simple, else wrap backend with AOT_autograd, required for distributed tensors
     Returns:
         bytes: Serialized TensorRT engine, can either be saved to a file or deserialized via TensorRT APIs
     """
@@ -1129,6 +1137,7 @@ def convert_exported_program_to_serialized_trt_engine(
         "strip_engine_weights": strip_engine_weights,
         "immutable_weights": immutable_weights,
         "enable_weight_streaming": enable_weight_streaming,
+        "use_aot_joint_export": use_aot_joint_export,
     }
 
     settings = CompilationSettings(**compilation_options)
--- a/py/torch_tensorrt/dynamo/conversion/converter_utils.py
+++ b/py/torch_tensorrt/dynamo/conversion/converter_utils.py
@@ -3,6 +3,8 @@ import ctypes
 import functools
 import logging
 import os
+import subprocess
+import sys
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union, overload
 
 import numpy as np
@@ -12,6 +14,7 @@ import torch_tensorrt.dynamo.conversion.impl as impl
 from torch.fx.node import Argument, Target
 from torch.fx.passes.shape_prop import TensorMetadata
 from torch_tensorrt import _enums
+from torch_tensorrt._enums import Platform
 from torch_tensorrt.dynamo._settings import CompilationSettings
 from torch_tensorrt.dynamo._SourceIR import SourceIR
 from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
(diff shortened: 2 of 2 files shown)


py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -930,57 +933,91 @@ def load_tensorrt_llm() -> bool:
...
-            # Load the shared library
-            handle = ctypes.CDLL(plugin_lib_path)
-            _LOGGER.info(f"Successfully loaded plugin library: {plugin_lib_path}")
-        except OSError as e_os_error:
-            _LOGGER.error(
-                f"Failed to load libnvinfer_plugin_tensorrt_llm.so from {plugin_lib_path}"
-                f"Ensure the path is correct and the library is compatible",
-                exc_info=e_os_error,
+        else:
+            py_version = f"cp{sys.version_info.major}{sys.version_info.minor}"
```
narendasan, line 954: Why do we restrict to cp310 and cp312, It shouldnt matter if we are pulling the whl and unzipping ourselves

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -930,57 +933,91 @@ def load_tensorrt_llm() -> bool:
...
+            if py_version == "cp310" and platform == "linux_aarch64":
+                _LOGGER.warning("No available wheel for python3.10 with Linux aarch64")
+
+            base_url = "https://pypi.nvidia.com/tensorrt-llm/"
+            file_name = (
+                "tensorrt_llm-0.17.0.post1-{py_version}-{py_version}-{platform}.whl"
             )
-            return False
+            download_url = base_url + file_name
+            cmd = ["wget", download_url]
```
narendasan, line 973: What happens if wget is not installed?

py/torch_tensorrt/dynamo/conversion/converter_utils.py
apbose: https://pypi.nvidia.com/tensorrt-llm/ In this since I see the tags for only cp310 and cp312 I added the check

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -498,6 +502,7 @@ def compile(
         enable_weight_streaming (bool): Enable weight streaming.
         tiling_optimization_level (str): The optimization level of tiling strategies. A higher level allows TensorRT to spend more time searching for better tiling strategy. We currently support ["none", "fast", "moderate", "full"].
         l2_limit_for_tiling (int): The target L2 cache usage limit (in bytes) for tiling optimization (default is -1 which means no limit).
+       ç
```
narendasan, line 505: What is this character?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -812,3 +817,123 @@ def is_tegra_platform() -> bool:
...
+
+def download_plugin_lib_path(py_version: str, platform: str) -> str:
+    plugin_lib_path = None
+
+    # Downloading TRT-LLM lib
+    # TODO: check how to fix the 0.18.0 hardcode below
+    base_url = "https://pypi.nvidia.com/tensorrt-llm/"
+    file_name = f"tensorrt_llm-0.18.0-{py_version}-{py_version}-{platform}.whl"
+    download_url = base_url + file_name
+    cmd = ["wget", download_url]
```
narendasan, line 830: Is there not a library in python we can use for this?

apbose: There is wget or requests or urlib libraries which can be used

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -812,3 +819,126 @@ def is_tegra_platform() -> bool:
...
+        except urllib.error.URLError as e:
+            logger.error(
+                f"URL error when trying to download {download_url}: {e.reason}"
+            )
+        except OSError as e:
+            logger.error(f"Local file write error: {e}")
+
+    # Proceeding with the unzip of the wheel file
+    # This will exist if the filename was already downloaded
+    if os.path.exists("./tensorrt_llm/libs/libnvinfer_plugin_tensorrt_llm.so"):
```
narendasan, line 850: what if its windows?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -812,3 +816,128 @@ def is_tegra_platform() -> bool:
     if torch.cuda.get_device_capability() in [(8, 7), (7, 2)]:
         return True
     return False
+
+
+def download_plugin_lib_path(py_version: str, platform: str) -> str:
+    plugin_lib_path = None
+
+    # Downloading TRT-LLM lib
+    # TODO: check how to fix the 0.18.0 hardcode below
```
apbose, line 825: @narendasan any thoughts on where to get the version? Is it something like TRT and TRT-LLM versions fixed for a release? Also for 0.18.0 I do not see win distribution

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -817,3 +821,127 @@ def is_tegra_platform() -> bool:
...
+    plugin_lib_path = None
+
+    # Downloading TRT-LLM lib
+    base_url = "https://pypi.nvidia.com/tensorrt-llm/"
+    file_name = f"tensorrt_llm-{__tensorrt_llm_version__}-{py_version}-{py_version}-{platform}.whl"
+    download_url = base_url + file_name
+    if not (os.path.exists(file_name)):
+        try:
+            logger.debug(f"Downloading {download_url} ...")
+            urllib.request.urlretrieve(download_url, file_name)
```
narendasan: Is this just downloading to the current directory? Instead of polluting users working directories, why not use a temp dir? We use the name `temp_dir / f"torch_tensorrt_{os.username}"` for temporary files

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -817,3 +821,127 @@ def is_tegra_platform() -> bool:
...
+        except OSError as e:
+            logger.error(f"Local file write error: {e}")
+
+    # Proceeding with the unzip of the wheel file
+    # This will exist if the filename was already downloaded
+    if "linux" in platform:
+        lib_filename = "libnvinfer_plugin_tensorrt_llm.so"
+    else:
+        lib_filename = "libnvinfer_plugin_tensorrt_llm.dll"
+    plugin_lib_path = os.path.join("./tensorrt_llm/libs", lib_filename)
```
narendasan: Use a temp dir

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -817,3 +821,127 @@ def is_tegra_platform() -> bool:
...
+        return plugin_lib_path
+    try:
+        import zipfile
+    except ImportError as e:
+        raise ImportError(
+            "zipfile module is required but not found. Please install zipfile"
+        )
+    with zipfile.ZipFile(file_name, "r") as zip_ref:
+        zip_ref.extractall(".")  # Extract to a folder named 'tensorrt_llm'
+        plugin_lib_path = "./tensorrt_llm/libs/" + lib_filename
```
narendasan: Use a temp dir

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -817,3 +821,127 @@ def is_tegra_platform() -> bool:
...
+            "on",
+        )
+        if not use_trtllm_plugin:
+            logger.warning(
+                "Neither TRTLLM_PLUGIN_PATH is set nor is it directed to download the shared library. Please set either of the two to use TRT-LLM libraries in torchTRT"
+            )
+            return False
+        else:
+            # this is used as the default py version
+            py_version = "cp310"
```
narendasan: Pull this default out of the function

narendasan: You can just define it above like `_WHL_CPYTHON_VERSION=cp310`

narendasan (commented):
Use a temp directory to save the wheel and unzipped wheel

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -820,3 +839,176 @@ def is_tegra_platform() -> bool:
     if torch.cuda.get_device_capability() in [(8, 7), (7, 2)]:
         return True
     return False
+
+
+@contextmanager
```
narendasan: Why do we need this to be a context manager?

apbose: The contextmanager yields the path to the extracted .so file, keeping the temporary directory alive only until the plugin is loaded. Once the with block exits, the directory is automatically cleaned up

tests/py/dynamo/distributed/test_nccl_ops.py
```diff
@@ -70,6 +82,7 @@ def forward(self, x):
             use_dynamo_tracer=True,
             enable_passes=True,
         )
+        dist.destroy_process_group()
```
apbose: Wanted to clarify on this, will pytest always run the first and second in order? I see that in my local runs

py/torch_tensorrt/dynamo/utils.py
narendasan: Shouldnt the file be persistent in the temp dir so that the next time the file does not need to be redownloaded?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -820,3 +839,201 @@ def is_tegra_platform() -> bool:
...
+        plugin_lib_path (str): Path to the shared TensorRT-LLM plugin library.
+
+    Returns:
+        bool: True if successful, False otherwise.
+    """
+    try:
+        handle = ctypes.CDLL(plugin_lib_path)
+        logger.info(f"Successfully loaded plugin library: {plugin_lib_path}")
+    except OSError as e_os_error:
+        if "libmpi" in str(e_os_error):
```
narendasan, line 1027: These string matching error hints can be misleading. Maybe change the wording to something like 

```
"Failed to load libnvinfer_plugin_tensorrt_llm.so from {path}, got error {actual_error} (hint: libmpi.so is a necessary dependency, ensure that OpenMPI is installed on you system)"

File: tests/py/dynamo/distributed/test_nccl_ops.py (hunk 12 above)
Comment (narendasan) [reply tests/py/dynamo/distributed/test_nccl_ops.py]: You should not assume a run order

File: py/torch_tensorrt/dynamo/utils.py (hunk 11 above)
Comment (apbose) [reply py/torch_tensorrt/dynamo/utils.py]: The downloaded archive is cached persistently, but the extracted contents (like the .so plugin) are placed in a temporary directory, which is managed via the context manager to clean it up after use.
This prevents unbounded growth of large intermediate files (the extracted plugin is ~3GB for Linux and 2.5GB on windows). Since extraction is relatively fast and also a one time thing, I guess it’s a good trade-off for keeping the disk usage under control.

File: tests/py/dynamo/distributed/test_nccl_ops.py (hunk 12 above)
Comment (apbose) [reply tests/py/dynamo/distributed/test_nccl_ops.py]: Ok I will clean this up

File: py/torch_tensorrt/dynamo/utils.py (hunk 11 above)
Comment (narendasan) [reply py/torch_tensorrt/dynamo/utils.py]: > The downloaded archive is cached persistently, but the extracted contents (like the .so plugin) are placed in a temporary directory,

Can you explain more? 

Shouldnt we just cache the extracted contents and remove everything else?

Comment (apbose) [reply py/torch_tensorrt/dynamo/utils.py]: You're right — it's ultimately a design choice, there are two parts

The download from the URL, which can be slow depending on network conditions, so it makes sense to persist the downloaded wheel/archive to avoid redundant downloads across runs.

The extracted contents (e.g., .so from the wheel) which is relatively fast and deterministic with the unzip, so I've kept this temporary for now. This helps avoid unnecessary disk usage, since the extracted plugin is ~3GB.

That said, if we expect repeated usage or loading across sessions, I can switch to persistent caching for both the download and the extraction if that’s preferable —  let me know your thoughts.

Comment (narendasan) [reply py/torch_tensorrt/dynamo/utils.py]: What is the compressed size? 

My inclination is that if there isnt that big a difference we should keep the one file we need live and delete everything else even if its fairly large since that will mean that the runtime in later compiles will be the fastest

Comment (apbose) [reply py/torch_tensorrt/dynamo/utils.py]: The compressed size is 1.1GB. Yeah we need the .so, so I will delete the rest then

**Review by narendasan — COMMENTED** [review]
Seems like its close are there any tests for the downloader? like verifying the correct file is downloaded and available?

**Comment by apbose** [comment]
The test t`ests/py/dynamo/distributed/test_nccl_ops.py` should verify it. The `USE_TRTLLM_PLUGINS=1 pytest distributed/test_nccl_ops.py` is the command. It is not part of CI though

File: .github/scripts/install-torch-tensorrt.sh [hunk 14]
```diff
@@ -12,6 +12,12 @@ if [[ $(uname -m) == "aarch64" ]]; then
     install_cuda_aarch64
 fi
 
+if [[ "$(uname -s)" == "Linux" && "$(uname -m)" == "x86_64" ]]; then
```
narendasan: Can this just be called in the test script instead installing it for all tests?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -820,3 +837,215 @@ def is_tegra_platform() -> bool:
...
+        TensorRT-LLM plugins for NCCL backend are not supported on:
+        - Windows platforms
+        - Jetson devices (aarch64 architecture)
+    """
+    if "windows" in platform:
+        logger.info(
+            "TensorRT-LLM plugins for NCCL backend are not supported on Windows"
+        )
+        return False
+    if "aarch64" in platform:
```
narendasan: What about SBSA?

apbose: aarch64 should take care of it, right? I should change the log message to maybe "TensorRT-LLM plugins for NCCL backend are not supported on Jetson and sbsa devices (aarch64)"?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -820,3 +837,218 @@ def is_tegra_platform() -> bool:
...
+        - Orin, Xavier, or Tegra devices (aarch64 architecture)
+
+    """
+    if "windows" in platform:
+        logger.info(
+            "TensorRT-LLM plugins for NCCL backend are not supported on Windows"
+        )
+        return False
+    if torch.cuda.is_available():
+        device_name = torch.cuda.get_device_name().lower()
```
narendasan: Did you test this, I was recommending using this: https://docs.python.org/ro/3/library/platform.html#platform.release

We use it here: https://github.com/pytorch/TensorRT/blob/e760c3c9b479ab531738c95736dd7a38b135517c/pyproject.toml#L54

.github/workflows/build-test-linux-x86_64.yml
```diff
@@ -337,6 +337,39 @@ jobs:
         python -m pytest -ra -n 4 --junitxml=${RUNNER_TEST_RESULTS_DIR}/tests_py_core_test_results.xml .
         popd
 
+  tests-py-distributed:
```
narendasan, line 343: Shouldnt there be tests for sbsa? even though we dont run them right now, it should be there for when we get a sbsa runner

py/torch_tensorrt/dynamo/utils.py
apbose: ok changed to the above

apbose:
Closing in favor of #3829

Issue #3415, opened 2025-02-25
🐛 [Bug]  test_lowering_full_like_to_full_dynamic_module fails in CI


peri044:
##  Bug Description

```py
FAILED test_decompositions.py::TestLowering::test_lowering_full_like_to_full_dynamic_module - torch.fx.experimental.symbolic_shapes.ConstraintViolationError: Constraints violated (L['x'].size()[0])! For more information, run with TORCH_LOGS="+dynamic".
```

https://github.com/pytorch/TensorRT/actions/runs/13168805591/job/36755789477?pr=3382

## Environment

> Build information about Torch-TensorRT can be found by turning on debug messages

 - Torch-TensorRT Version (e.g. 1.0.0):
 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source):
 - Build command you used (if compiling from source):
 - Are you using local sources or building from archives:
 - Python version:
 - CUDA version:
 - GPU models and configuration:
 - Any other relevant information:

PR #3416, opened 2025-02-26
fix: Fix CI issues due to unintended fake tensor creation in torch.compile tests

peri044:
# Description
FakeTensors are being created instead of torch.Tensors within a torch.compile session. Although TRT engine gets compiled successfully, torch dynamo reports some strange guard failures due to these unused FakeTensors within the session. 

Fixes https://github.com/pytorch/TensorRT/issues/3415
Fixes https://github.com/pytorch/TensorRT/issues/3393

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/utils.py
```diff
@@ -256,48 +257,54 @@ def prepare_inputs(
...
+            return (
+                torchtrt_input_list
+                if isinstance(inputs, list)
+                else tuple(torchtrt_input_list)
             )
-            torchtrt_inputs_dict[key] = torchtrt_input
 
-        return torchtrt_inputs_dict
+        elif isinstance(inputs, dict):
+            torchtrt_inputs_dict: Dict[Any, Any] = dict()
```
narendasan, line 293: What is the intended use case for this? Can you add an example as a comment for each case?

narendasan, line 293: Should this be str: Any?

narendasan (approved):
Seems fine, small details

py/torch_tensorrt/dynamo/utils.py
peri044: Yes, I think so. This was not changed but the linter just restructured it. I'll modify it

PR #3420, opened 2025-02-27
Fix assertions in slice_scatter decomposition

kkimmk:
- [fix: allow slice_scatter decomposition with SymInt parameters for the case when returning the source tensor](https://github.com/pytorch/TensorRT/commit/d3454ddfab75569e1e4a343af24484ae7497edc2)

# Description

Originally, the assertions only allowed `int` types for `start`, `end`, and `step`.  However, `end` can be a `SymInt` when set to `dim_size`, which only causes issues latter with `range(start, end, step)`. This PR allows returning `src_tensor` when the condition is met with a `SymInt` typed `end`.

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -203,17 +203,17 @@ def slice_scatter_decomposition(
     if step is None:
         step = 1
 
-    # Ensure start, end, and step are all integers
-    assert isinstance(start, int), "start must be an integer"
-    assert isinstance(end, int), "end must be an integer"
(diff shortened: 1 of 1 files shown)


apbose (approved):

apbose:
Ok got it. Looks like it is for dynamic shape case.

As such yes, the decomposition at present is not meant for dynamic shaped input. This code modification covers a very specific dynamic use case. But this change should work and should not affect the rest of the cases. So LGTM. But we need to expand these to support dynamic cases.

narendasan:
@apbose what is the AI here? Should we merge the PR and work on broader cases later or is the expectation that this PR would have these changes?

apbose:
We can merge this and work on the broader cases later.

Issue #3409, opened 2025-02-24
🐛 [Bug] Error Code 3: API Usage Error on IConvolutionLayer::setPaddingNd (Paligemma2)


chohk88:
##  Bug Description
Attempting torch.compile (backend = torch_tensorrt) the `google/paligemma2-3b-pt-224` model, I encountered below message for both torch_tensorrt 2.6 and 2.7 dev. 

```
INFO:torch_tensorrt.dynamo.conversion._TRTInterpreter:Converted node _to_copy [aten._to_copy.default] (Inputs: (arg0_1: (1, 3, 224, 224)@torch.float16) | Outputs: (_to_copy: (1, 3, 224, 224)@torch.float16))
DEBUG:torch_tensorrt.dynamo.conversion._TRTInterpreter:Converting node arg1_1 (kind: arg1_1, args: ())
DEBUG:torch_tensorrt.dynamo.conversion._TRTInterpreter:Adding input to in-progress INetwork: arg1_1 [shape=[1152, 3, 14, 14], dtype=DataType.HALF]
INFO:torch_tensorrt.dynamo.conversion._TRTInterpreter:Converted node arg1_1 [arg1_1] (Inputs: () | Outputs: (arg1_1: (1152, 3, 14, 14)@torch.float16))
DEBUG:torch_tensorrt.dynamo.conversion._TRTInterpreter:Converting node arg2_1 (kind: arg2_1, args: ())
DEBUG:torch_tensorrt.dynamo.conversion._TRTInterpreter:Adding input to in-progress INetwork: arg2_1 [shape=[1152], dtype=DataType.HALF]
INFO:torch_tensorrt.dynamo.conversion._TRTInterpreter:Converted node arg2_1 [arg2_1] (Inputs: () | Outputs: (arg2_1: (1152,)@torch.float16))
DEBUG:torch_tensorrt.dynamo.conversion._TRTInterpreter:Converting node convolution (kind: aten.convolution.default, args: ('_to_copy <Node>', 'arg1_1 <Node>', 'arg2_1 <Node>', ['14 <int>', '14 <int>'], ['0 <int>'], ['1 <int>', '1 <int>'], 'False <bool>', ['0 <int>'], '1 <int>'))
DEBUG:torch_tensorrt.dynamo.conversion._ConverterRegistry:Converter options for aten.convolution.default: 1
[...]

PR #3428, opened 2025-03-05
fix: conv parameter check failure

chohk88:
## Description

This PR fixes an issue where padding values given as a single-element list (e.g., `[0]` and `[1]`) were not properly converted in `convNd`, leading to a TensorRT error:

```
ERROR:torch_tensorrt [TensorRT Conversion Context]:IConvolutionLayer::setPaddingNd: Error Code 3: API Usage Error (Parameter check failed, condition: (padding.nbDims == 2 || padding.nbDims == 3) && allDimsGtEq(padding, 0) && allDimsLtEq(padding, kMAX_PADDING). )
```

For the padding value `[1]`, this also caused shape mismatches, such as:

```
AssertionError: The values for attribute 'shape' do not match: torch.Size([1, 6, 32, 32]) != torch.Size([1, 6, 34, 34]).
```

If `padding` is `[0]`, it still triggers a TensorRT `paramater check failure error`, but the final output is correct. This is a minor issue that doesn’t affect functionality.

#### Fix
- Now, if `padding` is a list with one element, it is expanded properly (`(1,1)` for conv2d, `(1,1,1)` for conv3d, etc.).
- Verified with additional test cases

Fixes #([issue](https://github.com/pytorch/TensorRT/issues/3409))

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2468,13 +2468,14 @@ def aten_ops_convolution(
     name: str,
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     is_transposed = args[6]
+    is_conv1d = len(args[0].shape) == 3
     if not is_transposed:
         return impl.conv.convNd(
             ctx,
             target,
             source_ir=SourceIR.ATEN,
             name=name,
-            is_conv1d=len(args[3]) == 1,
+            is_conv1d=is_conv1d,
             input=args[0],
             weight=args[1],
             bias=args_bounds_check(args, 2, None),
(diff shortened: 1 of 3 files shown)


zewenli98 (commented):
Good catch! Can you confirm if `stride` or `dilation` need to do the same changes or not?

chohk88:
> Good catch! Can you confirm if `stride` or `dilation` need to do the same changes or not?

Good catch! I missed that.

The same error occurs for list_dilation, e.g., param("list_dilation", 2, dilation=[3]), just like with padding. I have updated the code to properly handle list inputs for dilation and stride, ensuring they are correctly converted into tuples for the conv layer. 

Additionally, I have revised the method for checking 1D convolution.

zewenli98 (approved):
cool, LGTM!

PR #3455, opened 2025-03-31
chore: reenable py313

zewenli98:
# Description

Reenabled Python 3.13 builds since TensorRT has 3.13 wheels available now for TensorRT 10.9.

## Type of change

# Checklist:

### DIFF
--- a/.github/scripts/filter-matrix.py
+++ b/.github/scripts/filter-matrix.py
@@ -4,7 +4,7 @@ import argparse
 import json
 import sys
 
-disabled_python_versions = "3.13"
+disabled_python_versions = ["3.13"]
 
 
 def main(args: list[str]) -> None:
--- a/.github/scripts/filter-matrix.py
+++ b/.github/scripts/filter-matrix.py
@@ -3,8 +3,9 @@
 import argparse
 import json
 import sys
+from typing import List
 
-disabled_python_versions = ["3.13"]
+disabled_python_versions: List[str] = []
 
 
 def main(args: list[str]) -> None:
--- a/.github/scripts/generate_binary_build_matrix.py
+++ b/.github/scripts/generate_binary_build_matrix.py
@@ -19,9 +19,9 @@ import sys
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
 PYTHON_ARCHES_DICT = {
-    "nightly": ["3.9", "3.10", "3.11", "3.12"],
-    "test": ["3.9", "3.10", "3.11", "3.12"],
-    "release": ["3.9", "3.10", "3.11", "3.12"],
+    "nightly": ["3.9", "3.10", "3.11", "3.12", "3.13"],
+    "test": ["3.9", "3.10", "3.11", "3.12", "3.13"],
+    "release": ["3.9", "3.10", "3.11", "3.12", "3.13"],
 }
 CUDA_ARCHES_DICT = {
     "nightly": ["11.8", "12.6", "12.8"],
--- a/.github/scripts/generate_binary_build_matrix.py
+++ b/.github/scripts/generate_binary_build_matrix.py
@@ -18,16 +18,16 @@ import os
 import sys
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
+PYTHON_VERSIONS_FOR_PR_BUILD = ["3.11"]
 PYTHON_ARCHES_DICT = {
     "nightly": ["3.9", "3.10", "3.11", "3.12", "3.13", "3.13t"],
     "test": ["3.9", "3.10", "3.11", "3.12", "3.13", "3.13t"],
     "release": ["3.9", "3.10", "3.11", "3.12", "3.13"],
 }
-PYTHON_VERSIONS_FOR_PR_BUILD = ["3.11"]
 CUDA_ARCHES_DICT = {
     "nightly": ["11.8", "12.6", "12.8"],
     "test": ["11.8", "12.6", "12.8"],
-    "release": ["11.8", "12.6", "12.8"],
+    "release": ["11.8", "12.4", "12.6"],
 }
 ROCM_ARCHES_DICT = {
     "nightly": ["6.1", "6.2"],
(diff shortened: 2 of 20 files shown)


.github/scripts/filter-matrix.py
```diff
@@ -4,7 +4,7 @@
 import json
 import sys
 
-disabled_python_versions = "3.13"
+disabled_python_versions = ["3.13"]
```
narendasan, line 7: Why is it still disabled?

zewenli98: I think python 3.13 is only supported in TRT 10.9 right? so I'm not sure if directly removing the version here would mess up the tests of other TRT versions.

.github/workflows/build-test-linux.yml
```diff
@@ -37,6 +37,10 @@ jobs:
       - uses: actions/checkout@v4
         with:
           repository: pytorch/tensorrt
+      - name: Install Rust
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+          source $HOME/.cargo/env
```
HolyWu, line 43: I think you don't need to install rust. Instead, you should try to bump the version of `transformers` in `tests/py/requirements.txt`. `transformers 4.40.2` depended on `tokenizers 0.19.x`, but `tokenizers 0.19.x` didn't have wheels for cp313. That's why pip downloaded the source archive of `tokenizers` and then compiled unsuccessfully.

.github/scripts/generate_binary_build_matrix.py
```diff
@@ -19,9 +19,9 @@
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
 PYTHON_ARCHES_DICT = {
-    "nightly": ["3.9", "3.10", "3.11", "3.12"],
-    "test": ["3.9", "3.10", "3.11", "3.12"],
-    "release": ["3.9", "3.10", "3.11", "3.12"],
+    "nightly": ["3.9", "3.10", "3.11", "3.12", "3.13"],
+    "test": ["3.9", "3.10", "3.11", "3.12", "3.13"],
```
HolyWu, line 23: Torch has separate "3.13" and "3.13t" values for `nightly` and `test`. I'm not sure if you should add "3.13t" as well.

zewenli98: Thanks for the inputs! I think it's good to give it a chance.

.github/workflows/build-test-linux.yml
zewenli98: The updated `transformers` doesn't match the results of BERT, compared with old transformers and TRT, so we fixed the version here. Any idea?

HolyWu: The `test_bert_base_uncased` in `test_models` and `test_models_export` was flawed. The inputs were defined as `torch.randint(0, 1, ...)`, which returned a meaningless tensor filled with all zeros because the value of the second arg `high` is exclusive. To get a useful tensor the `high` arg should be at least 2, like https://pytorch.org/TensorRT/tutorials/_rendered_examples/dynamo/torch_compile_transformers_example.html.

.github/scripts/generate_binary_build_matrix.py
narendasan: Take a look at pytorch/test_infra and use their settings

tests/py/requirements.txt
```diff
@@ -8,6 +8,6 @@ pytest>=8.2.1
 pytest-xdist>=3.6.1
 pyyaml
 timm>=1.0.3
-transformers==4.40.2
-nvidia-modelopt[deploy,hf,torch]~=0.17.0
+transformers==4.40.2; python_version < "3.13"
```
narendasan, line 11: Id rather upgrade this if we can just because its easy to forget about settings like this

peri044 (approved):
LGTM

.github/scripts/generate_binary_build_matrix.py
```diff
@@ -18,15 +18,16 @@
...
+    "release": ["3.9", "3.10", "3.11", "3.12", "3.13"],
 }
 CUDA_ARCHES_DICT = {
     "nightly": ["11.8", "12.6", "12.8"],
     "test": ["11.8", "12.6", "12.8"],
-    "release": ["11.8", "12.6", "12.8"],
+    "release": ["11.8", "12.4", "12.6"],
```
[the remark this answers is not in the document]
zewenli98: I think we need to follow [pytorch test-infra](https://github.com/pytorch/test-infra/blob/729ba27b52ec12014fe287bee1eeda6e99d6eac2/tools/scripts/generate_binary_build_matrix.py#L29-L33) right?
@narendasan can you help confirm?

tests/py/requirements.txt
```diff
@@ -8,6 +8,6 @@ pytest>=8.2.1
 pytest-xdist>=3.6.1
 pyyaml
 timm>=1.0.3
-transformers==4.40.2
-nvidia-modelopt[deploy,hf,torch]~=0.17.0
+transformers==4.49.0
+nvidia-modelopt[deploy,hf,torch]~=0.17.0; python_version < "3.13"
```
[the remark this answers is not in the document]
zewenli98: right, I think we've already directly skip quantization tests like https://github.com/pytorch/TensorRT/blob/d3eb817b4368c0f77ba6cfea814891a5ddb63b85/tests/py/dynamo/models/test_models_export.py#L201-L220
Users may not be aware of it though

.github/scripts/generate_binary_build_matrix.py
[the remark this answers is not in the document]
narendasan: Yeah this is what is in test-infra

py/torch_tensorrt/_features.py
```diff
@@ -62,6 +64,22 @@ def not_implemented(*args: List[Any], **kwargs: Dict[str, Any]) -> Any:
     return wrapper
 
 
+def needs_refit(f: Callable[..., Any]) -> Callable[..., Any]:
```
narendasan, line 67: Maybe we can make this a bit more generic (like for any feature in the FeatureSet)

narendasan (commented):
Think it mostly looks good, I think we should think about making the requires decorator a bit more generic and unify `needs_torch_tensorrt_runtime` and `needs_refit`, maybe something like `needs_feature`. Maybe something for after 2.7

zewenli98:
> Think it mostly looks good, I think we should think about making the requires decorator a bit more generic and unify `needs_torch_tensorrt_runtime` and `needs_refit`, maybe something like `needs_feature`. Maybe something for after 2.7

sure, will think about it for the next release.

Issue #3410, opened 2025-02-24
🐛 [Bug] Unsupported ops : torch.ops.aten.masked_scatter.default (Paligemma2)


chohk88:
##  Bug Description
Attempting torch.compile (backend = torch_tensorrt) the `google/paligemma2-3b-pt-224` model, there is an unsupported ops like below:

```
DEBUG:torch_tensorrt.dynamo.backend.backends:Pre-AOT Autograd graph:
graph():
    %l_image_features_ : torch.Tensor [num_users=1] = placeholder[target=L_image_features_]
    %l_inputs_embeds_ : torch.Tensor [num_users=1] = placeholder[target=L_inputs_embeds_]
    %l_special_image_mask_ : torch.Tensor [num_users=1] = placeholder[target=L_special_image_mask_]
    %image_features : [num_users=2] = call_method[target=to](args = (%l_image_features_, cuda:0, torch.float16), kwargs = {})
    %inputs_embeds : [num_users=1] = call_method[target=masked_scatter](args = (%l_inputs_embeds_, %l_special_image_mask_, %image_features), kwargs = {})
    return (inputs_embeds, image_features)
DEBUG:torch_tensorrt.dynamo.lowering.passes.repair_input_aliasing:Inserted auxiliary clone nodes for placeholders:
graph():
    %l_image_features_ : torch.Tensor [num_users=1] = placeholder[target=L_image_features_]
    %l_inputs_embeds_ : torch.Tensor [num_users=1] = placeholder[target=L_inputs_embeds_]
    %l_special_image_mask_ : torch.Tensor [num_users=1] = placeholder[target=L_special_image_mask_]
    %clone_default_2 : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%l_special_image_mask_,), kwargs = {})
    %clone_default_1 : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%l_inputs_embeds_,), kwargs = {})
    %clone_default : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%l_image_features_,), kwargs = {})
    %image_features : [num_users=2] = call_method[target=to](args = (%clone_default, cuda:0, torch.float16), kwargs = {})
    %inputs_embeds : [num_users=1] = call_method[target=masked_scatter](args = (%clone_default_1, %clone_default_2, %image_features), kwargs = {})
    return (inputs_embeds, image_features)
[...]

PR #3476, opened 2025-04-15
fix: support masked_scatter by lowering path and corner case of maske…

chohk88:
# Description

Implemented support for `masked_scatter` in the lowering path, referring to [this implementation in PyTorch Inductor](https://github.com/pytorch/pytorch/blob/924a247fbbe32063966533535b1cc0cf474ef1ed/torch/_inductor/decomposition.py#L831).

Fixes # ([issue](https://github.com/pytorch/TensorRT/issues/3410))

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -196,8 +196,10 @@ def slice_scatter_decomposition(
 ) -> torch.Tensor:
     dim_size = input_tensor.shape[dim]
     device_input_tensor = input_tensor.device
+
+    start = 0 if start is None else start  # Ensure start is int
     start = get_positive_dim(start, input_tensor.shape[dim])
-    if end is None:
+    if end is None:  # Ensure end is int
         end = dim_size
     end = get_positive_dim(end, input_tensor.shape[dim])
     if step is None:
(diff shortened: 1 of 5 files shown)


py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -575,6 +577,53 @@ def cudnn_grid_sampler_decomposition(
...
+    gathered = source_flat.gather(0, safe_idx)
+
+    # 5) Replace masked positions in input with gathered values
+    replaced = torch.where(mask_flat, gathered, input_flat)
+
+    # 6) Reshape the result to match the original broadcasted shape
+    return replaced.view(input_b.shape)
+
+
 def get_decompositions(
```
apbose, line 627: @chohk88 the implementation looks perfect as such.'
Just a slight detail I was wondering. Torch mentions that it supports broadcasting between mask and self tensor, but I see this example working too, basically broadcasting between source and mask.

import torch
input = torch.zeros(3, 3)
mask = torch.tensor([[1, 0, 1], [0, 0, 1], [0, 0, 0]], dtype=torch.bool)
source = torch.tensor([2, 3, 4])
out = input.masked_scatter_(mask, source)
print("out is", out)

chohk88: Thanks @apbose, great question! I checked PyTorch’s behavior and it:

1. Broadcasts only input and mask (not source)
2. Flattens source as-is
3. Uses only the first mask.sum() elements of source, ignoring any extras

For example, with source=[2,3,4,5] and three True mask positions, PyTorch applies [2,3,4] and drops the 5 without error.

Our current decomposition does exactly that (broadcasts input/mask, flattens source, then cumsum→gather→where), so no code changes are needed. Let me know if you spot any other edge cases!

image
image
image

apbose: great! Thanks for the detailed explanation

PR #3486, opened 2025-04-22
Cross compile guard

apbose:
The PR addresses the following-
1. Enables the windows cross compile feature based on if platform linux and libnvinfer_builder_resource_win.so.* is present
2. Adds this as a decorator for the cross compile API
3. This utility is also added for skip tests

### DIFF
--- a/py/torch_tensorrt/_features.py
+++ b/py/torch_tensorrt/_features.py
@@ -43,8 +43,13 @@ _TS_FE_AVAIL = os.path.isfile(linked_file_full_path)
 _TORCHTRT_RT_AVAIL = _TS_FE_AVAIL or os.path.isfile(linked_file_runtime_full_path)
 _DYNAMO_FE_AVAIL = version.parse(sanitized_torch_version()) >= version.parse("2.1.dev")
 _FX_FE_AVAIL = True
+<<<<<<< HEAD
 _REFIT_AVAIL = True
 _WINDOWS_CROSS_COMPILE = check_cross_compile_trt_win_lib
+=======
+_REFIT_AVAIL = version.parse(sys.version.split()[0]) < version.parse("3.13")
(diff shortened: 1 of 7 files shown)


py/torch_tensorrt/_utils.py
```diff
@@ -9,3 +13,27 @@ def sanitized_torch_version() -> Any:
         if ".nv" not in torch.__version__
         else torch.__version__.split(".nv")[0]
     )
+
+
+def check_cross_compile_trt_win_lib():
+    if sys.platform.startswith("linux"):
+        LINUX_PATHS = ["/usr/local/cuda-12.8/lib64", "/usr/lib", "/usr/lib64"]
```
narendasan, line 20: I dont think this works generically, there needs to be a way to see of the library has been loaded aready

apbose: I took this code snippet from 
`py/torch_tensorrt/__init__.py` where we check the tensorrt libs while importing torch_tensorrt. Since this is included in features I am putting this code snippet in _utils, where it checks if the _win*.so is present or not to be loaded via ctypes.CDLL ( can use ctypes.cdll.LoadLibrary() too). What is your thoughts on how to do this

py/torch_tensorrt/_features.py
```diff
@@ -37,10 +41,21 @@
 _TORCHTRT_RT_AVAIL = _TS_FE_AVAIL or os.path.isfile(linked_file_runtime_full_path)
 _DYNAMO_FE_AVAIL = version.parse(sanitized_torch_version()) >= version.parse("2.1.dev")
 _FX_FE_AVAIL = True
+<<<<<<< HEAD
```
narendasan, line 44: Committed merge conflicts

PR #3487, opened 2025-04-22
disabling dla args for hope igx platform

apbose:
This PR disables the test for hope igx platform for TRT > 10.8


narendasan (approved):
LGTM

PR #3490, opened 2025-04-24
chore: Re-test BF16 fixes on main, refactor test suite

peri044:
# Description

Fixes # (issue)

## Type of change

# Checklist:


.github/workflows/build-test-linux.yml
```diff
@@ -173,7 +173,13 @@ jobs:
         cd tests/py
         python -m pip install -r requirements.txt
         cd dynamo
-        python -m pytest -ra --junitxml=${RUNNER_TEST_RESULTS_DIR}/dyn_models_export.xml --ir dynamo models/
+        python -m pytest -ra --junitxml=${RUNNER_TEST_RESULTS_DIR}/dynamo_models.xml --ir dynamo models/test_models.py
```
narendasan, line 176: Why do these need to be separated?

narendasan: Only issue is that people will forget to add their tests to the list here

narendasan (commented):
Seems mostly fine, I think if we need separate tests there are a few options, either we glob the files and run a loop in the yaml or we use something like -n 1 to run tests serially.

PR #3527, opened 2025-05-20
feat: Refactor LLM model zoo and add KV cache support

peri044:
# Description

This PR redesigns our LLM model compilation, unifies it, fixes output mismatch and performance issues. This PR also implements KV caching using native TensorRT. 

Fixes # (issue)

## Type of change

# Checklist:


tools/llm/cache_utils.py
```diff
@@ -0,0 +1,202 @@
...
+from torch._export.utils import _detect_fake_mode_from_gm
+from torch._ops import OpOverloadPacket
+from torch._subclasses.fake_tensor import FakeTensor, FakeTensorMode
+from torch.fx import Graph, GraphModule, Node
+from torch.fx.node import Target
+from torch.fx.passes.shape_prop import _extract_tensor_metadata
+from torch.utils._pytree import _LEAF_SPEC
+
+
+@torch_tensorrt.dynamo.conversion.dynamo_tensorrt_converter(
```
narendasan: Do we want to upstream this eventually?

tools/llm/sdpa_converter.py
```diff
@@ -62,25 +62,15 @@ def scaled_dot_product_attention(
 ) -> TRTTensor:
```
narendasan: I would put the torch_trt extensions in a subdirectory all together. like //tools/llm/torch_trt_ext/

tools/llm/test_qwen2.5_components.py
```diff
@@ -0,0 +1,193 @@
+import torch
```
narendasan, line 1: Should we create a `tests/py/tools/llm`?

tools/llm/cache_utils.py
```diff
@@ -0,0 +1,202 @@
...
+
+    return random_tensors
+
+
+def add_graph_input(
+    gm: GraphModule, name: str, val: Optional[torch.Tensor] = None, dynamic_shape=None
+) -> Node:
+    """Add a graph input to the given GraphModule and return the newly created node.
+
+    NOTE: function does NOT do any graph canonicalization. This is left to the user!
```
narendasan, line 110: If you dont want users to hit this directly consider prefixing the method name with `_` like `_add_graph_input`

tools/llm/cache_utils.py
peri044: eventually yes. But maybe not for this release.

tools/llm/test_qwen2.5_components.py
peri044: We can. These tests helped me debug accuracy/functionality issues during the development. It would take some time to get them working in CI. I can add them as a part of cherry-pick PR after this. What do you think ?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -837,6 +859,7 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
             continue
         submodule_node_dict[node.name] = node
 
+    preserve_module_specs(original_in_spec, original_out_spec, partitioned_module)
```
zewenli98, line 869: Should set spec for `partitioned_module`'s children?

examples/dynamo/weight_streaming_example.py
```diff
@@ -32,7 +32,43 @@
 import torch
 import torch_tensorrt
 from transformers import AutoModelForCausalLM
-from utils import export_llm
+
+
+def export_llm(model, inputs, min_seq_len=1, max_seq_len=16):
```
zewenli98, line 37: why not directly call tools/llm/utils.py?

zewenli98 (commented):
Mostly LGTM. just some questions

examples/dynamo/weight_streaming_example.py
peri044: We have to use sys.append to include the path to tools/llm/utils.py to do that. I wanted this example to be standalone so I just copied this function.

tools/llm/cache_utils.py
peri044: removed it now

tools/llm/sdpa_converter.py
peri044: done

py/torch_tensorrt/dynamo/_compiler.py
peri044: The spec for children is actually set by the partitioning. I'm only trying to preserve the output spec for the entire partitioned_module (It should be the same as original module unless we change it explicitly which is the case in KV caching)

tools/llm/utils.py
```diff
@@ -0,0 +1,244 @@
...
+        start_time = timeit.default_timer()
+        _ = generate_fn(model, inputs, output_seq_length, eos_token_id)
+        torch.cuda.synchronize()
+        end_time = timeit.default_timer()
+        timings.append(end_time - start_time)
+
+    return timings
+
+
+def recordStats(backend, timings, precision, batch_size=1, compile_time_s=None):
```
narendasan: Super minor but everything else is snake case

narendasan (approved):
LGTM only suggestion is moving tests into a tests subdirectory or into the tests directory. Is optional.

PR #3532, opened 2025-05-25
Add fp4 support(Currently static double quantization is not working)

lanluo-nvidia:
# Description

Add fp4 support

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -581,13 +581,13 @@ def compile(
             "\nThis feature is unimplemented in Torch-TRT Dynamo currently."
         )
 
-    if use_explicit_typing:
-        if len(enabled_precisions) != 1 or not any(
-            x in enabled_precisions for x in {torch.float32, dtype.f32}
-        ):
-            raise AssertionError(
-                f"When use_explicit_typing is enabled, only torch.float32 is allowed in the enabled_precisions but found {enabled_precisions}"
-            )
+    # if use_explicit_typing:
+    #     if len(enabled_precisions) != 1 or not any(
+    #         x in enabled_precisions for x in {torch.float32, dtype.f32}
+    #     ):
+    #         raise AssertionError(
+    #             f"When use_explicit_typing is enabled, only torch.float32 is allowed in the enabled_precisions but found {enabled_precisions}"
+    #         )
 
     if use_fp32_acc:
         logger.debug(
--- a/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
+++ b/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
@@ -274,13 +274,13 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
                 self.compilation_settings.dla_global_dram_size,
             )
 
-        if dtype.float16 in self.compilation_settings.enabled_precisions:
+        if not self.compilation_settings.use_explicit_typing and dtype.float16 in self.compilation_settings.enabled_precisions:
             builder_config.set_flag(trt.BuilderFlag.FP16)
 
         if dtype.int8 in self.compilation_settings.enabled_precisions:
             builder_config.set_flag(trt.BuilderFlag.INT8)
 
-        if dtype.fp8 in self.compilation_settings.enabled_precisions:
+        if not self.compilation_settings.use_explicit_typing and dtype.fp8 in self.compilation_settings.enabled_precisions:
             builder_config.set_flag(trt.BuilderFlag.FP8)
 
         if dtype.bfloat16 in self.compilation_settings.enabled_precisions:
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -617,6 +617,39 @@ else:
         )
 
 
+try:
+    import modelopt.torch.quantization as mtq  # noqa: F401
+
+    assert torch.ops.tensorrt.dynamic_block_quantize_op.default
+except Exception as e:
+    _LOGGER.warning(
+        "Unable to import dynamic block quantize op. Please install modelopt library (https://github.com/NVIDIA/TensorRT-Model-Optimizer?tab=readme-ov-file#installation) to add support for compiling dynamic blockquantized models"
+    )
+else:
+
+    @dynamo_tensorrt_converter(torch.ops.tensorrt.dynamic_block_quantize_op.default)
+    def aten_ops_dynamic_block_quantize_op(
+        ctx: ConversionContext,
+        target: Target,
+        args: Tuple[Argument, ...],
+        kwargs: Dict[str, Argument],
+        name: str,
+    ) -> Union[TRTTensor, Sequence[TRTTensor]]:
+        return impl.nvfp4_quantize.nvfp4_quantize(
+            ctx,
+            target,
+            SourceIR.ATEN,
+            name,
+            args[0],
+            args[1],
+            args[2],
+            args[3],
+            args[4],
+            args[5],
+            args[6],
+        )
+
+
 @dynamo_tensorrt_converter(torch.ops.aten.squeeze.dim, supports_dynamic_shapes=True)
 @dynamo_tensorrt_converter(torch.ops.aten.squeeze.dims, supports_dynamic_shapes=True)
 def aten_ops_squeeze(
--- a/py/torch_tensorrt/dynamo/conversion/converter_utils.py
+++ b/py/torch_tensorrt/dynamo/conversion/converter_utils.py
@@ -361,12 +361,37 @@ def create_constant(
             shape = list(torch_value.shape)
 
         if torch_value is not None:
+            if torch_value.dtype == torch.float8_e4m3fn:
+                weights = trt.Weights(
+                    type=trt.DataType.FP8,
+                    ptr=torch_value.data_ptr(),
+                    count=torch_value.numel(),
+                )
+                constant = ctx.net.add_constant(
+                    shape,
+                    weights,
+                )
+                constant.name = name
+                return constant.get_output(0)
+            # Iconstant layer does not support Uint8, it only support that FP4 data packed in uint8
+            if torch_value.dtype == torch.uint8:
+                weights = trt.Weights(
+                    type=trt.DataType.FP4,
+                    ptr=torch_value.data_ptr(),
+                    count=torch_value.numel() * 2,
+                )
+                shape[-1] = shape[-1] * 2
+                constant = ctx.net.add_constant(
+                    shape,
+                    weights,
+                )
+                constant.name = name
+                return constant.get_output(0)
             if torch_value.dtype == torch.bfloat16:
                 torch_value_fp32 = torch_value.to(torch.float32)
                 numpy_value = torch_value_fp32.numpy()
             else:
                 numpy_value = torch_value.numpy()
-
             ctx.mapping[name + " CONSTANT"] = numpy_value.reshape(-1)
             constant = ctx.net.add_constant(
                 shape,
@@ -381,7 +406,6 @@ def create_constant(
                     trt.DataType.BF16,
                     name + "_bf16_cast",
                 )
-
             return constant.get_output(0)
         else:
             raise ValueError(
--- a/tests/py/dynamo/models/test_models_export.py
+++ b/tests/py/dynamo/models/test_models_export.py
@@ -3,7 +3,6 @@ import importlib
 import platform
 import unittest
 from importlib import metadata
-
 import pytest
 import timm
 import torch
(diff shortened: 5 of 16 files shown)


py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -250,13 +250,10 @@ def cross_compile_for_windows(
...
-    if use_explicit_typing:
-        if len(enabled_precisions) != 1 or not any(
-            x in enabled_precisions for x in {torch.float32, dtype.f32}
-        ):
-            raise AssertionError(
-                f"When use_explicit_typing is enabled, only torch.float32 is allowed in the enabled_precisions but found {enabled_precisions}"
-            )
+    if use_explicit_typing and len(enabled_precisions) != 1:
+        raise AssertionError(
+            f"When use_explicit_typing is enabled, allow only 1 precision in the enabled_precisions but found {enabled_precisions}"
```
narendasan: What does this error mean?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -581,13 +578,10 @@ def compile(
...
 
-    if use_explicit_typing:
-        if len(enabled_precisions) != 1 or not any(
-            x in enabled_precisions for x in {torch.float32, dtype.f32}
-        ):
-            raise AssertionError(
-                f"When use_explicit_typing is enabled, only torch.float32 is allowed in the enabled_precisions but found {enabled_precisions}"
-            )
+    if use_explicit_typing and len(enabled_precisions) != 1:
+        raise AssertionError(
```
narendasan: Same here

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -39,7 +39,7 @@
     CallingConvention,
 )
 from torch_tensorrt.dynamo.conversion._TRTBuilderMonitor import TRTBulderMonitor
-from torch_tensorrt.dynamo.conversion.converter_utils import (
+from torch_tensorrt.dynamo.conversion.converter_utils import (  # global_reference_holder,
```
narendasan: What is this comment?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -274,16 +274,28 @@ def _populate_trt_builder_config(
                 self.compilation_settings.dla_global_dram_size,
             )
 
-        if dtype.float16 in self.compilation_settings.enabled_precisions:
+        if (
```
narendasan: Do you have the expected semantics of these models written out somewhere?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -737,7 +756,8 @@ def run(
             self.ctx.net, builder_config
         )
         assert serialized_engine
-
+        # clear the global reference holder after engin is built
+        # global_reference_holder.clear()
```
narendasan: What are these comments?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -617,6 +617,42 @@ def aten_ops_quantize_op(
         )
 
 
+try:
+    import modelopt.torch.quantization as mtq  # noqa: F401
+
+    assert torch.ops.tensorrt.dynamic_block_quantize_op.default
+except Exception as e:
+    _LOGGER.warning(
+        "Unable to import dynamic block quantize op. Please install modelopt library (https://github.com/NVIDIA/TensorRT-Model-Optimizer?tab=readme-ov-file#installation) to add support for compiling dynamic blockquantized models"
```
narendasan: Are dynamic block quantized models something that people will know?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -617,6 +617,42 @@ def aten_ops_quantize_op(
...
+    )
+else:
+
+    @dynamo_tensorrt_converter(
+        torch.ops.tensorrt.dynamic_block_quantize_op.default,
```
narendasan, line 633: Is the only use of this op for fp4 or will it be used for other types in the future?

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -25,6 +25,9 @@
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
+# global reference holder for the objects that should not be garbage collected before the engine is built
+global_reference_holder: List[torch.Tensor] = []
```
narendasan: You should make this a member of the conversion context not just some global buffer. When the conversion context is cleaned up it will release these

narendasan: Really we should be explicitly cleaning these so maybe as a post compilation clean up step

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -361,12 +364,37 @@ def create_constant(
             shape = list(torch_value.shape)
 
         if torch_value is not None:
+            if torch_value.dtype in (torch.float8_e4m3fn, torch.uint8):
```
narendasan: Why do we handle both fp8 and fp4 in this one conditional instead of separate ones?

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -361,12 +364,37 @@ def create_constant(
...
+                    ptr=torch_value.data_ptr(),
+                    count=count,
+                )
+                constant = ctx.net.add_constant(
+                    shape,
+                    weights,
+                )
+                constant.name = name
+                # TODO: confirm with @dheeraj @naren whether i can use ctx.mapping as the reference holder to prevent the torch tensor being garbage collected.
+                ctx.mapping[name + " CONSTANT"] = torch_value.reshape(-1)
```
narendasan: This isnt really the point of this dictionary. It is used to create a mapping for weight refitting. It might be a side effect that it serves as a reference holder. @cehongwang Is this the correct usage for weight refitting?

narendasan: We can either create two dictionaries because supporting weight refitting is probably good and we can have another which actually does the reference holding. so @cehongwang can provide guidance on correct usage of this map. 

also @cehongwang please rename this to something that is more descriptive like, `weight_src_to_constant_map` or something so people dont get confused

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
lanluo-nvidia: modified to be more generic:
`"Unable to import quantize op. Please install modelopt library (https://github.com/NVIDIA/TensorRT-Model-Optimizer?tab=readme-ov-file#installation) to add support for compiling quantized models"`

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
lanluo-nvidia: currently it is only for fp4, but in future it can add more dynamic block quantization types.
changed file name from nvfp4_quantize.py to dynamic_block_quantize.py

py/torch_tensorrt/dynamo/conversion/converter_utils.py
lanluo-nvidia: removed from here, added weights_reference_holder in ctx: ConversionContext

py/torch_tensorrt/dynamo/conversion/converter_utils.py
lanluo-nvidia: split into separated if

py/torch_tensorrt/dynamo/conversion/converter_utils.py
lanluo-nvidia: removed from here, added weights_reference_holder in ctx: ConversionContext

py/torch_tensorrt/dynamo/conversion/impl/dynamic_block_quantize.py
```diff
@@ -0,0 +1,250 @@
...
+    on the output_type set and dequantizes them back.
+    """
+    if len(input_tensor.shape) not in (2, 3):
+        raise ValueError(
+            f"dynamic_block_quantize converter received an input of {input_tensor.shape} shape. Supported shapes: 2D or 3D"
+        )
+    with unset_fake_temporarily():
+        axis = -1
+        global_scale = _calculate_global_scale(ctx, name, amax)
+        if ".weight_quantizer" in name:
```
narendasan, line 41: Is this really the way we determine which quantization scheme to use? What metadata is associated with these nodes?

tests/py/dynamo/models/test_models_export.py
```diff
@@ -199,6 +199,133 @@ def test_resnet18_half(ir):
     torch._dynamo.reset()
 
 
+# @unittest.skipIf(
+#     torch.cuda.get_device_capability() < (10, 0),
+#     "FP4 quantization requires compute capability 10.0 or later",
+# )
+@unittest.skipIf(
+    not importlib.util.find_spec("modelopt"),
```
narendasan, line 207: Wont we need blackwell as well?

py/torch_tensorrt/dynamo/_compiler.py
peri044: yes, the old message seems right. Is this required ?

py/torch_tensorrt/dynamo/conversion/_ConversionContext.py
```diff
@@ -21,3 +23,6 @@ class ConversionContext:
     )
     requires_output_allocator: bool = False
     mapping: dict[str, np.array] = field(default_factory=dict)
+    weights_reference_holder: dict[str, Union[torch.Tensor, np.array]] = field(
```
peri044: consider naming this to cpu_weights_reference_holder for more clarity

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -361,12 +361,41 @@ def create_constant(
...
+                    type=trt.DataType.FP4,
+                    ptr=torch_value.data_ptr(),
+                    count=torch_value.numel() * 2,
+                )
+                constant = ctx.net.add_constant(
+                    shape,
+                    weights,
+                )
+                constant.name = name
+                ctx.weights_reference_holder[name + " FP4_CONSTANT"] = torch_value
```
peri044: how do we distinguish a `uint8` constant (stored from FP4 workflow) vs a `uint8` constant manually inserted in the graph. (eg: the graph can have something like `torch.add(x, torch.tensor(2, dtype=torch.uint8))` ?

py/torch_tensorrt/dynamo/conversion/impl/dynamic_block_quantize.py
```diff
@@ -0,0 +1,250 @@
...
+def _double_dequantize(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    quantized_data_in_fp4: TRTTensor,
+    quantized_scale_in_fp8: TRTTensor,
+    global_scale: torch.Tensor,
+    axis: int = -1,
+    output_type: trt.DataType = trt.DataType.FLOAT,
```
peri044, line 148: minor : add a docstring and a little bit of explanation about the function wherever possible. Any references to this quantization (eg: TRT docs) will be useful

narendasan (commented):
Add an end to end example for FP4 using modelopt and tensorrt

py/torch_tensorrt/dynamo/conversion/converter_utils.py
narendasan: have an optional argument that sets the target type and in the case of uint8 without fp4 then throw error

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
narendasan: Just check it once

py/torch_tensorrt/dynamo/conversion/impl/dynamic_block_quantize.py
lanluo-nvidia: below are all the information I can get from the aten ops,
I don't see any better way to know this is the weight or input
```
    @dynamo_tensorrt_converter(
        torch.ops.tensorrt.dynamic_block_quantize_op.default,
        supports_dynamic_shapes=True,
    )
    def aten_ops_dynamic_block_quantize_op(
        ctx: ConversionContext,
        target: Target,
        args: Tuple[Argument, ...],
        kwargs: Dict[str, Argument],
        name: str,
```

lanluo-nvidia:
>Add an end to end example for FP4 using modelopt and tensorrt

There is a simple linear end to end FP4 example in this PR.
There is another FP4 flux example in a seperate PR: https://github.com/pytorch/TensorRT/pull/3537
Which based on Adrian's flux PR.

PR #3535, opened 2025-05-28
fix: wrong dtype and device in `aten.full_like` decomposition

junstar92:
# Description

This PR addresses a bug in the Torch-TensorRT decomposition of `torch.ops.aten.full_like`.

In the current implementation, the decomposition incorrectly overrides the `dtype` and `device` arguments, ignoring explicitly set `dtype` values and assigning all tensors to the `default_device()` (typically `cuda:0`), regardless of the inputs' actual device.

Specifically, the issue occurs in the following decomposition function:
https://github.com/pytorch/TensorRT/blob/5c37931f1b52fc340e2231373a5af3a92ff727cb/py/torch_tensorrt/dynamo/lowering/_decompositions.py#L428-L437

This implementation causes two main issues:

1. **Incorrect dtype propagation**: Even when `torch.full_like(..., dtype=torch.bool)` is used in the model, the decomposition overwrites the `dtype` with `input.dtype` (e.g., `float16`), resulting in an incorrect output type.
2. **Device mismatch**: When exporting and running models on devices other than `cuda:1` (e.g., `cuda:1`), the decomposition forces outputs to be on `cuda:0`, causing runtime errors or silent bugs due to device mismatch.

To demonstrate the issue, the following test cases are included in this PR:
```Python
import torch
from torch.export._trace import _export
from torch_tensorrt.dynamo.lowering import get_decompositions


class MyModel(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        return torch.ones_like(x, dtype=torch.bool)


def test1() -> tuple[bool, str]:
    model = MyModel()
    x = torch.randn(1, 10, dtype=torch.float16)
    y = model(x)
    return y.dtype == torch.bool, f"expected dtype {torch.bool}, and got {y.dtype}"


def test2() -> tuple[bool, str]:
    model = MyModel()
    x = torch.randn(1, 10, dtype=torch.float16)
    ep = _export(model, (x,))
    ep = ep.run_decompositions(get_decompositions(False))
    gm = ep.module()
    y = gm(x)
    return y.dtype == torch.bool, f"expected dtype {torch.bool}, and got {y.dtype}"

def test3() -> tuple[bool, str]:
    device = torch.device("cuda", index=1)
    model = MyModel().to(device)
    x = torch.randn(1, 10, dtype=torch.float16).to(device)
    ep = _export(model, (x,))
    ep = ep.run_decompositions(get_decompositions(False))
    gm = ep.module()
    y = gm(x)
    return y.device == device, f"expected device {device}, and got {y.device}"
    

for test in (test1, test2, test3):
    success, msg = test()
    print(f"{test.__name__}: {'Success' if success else 'Failed'} - {msg}")

```

Results:
```
test1: Success - expected dtype torch.bool, and got torch.bool
test2: Failed - expected dtype torch.bool, and got torch.float16
test3: Failed - expected device cuda:1, and got cuda:0
```

- `test1`: Verifies that `torch.ones_like` returns a tensor with the correct dtype.
- `test2`: Shows that the exported model via `torch.export(...).run_decompositions(...)` fails to preserve `dtype`.
- `test3`: Demonstrates the incrroect `device` assignment after decomposition when using non-default CUDA devices.

This PR fixes the decomposition logic to correctly respect the explicitly passed `dtype` and `device` values, or fall back to those inferred from the input tensor only if not explicitly provided.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -9,6 +9,7 @@ from torch._export.utils import (
     _get_decomp_for_cia,
 )
 from torch._ops import OpOverload
+
 from torch_tensorrt.dynamo._defaults import default_device
 from torch_tensorrt.dynamo.conversion.converter_utils import get_positive_dim
 from torch_tensorrt.dynamo.utils import to_torch_device
@@ -432,8 +433,8 @@ def full_like_decomposition(*args, **kwargs) -> torch.Tensor:
     input = args[0]
     shape = args[0].shape
     fill_value = args[1]
-    kwargs["dtype"] = input.dtype
-    kwargs["device"] = to_torch_device(default_device())
+    kwargs["dtype"] = kwargs.get("dtype", None) or input.dtype
+    kwargs["device"] = input.device
     return torch.full(shape, fill_value, dtype=kwargs["dtype"], device=kwargs["device"])
 
 


junstar92:
@peri044 Thanks for quick review.

But, I missed that `device` needs to be treated like `dtype` because `full_like` also accepts `device` as an argument.
```Python
kwargs["device"] = kwargs.get("device", None) or input.device
```
Without handling it, a `device` different from the input's may be used, causing mismatches.
Should I open a new one to address the remaining issue?

PR #3539, opened 2025-05-29
feat: Hierarchical Partitioner to support multi-backends

zewenli98:
# Description

Hierarchical Partitioner to support multi-backends.

## Type of change

# Checklist:


docs/hierarchical_partitioner.md
```diff
@@ -0,0 +1,204 @@
+# Hierarchical Partitioner for TensorRT
```
narendasan, line 1: Lets add this to the contributor docs for now

examples/hierarchical_partitioner_example.py
```diff
@@ -0,0 +1,114 @@
...
+                torch.ops.aten.conv2d.default,
+                torch.ops.aten.convolution.default,
+            },
+            "tensorrt": set(DYNAMO_ATEN_CONVERTERS.keys()),
+        },
+        torch_executed_ops=[
+            torch.ops.aten._native_batch_norm_legit_no_training.default
+        ],
+        require_full_compilation=False,
+        skip_fusion=False,
```
peri044: skip_fusion=False slows down the partitioning a lot. Can you check if it's really needed ?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -913,26 +945,61 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
...
+                            torch.randn(input.shape)
+                            .to(dtype.to(input.dtype, t=torch.dtype))
+                            .cuda()
+                        )
+                        sub_inputs.append(sub_input)
+
+                    compiled_func = torch._inductor.compile(
+                        submodule,
+                        sub_inputs,
+                    )
```
peri044: `torch.dynamo.mark_dynamic` API  is used to set dynamic shapes for torch.compile workflow. Reference: https://docs.pytorch.org/TensorRT/user_guide/dynamic_shapes.html. So you can use the construct_submodule_inputs() API to give you dynamic inputs (if that's the case) and set them to the inductor segment accordingly.

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -913,26 +945,61 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
...
+                elif "_run_on_acc_tensorrt" in name:
+                    compiled_submodule = convert_module(
+                        submodule,
+                        submodule_inputs,
+                        settings=settings,
+                        name=name,
+                        engine_cache=engine_cache,
+                    )
+                else:
+                    raise ValueError(f"Unknown backend for submodule: {name}")
```
peri044: should there be a `_run_on_gpu` segment here ? How would the torch.ops.aten._native_batch_norm_legit_no_training.default fallback to native pytorch in your example from slack ?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1276,3 +1343,12 @@ def load_cross_compiled_exported_program(file_path: str = "") -> Any:
...
     return replace_execute_engine_no_op_node(exp_program)
+
+
+class FunctionWrapper(torch.nn.Module):
+    def __init__(self, func):
+        super().__init__()
+        self.func = func
+
+    def forward(self, *args, **kwargs):
+        return self.func(*args, **kwargs)
```
peri044: Consider naming this to `InductorModule` and moving to utils

py/torch_tensorrt/dynamo/partitioning/_hierarchical_partitioner.py
```diff
@@ -0,0 +1,574 @@
...
+    partitioner = HierarchicalAdjacencyPartitioner(
+        gm,
+        supported_ops,
+        backend_support_map=backend_support_map,
+        backend_priority=backend_priority,
+        min_block_size=min_block_size,
+        require_full_compilation=require_full_compilation,
+        skip_fusion=skip_fusion,
+    )
+
```
peri044, line 580: Do you think copying all this code is necessary ? or inheriting only some components would be sufficient ?

peri044 (commented):
Considering how this feature will be made available, the min_block_size and torch_executed_ops need to be re-thought or deprecated as the min_block_size can apply to all backends and `torch_executed_ops` will be replaced by `backend_support_map`

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -913,26 +945,61 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
...
+                        sub_inputs.append(sub_input)
+
+                    compiled_func = torch._inductor.compile(
+                        submodule,
+                        sub_inputs,
+                    )
+                    # Wrap the compiled function to be a torch.nn.Module
+                    compiled_submodule = FunctionWrapper(compiled_func)
+
+                elif "_run_on_acc_tensorrt" in name:
```
narendasan: Is there some sort of design where the capability and conversion parts can be grouped and registered? We can add this concept to the RFC for later

narendasan: That way we dont have a long conditional case set we just look up the appropriate conversion function on a standardized API

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -913,26 +945,61 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
...
     parse_graph_io(gm, dryrun_tracker)
 
     # Replace all FX Modules with TRT Modules
-    for name, trt_module in trt_modules.items():
-        setattr(partitioned_module, name, trt_module)
+    for name, compiled_module in compiled_modules.items():
+        setattr(partitioned_module, name, compiled_module)
         if settings.lazy_engine_init and not settings.enable_cross_compile_for_windows:
-            getattr(partitioned_module, name).setup_engine()
+            if use_hierarchical_partitioner:
```
narendasan: Similar here, we should standardize post processing as well

narendasan (commented):
What would be the merge-able version of this PR since we are going to expose the new partitioner yet?

examples/hierarchical_partitioner_example.py
zewenli98: > the min_block_size and torch_executed_ops need to be re-thought or deprecated as the min_block_size can apply to all backends 

My understanding is that, if the num of ops of a GM is less than `min_block_size`, no matter it's TRT or other backend, the GM would not be compiled. Do you mean it shouldn't apply to other backends? 

> `torch_executed_ops` will be replaced by `backend_support_map`

My understanding is that `torch execute` is not considered as a backend because it doesn't need any compilation, it just runs ops in eager mode. So, if an op was in `torch_executed_ops`, it would ignore `backend_support_map` and run in torch eager anyway.

py/torch_tensorrt/dynamo/_compiler.py
zewenli98: Nope, as I mentioned above, `_run_on_gpu` is not considered as a backend. We just keep the module as is. Only `_run_on_acc_backend` modules need to be replaced with the compiled module. This aligns with our existing implementation.

examples/hierarchical_partitioner_example.py
zewenli98: > skip_fusion=False slows down the partitioning a lot. Can you check if it's really needed ?

Since adjacency partitioner uses this flag, I just keep it here. yeah I can definitely switch it to True in the example.

py/torch_tensorrt/dynamo/_compiler.py
zewenli98: yeah I agree that the conversion part of different backends should be grouped, but for now I don't have too much info about other backends (like how to convert an op to that backend). We can definitely do this when we are ready to support other backends.

zewenli98:
> What would be the merge-able version of this PR since we are going to expose the new partitioner yet?

@narendasan As you said that we don't want to expose APIs of this feature to users for now, the current code doesn't change any behavior. It can be merged after I resolve the issues above.

py/torch_tensorrt/dynamo/_compiler.py
zewenli98: `submodule_inputs` is already the return of `construct_submodule_inputs()`. What I did here is to convert torch-trt Input to torch Tensor. I changed to:
```
sub_input = (
    input.torch_tensor
    .to(dtype.to(input.dtype, t=torch.dtype))
    .cuda()
)
```

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -788,20 +798,47 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
             "Some nodes do not have metadata (shape and dtype information). This could lead to problems sometimes if the graph has PyTorch and TensorRT segments."
         )
 
+    ############ TODO: testing only ############
```
narendasan: Should get removed right?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -788,20 +798,47 @@ def contains_metadata(gm: torch.fx.GraphModule) -> bool:
...
-            logger.info("Partitioning the graph via the fast partitioner")
-            partitioned_module, supported_ops = partitioning.fast_partition(
-                gm,
-                verbose=settings.debug,
-                min_block_size=settings.min_block_size,
-                torch_executed_ops=settings.torch_executed_ops,
-                require_full_compilation=settings.require_full_compilation,
-                skip_fusion=(num_supported_ops == total_ops),
-            )
+            if use_hierarchical_partitioner:
```
narendasan: Should this be removed as well or should we just tie it to a set of hidden settings?

py/torch_tensorrt/dynamo/_compiler.py
narendasan: Lets remove the _hierarchical partitioner from _compile.py then and put together an RFC

narendasan (commented):
Looks mostly good to me, just remove the stuff in _compile and make sure the example comprehensively demonstrates how to use the feature

zewenli98:
> Looks mostly good to me, just remove the stuff in _compile and make sure the example comprehensively demonstrates how to use the feature

Should we expose APIs to users? If not, how can they use the feature?

py/torch_tensorrt/dynamo/_compiler.py
zewenli98: If these variables are removed here, how to provide backend_priority and backend_support_map if we don't expose APIs?

examples/hierarchical_partitioner_example.py
```diff
@@ -0,0 +1,116 @@
...
+    partitioned_model, op_support = hierarchical_adjacency_partition(
+        gm,
+        verbose=True,
+        min_block_size=1,
+        backend_priority=["inductor", "tensorrt"],
+        backend_support_map={
+            "inductor": {
+                "torch.ops.aten.convolution.default",
+            },
+            "tensorrt": CONVERTERS.keys(),
```
peri044: is this needed ? this will also have the convolution op right ?

zewenli98: This is a demo just for showing the functionality. Since both inductor and tensorrt support conv and `backend_priority=["inductor", "tensorrt"]`, conv will run on inductor. The module structure:
```
Partitioned Model Structure:
 GraphModule(
  (_run_on_acc_inductor_0): GraphModule(
    (conv1): Module()
  )
  (_run_on_gpu_1): GraphModule(
    (bn1): Module()
  )
  (_run_on_acc_tensorrt_2): GraphModule()
  (_run_on_acc_inductor_3): GraphModule(
    (conv2): Module()
  )
  (_run_on_gpu_4): GraphModule(
    (bn2): Module()
  )
  (_run_on_acc_tensorrt_5): GraphModule()
)
```

py/torch_tensorrt/dynamo/partitioning/_hierarchical_partitioner.py
zewenli98: I revisited the code where it has only two minor functions are the same and one of them calls a customized function. Considering the maintainability and suggestions from Cursor, I think it would be better to separate them.

narendasan (approved):
LGTM

Issue #3522, opened 2025-05-16
❓ [Question] Quantization IR w/ Dynamo


patrick-botco:
## ❓ Manually Annotate Quantization Parameters in FX Graph

is there a way to manually annotate quantization parameters that will be respected throughout torch_tensorrt conversion (e.g. manually adding q/dq nodes, or specifying some tensor metadata) via dynamo?

patrick-botco:
cc @narendasan @peri044 maybe? 🙏

narendasan:
This should be possible as this is what the tensorrt model optimizer toolkit effectively does. @peri044 or @lanluo-nvidia could maybe give more specific guidance.

peri044:
We currently use NVIDIA Model optimizer toolkit which inserts quantization nodes within the torch model using quantize API 
1) https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/9c54aa1c47871d0541801a20962996461d805162/modelopt/torch/quantization/model_quant.py#L126
2) https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/9c54aa1c47871d0541801a20962996461d805162/modelopt/torch/quantization/tensor_quant.py#L229-L243 (definition of custom ops which do the quantization). We have converters for these quantization custom ops (which call Q & DQ apis in TensorRT). 

You can also manually insert a quantization custom op by implementing a  lowering pass by adding nodes to the `torch.fx.GraphModule` and implement/register a custom converter for it. You can append custom metadata to this node by updating `node.meta["val"]`
1) https://docs.pytorch.org/TensorRT/contributors/writing_dynamo_aten_lowering_passes.html ([existing lowering passes](https://github.com/pytorch/TensorRT/tree/main/py/torch_tensorrt/dynamo/lowering/passes))
2) https://docs.pytorch.org/TensorRT/contributors/dynamo_converters.html
This can be done outside Torch-TRT codebase using the decorations listed above to register your lowering pass/ converter. 

Please let me know if you have any further questions.

patrick-botco:
hey @peri044 , thanks for the response. i tried modelopt -> export on a simple model below. am i using this wrong or missing something obvious? im using non-strict export (strict runs into `torch._dynamo.exc.Unsupported: reconstruct:  UserDefinedObjectVariable(_DMAttributeManager)`), but hitting `ValueError: Node type mismatch; expected <class 'tuple'>, but got <class 'torch.Size'>.` thanks!

```python
[...]

PR #3543, opened 2025-06-01
fix int8/fp8 constant folding issue

lanluo-nvidia:
# Description

https://github.com/pytorch/TensorRT/issues/3522

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/impl/quantize.py
```diff
@@ -28,42 +29,59 @@ def quantize(
...
+                    f"quantize converter received an input of {input_tensor.dtype} type. Supported types: float32 | float16"
+                )
             if num_bits != 8 or exponent_bits not in (0, 4):
                 raise ValueError(
                     f"quantize converter currently only accept INT8 or FP8 based quantize, got {num_bits=}, {exponent_bits=}"
                 )
+        else:
+            raise ValueError(
+                f"quantize converter received an input of {type(input_tensor)} type. Supported types: torch.Tensor | TRTTensor"
+            )
```
patrick-botco, line 50: can this also handle when `input_tensor` is a `np.ndarray`?

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
```diff
@@ -28,42 +29,59 @@ def quantize(
...
+        if not isinstance(amax, trt.ITensor):
+            # for the int8/fp8 quantization, it is per tensor quantization, so amax should be a singlular element
+            # TODO: to confirm with ModelOpt team, why I am getting a amax with shape > 1
+            if len(amax.shape) >= 1:
+                amax = amax.abs().amax()
+                if amax == 0.0:
+                    amax = 1.0
+            amax = to_torch(amax, None)
+            scale = torch.divide(amax, max_bound)
+            scale = get_trt_tensor(ctx, amax, name + "_scale")
```
patrick-botco:
```suggestion
            amax = to_torch(amax, None)
            if len(amax.shape) >= 1:
                amax = amax.abs().amax()
                if amax == 0.0:
                    amax = 1.0
            scale = torch.divide(amax, max_bound)
            scale = get_trt_tensor(ctx, scale, name + "_scale")
```

`amax` needs to be a `torch.Tensor` before calling `abs()`
think you meant to do `get_trt_tensor(ctx, scale, name + "_scale")`?

lanluo-nvidia: confirmed with modelopt interface, the input can only be a torch.Tensor, I have changed the interface to be torch.Tensor only removed np.ndarray

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
```diff
@@ -28,42 +28,59 @@ def quantize(
...
```
patrick-botco: do we want to delete the `to_torch` call entirely then? im seeing a `np.ndarray` on torch-tensorrt 2.6.0. in that case, we'll still need `to_torch` before the call to `amax.abs()` or it'll break.

separately, `scale` still seems off, its being overwritten by amax
```suggestion
            scale = get_trt_tensor(ctx, scale, name + "_scale")
```

patrick-botco (approved):

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
```diff
@@ -28,42 +28,73 @@ def quantize(
...
+            dtype = trt.DataType.INT8
             max_bound = 127
         elif num_bits == 8 and exponent_bits == 4:
+            dtype = trt.DataType.FP8
             max_bound = 448
 
         amax = to_torch(amax, None)
+        axis = None
+        # int8 weight quantization is per-channel quantization(it can have one or multiple amax values)
+        if dtype == trt.DataType.INT8 and ".weight_quantizer" in name:
```
peri044: Is this needed ? - ".weight_quantizer" in name

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
```diff
@@ -28,42 +28,73 @@ def quantize(
...
+                assert (
+                    axis == 0
+                ), f"{name=} {amax=} is per-channel quantization, expected axis to be 0, but got {axis=}"
+        else:
+            # int8 activation and fp8 weight/activation quantization is per-tensor quantization, it can only have single amax value
+            assert (
+                amax.numel() == 1
+            ), f"{name=} is per-tensor quantization, expected amax is a singular value, but got {amax.shape=}"
         scale = torch.divide(amax, max_bound)
+        scale.masked_fill_(scale == 0, 1.0)
```
peri044, line 79: what is this doing ? Also is there a scenario where amax can be an ITensor ?

lanluo-nvidia: if scale is 0, then set as 1.0,
because quantize will do input_tensor/scale, we need to make sure the scale is not zero.

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
lanluo-nvidia: int8 weight_quantization is per-channel, int8 activation quantization is per-tensor, hence we need the name check.

peri044: I think the check should be based on the number of scale values (and if it matches with the dimension of the channel axis in the input) and not include name.

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
peri044: But if it is zero, we should error out right ? Otherwise the scale of 1.0 will be incorrect and would result in a silent accuracy issue ?

py/torch_tensorrt/dynamo/conversion/impl/quantize.py
lanluo-nvidia: modified as suggested.

peri044 (approved):
LGTM

Issue #3564, opened 2025-06-11
Constant folding crashes if model opt is not installed


narendasan:
There was an implicit dependency added to constant folding on model opt operations that causes a failure if model opt is not installed.

When running a basic example: 

```
Traceback (most recent call last):
  File "/home/naren/pytorch_org/tensorrt/examples/dynamo/torch_compile_resnet_example.py", line 116, in <module>
    trt_model = torch_tensorrt.compile(model, **compile_spec)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/naren/pytorch_org/tensorrt/py/torch_tensorrt/_compile.py", line 289, in compile
    trt_graph_module = dynamo_compile(
                       ^^^^^^^^^^^^^^^
  File "/home/naren/pytorch_org/tensorrt/py/torch_tensorrt/dynamo/_compiler.py", line 697, in compile
    gm = post_lowering(gm, settings)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/naren/pytorch_org/tensorrt/py/torch_tensorrt/dynamo/lowering/passes/_aten_lowering_pass.py", line 102, in post_lowering
    gm = ATEN_POST_LOWERING_PASSES(gm, settings)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/naren/pytorch_org/tensorrt/py/torch_tensorrt/dynamo/lowering/passes/pass_manager.py", line 56, in __call__
    out = _pass(out, settings)
          ^^^^^^^^^^^^^^^^^^^^
  File "/home/naren/.local/share/uv/python/cpython-3.11.12-linux-x86_64-gnu/lib/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/naren/pytorch_org/tensorrt/py/torch_tensorrt/dynamo/lowering/passes/constant_folding.py", line 34, in constant_fold
    cf.run()
  File "/home/naren/pytorch_org/tensorrt/.venv/lib/python3.11/site-packages/torch/_inductor/constant_folding.py", line 295, in run
    return super().run(initial_env=env)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3565, opened 2025-06-11
fix: Fix constant folding failure due to modelopt

peri044:
# Description

Fixes https://github.com/pytorch/TensorRT/issues/3564

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/passes/constant_folding.py
+++ b/py/torch_tensorrt/dynamo/lowering/passes/constant_folding.py
@@ -101,6 +101,12 @@ class _TorchTensorRTConstantFolder(ConstantFolder):  # type: ignore[misc]
 
     # TODO: Update this function when quantization is added
     def is_impure(self, node: torch.fx.node.Node) -> bool:
-        if node.target in (torch.ops.tensorrt.quantize_op.default,):
+        quantization_ops = {}
+        try:
+            assert torch.ops.tensorrt.quantize_op.default
+            quantization_ops.add(torch.ops.tensorrt.quantize_op.default)
+        except Exception as e:
+            pass
+        if quantization_ops and node.target in quantization_ops:
             return True
         return False


py/torch_tensorrt/dynamo/lowering/passes/constant_folding.py
```diff
@@ -101,6 +101,12 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
 
     # TODO: Update this function when quantization is added
     def is_impure(self, node: torch.fx.node.Node) -> bool:
-        if node.target in (torch.ops.tensorrt.quantize_op.default,):
+        quantization_ops = {}
+        try:
+            assert torch.ops.tensorrt.quantize_op.default
```
narendasan, line 109: Should we verify on model opt import or should we just check if the op was registered externally?

narendasan: Also shouldnt it be like do the `return True` if `quantize_op` is available otherwise return `return False`?

peri044: > Should we verify on model opt import or should we just check if the op was registered externally?

Actually, thinking more about it, we should do both. importing modelopt ensures this op is only registered by modelopt so that we don't return True if user registers the op with same name externally.

peri044: > Also shouldnt it be like do the return True if quantize_op is available otherwise return return False?

we should return True if the node.target matches the quantization node target right ? Could you elaborate ?

narendasan: Oh I see i misread the code, i think this is right but we need to figure out the correct way to check if an op has been registered. I think pytorch has a formal api for this

lanluo-nvidia (approved):
LGTM

PR #3570, opened 2025-06-13
adding rotary embedding example, with graph rewrite for complex subgraph [WIP]

apbose:
This PR-
1. Adds an example for parallel rotary embedding
2. Adds logic for complex graph detection
3. Adds a pass for complex graph rewrite in aten_lowering_pass
Please note that this PR is currently for the single GPU case where there is no DTensor in the inputs of the torch module. Ideally this should not require runtime changes.
This should avoid the graph breaks caused due to view_as_complex and view_as_real nodes.

### DIFF
--- a/examples/distributed_inference/rotary_embedding.py
+++ b/examples/distributed_inference/rotary_embedding.py
@@ -1,0 +1,112 @@
+import time
+
+import tensorrt as trt
+import torch
+import torch.distributed as dist
+import torch.nn as nn
+import torch_tensorrt
+from tensor_parallel_initialize_dist import initialize_distributed_env
+from torch.distributed._tensor import Shard
+from torch.distributed.tensor.parallel import (
+    ColwiseParallel,
+    RowwiseParallel,
+    parallelize_module,
+)
+
+"""
+This example covers the rotary embedding and rotary attention case for tensor parallel
+"""
+
+
+def precompute_freqs_cis(
+    dim: int, end: int, theta: float = 10000.0, n_parallel=1
+) -> torch.Tensor:
+    """Precompute the frequency tensor for complex exponentials (cis) with given dimensions.
+    This function calculates a frequency tensor with complex exponentials using the given dimension 'dim'
+    and the end index 'end'. The 'theta' parameter scales the frequencies.
+    The returned tensor contains complex values in complex64 data type.
+    Args:
+        dim (int): Dimension of the frequency tensor.
+        end (int): End index for precomputing frequencies.
+        theta (float, optional): Scaling factor for frequency computation. Defaults to 10000.0.
+        n_parallel (int, optional): Number of GPUs for parallel computation. Defaults to 1.
+    Returns:
+        torch.Tensor: Precomputed frequency tensor with complex exponentials.
+    """
+    freqs = 1.0 / (theta ** (torch.arange(0, dim // n_parallel, 2).float() / dim))
+    t = torch.arange(end, device=freqs.device)
+    freqs = torch.outer(t, freqs).float()
+    return torch.polar(torch.ones_like(freqs), freqs)
+
+
+def rotary_embedding(xq, xk, dim, freqs_cis=None):
+    """This calculates the rotary embedding for the query and key tensors.
+    Args:
+        xq (torch.Tensor): Query tensor.
+        xk (torch.Tensor): Key tensor.
+        dim (int): Dimension of the query and key tensors.
+        freqs_cis (torch.Tensor, optional): Precomputed frequency tensor. Defaults to None.
+    Returns:
+        tuple: Tuple containing the rotated query and key tensors.
+    """
+    freqs_cis = freqs_cis[None, :, None, :]  # [1, seq, 1, dim/2]
+    xq_ = torch.view_as_complex(xq.float().reshape(*xq.shape[:-1], -1, 2))
+    xk_ = torch.view_as_complex(xk.float().reshape(*xk.shape[:-1], -1, 2))
+
+    xq_out = torch.view_as_real(xq_ * freqs_cis).flatten(3)
+    xk_out = torch.view_as_real(xk_ * freqs_cis).flatten(3)
+    return xq_out.type_as(xq), xk_out.type_as(xk)
+
+
+########Tensor Parallel########
+def parallel_rotary_block(rotary_block, tp_mesh):
+    """Parallel rotary block for tensor parallel
+    Args:
+        rotary_block: Rotary block to parallelize
+        tp_mesh: Tensor parallel mesh
+    """
+    if tp_mesh.size() <= 1:
+        return
+
+    plan = {
+        "wq": ColwiseParallel(),
+        "wk": ColwiseParallel(),
+        "wo": RowwiseParallel(output_layouts=Shard(0)),
+    }
+    rotary_block.n_parallel = 2
+
+    parallelize_module(rotary_block, tp_mesh, plan)
+
+
+class RotaryAttention(nn.Module):
+    def __init__(self, dim: int, seq_len: int):
+        super().__init__()
+        self.dim = dim
+        # self.rotary = RotaryEmbedding(dim)
+        self.wq = nn.Linear(dim, dim)
+        self.wk = nn.Linear(dim, dim)
+        self.wo = nn.Linear(dim, dim)
+        self.seq_len = seq_len
+        self.n_parallel = 1
+        self.register_buffer(
+            "freqs_cis",
+            self._precompute_freqs_cis(),
+            persistent=True,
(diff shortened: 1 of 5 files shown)


py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
peri044, line 1: add tests for this pass

py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
```diff
@@ -0,0 +1,271 @@
...
+]:
+    """Constructs the original and replacement functions for complex multiplication.
+
+    The original functions correspond to native complex multiplication
+    via torch.mul or operator.mul on complex tensors.
+
+    The replacement function assumes x and y are real tensors with the last
+    dimension size 2 representing real and imaginary parts, and performs
+    complex multiplication manually returning the same shaped tensor.
+    """
```
peri044, line 310: if you know the limitations of this pass, please add for future reference

py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
```diff
@@ -0,0 +1,271 @@
...
+            f"ComplexOpSubGraphInfo(anchor_nodes={[n.name for n in self.anchor_nodes]}, "
+            f"subgraph={[n.name for n in self.subgraph_nodes]}, "
+            f"inputs={[n.name for n in self.input_nodes]})"
+        )
+
+
+class ComplexOpDetector:
+    def __init__(self, logger):
+        self.logger = logger
+        pass
```
peri044, line 37: is this needed ?

apbose: you mean the pass? I moved the logger out of the __init__ func

apbose:
I see lint error here would reformat /home/runner/work/TensorRT/TensorRT/py/torch_tensorrt/dynamo/lowering/passes/constant_folding.py which is unrelated to my change.  Not sure why is it failing here

py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
apbose: I added this below in line 307

examples/distributed_inference/rotary_embedding.py
```diff
@@ -0,0 +1,106 @@
...
+from torch.distributed._tensor import Shard
+from torch.distributed.tensor.parallel import (
+    ColwiseParallel,
+    RowwiseParallel,
+    parallelize_module,
+)
+
+"""
+This example covers the rotary embedding and rotary attention case for tensor parallel
+"""
```
peri044, line 29: can you build the docs and see if the doc build passes ? Previously, docs build failed because our examples did not have the right sphnix formatting

examples/distributed_inference/tensor_parallel_rotary_embedding.py
```diff
@@ -0,0 +1,58 @@
...
+    parallel_rotary_block(model, device_mesh)
+    device = torch.device("cuda", device_mesh.get_rank())
+    model.to(device)
+    x = torch.randn(BATCH, SEQ_LEN, HEADS, DIM).to(device)
+
+    python_result = model(x)
+
+    logger.info("Torch-tensorrt compilation for rotary embedding")
+
+    model = torch.compile(model, backend="torch_tensorrt", options={"debug": True})
```
peri044: remove debug=True as it is deprecated

examples/distributed_inference/tensor_parallel_rotary_embedding.py
```diff
@@ -0,0 +1,58 @@
...
+                assert (
+                    python_result - output
+                ).std() < 0.01, "Compilation result is not correct."
+            elif _rank == 0:
+                logger.info(f"Inference time is {end-start}")
+    except Exception as e:
+        logger.error(f"Error: {e}")
+        raise e
+    finally:
+        cleanup_distributed_env()
```
peri044: Do we need a try/except and finally here for the example ? And also why are we looping ? If you want to display any results of this block, please use the right formatting https://github.com/pytorch/TensorRT/blob/main/examples/dynamo/torch_export_sam2.py#L282-L297

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
```diff
@@ -23,7 +23,10 @@ def getitem_validator(getitem_node: Node, settings: CompilationSettings = None)
     from torch_tensorrt.dynamo.conversion._ConverterRegistry import DYNAMO_CONVERTERS
 
     # Getitem nodes can only be converted if their parent node also can
-    return getitem_node.args[0] in DYNAMO_CONVERTERS
+    return (
+        getitem_node.args[0] in DYNAMO_CONVERTERS
+        or getitem_node.args[0].op == "get_attr"
```
peri044, line 28: why is this needed ?

py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
```diff
@@ -0,0 +1,319 @@
...
+            ],
+            -1,
+        )
+
+    return (original_mul, replacement)
+
+
+# This lowering pass is used to detect and rewrite complex subgraphs in the graph
+# This lowering pass works for complex tensor in mul which are parameter or buffers in the graph
+def complex_graph_detection(
```
peri044: add this as a docstring

peri044 (changes requested):

examples/distributed_inference/tensor_parallel_rotary_embedding.py
apbose: The try/except block is for the case where in we check for the inference time improvement over a no of iterations after the graph compilation to see performance improvement without the graph breaks. I am not clear about the formatting part pointed above in the link. Did you point it out because the try loop wont be rendered? I see that the rendering is correct.

I could as such remove the loop too

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
apbose: This is needed because the complex tensor is at present a buffer, wherein we extract the real and imag part through input[..., 0], input[..., 1]. This leads to getitem node with arg being a get_attr node.
If I don't include the above it leads to graph break for this part in GPU, which is unnecessary if we support get_attr.
```
graph():
   %_frozen_param3_reshaped : [num_users=2] = get_attr[target=_frozen_param3_reshaped]
   %getitem : [num_users=1] = call_function[target=operator.getitem](args = (%_frozen_param3_reshaped, (Ellipsis, 0)), kwargs = {})
   %getitem_1 : [num_users=1] = call_function[target=operator.getitem](args = (%_frozen_param3_reshaped, (Ellipsis, 1)), kwargs = {})
   return (getitem, getitem_1)
```

tests/py/dynamo/lowering/test_aten_lowering_passes.py
```diff
@@ -237,5 +237,98 @@ def forward(self, input, mat1, mat2):
...
+        torch._dynamo.reset()
+
+        # Validate that the results between Torch and Torch-TRT are similar
+        optimized_model = torch_tensorrt.compile(
+            model,
+            "torch_compile",
+            inputs,
+            min_block_size=1,
+            pass_through_build_failures=True,
+            debug=True,
```
apbose: Will remove this debug=True

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
peri044: Okay. What does `getitem_node.args[0]` return ? Does it return the op name ?

examples/distributed_inference/tensor_parallel_rotary_embedding.py
peri044: Why do we need to loop here ? Is the goal of this example to compare output of rotary embedding block b/w TRT and pytorch model ?

py/torch_tensorrt/dynamo/conversion/ops_evaluators.py
apbose: getitem_node.args[0] returns the node which is the key in the ConverterRegistry. 
```
def __getitem__(
        self, node: Node
    )
``` with the value being converter, calling convention and dictionary containing supports_dynamic shape, requires_output_allocator.

peri044 (approved):

PR #3573, opened 2025-06-13
Changed weight map to tensor and fix the refit bug

cehongwang:
# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/_ConversionContext.py
+++ b/py/torch_tensorrt/dynamo/conversion/_ConversionContext.py
@@ -1,7 +1,6 @@
 from dataclasses import dataclass, field
 from typing import Union
 
-import numpy as np
 import torch
 from torch_tensorrt.dynamo._settings import CompilationSettings
 from torch_tensorrt.dynamo.types import TRTNetwork
@@ -24,10 +23,14 @@ class ConversionContext:
         default_factory=CompilationSettings
     )
     requires_output_allocator: bool = False
-    weight_refit_map: dict[str, np.array] = field(default_factory=dict)
+    weight_refit_map: dict[str, torch.Tensor] = field(default_factory=dict)
     cpu_weights_reference_holder: dict[str, Union[torch.Tensor]] = field(
         default_factory=dict
     )
+
+    def record_weight(self, name: str, weight: torch.Tensor) -> None:
+        self.weight_refit_map[name] = weight
+        self.cpu_weights_reference_holder[name + " CPU_REFERENCE"] = weight
 
     def clear_cpu_weights_reference_holder(self) -> None:
         self.cpu_weights_reference_holder.clear()
--- a/py/torch_tensorrt/dynamo/conversion/_ConversionContext.py
+++ b/py/torch_tensorrt/dynamo/conversion/_ConversionContext.py
@@ -1,5 +1,4 @@
 from dataclasses import dataclass, field
-from typing import Union
 
 import torch
 from torch_tensorrt.dynamo._settings import CompilationSettings
@@ -24,9 +23,7 @@ class ConversionContext:
     )
     requires_output_allocator: bool = False
     weight_refit_map: dict[str, torch.Tensor] = field(default_factory=dict)
-    cpu_weights_reference_holder: dict[str, Union[torch.Tensor]] = field(
-        default_factory=dict
-    )
+    cpu_weights_reference_holder: list[torch.Tensor] = field(default_factory=list)
 
     def record_weight(self, name: str, weight: torch.Tensor) -> None:
         """
@@ -39,7 +36,7 @@ class ConversionContext:
             weight: Weight to record
         """
         self.weight_refit_map[name] = weight
-        self.cpu_weights_reference_holder[name + " CPU_REFERENCE"] = weight
+        self.cpu_weights_reference_holder.append(weight)
 
     def clear_cpu_weights_reference_holder(self) -> None:
         self.cpu_weights_reference_holder.clear()
--- a/py/torch_tensorrt/dynamo/conversion/_ConversionContext.py
+++ b/py/torch_tensorrt/dynamo/conversion/_ConversionContext.py
@@ -29,6 +29,15 @@ class ConversionContext:
     )
 
     def record_weight(self, name: str, weight: torch.Tensor) -> None:
+        """
+        Record the weight and name for refitting and CPU reference.
+        For the refit map, the key is the weight name that appears in the TRT engine and the value is the weight tensor.
+        For the CPU reference holder, we need to hold the reference to the weight tensor until the whole compilation process is complete.
+
+        Args:
+            name: Name of the weight
+            weight: Weight to record
+        """
         self.weight_refit_map[name] = weight
         self.cpu_weights_reference_holder[name + " CPU_REFERENCE"] = weight
 
--- a/py/torch_tensorrt/dynamo/conversion/converter_utils.py
+++ b/py/torch_tensorrt/dynamo/conversion/converter_utils.py
@@ -418,15 +418,8 @@ def create_constant(
                 ctx.cpu_weights_reference_holder[name + " FP4_CONSTANT"] = torch_value
                 return constant.get_output(0)
 
-            # TODO: Refit map uses numpy arrays. Remove this once refit is updated to use torch.Tensor
-            if torch_value.dtype == torch.bfloat16:
-                torch_value_fp32 = torch_value.to(torch.float32)
-                numpy_value = torch_value_fp32.numpy()
-            else:
-                numpy_value = torch_value.numpy()
-
             # Used for refit
-            ctx.weight_refit_map[name + " CONSTANT"] = numpy_value.reshape(-1)
+            ctx.weight_refit_map[name + " CONSTANT"] = torch_value
 
             # This is a buffer to hold the torch.Tensor so that they are alive during the course of TRT compilation.
             ctx.cpu_weights_reference_holder[name] = torch_value
--- a/py/torch_tensorrt/dynamo/conversion/converter_utils.py
+++ b/py/torch_tensorrt/dynamo/conversion/converter_utils.py
@@ -345,7 +345,7 @@ def to_trt_weights(
         - Input tensors are made contiguous before conversion
         - Data type is preserved from the original tensor/array
     """
-    if isinstance(value, np.ndarray):
+    if not isinstance(value, torch.Tensor):
         raise AssertionError(
             f"to_trt_weights can only be called on torch.Tensor, got an object of type: {type(value)}"
         )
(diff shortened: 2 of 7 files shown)


py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -418,15 +418,8 @@ def create_constant(
...
-            # TODO: Refit map uses numpy arrays. Remove this once refit is updated to use torch.Tensor
-            if torch_value.dtype == torch.bfloat16:
-                torch_value_fp32 = torch_value.to(torch.float32)
-                numpy_value = torch_value_fp32.numpy()
-            else:
-                numpy_value = torch_value.numpy()
-
             # Used for refit
-            ctx.weight_refit_map[name + " CONSTANT"] = numpy_value.reshape(-1)
+            ctx.weight_refit_map[name + " CONSTANT"] = torch_value
```
cehongwang: Comment the reason of why adding the " Constant"

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -321,7 +321,15 @@ def cast_int_or_float_to_bool(
 
 
 def to_trt_weights(
-    value: Any, target_quantized_type: Optional[trt.DataType] = None
+    value: Any,
+    record_weight: bool = False,
```
narendasan: Are there usecases for non recorded weights? @peri044 thoughts?

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -321,7 +321,15 @@ def cast_int_or_float_to_bool(
...
+    name: Optional[str] = None,
+    ctx: Optional[ConversionContext] = None,
```
narendasan: Id make ctx the first arg just for the sake of convention

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -321,7 +321,15 @@ def cast_int_or_float_to_bool(
 
 
 def to_trt_weights(
-    value: Any, target_quantized_type: Optional[trt.DataType] = None
+    value: Any,
```
narendasan: Does this need to be a torch.Tensor or do we handle other types?

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -321,7 +321,15 @@ def cast_int_or_float_to_bool(
 
 
 def to_trt_weights(
```
narendasan, line 334: I think we can streamline some arguments like why do we need target, name, layer name and weight name? can we derive some of these from others?

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -351,6 +388,15 @@ def to_trt_weights(
         )
 
 
+def record_weight_in_ctx(
```
narendasan: This could either be a method of the ConversionCtx or it can just be done in to_trt_weights

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -351,6 +388,15 @@ def to_trt_weights(
...
+    ctx: ConversionContext,
+    name: str,
+    value: torch.Tensor,
+) -> None:
+    ctx.weight_refit_map[name] = value
```
narendasan: Have this method do the name formating specific to each map, and the user just passes a general prefix for it

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -415,24 +461,14 @@ def create_constant(
                     weights,
                 )
                 constant.name = name
-                ctx.cpu_weights_reference_holder[name + " FP4_CONSTANT"] = torch_value
+                record_weight_in_ctx(ctx, name + " FP4_CONSTANT", torch_value)
```
narendasan: Why arent we using trt.Weights here>

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -415,24 +461,14 @@ def create_constant(
...
-            else:
-                numpy_value = torch_value.numpy()
-
-            # Used for refit
-            ctx.weight_refit_map[name + " CONSTANT"] = numpy_value.reshape(-1)
-
-            # This is a buffer to hold the torch.Tensor so that they are alive during the course of TRT compilation.
-            ctx.cpu_weights_reference_holder[name] = torch_value
+            # Record the weight in ctx for refit and cpu memory reference
+            record_weight_in_ctx(ctx, name + " CONSTANT", torch_value)
```
narendasan: Same here? We are recording the weight twice now

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -415,24 +461,14 @@ def create_constant(
...
 
             # Convert the torch.Tensor to a trt.Weights object
-            trt_weights = to_trt_weights(torch_value)
+            trt_weights = to_trt_weights(torch_value, record_weight=False)
```
narendasan: The recording should be done once in here and then this function does not need to worry about anything re weights

py/torch_tensorrt/dynamo/conversion/impl/conv.py
```diff
@@ -79,13 +79,23 @@ def convNd(
...
         num_output_maps = weight.shape[0]
         kernel_shape = weight.shape[2:]
-        weight = to_trt_weights(weight)
+        weight = to_trt_weights(
+            weight,
+            record_weight=True,
+            name=name,
+            ctx=ctx,
+            target=target,
+            layer_type_name="CONVOLUTION",
```
narendasan, line 100: Yeah so we can pull the type from the node right?, we just need to know its a KERNEL weight

py/torch_tensorrt/dynamo/conversion/converter_utils.py
cehongwang: Only torch.Tensor. Erroring out other types

py/torch_tensorrt/dynamo/conversion/_ConversionContext.py
```diff
@@ -24,10 +23,14 @@ class ConversionContext:
...
     )
     requires_output_allocator: bool = False
-    weight_refit_map: dict[str, np.array] = field(default_factory=dict)
+    weight_refit_map: dict[str, torch.Tensor] = field(default_factory=dict)
     cpu_weights_reference_holder: dict[str, Union[torch.Tensor]] = field(
         default_factory=dict
     )
 
+    def record_weight(self, name: str, weight: torch.Tensor) -> None:
+        self.weight_refit_map[name] = weight
```
peri044, line 38: add a docstring explaining why we are doing this especially the comment related to `self.cpu_weights_reference_holder[name + " CPU_REFERENCE"] = weight`

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -336,20 +345,51 @@ def to_trt_weights(
...
-        # Tensor must be contiguous before conversion
-        value = value.contiguous()
-        value_trt_dtype = _enums.dtype._from(value.dtype).to(trt.DataType)
-        return trt.Weights(value_trt_dtype, value.data_ptr(), value.nelement())
-    elif isinstance(value, np.ndarray):
-        value = np.ascontiguousarray(value)
-        value_np_dtype = _enums.dtype._from(value.dtype).to(np.dtype, use_default=True)
-        return trt.Weights(value_np_dtype, value.data, value.size)
-    else:
+    if isinstance(value, np.ndarray):
```
peri044: consider checking `if not isinstance(value, torch.Tensor)` to cover all cases

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -336,20 +345,51 @@ def to_trt_weights(
...
-            f"to_trt_weights can only be called on torch.Tensor or np.ndarray, got an object of type: {type(value)}"
+            f"to_trt_weights can only be called on torch.Tensor, got an object of type: {type(value)}"
+        )
+
+    # Weight Recording
+    supported_layer_types = ["CONVOLUTION", "DECONVOLUTION", "CONSTANT"]
+    supported_weight_types = ["KERNEL", "BIAS", "CONSTANT"]
+    assert (
+        layer_type_name in supported_layer_types
+    ), f"Unsupported layer type: {layer_type_name}. Please add the layer type to this function to enable refitting."
```
peri044: Please add the layer type to this function to enable refitting. - what does this mean ? How do we add this ?

py/torch_tensorrt/dynamo/conversion/impl/conv.py
```diff
@@ -85,7 +85,15 @@ def convNd(
 
         num_output_maps = weight.shape[0]
         kernel_shape = weight.shape[2:]
-        weight = to_trt_weights(weight)
+        weight = to_trt_weights(
+            ctx,
+            weight,
```
peri044, line 98: modify the bias in the same way for conv and deconv as well

py/torch_tensorrt/dynamo/conversion/impl/deconv.py
```diff
@@ -85,7 +85,15 @@ def deconvNd(
             weight = torch.unsqueeze(weight, -1)
         num_output_maps = weight.shape[1]
         kernel_shape = weight.shape[2:]
-        weight = to_trt_weights(weight)
+        weight = to_trt_weights(
+            ctx,
+            weight,
+            name,
+            layer_type_name="CONVOLUTION",
```
peri044, line 100: use the same `to_trt_weights` for bias as well.

peri044 (changes requested):

py/torch_tensorrt/dynamo/conversion/_ConversionContext.py
cehongwang: The name " CPU_REFERENCE" is a bit random. Any name could work because all we need is to hold it on CPU. Moreover, since we have refit map, this is actually a bit redundant.

narendasan: Do we need a suffix here? Its just holding references, it should not be inspected later. I dont think it should even be a dictionary

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -321,7 +321,16 @@ def cast_int_or_float_to_bool(
 
 
 def to_trt_weights(
-    value: Any, target_quantized_type: Optional[trt.DataType] = None
+    ctx: ConversionContext,
+    value: torch.Tensor,
+    name: str,
+    layer_type_name: str,
```
narendasan: We should use those literal type annotations

py/torch_tensorrt/dynamo/conversion/converter_utils.py
```diff
@@ -336,20 +345,51 @@ def to_trt_weights(
...
+    supported_layer_types = ["CONVOLUTION", "DECONVOLUTION", "CONSTANT"]
+    supported_weight_types = ["KERNEL", "BIAS", "CONSTANT"]
+    assert (
+        layer_type_name in supported_layer_types
+    ), f"Encountered unsupported layer type: {layer_type_name}. Supported types are: {supported_layer_types}. Manually calling to_trt_weights with a custom layer type is not intended for general use."
+    assert (
+        weight_type_name in supported_weight_types
+    ), f"Encountered unsupported weight type: {weight_type_name}. Supported types are: {supported_weight_types}. Manually calling to_trt_weights with a custom weight type is not intended for general use."
+
+    if weight_type_name == "CONSTANT" and layer_type_name == "CONSTANT":
```
narendasan, line 374: What is the difference between a weight type and a layer type?

cehongwang: Layer type is the type of the layer i.e. CONVOLUTION/DECONVOLUTION. Weight type is the name of the weight attribute i.e. KERNEL, BIAS...

PR #3595, opened 2025-06-16
Fix: fix aot plugin example docstring issue

bowang007:
# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/examples/dynamo/aot_plugin.py
+++ b/examples/dynamo/aot_plugin.py
@@ -1,3 +1,34 @@
+"""
+.. aot_plugin:
+Automatically Generate a TensorRT AOT Plugin 
+===================================================================
+We are going to demonstrate how to automatically generate a plugin for a custom kernel using Torch-TensorRT using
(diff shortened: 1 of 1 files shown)


examples/dynamo/aot_plugin.py
```diff
@@ -1,3 +1,34 @@
+"""
+.. aot_plugin:
```
narendasan: I think you need an _ infront of the aot_plugin like: `_aot_plugin`. Also do we want to render this example? It may need to get added to `index.rst`

bowang007: Hi @narendasan , thanks for the comment!
Do we want to render this example? I was thinking rendering the RMSNorm example instead of this one since this one is a little bit simple.

narendasan: Ok, I still think you need the prefix but we dont need to necessarily render it.

PR #3602, opened 2025-06-17
Cpu memory optimization

cehongwang:
# Description

Delete the extra copy INetworkDefinition creates

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -65,7 +65,7 @@ class UnsupportedOperatorException(RuntimeError):
 
 
 class TRTInterpreterResult(NamedTuple):
-    serialized_engine: bytes
+    engine: trt.ICudaEngine | bytes
```
cehongwang: @narendasan Should we set it to be engine when it is python runtime and serialized engine if it is cpp runtime? In this way we can do serialization in Interpreter.

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -103,8 +109,8 @@ def convert_module(
             "Since Torch-TensorRT runtime is not available, using Python Runtime, some features may not be available"
         )
 
-    return rt_cls(
-        serialized_engine=interpreter_result.serialized_engine,
+    return PythonTorchTensorRTModule(
```
narendasan: This should be the rt_class still but there should be some preprocessing modifying the arguments

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
narendasan: It can just be `trt.ICudaEngine`and post processing for the cpp runtime

narendasan: Like I would have the CPP runtime determine if the engine is live or serialized, and if its live serialize the engine.

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -592,13 +591,11 @@ def _save_weight_mapping(self) -> None:
         torch.cuda.empty_cache()
 
     @needs_refit  # type: ignore[misc]
-    def _insert_engine_to_cache(self, hash_val: str, serialized_engine: bytes) -> None:
+    def _insert_engine_to_cache(self, hash_val: str, engine: bytes) -> None:
+        serialized_engine = engine.serialize()
```
narendasan, line 595: If we are doing this don't we end up paying the serialization cost again?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -772,19 +768,31 @@ def run(
...
+        if self.compilation_settings.use_python_runtime:
+            return TRTInterpreterResult(
+                cuda_engine,
+                self._input_names,
+                self._output_names,
+                self.weight_name_map,
+                self.ctx.requires_output_allocator,
+            )
+        else:
+            serialized_engine = cuda_engine.serialize()
```
narendasan: Just have the TRTInterpreter return a live engine in all cases

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -123,7 +122,7 @@ class PythonTorchTensorRTModule(Module):  # type: ignore[misc]
 
     def __init__(
         self,
-        serialized_engine: Optional[bytes] = None,
+        cuda_engine: trt.ICudaEngine = None,
```
narendasan, line 126: The API should support both since uses may use this runtime independently, its just now we either support `cuda_engine` or `serialized_engine`

narendasan: Same thing for the CPP runtime

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
cehongwang: engine.serialize() does not take any memory. 

```
with io.BytesIO() as engine_bytes:
```
This does

Issue #3588, opened 2025-06-14
🐛 [Bug] Reexport test failures


narendasan:
##  Bug Description

FAILED models/test_reexport.py::test_base_full_compile - AssertionError: False is not true : test_base_model_full_compile TRT outputs don't match with the original model. Cosine sim score: nan Threshold: 0.99
FAILED models/test_reexport.py::test_base_full_compile_multiple_outputs - AssertionError: False is not true : test_base_full_compile_multiple_outputs TRT outputs don't match with the original model. Cosine sim score: nan Threshold: 0.99
FAILED models/test_reexport.py::test_hybrid_relu_fallback - AssertionError: False is not true : test_hybrid_relu_fallback TRT outputs don't match with the original model. Cosine sim score: nan Threshold: 0.99
FAILED models/test_reexport.py::test_resnet18 - AssertionError: False is not true : test_resnet18 TRT outputs don't match with the original model. Cosine sim score: 0.048845380544662476 Threshold: 0.99
FAILED models/test_reexport.py::test_resnet18_dynamic - KeyError: u0
FAILED models/test_reexport.py::test_resnet18_dynamic_fallback - KeyError: u0
FAILED models/test_reexport.py::test_random_dynamic_fallback - KeyError: u0

## To Reproduce

1.
2.
3.

## Environment

> Build information about Torch-TensorRT can be found by turning on debug messages

 - Torch-TensorRT Version (e.g. 1.0.0):
 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source):
 - Build command you used (if compiling from source):
 - Are you using local sources or building from archives:
 - Python version:
 - CUDA version:
 - GPU models and configuration:
 - Any other relevant information:

PR #3617, opened 2025-06-25
fix: Fix unbacked sym int not found issue

peri044:
# Description

Fixes https://github.com/pytorch/TensorRT/issues/3588

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/_tracer.py
```diff
@@ -70,6 +70,9 @@ def trace(
     if kwarg_inputs is None:
         kwarg_inputs = {}
 
+    # After validation above, arg_inputs is guaranteed to be non-None
+    assert arg_inputs is not None, "arg_inputs should not be None after validation"
```
narendasan: What would the action for the user be here if they hit this assert?

peri044: So, this is actually a fix for mypy type error. This assertion will not be hit because the error would be caught at https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/_tracer.py#L60-L67. 
If we should handle the mypy issue in a different way, I can investigate.

narendasan: I would just add a comment that says this is for type narrowing then

zewenli98 (approved):
LGTM. just a quick question: `add_var_to_val(k, v)` would fail but `set_unbacked_var_to_val(k, v)` would work if `k` already exists right?   The tests worked well in the previous. Any changes caused the issue?

peri044:
> LGTM. just a quick question: `add_var_to_val(k, v)` would fail but `set_unbacked_var_to_val(k, v)` would work if `k` already exists right? The tests worked well in the previous. Any changes caused the issue?

Yes, the tests failed with recent versions of Pytorch which went undetected. `add_var_to_val` wasn't adding the unbacked symint  to `var_to_val` anymore throughout the torch.export tracing and it resulted in KeyError: u0. The recommended way by Meta to add unbacked symints is through `set_unbacked_var_to_val` and this ensures unbacked symints are available via `node.shape_env.unbacked_var_to_val` attribute.

PR #3619, opened 2025-06-25
fix ts fe test error.

lanluo-nvidia:
# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/tests/modules/hub.py
+++ b/tests/modules/hub.py
@@ -19,10 +19,66 @@ if not torch.cuda.is_available():
         "No GPU found. Please check if installed torch version is compatible with CUDA version"
     )
 
+if not importlib.util.find_spec("torchvision"):
+    print(f"torchvision is not installed, skip models download")
+    exit(0)
+
 # Downloads all model files again if manifest file is not present
 MANIFEST_FILE = "model_manifest.json"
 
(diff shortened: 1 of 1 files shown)


tests/modules/hub.py
```diff
@@ -19,10 +19,66 @@
         "No GPU found. Please check if installed torch version is compatible with CUDA version"
     )
 
+if not importlib.util.find_spec("torchvision"):
+    print(f"torchvision is not installed, skip models download")
+    exit(0)
+
```
peri044: can you merge this in an else block with already existing snippet at the top ? 
```py
if importlib.util.find_spec("torchvision"):
    import timm
    import torchvision.models as models
```

peri044: also does `exit(0)` stop running this test and label it as failure ?

lanluo-nvidia: good point, exit(0) will not mark the test as failure.
but reminded me that I should not completely exit(0) if torchvision is not installed.
Because we have some custom models located in the tests/modules/custom_models.py which should be downloaded via hub.py even without torchvision installed.

peri044 (approved):
LGTM

PR #3629, opened 2025-06-27
Closed the perf gap of resnet and enabled refit

cehongwang:
# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_refit.py
+++ b/py/torch_tensorrt/dynamo/_refit.py
@@ -22,6 +22,9 @@ from torch_tensorrt.dynamo.conversion._ConverterRegistry import (
     DYNAMO_CONVERTERS as CONVERTERS,
 )
 from torch_tensorrt.dynamo.conversion._TRTInterpreter import TRTInterpreter
+from torch_tensorrt.dynamo.conversion.impl.normalization.ops import (
+    batch_norm_constant_folding,
+)
 from torch_tensorrt.dynamo.conversion.truncate_double import repair_double_inputs
 from torch_tensorrt.dynamo.lowering import (
     get_decompositions,
@@ -91,13 +94,15 @@ def construct_refit_mapping_from_weight_name_map(
 ) -> dict[Any, Any]:
     engine_weight_map = {}
     for engine_weight_name, (sd_weight_name, np_weight_type) in weight_name_map.items():
+        # Add more constant folding converters here
         if engine_weight_name.split(" ")[-1] in ["SCALE", "SHIFT"]:
             # Batch Norm Layer
             params = {}
             for w in sd_weight_name:
                 params[w.split(".")[-1]] = state_dict[w].cuda()
-            scale = params["weight"] / torch.sqrt(params["running_var"] + 1e-7)
-            shift = params["bias"] - params["running_mean"] * scale
+            # Batch norm constant folding
+
+            scale, shift = batch_norm_constant_folding(**params, eps=1e-7)
             # Set scale to scale or shift to shift
             engine_weight_map[engine_weight_name] = eval(
                 engine_weight_name.split(" ")[-1].lower()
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -48,89 +49,160 @@ def batch_norm(
...
+            isinstance(running_var, torch.Tensor),
+        ]
+    ):
+        if weight is None:
+            weight = 1.0
+
+        if bias is None:
+            bias = 0.0
+
+        if running_mean is None:
```
peri044: add a docstring in this function explaining the batchnorm folding and why it is the default behavior for performance reasons for future reference.

peri044 (commented):
minor comment. LGTM

py/torch_tensorrt/dynamo/_refit.py
```diff
@@ -90,7 +91,18 @@ def construct_refit_mapping_from_weight_name_map(
 ) -> dict[Any, Any]:
     engine_weight_map = {}
     for engine_weight_name, (sd_weight_name, np_weight_type) in weight_name_map.items():
-        if sd_weight_name not in state_dict:
+        if engine_weight_name.split(" ")[-1] in ["SCALE", "SHIFT"]:
```
narendasan, line 98: We should abstract this imo. Like if there are any weight types that require constant folding in converter this should be associated with the converter. Then the refit system will just iterate through all these constant fold operations. Ideally the converter can use the same implementation

cehongwang: Currently BN is the only one. Do you think we should have a constant_fold function and have `refit` and `conversion` call that function?

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -47,90 +48,163 @@ def batch_norm(
...
+    # In this way, the batch norm layer will be fused with the Convolution layer and get a performance boost.
+    if all(
+        [
+            isinstance(weight, torch.Tensor),
+            isinstance(bias, torch.Tensor),
+            isinstance(running_mean, torch.Tensor),
+            isinstance(running_var, torch.Tensor),
+        ]
+    ):
+        if weight is None:
```
HolyWu: I think this check is redundant. If `weight` was None, then the `isinstance(weight, torch.Tensor)` above would already return `False`, and hence it wouldn't enter this branch at all.

py/torch_tensorrt/dynamo/_refit.py
narendasan: yeah we should think of some sort of abstraction where the constant folding process is defined once for both refit and conversion

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
narendasan: Think the correct check on L54 is more like if all parameters are not ITensors. If one is none and the rest are torch tensors or floats we can still do constant folding. but if one of the parameters is a ITensor then we cant

cehongwang: IDK when I don't have the check the pre-commit gives me errors

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -47,90 +48,164 @@ def batch_norm(
 
     # Save the original output shape for later use
     output_shape = input.shape
+    # We perform constant folding for batch norm when the weight, bias, running_mean, and running_var are all tensors.
+    # Batch norm operation can be fused into a single layer, which is more efficient than the original implementation.
+    # In this way, the batch norm layer will be fused with the Convolution layer and get a performance boost.
+    if not all(
```
narendasan: @cehongwang, weight, bias, etc can be none but we can still do constant folding. We just need to check that they are not ITensors: https://github.com/pytorch/pytorch/blob/f16053f0c9a09fa337fbf85aaf64f88712b8dcdb/aten/src/ATen/native/native_functions.yaml#L1093

narendasan: So this check should probably be `if any([isinstance(weight, trt.ITensor), ...])`

narendasan (approved):
LGTM

PR #3615, opened 2025-06-25
Remove Bazel files from wheel

mering:
# Description

Once the wheel is built, the Bazel files are no longer needed.

Fixes #3477

## Type of change

Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #3641, opened 2025-07-01
Remove BAZEL build files from python package and changes to make cpp tests work

apbose:
Follow up from https://github.com/pytorch/TensorRT/pull/3615

### DIFF
--- a/docker/MODULE.bazel.ngc
+++ b/docker/MODULE.bazel.ngc
@@ -24,13 +24,16 @@ git_override(
 
 local_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:local.bzl", "local_repository")
 
-# External dependency for torch_tensorrt if you already have precompiled binaries.
-local_repository(
-    name = "torch_tensorrt",
-    path = "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/",
-)
 
 new_local_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:local.bzl", "new_local_repository")
+
+
+# External dependency for torch_tensorrt if you already have precompiled binaries.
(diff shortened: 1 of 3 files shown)


third_party/BUILD
```diff
@@ -0,0 +1,53 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
```
narendasan: Move this to `//third_party/torch_tensorrt/BUILD

docker/MODULE.bazel.ngc
```diff
@@ -24,14 +24,17 @@ git_override(
...
 
+
+new_local_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:local.bzl", "new_local_repository")
+
+
 # External dependency for torch_tensorrt if you already have precompiled binaries.
-local_repository(
+new_local_repository(
     name = "torch_tensorrt",
+    build_file = "@//third_party:BUILD",
```
narendasan: Adjust this path accordingly

setup.py
```diff
@@ -613,8 +613,6 @@ def run(self):
     package_data.update(
         {
             "torch_tensorrt": [
-                "BUILD",
-                "WORKSPACE",
                 "include/torch_tensorrt/*.h",
```
narendasan, line 616: Should we just delete the bzl folder now?

docker/MODULE.bazel.ngc
```diff
@@ -55,8 +58,26 @@ new_local_repository(
     build_file = "third_party/libtorch/BUILD"
 )
 
+new_local_repository(
```
narendasan, line 64: Add comments that these are WARs and not functional

PR #3650, opened 2025-07-02
Upgrade perf_run script to support TRT 10 and fix some issues

### LINKED ISSUE #3634: 🐛 [Bug] Changing input size would affect the TRT engine size, testing on BERT
##  Bug Description

I'm testing BERT with commands `python perf_run.py --backends=dynamo --inputs="(512, 128)@int32;(512, 128)@int32" ...` and `python perf_run.py --backends=dynamo --inputs="(256, 128)@int32;(256, 128)@int32" ...`. I only changed the inputs here and saved their TRT engines but the two engines have different sizes:

```
Using --inputs="(256, 128)@int32;(256, 128)@int32":
Number of layers: 1277
Number of inputs: 2
Number of outputs: 2
Input 0: input_ids, shape: (256, 128), dtype: DataType.INT32
Input 1: attention_mask, shape: (256, 128), dtype: DataType.INT32
Output 0: output0, shape: (256, 128, 768), dtype: DataType.FLOAT
Output 1: output1, shape: (256, 768), dtype: DataType.FLOAT
TRT Engine uses: 516.5105247497559 Mb of Memory

Using --inputs="(512, 128)@int32;(512, 128)@int32":
Number of layers: 1277
Number of inputs: 2
Number of outputs: 2
Input 0: input_ids, shape: (512, 128), dtype: DataType.INT32
Input 1: attention_mask, shape: (512, 128), dtype: DataType.INT32
Output 0: output0, shape: (512, 128, 768), dtype: DataType.FLOAT
Output 1: output1, shape: (512, 768), dtype: DataType.FLOAT
TRT Engine uses: 612.4900169372559 Mb of Memory
```

## To Reproduce

1. `cd tools/perf`
2. `python perf_run.py ...`

## Expected behavior
Torch-TensorRT Dynamo backend looks faster than Inductor, but a little bit slower than ONNX path. I was wondering the reason, so I pulled out the engines and directly ran them. When I check the engines, I found the engine size got larger and larger as inputs increase. However, the engine sizes exported from ONNX path keep the same.

zewenli98:
# Description

1. upgraded TRT API to execute_async_v3
2. added `optimization_level` arg and set highest optimization 3 as default
3. fixed some known issues

Fixes #3634

## Type of change

# Checklist:


peri044 (commented):
1) remove lower_linear 
2) consider try/catch for onnx export 
3) Update perf_run readme with the latest arguments
4) Run llm example with the constant_folding to test accuracy

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3579,3 +3579,23 @@ def aten_ops_nonzero(
         name,
         args[0],
     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.linear.default, supports_dynamic_shapes=True)
+@dynamo_tensorrt_converter(torch.ops.aten.linear, supports_dynamic_shapes=True)
```
HolyWu: I think that registering a converter for OpOverloadPacket has no effect.

py/torch_tensorrt/dynamo/conversion/impl/linear.py
```diff
@@ -0,0 +1,54 @@
+from typing import Optional, Union
+
+import numpy as np
+import tensorrt as trt
+import torch
+from torch.fx.node import Target
+from torch_tensorrt.dynamo.conversion import impl
+from torch_tensorrt.dynamo.conversion._ConversionContext import ConversionContext
+from torch_tensorrt.dynamo.conversion.converter_utils import SourceIR, get_trt_tensor
+from torch_tensorrt.fx.types import TRTTensor
```
HolyWu: torch_tensorrt.dynamo.types

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: I found for some models in fp16, for example, bert, registering a linear op can reduce latency. It seems no effect for fp32 though.

zewenli98: Could you double check in your case if the linear converter affects perf in fp16?

HolyWu: I meant that you should only register `torch.ops.aten.linear.default` (an OpOverload) and remove the second line because registering `torch.ops.aten.linear` (an OpOverloadPacket) was redundant.

tools/perf/custom_models.py
```diff
@@ -33,3 +33,17 @@ def StableDiffusionUnet():
         "CompVis/stable-diffusion-v1-4", revision="fp16", torch_dtype=torch.float16
     )
     return pipe.unet
+
+
+def UNet():
+    from monai.networks.nets import UNet
```
peri044, line 39: Consider adding monai as a dependency or instruction to install in README.

tools/perf/requirements.txt
```diff
@@ -4,6 +4,5 @@ pyyaml
 onnx
 pandas
 transformers
```
peri044: Let's try to pin transformers and diffusers version that we usually support. transformers==4.51.3. Please list whatever diffusers version works for you.

Issue #3647, opened 2025-07-01
🐛 [Bug] index and index_put don't support dynamic shapes


peri044:
##  Bug Description

index and index_put converter cause graph breaks due to no `supports_dynamic_shapes=True`. 
1) https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py#L394-L401
2) https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py#L851-L859

## Environment

> Build information about Torch-TensorRT can be found by turning on debug messages

 - Torch-TensorRT Version (e.g. 1.0.0):
 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source):
 - Build command you used (if compiling from source):
 - Are you using local sources or building from archives:
 - Python version:
 - CUDA version:
 - GPU models and configuration:
 - Any other relevant information:

apbose:
I took a look at index converter. It needs to have `supports_dynamic_shapes=True`. 
1. This is a side question. Index converter had a test case with input dynamic. This part in `_ConverterRegistry.py`
```
if is_valid := candidate.capability_validator(
                            node, self.compilation_settings
                        ) and (
                            assume_dynamic_shape_support
                            or not node_has_dynamic_shapes(node)
                            or candidate.supports_dynamic_shapes
                        ):
```
returned true in that case, since `node_has_dynamic_shapes(node)` returned False because the meta data of the nodes are not filled up during compile time. Would that be a desirable behavior?

2. The test cases have case where input is dynamic, but not with index dynamic. I am trying something like below which works in pytorch. The index signature is https://github.com/zdevito/ATen/blob/master/aten/src/ATen/native/native_functions.yaml#L1324C3-L1324C63
where indices is a tuple of Tensor. However, in this case the index argument is a tuple of dynamic tensors. That means the second argument in the aten.index.Tensor node is not a List containing dynamic tensors (which can be covered with `torch.where` with some hardcoding) or a static list/tuple, but rather a dynamic FX node (from nonzero which was earlier not supported in TRT). Currently, the index converter validator doesn't handle that, which causes the converter to be skipped. The converter is long and complicated :( but working on the fix.
```
def test_index_input_dynamic_index_dynamic(self):
[...]

PR #3694, opened 2025-07-17
Index converter dynamic cases fix

apbose:
This PR-
1. Addresses https://github.com/pytorch/TensorRT/issues/3647 for index
2. Adds a test case for dynamic index

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/select.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/select.py
@@ -50,16 +50,22 @@ def select(
     return layer.get_output(0)
 
 
-def is_boolean_tensor(tensor: Union[TRTTensor, np.ndarray, torch.Tensor]) -> bool:
-    if isinstance(tensor, (torch.Tensor, np.ndarray, TRTTensor)):
+def is_boolean_tensor(
+    tensor: Union[TRTTensor, np.ndarray, torch.Tensor, torch.fx.Node],
+) -> bool:
+    if isinstance(tensor, torch.Tensor):
         return bool(tensor.dtype == torch.bool)
+    elif isinstance(tensor, np.ndarray):
+        return bool(tensor.dtype == np.bool_)
+    elif isinstance(tensor, TRTTensor):
+        return bool(tensor.dtype == trt.DataType.BOOL)
     # when index is a node
     else:
         val = tensor.meta.get("val")
         if val is not None and val.dtype is torch.bool:
             return True
 
-    return isinstance(tensor, (torch.Tensor, np.ndarray)) and tensor.dtype == torch.bool
+    return False
 
 
 def expand_boolean_indices(
--- a/tests/py/dynamo/conversion/test_index_aten.py
+++ b/tests/py/dynamo/conversion/test_index_aten.py
@@ -125,6 +125,17 @@ class TestIndexConverter(DispatchTestCase):
             [input, index0],
         )
 
+    def test_index_zero_two_dim_ITensor_mask(self):
+        class TestModule(nn.Module):
+            def forward(self, x, index0):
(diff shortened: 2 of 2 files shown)


peri044:
The index dtype validator doesn't support bool type 
```py
(Pdb) node.args
(_reshape_copy_597, [eq_1427])
(Pdb) node.args
(_reshape_copy_597, [eq_1427])
(Pdb) node.args[0].dtype
*** AttributeError: 'Node' object has no attribute 'dtype'
(Pdb) node.args[0].meta['val']
FakeTensor(..., device='cuda:0', size=(s43*s53, 2048), dtype=torch.float16)
(Pdb) node.args[0].meta['val'].dtype
torch.float16
(Pdb) node.args[1][0].meta['val'].dtype
torch.bool
```
The inputs to index are FP16 and bool. In this case, converter is not being invoked and falls back to PyTorch

apbose:
oh ok, looks like in https://github.com/zdevito/ATen/blob/master/aten/src/ATen/native/native_functions.yaml#L1324 tensor indices as bool should denote mask then. I think a non zero layer to get the corresponding indices should work. I will push in the change

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -53,6 +53,65 @@ def select(
     return layer.get_output(0)
 
 
+def is_boolean_tensor(tensor: Union[TRTTensor, np.ndarray, torch.Tensor]) -> bool:
+    if isinstance(tensor, (TRTTensor)):
+        val = tensor.meta.get("val")
```
lanluo-nvidia, line 64: if getattr(tensor, “meta”, None) is None:
            return tensor.dtype == trt.DataType.BOOL
I hit with issue where tensor has no meta in gemma3 model.

lanluo-nvidia (approved):
One comment, please take a look, others LGTM

apbose:
I think the tests are passing related to this test. @peri044 could you please review?

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -51,6 +50,71 @@ def select(
     return layer.get_output(0)
 
 
+def is_boolean_tensor(tensor: Union[TRTTensor, np.ndarray, torch.Tensor]) -> bool:
+    if isinstance(tensor, (torch.Tensor, np.ndarray, TRTTensor)):
+        return bool(tensor.dtype == torch.bool)
```
peri044, line 57: I think this condition might fail when tensor is an instance of `TRTTensor` but it checks equality with `torch.bool`

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -51,6 +50,71 @@ def select(
     return layer.get_output(0)
 
 
+def is_boolean_tensor(tensor: Union[TRTTensor, np.ndarray, torch.Tensor]) -> bool:
```
peri044: if tensor can be a node, then the type hint should also add this

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -51,6 +50,71 @@ def select(
...
+    if isinstance(tensor, (torch.Tensor, np.ndarray, TRTTensor)):
+        return bool(tensor.dtype == torch.bool)
+    # when index is a node
+    else:
+        val = tensor.meta.get("val")
+        if val is not None and val.dtype is torch.bool:
+            return True
+
+    return isinstance(tensor, (torch.Tensor, np.ndarray)) and tensor.dtype == torch.bool
```
peri044: does this line ever get executed ? consider removing it if it's not

PR #3703, opened 2025-07-22
feat: revert linear converter

### LINKED ISSUE #3702: 🐛 [Bug] perf gap reduce on BERT
##  Bug Description

Compare the perf of Torch-TRT against ONNX-TRT.

In fp32:
1. Skipping constant folding of embedding layers can reduce engine size. It doesn't affect latency or precision
2. Disabling linear decomposition + adding linear converter doesn't affect latency
3. opt_level=3 or 5 get almost same latency
4. onnx-trt takes much longer in compile time
5. torch-trt is ~2.5% slower than onnx-trt

In fp16:
1. Skipping constant folding of embedding layers can reduce engine size. It doesn't affect latency or precision
2. Disabling linear decomposition + adding linear converter reduces ~18% latency
3. opt_level=3 or 5 get almost same latency
4. onnx-trt takes much longer in compile time
5. torch-trt is ~11% slower than onnx-trt

## To Reproduce
run `perf_run.py` script

zewenli98:
# Description

This PR reverts back **linear converter** because it shows perf improvements on BERT in fp16 if we take linear op as a whole. I didn't observe significant improvements on BERT fp32 though.

Fixes #3702

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -3579,3 +3579,22 @@ def aten_ops_nonzero(
         name,
         args[0],
     )
+
+
+@dynamo_tensorrt_converter(torch.ops.aten.linear.default, supports_dynamic_shapes=True)
+def aten_ops_linear(
+    ctx: ConversionContext,
+    target: Target,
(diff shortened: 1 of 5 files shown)


cehongwang (approved):
lgtm

PR #3705, opened 2025-07-23
addressing ngc aarch64 error

apbose:


narendasan (commented):
Can you run the buildifer linter?

PR #3717, opened 2025-07-24
fix: prelu perf gap on Unet

### LINKED ISSUE #3715: 🐛 [Bug] Torch-TRT's prelu is slower than ONNX-TRT
##  Bug Description

In Monai/Unet, I observed that Torch-TRT's prelu is slower than ONNX-TRT, and what is worse is that using prelu converter is even slower than decomposing it. The reason seems to be additional layers are inserted into the graph:

```
ONNX-TRT:
{ "name" : "PWN(val_9, PWN(node_prelu))", "timeMs" : 1.10378, "averageMs" : 0.00951531, "medianMs" : 0.009216, "percentage" : 0.0934733 }

Torch-TRT w/o decomposition:
{ "name" : "__myl_Cast_myl3_1", "timeMs" : 0.95728, "averageMs" : 0.00878239, "medianMs" : 0.009216, "percentage" : 0.0743985 }
{ "name" : "model.0.conv.unit0.adn.A/_prelu_kernel_constant_1", "timeMs" : 0, "averageMs" : 0, "medianMs" : 0, "percentage" : 0 }
{ "name" : "PWN([PARAMETRIC_RELU]-[aten_ops._prelu_kernel.default]-[model.0.conv.unit0.adn.A/_prelu_kernel])", "timeMs" : 2.87757, "averageMs" : 0.0263997, "medianMs" : 0.026624, "percentage" : 0.223641 }

Torch-TRT w/ decomposition:
{ "name" : "__myl_MulGtrSele_myl3_1", "timeMs" : 0.946529, "averageMs" : 0.00884606, "medianMs" : 0.009216, "percentage" : 0.0707629 }
```

zewenli98:
# Description

In Monai/Unet, I observed that Torch-TRT's prelu is slower than ONNX-TRT, and what is worse is that using prelu converter is even slower than decomposing it.

Update: The perf issue is because `to_copy` forces to insert Cast Layer.

Fixes #3715 #3723 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
+++ b/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
@@ -440,7 +440,7 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
             except Exception:
                 return torch.all(sd_weight == network_weight)
 
-    @needs_refit
+    @needs_refit  # type: ignore[misc]
     def _save_weight_mapping(self) -> None:
         """
         Construct the weight name mapping from engine weight name to state_dict weight name.
@@ -577,7 +577,7 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
         gc.collect()
         torch.cuda.empty_cache()
 
-    @needs_refit
+    @needs_refit  # type: ignore[misc]
     def _insert_engine_to_cache(self, hash_val: str, serialized_engine: bytes) -> None:
         # TODO: @Evan is waiting for TRT's feature to cache the weight-stripped engine
         # if not self.compilation_settings.strip_engine_weights:
@@ -605,7 +605,7 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
             ),
         )
 
-    @needs_refit
+    @needs_refit  # type: ignore[misc]
     def _pull_cached_engine(self, hash_val: str) -> Optional[TRTInterpreterResult]:
         # query the cached TRT engine
         cached_data = self.engine_cache.check(hash_val)  # type: ignore[union-attr]
@@ -941,7 +941,14 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
                 f"Specified output dtypes ({len(self.output_dtypes)}) differ from number of outputs ({len(outputs)})"
             )
 
+        marked_outputs_ids = []
         for i, output in enumerate(outputs):
+            # In some cases, the same output tensor may be marked multiple times, such as _to_oppy,
+            # so we skip marking if the output is already marked
+            if id(output) in marked_outputs_ids:
+                continue
+            marked_outputs_ids.append(id(output))
+
             name = f"output{i}"
 
             output_dtype = dtype.unknown
--- a/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
+++ b/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
@@ -943,7 +943,7 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
 
         marked_outputs_ids = []
         for i, output in enumerate(outputs):
-            # In some cases, the same output tensor may be marked multiple times, such as _to_oppy,
+            # In some cases, the same output tensor may be marked multiple times, such as _to_copy,
             # so we skip marking if the output is already marked
             if id(output) in marked_outputs_ids:
                 continue
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1123,7 +1123,7 @@ def aten_ops_clone_copy_placeholder(
         name,
         args[0],
         kwargs.get("dtype", args[0].dtype),
-        force_layer=True,
+        force_layer=False,
     )
 
 
@@ -1226,7 +1226,7 @@ def aten_ops_sum(
             name,
             sum_,
             kwargs["output_dtype"],
-            force_layer=True,
+            force_layer=False,
         )
     else:
         return sum_
@@ -3229,7 +3229,7 @@ def aten_ops_copy(
         name,
         src,
         src.dtype,
-        force_layer=True,
+        force_layer=False,
     )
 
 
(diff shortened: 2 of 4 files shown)


narendasan (approved):
LGTM

narendasan:
If possible consider adding a test, we talked a bit about a new INetwork verification test based on expected numbers of ops. Might be a simple case to build the harness for

zewenli98:
@narendasan I'll merge this PR now and cherrypick to the release 2.8, and then work on the number of ops check in another PR for the next release. sounds like a plan?

zewenli98:
there's an error. will merge after fixing

py/torch_tensorrt/dynamo/conversion/impl/prelu.py
```diff
@@ -15,6 +16,12 @@ def prelu(
     input: TRTTensor,
     weight: TRTTensor,
 ) -> TRTTensor:
-    layer = ctx.net.add_parametric_relu(input, weight)
+    # TRT requires that the slopes tensor must be unidirectional broadcastable to the input tensor:
+    # the rank of the two tensors must be the same, and all dimensions of the slopes tensor must
+    # either equal the input tensor or be 1. The output tensor has the same shape as the input tensor.
+    input, weight = impl.elementwise.broadcast(
+        ctx, input, weight, f"{name}_broadcast_input", f"{name}_broadcast_weight"
+    )
```
HolyWu: This is probably redundant because PyTorch already reshapes the weight beforehand so we already receive the weight of the required shape.


```py
import torch
import torch_tensorrt


class MyModule(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()

    def forward(self, x: torch.Tensor, weight: torch.Tensor) -> torch.Tensor:
        return torch.nn.functional.prelu(x, weight)


with torch.inference_mode():
    model = MyModule().eval().cuda().half()

    inputs = (
        torch.randn(1, 3, 224, 224, dtype=torch.half, device="cuda"),
        torch.randn(3, dtype=torch.half, device="cuda"),
    )

    exported_program = torch.export.export(model, inputs)

    with torch_tensorrt.dynamo.Debugger():
        trt_model = torch_tensorrt.dynamo.compile(
            exported_program, inputs, enabled_precisions={torch.half}, min_block_size=1
        )

    torch.testing.assert_close(trt_model(*inputs), model(*inputs), rtol=5e-3, atol=5e-3)
```

```py
10:39:52 - DEBUG - Invoking DynamoPassManager and applying lowering passes: [<function remove_detach at 0x000001A0385E4EA0>]
10:39:52 - DEBUG - Removed 0 detach nodes:
graph():
    %weight : [num_users=1] = placeholder[target=weight]
    %prelu : [num_users=1] = call_function[target=torch.ops.aten.prelu.default](args = (%x, %weight), kwargs = {})
    return (prelu,)
10:39:52 - DEBUG - Input graph: graph():
    %weight : [num_users=1] = placeholder[target=weight]
    %_reshape_copy : [num_users=1] = call_function[target=torch.ops.aten._reshape_copy.default](args = (%weight, [1, 3, 1, 1]), kwargs = {})
    %_prelu_kernel : [num_users=1] = call_function[target=torch.ops.aten._prelu_kernel.default](args = (%x, %_reshape_copy), kwargs = {})
    return (_prelu_kernel,)
```

zewenli98:
```
def forward(self, x):
    x_1 = x + 1
    y = torch.ops.aten._to_copy.default(x_1, dtype=torch.float)
    z = torch.ops.aten._to_copy.default(x_1, dtype=torch.float)
    return y, z
```
this case fails because `self.engine.num_io_tensors=2`, `len(self.input_names)=1`, `len(self.output_names)=2`, which doesn't meet the assertion:
```
assert self.engine.num_io_tensors == len(self.input_names) + len(self.output_names)
```

self.input_names: ['x']
self.output_names: ['output0', 'output1']

Since `output0` is invalid, it gets `self.engine.num_io_tensors=2`.

py/torch_tensorrt/dynamo/conversion/impl/prelu.py
zewenli98: in case that _prelu_kernel is directly called?

HolyWu: then you still get the wrong shape of weight because `broadcast` prepend 1s to the beginning and hence you get `(1, 1, 1, 3)` rather than `(1, 3, 1, 1)` with the above example.

zewenli98: thanks for the bug report. I'll revert and leave pytorch to do the reshape

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -941,7 +941,14 @@ def output(self, target: str, args: Any, kwargs: Any) -> List[Any]:
                 f"Specified output dtypes ({len(self.output_dtypes)}) differ from number of outputs ({len(outputs)})"
             )
 
+        marked_outputs_ids = []
         for i, output in enumerate(outputs):
+            # In some cases, the same output tensor may be marked multiple times, such as _to_oppy,
```
narendasan: Small typo

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -941,7 +941,14 @@ def output(self, target: str, args: Any, kwargs: Any) -> List[Any]:
...
+            # so we skip marking if the output is already marked
+            if id(output) in marked_outputs_ids:
```
narendasan, line 948: Where does this id function come from?

zewenli98: `id()` is a built-in function that returns the "identity" of an object.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1094,7 +1094,7 @@ def aten_ops_clone_copy_dtype(
         name,
         args[0],
         kwargs.get("dtype", args[0].dtype),
-        force_layer=True,
+        force_layer=False,
```
peri044: I think there might be cases where we might need to actually force_layer=True. Do you know when that would be useful?. Also consider adding a comment here conveying that force_layer=False results in better performance.

zewenli98: Do you know when we need force_layer=True? My understanding is that 1) Since `cast_trt_tensor` would be called even if the force_layer=False, it only adds Cast Layer if necessary. 2) If an operator satisfies `not is_only_operator_on_placeholder`, that means there's other operator on the placeholder or the lowering pass `repair_input_as_output` to promise "input is not output", so we don't have to explicitly add Cast Layer.

PR #3732, opened 2025-07-31
feat: check the num of ILayer types and compare with ONNX-TRT in converter tests

zewenli98:
# Description

This PR enables checking the num of ILayer types and comparing with ONNX-TRT in converter tests. 
To improve perf, we hope Torch-TRT to get the expected ILayer types for every single converter.
Since we observed on many models that ONNX-TRT generated TRT engines typically have better perf than Torch-TRT generated ones, to reduce perf gaps between Torch-TRT and ONNX-TRT, we introduce the mechanism to make sure Torch-TRT has the same num of ILayer types as ONNX-TRT.

## Type of change

# Checklist:

### DIFF
--- a/tests/py/dynamo/conversion/harness.py
+++ b/tests/py/dynamo/conversion/harness.py
@@ -1,10 +1,13 @@
 # type: ignore
 
 import logging
+import os
+import tempfile
 import time
 import unittest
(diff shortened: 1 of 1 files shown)


zewenli98:
After discussion, we decided not to match with ONNX-TRT for every single converter. Instead, we are going to optimize in sub-graph level.

PR #3740, opened 2025-08-01
Feat: Pre-quantized LLM model support

keehyuna:
# Description

Support pre-quantized HF models and post-training quantization (PTQ) option for [run_llm.py](https://github.com/pytorch/TensorRT/blob/main/tools/llm/run_llm.py)

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/tests/py/dynamo/llm/test_llm_models.py
+++ b/tests/py/dynamo/llm/test_llm_models.py
@@ -23,15 +23,15 @@ def test_llm_decoder_layer(precision):
         args.debug = False
         args.num_tokens = 128
         args.model = "Qwen/Qwen2.5-0.5B-Instruct"
-        args.precision = precision
+        args.model_precision = precision
         args.min_block_size = 1
         args.prompt = "What is parallel programming ?"
-        if args.precision == "FP16":
+        if args.model_precision == "FP16":
             dtype = torch.float16
-        elif args.precision == "BF16":
+        elif args.model_precision == "BF16":
             dtype = torch.bfloat16
         else:
-            args.precision = "FP32"
+            args.model_precision = "FP32"
             dtype = torch.float32
 
         model = (


tools/llm/quantize_utils.py
```diff
@@ -0,0 +1,267 @@
...
+    calibrate_loop = create_forward_loop(dataloader=calib_dataloader)
+
+    model = mtq.quantize(model, quant_cfg, forward_loop=calibrate_loop)
+    if args.debug:
+        mtq.print_quant_summary(model)
+
+    return model
+
+
+class TensorRTQuantizedLinear(torch.nn.Module):
```
narendasan, line 74: @peri044 Is this something we might want to upstream to ModelOpt in the future?

tools/llm/quantize_utils.py
```diff
@@ -0,0 +1,267 @@
...
+    # Load and parse quantization configuration
+    hf_quant_config_path = f"{hf_folder}/hf_quant_config.json"
+    if os.path.exists(hf_quant_config_path):
+        with open(hf_quant_config_path, "r") as f:
+            hf_quant_config = json.load(f)
+            hf_quant_config = hf_quant_config["quantization"]
+
+        hf_quant_algo = hf_quant_config.pop("quant_algo", None)
+        if hf_quant_algo != "FP8" and hf_quant_algo != "NVFP4":
+            raise RuntimeError("Only FP8 or NVFP4 quantization is supported")
```
narendasan: How would it be different for MXFP4?

tools/llm/quantize_utils.py
narendasan: Or pull into main torch-tensorrt as a pass?

narendasan: I guess its somewhat HF specific, so remaining in this tool would make sense but are there some parts we could make generic for any sort of quantization workflow (e.g. torchao)?

tools/llm/run_llm.py
```diff
@@ -54,10 +55,13 @@ def get_model(args):
                 args.model,
                 use_cache=False,
                 attn_implementation="sdpa",
+                ignore_mismatched_sizes=True,
             )
             .eval()
             .cuda()
         )
+    if args.pre_quantized:
```
narendasan: Is this something we could determine automatically?

tools/llm/run_llm.py
```diff
@@ -234,13 +239,36 @@ def measure_perf(trt_model, input_signature, backend_name):
     arg_parser.add_argument(
         "--benchmark", action="store_true", help="Enable benchmark (default: False)"
     )
-
+    arg_parser.add_argument(
+        "--qformat",
```
narendasan: For the sake of clarity, this should be something like `quant-[scheme/format]`

tools/llm/quantize_utils.py
keehyuna: Thanks. I think quantize_model() can be moved to function like torch_tensorrt.dynamo.quantize(). Currently investigating how to separate the calibration data path from the quantization logic

tools/llm/quantize_utils.py
keehyuna: looked at quantization cfg in modelopt

[NVFP4_DEFAULT_CFG](https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/main/modelopt/torch/quantization/config.py#L360) NVFP4 has E4M3 scales and a block size is 16.

[MXFP4_DEFAULT_CFG](https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/main/modelopt/torch/quantization/config.py#L289) MXFP4 has E8M0 scales and a block size is 32.

tools/llm/README.md
```diff
@@ -39,11 +40,38 @@ python run_llm.py --model meta-llama/Llama-3.2-1B-Instruct --prompt "What is par
...
+### Quantization
+
+Torch-TensorRT supports quantization to reduce model memory footprint and improve inference performance:
+
+#### Using Pre-quantized Models
+
+To use pre-quantized models from HuggingFace:
+
+```bash
+python run_llm.py --model nvidia/Llama-3.1-8B-Instruct-FP8 --prompt "What is parallel programming?" --precision FP16 --num_tokens 128
```
lanluo-nvidia: I feel like --precision FP16 will be confusing to user, 
user will think I already provided the fp8 quantized model, why do I need to provide --precision FP16?
(because user won't know that for the pre-quantized model, we actually dequantize back to the precision which user provided via --precision and call modelopt quantize to fp8 again)

lanluo-nvidia:
modelopt has changed their code structure in 0.35.0:
please make the same changes as here: https://github.com/pytorch/TensorRT/commit/9c520f8a78303f02c551437fc2b5d03093934790

tools/llm/quantize_utils.py
```diff
@@ -0,0 +1,272 @@
...
+
+            if hf_quant_algo == "FP8":
+                # FP8 E4M3 format has a maximum representable value of 448.0
+                # Scale the quantization parameters accordingly
+                weight_scale = tensors.pop(weight_scale_name)
+                weight_amax = weight_scale * 448.0
+                input_amax = tensors.pop(input_scale_name) * 448.0
+
+                # Dequantize the weight using the scale factor
+                dequantized_weight_data = module.weight.to(torch.float32) * weight_scale
```
lanluo-nvidia: should we check if precison is fp16 then .to(torch.float16) otherwise float32?

py/torch_tensorrt/dynamo/_quantization.py
peri044, line 1: Sorry for the late review. I think we should have this in tools for now and ideally in modelopt in the future. As more configurations come up, this would increase the cost of maintainence (when modelopt would already be doing this)

tools/llm/quantize_utils.py
```diff
@@ -0,0 +1,272 @@
...
+        super().__init__()
+
+        # Store reference to original linear layer for weight access
+        self.original_linear = original_linear
+
+        # Copy bias from original layer if it exists
+        if original_linear.bias is not None:
+            self.bias = torch.nn.Parameter(original_linear.bias.clone()).cuda()
+        else:
+            self.bias = None
```
peri044: why do we clone only bias ? But not the weight ?

tools/llm/quantize_utils.py
```diff
@@ -0,0 +1,272 @@
...
+                # NVFP4 format requires additional scale tensor and different configuration
+                weight_name = name + ".weight"
+                weight_scale2_name = name + ".weight_scale_2"
+                weight_scale = tensors.pop(weight_scale_name)
+                input_scale = tensors.pop(input_scale_name)
+                weight_scale2 = tensors.pop(weight_scale2_name)
+
+                # Calculate amax values with additional scaling factor for NVFP4
+                input_amax = input_scale * 448.0 * 6.0
+                weight_amax = weight_scale2 * 448.0 * 6.0
```
peri044: Consider storing these constants in a variable at the top of the file and add description about these constants ?

tools/llm/README.md
peri044: +1. @keehyuna  It looks like we dequantize FP8 weights back to FP32. Do you know what happens if we provide --precision=FP16 here ? If FP16+FP8 isn't supported, consider throwing an error

tools/llm/README.md
```diff
@@ -39,11 +40,38 @@ python run_llm.py --model meta-llama/Llama-3.2-1B-Instruct --prompt "What is par
...
+```
+
+#### Applying quantization by ModelOpt
+
+Apply fp8 quantization from HuggingFace:
+
+```bash
+python run_llm.py --model meta-llama/Llama-3.1-8B --quant_format fp8 --prompt "What is parallel programming?" --precision FP16 --num_tokens 128
```
peri044: could you add a description on when --quant_format should be used and info on how we auto-detect if the model is pre-quantized or not ?

tools/llm/run_llm.py
```diff
@@ -54,11 +61,17 @@ def get_model(args):
...
+                ignore_mismatched_sizes=True,
             )
             .eval()
             .cuda()
         )
 
+    hf_quant_config = load_quantization_config(args.model)
+    if hf_quant_config:
+        model = convert_linear_to_tensorrt_quantized(model, hf_quant_config).cuda()
+        print(f"Model converted to TensorRT quantized")
```
peri044: Consider changing this to a more informative message

peri044 (changes requested):
Functionality looks good to me. Posted some comments on code restructuring

py/torch_tensorrt/dynamo/_quantization.py
keehyuna: Thanks I agree with it

tools/llm/README.md
keehyuna: @lanluo-nvidia `--precision FP16 --quant_format fp8` option will convert model weight to float 16 and quantize this weight with fp8 algorithm. I will rename precision with model_precision.
@peri044 I think model weight and quant algo can differ and all are valid combination as quantization precision is smaller bit depth than weight(FP32/FP16/BF16 > fp8/nvfp4). I have verified that the following combinations are all functional
--precision FP32 --quant_format fp8
--precision FP16 --quant_format fp8
--precision BF16 --quant_format fp8
--precision FP32 --quant_format nvfp4
--precision FP16 --quant_format nvfp4
--precision BF16 --quant_format nvfp4

tools/llm/quantize_utils.py
keehyuna: Using clone() for weights doubles GPU memory consumption. While clone() for bias isn't necessary, I'll incorporate it into original_linear for consistency.

tools/llm/quantize_utils.py
keehyuna: Thanks, that makes sense. I've updated it to use the same model precision.

PR #3758, opened 2025-08-07
fix: batch norm issue encountered in RAFT

### LINKED ISSUE #3731: 🐛 [Bug] perf gap reduce on RAFT
##  Bug Description

A user reported Torch-TRT has slower inference time on RAFT-small compared to the original Pytorch model:

```
Backend                        Time (ms)      Speedup   
----------------------------------------------------------------------
Original                         7.12           -         
Torch-TensorRT                 20.35         0.35x   
ONNX-TRT                         2.96          2.41x
```

I printed out the engine profiles of Torch-TensorRT and ONNX-TRT where the num of layers of Torch-TRT are twice more than ONNX-TRT.

zewenli98:
# Description

Pytorch decomposes `torch.ops.aten._native_batch_norm_legit.no_stats` into a bunch of aten ops, which makes it way slow. See the issue below for details.

Fixes #3731 

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
@@ -16,6 +16,7 @@ from torch_tensorrt.dynamo.conversion.converter_utils import (
     get_trt_tensor,
     has_dynamic_shape,
     set_layer_name,
+    to_torch,
     to_trt_weights,
 )
 from torch_tensorrt.dynamo.conversion.impl.cat import cat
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -60,22 +62,30 @@ def batch_norm(
     ):
         # We name the weight here according to the state_dict name
         weight = (
-            get_trt_tensor(ctx, 1.0, f"{name}_weight", dtype=input.dtype)
+            get_trt_tensor(
+                ctx, np.ones((feature_num,)), f"{name}_weight", dtype=input.dtype
```
narendasan: @peri044 should we use torch here?

zewenli98:
> looks fine, is there a reason we use numpy instead of torch?

because it would be FakeTensor

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
zewenli98: Within `get_trt_tensor`, both np and torch practically work. Any reason for using torch?

zewenli98: if using torch, will it be fakeTensor?
```
RuntimeError: Cannot access data pointer of Tensor (e.g. FakeTensor, FunctionalTensor). If you're using torch.compile/export/fx, it is likely that we are erroneously tracing into a custom kernel. To fix this, please wrap the custom kernel into an opaque custom op. Please see the following for details: https://pytorch.org/tutorials/advanced/custom_ops_landing_page.html
```

narendasan (approved):
Lgtm

PR #3829, opened 2025-09-22
TRT-LLM installation tool

apbose:
1. TRT-LLM download utility. USE_TRTLLM_PLUGINS=1 should trigger the download mechanism. 
2. wheel downloaded to /tmp/torch_tensorrt_<username>/ and unzipped
3. plugin_lib_path is set to here - /tmp/torch_tensorrt_<username>/trtllm/0.17.0.post1_linux_x86_64/tensorrt_llm/libs/libnvinfer_plugin_tensorrt_llm.so.
Across runs wheel is removed, while  .so file is retained

### DIFF
--- a/.github/workflows/build-test-linux-x86_64.yml
+++ b/.github/workflows/build-test-linux-x86_64.yml
@@ -315,39 +315,6 @@ jobs:
         python -m pytest -ra --junitxml=${RUNNER_TEST_RESULTS_DIR}/l1_ts_models_tests_results.xml -n auto  models/
         popd
 
-  L1-dynamo-distributed-tests:
-    name: L1 dynamo distributed tests
-    needs: [filter-matrix, build,L0-dynamo-core-tests, L0-dynamo-converter-tests, L0-py-core-tests, L0-torchscript-tests]
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - repository: pytorch/tensorrt
-            package-name: torch_tensorrt
-            pre-script: packaging/pre_build_script.sh
-            post-script: packaging/post_build_script.sh
-            smoke-test-script: packaging/smoke_test_script.sh
-    uses: ./.github/workflows/linux-test.yml
-    with:
-      job-name: L1-dynamo-distributed-tests
-      repository: "pytorch/tensorrt"
-      ref: ""
-      test-infra-repository: pytorch/test-infra
-      test-infra-ref: main
-      build-matrix: ${{ needs.filter-matrix.outputs.matrix }}
-      pre-script: ${{ matrix.pre-script }}
-      script: |
-        set -euo pipefail
-        export USE_HOST_DEPS=1
-        export CI_BUILD=1
-        export USE_TRTLLM_PLUGINS=1
-        dnf install -y mpich mpich-devel openmpi openmpi-devel
-        pushd .
-        cd tests/py
-        cd dynamo
-        python -m pytest -ra --junitxml=${RUNNER_TEST_RESULTS_DIR}/dynamo_distributed_test_results.xml distributed/test_nccl_ops.py
-        popd
-
   L2-torch-compile-tests:
     name: L2 torch compile tests
     needs: [filter-matrix, build, L1-torch-compile-tests, L1-dynamo-compile-tests, L1-dynamo-core-tests, L1-torchscript-tests]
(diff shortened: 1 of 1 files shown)


py/torch_tensorrt/dynamo/conversion/custom_ops_converters.py
```diff
@@ -11,15 +11,15 @@
...
+from torch_tensorrt.dynamo.utils import load_tensorrt_llm_for_nccl
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
-if load_tensorrt_llm():
-    from torch_tensorrt.dynamo.lowering.passes.fuse_distributed_ops import (
-        tensorrt_fused_nccl_all_gather_op,
-        tensorrt_fused_nccl_reduce_scatter_op,
-    )
+if load_tensorrt_llm_for_nccl():
```
narendasan: I would like to use the enabled features system for this rather than a stand alone function

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -858,3 +874,243 @@ def is_thor() -> bool:
...
+def is_platform_supported_for_trtllm() -> bool:
+    """
+    Checks if the current platform supports TensorRT-LLM plugins for the NCCL backend.
+
+    Returns:
+        bool: True if supported, False otherwise.
+
+    Unsupported:
+        - Windows platforms
+        - Jetson/Orin/Xavier (aarch64 architecture + 'tegra' in platform release)
```
narendasan: Thor also not supported by TRT-LLM right?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -858,3 +874,243 @@ def is_thor() -> bool:
...
+
+    if "windows" in system:
+        logger.info(
+            "TensorRT-LLM plugins for NCCL backend are not supported on Windows."
+        )
+        return False
+
+    if machine == "aarch64" and "tegra" in release:
+        logger.info(
+            "TensorRT-LLM plugins for NCCL backend are not supported on Jetson/Orin/Xavier (Tegra) devices."
```
narendasan: Edit the error message here to include thor

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -858,3 +874,243 @@ def is_thor() -> bool:
...
+        )
+        return False
+
+    try:
+        cuda_version = torch.version.cuda  # e.g., "12.4" or "13.0"
+        if cuda_version is None:
+            logger.warning("No CUDA runtime detected — TRT-LLM plugins unavailable.")
```
narendasan: This is somewhat misleading because the actual error is that the pytorch install does not support cuda.

narendasan: Also if that is the case would this be an error? What invokes this function? Should the user continue to be able to run? Would they be under the assumption that TRT-LLM plugins would be available?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -858,3 +874,243 @@ def is_thor() -> bool:
...
+            return False
+
+        major, minor = map(int, cuda_version.split("."))
+        if major != 12:
+            logger.warning("CUDA 13 is not supported for TRT-LLM plugins.")
```
narendasan: not *currently* supported. Same comment as above though, Seems to me this is at least log error, but the question is if we should kill the process. If the program will not run as intended we should otherwise its still an error but we can continue

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -858,3 +874,243 @@ def is_thor() -> bool:
...
+    except Exception as e_initialization_error:
+        logger.warning(
+            "Exception occurred during TensorRT-LLM plugin library initialization",
+            exc_info=e_initialization_error,
+        )
+        return False
+    return False
+
+
+def load_tensorrt_llm_for_nccl() -> bool:
```
narendasan: This function should be in the enabled features system. And should register the feature for other parts of the library to query against

narendasan:
Yes you should be able to run it on GB200, I think there is just not a thor distribution of TRT-LLM for now.

py/torch_tensorrt/dynamo/utils.py
apbose: yeah Thor and sbsa should support NCCL, but TRT-LLM I am not aware. Will include Thor in the list of unsupported platform.
A followup question, what about sbsa? I see on TRT-LLM page that they are supported on Blackwell, but that does not imply sbsa support right (can be supported on B200 - non sbsa vs GB200 - sbsa).

py/torch_tensorrt/dynamo/utils.py
apbose: yes will change the error message. 
In that case cuda runtime is not available, but I assume we would hit an error before only before reaching this point. Wrt to this function we won't be able to verify if the  CUDA is 12.X or 13.X. Should I remove this check altogether?

py/torch_tensorrt/dynamo/utils.py
apbose: Will change the error message to add `currently`
Its more like then this function will return a false
`load_tensorrt_llm_for_nccl()` calls `is_platform_supported_for_trtllm()` which will return false and the converter will be unsupported.

py/torch_tensorrt/dynamo/utils.py
apbose: Will make this change

py/torch_tensorrt/dynamo/utils.py
narendasan: Its fine to have redundant checks as long as they are clear

.github/workflows/build-test-linux-x86_64.yml
```diff
@@ -315,6 +315,39 @@ jobs:
         python -m pytest -ra --junitxml=${RUNNER_TEST_RESULTS_DIR}/l1_ts_models_tests_results.xml -n auto  models/
         popd
 
+  L1-dynamo-distributed-tests:
```
narendasan: Can we make this L2 for now?

PR #3830, opened 2025-09-22
Changes to TRT-LLM download tool for multigpu distributed case 

apbose:
TRT-LLM installation tool for distributed
1. The download is to be done by only one GPU to avoid unnecessary downloads
2. Use of lock files in the tool for the above purpose


py/torch_tensorrt/dynamo/distributed/utils.py
```diff
@@ -0,0 +1,309 @@
...
+from torch_tensorrt._version import __tensorrt_llm_version__
+
+_WHL_CPYTHON_VERSION = "cp310"
+
+logger = logging.getLogger(__name__)
+
+
+def check_tensor_parallel_device_number(world_size: int) -> None:
+    if world_size % 2 != 0:
+        raise ValueError(
```
narendasan: Why would it matter what the examples need? This is supposed to be user facing code.

py/torch_tensorrt/dynamo/distributed/utils.py
```diff
@@ -0,0 +1,309 @@
...
+    device_id = (
+        rank % torch.cuda.device_count()
+    )  # Ensure each rank gets a unique device
+    torch.cuda.set_device(device_id)
+
+    return device_mesh, world_size, rank
+
+
+def initialize_logger(rank: int, logger_file_name: str) -> logging.Logger:
+    logger = logging.getLogger()
```
narendasan: Where is this logger used? Typically this is a user's responsibility. we should not be creating the actual logger handlers in our library. We can do this in the examples code tho

apbose: ok was meant for the use case that if we want separate log outputs for the two different GPUs. But I agree these all can be moved to the user facing code

tests/py/dynamo/distributed/distributed_utils.py
```diff
@@ -8,24 +9,25 @@
 from torch.distributed._tensor.device_mesh import init_device_mesh
 
 
-def set_environment_variables_pytest():
+# the below two functions are used to set the environment variables for the pytest single and multi process
+# this is for the github CI where we use pytest
+def set_environment_variables_pytest_single_process():
+    port = 29500 + random.randint(1, 1000)
```
narendasan, line 16: Why does it need to be random?

apbose: I have encountered cases locally where in the port was already in use and then it led to address bus error. I have changed the logic a bit to handle the single process and multi process differently. Since multi would need same port for the two processes. In multi we shall set it via the env variable. Added a warning for the same.

PR #3831, opened 2025-09-22
Llama distributed example

apbose:
The PR addresses
1. Llama3 end to end example with complex graph lowering
2. Removal of hardcoded components of rotary embedding example


py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -713,16 +709,13 @@ def compile(
...
-        gm, trt_arg_inputs, trt_kwarg_inputs, settings, engine_cache
-    )
+    trt_gm = compile_module(gm, trt_kwarg_inputs, settings, engine_cache)
     return trt_gm
 
 
 @fn_supports_debugger  # type: ignore[misc]
 def compile_module(
     gm: torch.fx.GraphModule,
-    sample_arg_inputs: Sequence[Input],
```
narendasan, line 725: Why is arg_inputs being deleted?

apbose: looked like the arg_inputs was not being used in the compile_module, hence I just removed it

Issue #3783, opened 2025-08-19
❓ [Question] Error when exporting TorchScript model with dynamic input using Torch-TensorRT (aten::clamp issue)


Mmmyyym:
## Environment
Libtorch 2.5.0.dev (latest nightly) (built with CUDA 12.4)
CUDA 12.4
TensorRT 10.1.0.27
PyTorch 2.4.0+cu124
Torch-TensorRT 2.4.0
Python 3.12.8
Windows 10

## Code
```
import torch
import torch_tensorrt

model = DeepLabv3+(num_classes=8,  backbone='mobilenet', output_stride=16)
model.eval().cuda()
dynamic_input = torch_tensorrt.Input(
        min_shape=[1, 3, 20, 20],
        opt_shape=[1, 3, 512, 512],
        max_shape=[1, 3, 2448, 2048],
        dtype=torch.float32 )
static_input = [torch.randn((1, 3, 512, 512)).cuda()]
trt_gm = torch_tensorrt.compile(model, ir="dynamo", inputs = dynamic_input)
torch_tensorrt.save(trt_gm, "trt_dynamic.ts", output_format="torchscript",inputs=static_input)
```
## ERROR
**torch_tensorrt.compile succeeds, but torch_tensorrt.save fails.**

`torch_tensorrt.save(trt_gm, "trt_dynamic.ts", output_format="torchscript",inputs=static_input)`
```
    def __call__(self_, *args, **kwargs):  # noqa: B902
        # use `self_` to avoid naming collide with aten ops arguments that
        # are named "self". This way, all the aten ops can be called by kwargs.
        return self_._op(*args, **kwargs)
```

<img width="988" height="180" alt="Image" src="https://github.com/user-attachments/assets/ccbbae26-e328-4543-acdf-7ec9c8a94730" />

```
RuntimeError: aten::clamp() Expected a value of type 'Optional[number]' for argument 'max' but instead found type 'Tensor'.
Position: 2
Value: tensor([31], device='cuda:0')
Declaration: aten::clamp(Tensor self, Scalar? min=None, Scalar? max=None) -> Tensor
Cast error details: Cannot cast tensor([31], device='cuda:0') to number
```

narendasan:
@apbose Please take a look at this, when I run with this implementation of DeepLabV3+ (https://github.com/VainF/DeepLabV3Plus-Pytorch) with dynamic shape I get the following converter error 

```sh
[...]

PR #3841, opened 2025-09-27
addresses the case when shape of upsample tensor contains ITensor

apbose:
Addresses https://github.com/pytorch/TensorRT/issues/3783

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/shape.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/shape.py
@@ -123,3 +123,41 @@ def get_shape_with_dynamic_shape(
     select_layer = ctx.net.add_select(condition_val, input_shape, scale_res)
     set_layer_name(select_layer, target, f"{name}_select")
     return select_layer.get_output(0)
+
+
+def to_trt_shape_tensor(
+    ctx: ConversionContext, target: Target, name: str, shape_list: List[int | TRTTensor]
+) -> TRTTensor:
+    """
+    Convert a mixed shape list (ints + ITensors) into a single ITensor.
+
+    Args:
+        ctx: ConversionContext
+        target: fx node target (used for naming).
+        name (str): base name for layer naming.
+        shape_list (list[int | ITensor]): list containing static ints and/or ITensors.
+
+    Returns:
+        ITensor if shape_list contains any ITensors, else plain Python list of ints.
+    """
+    trt_tensors = []
+
+    for i, s in enumerate(shape_list):
+        if isinstance(s, int):
+            const = ctx.net.add_constant((1,), np.array([s], dtype=np.int32))
+            set_layer_name(const, target, f"{name}_dim{i}_const")
+            trt_tensors.append(const.get_output(0))
+        else:
+            # Assume it's already an ITensor
+            trt_tensors.append(s)
+
+    if trt_tensors:
+        if any(not isinstance(s, int) for s in shape_list):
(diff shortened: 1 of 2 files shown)


apbose:
yeah thats true. cat does it for inputs while the above is for shape tensor. Yeah I guess we could unify this. Should 
I merge this for now since it addresses user error? Or do you think I should merge after unifying?

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -16,6 +16,63 @@
...
+        return trt_tensors  # all ints
+
+    # promote remaining ints to TRT consts before concat
+    for i, t in enumerate(trt_tensors):
+        if isinstance(t, int):
+            const = ctx.net.add_constant((1,), np.array([t], dtype=np.int32))
+            set_layer_name(const, target, f"{name}_static_{i}_const")
+            trt_tensors[i] = const.get_output(0)
+
+    concat = ctx.net.add_concatenation(trt_tensors)
```
peri044, line 104: If trt_tensors have a mix of scalar integers and ITensors of dtype int64, would this work (because you're casting the scalar integers to int32 explicitly) ?

apbose: In the case of shape tensors int will always be int32, so in that case this should work.
Coming to cat case. concat tensors will be either torch.Tensor or TRTTensor. They cannot be int. So I think the above should cover all the cases. Can you think of any other case?

apbose:
Embedding bag looks like is failing. Need to look into

py/torch_tensorrt/dynamo/conversion/impl/cat.py
peri044: So my thought is how are we ensuring all trt_tensors have same datatypes explicitly before concatenating here because that will error out ? 
This check could either be an assertion check or explicit type promotion of tensors within trt_tensor

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -11,11 +11,107 @@
...
+
+    final_dtype = None
+    if cast_dtype:
+        # Explicit cast requested
+        if isinstance(cast_dtype, _enums.dtype):
+            final_dtype = cast_dtype.to(trt.DataType)
+        elif isinstance(cast_dtype, np.dtype):
+            final_dtype = _enums.dtype._from(cast_dtype).to(trt.DataType)
+        else:
+            final_dtype = cast_dtype  # already trt.DataType
```
peri044, line 71: should we also check torch.dtype case ?

py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -11,11 +11,107 @@
...
+                const_arr = np.array(x, dtype=np.int32)
+                shape = (x.numel(),)
+
+            layer = ctx.net.add_constant(shape, const_arr)
+            set_layer_name(layer, target, f"{name}_dim{i}_const")
+            t = layer.get_output(0)
+
+        # optional cast
+        if cast_dtype and isinstance(t, TRTTensor):
+            t = cast_trt_tensor(ctx, t, cast_dtype, f"{name}_cast_{i}")
```
peri044: is this necessary if we are also casting at line 69 onwards ?

peri044 (commented):
@apbose  I have 2 minor questions. Rest LGTM

PR #3845, opened 2025-10-02
Cpu memory optimization

cehongwang:
# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -42,6 +42,7 @@ from torch_tensorrt.dynamo.lowering import (
 )
 from torch_tensorrt.dynamo.utils import (
     deallocate_module,
+    get_cpu_memory_usage,
     get_flat_args_with_check,
     get_output_metadata,
     parse_graph_io,
@@ -675,7 +676,7 @@ def compile(
         "l2_limit_for_tiling": l2_limit_for_tiling,
         "offload_module_to_cpu": offload_module_to_cpu,
     }
-
+    logger.debug(f"CPU memory usage before lowering: {get_cpu_memory_usage()} MB")
     settings = CompilationSettings(**compilation_options)
     logger.info("Compilation Settings: %s\n", settings)
     exported_program = pre_export_lowering(exported_program, settings)
@@ -689,6 +690,7 @@ def compile(
 
     # Apply lowering on the graph module
     gm = post_lowering(gm, settings)
+    logger.debug(f"CPU memory usage after post_lowering: {get_cpu_memory_usage()} MB")
     logger.debug("Lowered Input graph: " + str(gm.graph))
 
     # Move the weights in the state_dict to CPU
@@ -698,6 +700,7 @@ def compile(
         logger.info(
             "The PyTorch model was moved to the CPU to allocate all GPU memory to TensorRT. To retain the model on the GPU, set offload_module_to_cpu=False"
         )
+        logger.debug(f"CPU memory usage after CPU offload: {get_cpu_memory_usage()} MB")
     else:
         remaining_memory, total_memory = torch.cuda.mem_get_info()
         if remaining_memory < total_memory // 2:
@@ -859,6 +862,9 @@ def compile_module(
     # Iterate over all components that can be accelerated
     # Generate the corresponding TRT Module for those
 
+    # Here we delete the frozen parameters from the graph module. Note this does not affect the submodules. We are going to delete the frozen parameters from the submodules in the convert_module function.
+    # This is done to release CPU memory.
+    [delattr(gm, attr) for attr in dir(gm) if attr.startswith("_frozen_param")]
     for name, _ in partitioned_module.named_children():
         submodule = getattr(partitioned_module, name)
         # filter on the GraphModule
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -693,7 +693,8 @@ def compile(
 
     # Move the weights in the state_dict to CPU
     if offload_module_to_cpu:
-        deallocate_module(exported_program.module(), delete_module=False)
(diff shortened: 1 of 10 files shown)


py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -65,6 +70,29 @@ def interpret_module_to_result(
     )
 
     interpreter_result = interpreter.run()
+    # Delete the frozen parameters from the module to release CPU memory
```
narendasan, line 78: Can we gate this by the same env variable as the malloc_trim?

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -104,7 +132,7 @@ def convert_module(
         )
 
     return rt_cls(
-        serialized_engine=interpreter_result.serialized_engine,
+        serialized_engine=interpreter_result.engine,
```
narendasan: Will this be the serialized or live engine?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -733,6 +735,9 @@ def run(
                         return interpreter_result  # type: ignore[no-any-return]
 
         self._construct_trt_network_def()
+        _LOGGER.info(
```
narendasan: Debug level?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -689,14 +690,17 @@ def compile(
 
     # Apply lowering on the graph module
     gm = post_lowering(gm, settings)
+    logger.debug(f"CPU memory usage after post_lowering: {get_cpu_memory_usage()} MB")
     logger.debug("Lowered Input graph: " + str(gm.graph))
 
     # Move the weights in the state_dict to CPU
     if offload_module_to_cpu:
+        deallocate_module(gm, delete_module=False)
```
narendasan, line 698: Arent these the same?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -858,6 +862,11 @@ def preserve_module_specs(
     # Iterate over all components that can be accelerated
     # Generate the corresponding TRT Module for those
 
+    # Here we delete the frozen parameters from the graph module. Note this does not affect the submodules. We are going to delete the frozen parameters from the submodules in the convert_module function.
+    # This is done to release CPU memory.
+    for attr in dir(gm):
```
narendasan, line 867: Lets make this opt in similar to malloc trim

narendasan (commented):
We need the env variable control

py/torch_tensorrt/dynamo/_compiler.py
cehongwang: Shouldn't this be cleared no matter what?

py/torch_tensorrt/dynamo/conversion/_conversion.py
cehongwang: Do you mean something like
`export low_RAM_mode=1 python flux.py` or build env variable?

narendasan: I would say something like `export TORCHTRT_USE_MALLOC_TRIM=1 python flux.py` and the goal would be for this to go away in 2.11 and be merged into `prioritize_host_memory_consumption=True`

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
```diff
@@ -225,13 +226,80 @@ def partition_graph(self) -> torch.fx.GraphModule:
...
+        subgraphs = self.break_subgraphs(subgraphs, num_of_break=num_of_break)
+
         # Set the number of TRT engines to be generated
         self.num_trt_accelerated_subgraphs = len([s for s in subgraphs if s.is_acc])
 
         # Tag the accelerated nodes and split the graph accordingly
         self.tag(subgraphs)
         return self.split()
 
+    def calculate_num_of_break(self, subgraphs: List[Subgraph]) -> int:
```
narendasan, line 239: I feel like this is too much of an heuristic based system. A better approach IMO is to calculate a graph size budget based on available memory (or eventually this could be user specified). Then of the TRT blocks we estimate its size and then decide how many subgraphs it should be split into to meet the budget

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
```diff
@@ -225,13 +226,80 @@ def partition_graph(self) -> torch.fx.GraphModule:
...
+            [node for node in self.acc_nodes if "scaled_dot" in str(node.target)]
+        )
+        break_period = num_of_sdpa_node // num_of_break + 1
+        current_break_idx = 0
+        current_num_break = 0
+        new_subgraphs = []
+        for subgraph in subgraphs:
+            if subgraph.is_acc:
+                for i, node in enumerate(subgraph.nodes):
+                    if "scaled_dot" in str(node.target):
```
narendasan, line 280: Its fine if we do this for testing but really we should be taking a much more generic approach rather than assuming only sdpa is a viable break point.

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -750,16 +755,19 @@ def run(
         self._create_timing_cache(
             builder_config, self.compilation_settings.timing_cache_path
         )
-        serialized_engine = self.builder.build_serialized_network(
+
+        cuda_engine = self.builder.build_engine_with_config(
```
lanluo-nvidia, line 759: Curious what is the benefit in CPU memory, if we return cuda_engine instead of serialized_engine?
because interpret_module_to_result() will anyway deserialize the cuda_engine to serialized_engine str.

PR #3849, opened 2025-10-07
add capture and replay feature in tensorrt

lanluo-nvidia:
# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/debug/_Debugger.py
+++ b/py/torch_tensorrt/dynamo/debug/_Debugger.py
@@ -171,7 +171,6 @@ class Debugger:
         ]
 
         if self.cfg.capture_shim:
-            self.original_environ_dict = {}
             shim_lib_name = "libtensorrt_shim.so"
             nvinfer_lib_name = "libnvinfer.so"
 
@@ -189,43 +188,31 @@ class Debugger:
                         f"TRT_SHIM_NVINFER_LIB_NAME is not set, please add the {nvinfer_lib_name} with full path to the TRT_SHIM_NVINFER_LIB_NAME environment variable"
                     )
                     is_valid = False
+                if os.environ.get("TRT_SHIM_OUTPUT_JSON_FILE") is None:
+                    _LOGGER.error(
+                        "TRT_SHIM_OUTPUT_JSON_FILE is not set, please add the shim output json file name with full path to the TRT_SHIM_OUTPUT_JSON_FILE environment variable"
+                    )
+                    is_valid = False
+                else:
+                    shim_output_json_file = os.environ["TRT_SHIM_OUTPUT_JSON_FILE"]
+                    shim_output_dir = os.path.dirname(shim_output_json_file)
+                    if len(shim_output_dir) > 0 and not os.path.exists(shim_output_dir):
+                        _LOGGER.debug(
+                            f"shim output directory {shim_output_dir} does not exist, creating it now"
+                        )
+                        os.makedirs(shim_output_dir)
                 return is_valid
 
             if not validate_setting():
                 return
-
-            self.original_environ_dict["TRT_SHIM_OUTPUT_JSON_FILE"] = os.environ.get(
-                "TRT_SHIM_OUTPUT_JSON_FILE"
+            json_file_name = os.environ["TRT_SHIM_OUTPUT_JSON_FILE"]
+            _LOGGER.info(
+                f"capture_shim feature is enabled, shim output file is set to {json_file_name}"
             )
-            if os.environ.get("TRT_SHIM_OUTPUT_JSON_FILE") is None:
-                # TRT_SHIM_OUTPUT_JSON_FILE is not set, set it to the default shim json file path
-                shim_output_dir = os.path.join(self.cfg.logging_dir, "shim")
-                shim_output_json_file = os.path.join(shim_output_dir, "shim.json")
-            else:
-                shim_output_json_file = os.environ["TRT_SHIM_OUTPUT_JSON_FILE"]
-                # validate the shim_output_dir
-                # split the path from the full path of shim_output_dir
-                shim_output_dir = os.path.dirname(shim_output_json_file)
-                if len(shim_output_dir) == 0:
-                    shim_output_dir = os.path.join(self.cfg.logging_dir, "shim")
-                    shim_output_json_file = os.path.join(shim_output_dir, "shim.json")
-
-            if not os.path.exists(shim_output_dir):
-                os.makedirs(shim_output_dir, exist_ok=True)
-            # if file alaredy exists, delete it first, so that we can create a fresh new one
-            if os.path.exists(shim_output_json_file):
-                os.remove(shim_output_json_file)
-            os.environ["TRT_SHIM_OUTPUT_JSON_FILE"] = shim_output_json_file
 
     def __exit__(self, exc_type: Any, exc_value: Any, exc_tb: Any) -> None:
 
         dictConfig(self.get_logging_config(None))
-        if self.cfg.capture_shim:
-            for k, v in self.original_environ_dict.items():
-                if v is None:
-                    os.environ.pop(k, None)
-                else:
-                    os.environ[k] = v
         if ENABLED_FEATURES.torch_tensorrt_runtime:
             torch.ops.tensorrt.set_logging_level(self.rt_level)
         if self.capture_fx_graph_before or self.capture_fx_graph_after:
--- a/py/torch_tensorrt/dynamo/debug/_Debugger.py
+++ b/py/torch_tensorrt/dynamo/debug/_Debugger.py
@@ -2,6 +2,7 @@ import contextlib
 import functools
 import logging
 import os
+import sys
 import tempfile
 from logging.config import dictConfig
 from typing import Any, List, Optional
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -32,6 +34,7 @@ def __init__(
         capture_fx_graph_before: Optional[List[str]] = None,
         capture_fx_graph_after: Optional[List[str]] = None,
         save_engine_profile: bool = False,
+        capture_shim: bool = False,
```
narendasan: Rename this something like `capture_tensorrt_api_recording`.

py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -65,6 +70,7 @@ def __init__(
         self.cfg = DebuggerConfig(
             log_level=log_level,
             save_engine_profile=save_engine_profile,
+            capture_shim=capture_shim,
```
narendasan: Same here

py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -92,6 +98,26 @@ def __init__(
         self.capture_fx_graph_before = capture_fx_graph_before
         self.capture_fx_graph_after = capture_fx_graph_after
 
+        if self.cfg.capture_shim:
+            if not sys.platform.startswith("linux"):
+                _LOGGER.warning(
+                    "capture_shim featureis only supported on linux, will disable it"
```
narendasan: `Capturing TensorRT API calls is only supported on Linux, therefore ignoring the capture_tensorrt_api_recording setting`

py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -92,6 +98,26 @@ def __init__(
...
+                )
+                self.cfg.capture_shim = False
+                return
+            if ENABLED_FEATURES.tensorrt_rtx:
+                _LOGGER.warning(
+                    "capture_shim feature is not supported on TensorRT-RTX, will disable it"
```
narendasan: Capturing TensorRT API calls is only supported with standard TensorRT, therefore ignoring the capture_tensorrt_api_recording setting

py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -92,6 +98,26 @@ def __init__(
...
+                )
+                self.cfg.capture_shim = False
+                return
+            if not is_tensorrt_version_supported("10.13.0"):
+                _LOGGER.warning(
+                    "capture_shim feature is only supported on TensorRT 10.13 and above, will disable it"
```
narendasan: Capturing TensorRT API calls is only supported for TensorRT versions 10.13 and above, therefore ignoring the capture_tensorrt_api_recording setting

py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -143,6 +169,46 @@ def __enter__(self) -> None:
...
+                    _LOGGER.error(
+                        f"LD_PRELOAD is not set, please add the {shim_lib_name} with full path to the LD_PRELOAD environment variable"
+                    )
+                    is_valid = False
+                if os.environ.get("TRT_SHIM_NVINFER_LIB_NAME") is None:
+                    _LOGGER.error(
+                        f"TRT_SHIM_NVINFER_LIB_NAME is not set, please add the {nvinfer_lib_name} with full path to the TRT_SHIM_NVINFER_LIB_NAME environment variable"
+                    )
+                    is_valid = False
+                if os.environ.get("TRT_SHIM_OUTPUT_JSON_FILE") is None:
```
narendasan: This should just go into the debugging dir

py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -143,6 +169,46 @@ def __enter__(self) -> None:
...
+                    _LOGGER.error(
+                        "TRT_SHIM_OUTPUT_JSON_FILE is not set, please add the shim output json file name with full path to the TRT_SHIM_OUTPUT_JSON_FILE environment variable"
+                    )
+                    is_valid = False
+                else:
+                    shim_output_json_file = os.environ["TRT_SHIM_OUTPUT_JSON_FILE"]
+                    shim_output_dir = os.path.dirname(shim_output_json_file)
```
narendasan: Same thing here, we have a debugging directory that gets created for this stuff

tools/debug/capture_replay/README.md
```diff
@@ -0,0 +1,66 @@
+## Introduction
```
narendasan: Can we add this to the web documentation?

py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -91,6 +96,10 @@ def __init__(
 
         self.capture_fx_graph_before = capture_fx_graph_before
         self.capture_fx_graph_after = capture_fx_graph_after
+        if os.environ.get("TORCHTRT_ENABLE_TENSORRT_API_CAPTURE") == "1":
```
narendasan: This will get weird, not sure what the right way to go is. 

I think the correct behavior would be 

#### Case 1

```py
with torch_tensorrt.Debugger(capture_tensorrt_api_recording=True):
```

but the script was invoked with 

```sh
python my_script.py
```

Should result in a warning like `WARNING: In order to capture TensorRT API calls, please invoke the script with environment variable TORCHTRT_ENABLE_TENSORRT_API_CAPTURE=1` 

And if the user does this: 

#### Case 2

```py
with torch_tensorrt.Debugger(capture_tensorrt_api_recording=False):
```

but the script was invoked with 

```sh
TORCHTRT_ENABLE_TENSORRT_API_CAPTURE=1 python my_script.py
```

Then we do nothing. Don't override their setting just a log INFO message in `torch_tensorrt.__init__` that we loaded the shim library. 

And if the user does this 

#### Case 3 

```py
with torch_tensorrt.Debugger(capture_tensorrt_api_recording=1):
```

but the script was envoked with 

```sh
TORCHTRT_ENABLE_TENSORRT_API_CAPTURE=1 python my_script.py
```

Then it should run normally

py/torch_tensorrt/__init__.py
```diff
@@ -24,6 +26,69 @@
 
 import torch
 
+
```
narendasan: Do you think this should be with the other TRT proxy code?

narendasan: That way we can directly control the import order

py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -91,6 +96,10 @@ def __init__(
...
+            self.cfg.capture_tensorrt_api_recording = True
+        else:
+            self.cfg.capture_tensorrt_api_recording = False
```
narendasan, line 117: We still need the behavior to move the default temp location for the location to the debugging folder

py/torch_tensorrt/dynamo/debug/_Debugger.py
```diff
@@ -92,6 +98,18 @@ def __init__(
         self.capture_fx_graph_before = capture_fx_graph_before
         self.capture_fx_graph_after = capture_fx_graph_after
 
+        if self.cfg.capture_tensorrt_api_recording:
+            env_flag = os.environ.get("TORCHTRT_ENABLE_TENSORRT_API_CAPTURE", None)
+            if env_flag is None or (env_flag != "1" and env_flag.lower() != "true"):
+                # currently this feature is only supported for TensorRT on Linux platform
+                if (
```
narendasan: I think we can make this more meaningful to the user, like errors for the unsupported conditions and the warning for the rest. 

So like:

```py
if self.cfg.capture_tensorrt_api_recording:
        _LOGGER.error("TensorRT API capture is not supported with TensorRT-RTX"
    if not sys.platform.startswith("linux"):
        _LOGGER.error("TensorRT API capture is only supported on Linux")
    if env_flag is None or (env_flag != "1" and env_flag.lower() != "true"):
        _LOGGER.warning(
                        "In order to capture TensorRT API calls, please invoke the script with environment variable TORCHTRT_ENABLE_TENSORRT_API_CAPTURE=1"
                    )
    else:
        print("...find the logs here...")
```

py/torch_tensorrt/_TensorRTProxyModule.py
```diff
@@ -28,6 +32,66 @@ def _find_lib(name: str, paths: List[str]) -> str:
...
+                ctypes.CDLL(
+                    os.path.join(path, "libtensorrt_shim.so"), mode=ctypes.RTLD_GLOBAL
+                )
+                tensorrt_lib_path = path
+                break
+            except Exception as e:
+                continue
+
+    if tensorrt_lib_path is None:
+        _LOGGER.error(
```
narendasan: Maybe this should be a warning?

tools/debug/capture_replay/README.md
```diff
@@ -0,0 +1,37 @@
...
+Run the engine with `trtexec`:
+
+```bash
+trtexec --loadEngine=/absolute/path/to/output_engine
+```
+
+### Notes
+- Ensure the `libnvinfer.so` used by the shim matches the TensorRT version in your environment.
+- If multiple TensorRT versions are installed, prefer absolute paths as shown above.
+- The capture is best-effort; if your program builds multiple engines, multiple captures may be produced.
```
narendasan: This isnt the behavior right? Also make this file into an rst and move it into docsrc

Issue #3865, opened 2025-10-16
🐛 [Bug]Torch distributed data parallel accelerate GPT2 example failing


apbose:
Torch distributed data parallel accelerate GPT2 example failing
```
cd examples/distributed_inference
CUDA_VISIBLE_DEVICES=0 accelerate launch data_parallel_gpt2.py <single GPU>
accelerate launch data_parallel_gpt2.py
```

torch                    2.9.0.dev20250821+cu129
torch_tensorrt           2.10.0.dev0+0
accelerate               1.10.1

PR #3866, opened 2025-10-16
addressing cat empty tensor case.Fixes gpt2 data distributed example

apbose:
Fixes https://github.com/pytorch/TensorRT/issues/3865

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2,7 +2,7 @@
 
 import logging
 import operator
-from typing import Callable, Dict, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/conversion/impl/cat.py
```diff
@@ -27,6 +30,13 @@ def cat(
 ) -> Union[TRTTensor, Sequence[TRTTensor]]:
     trt_inputs = []
     for i, each_input in enumerate(input):
+        if isinstance(each_input, torch.Tensor) and each_input.numel() == 0:
+            logger.warning(
+                f"Warning: empty tensor in cat input {i}, replacing with zeros"
```
narendasan, line 35: Can you make this warning much more specific? Print information like the current node, if you can where in the graph it comes from etc. Because users will not understand what you mean by this. Also where is the replacing with zeros?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -217,21 +217,67 @@ def aten_ops_native_group_norm(
...
+            input_tensors = list(args[0])
+        else:
+            input_tensors = list(args)
+
+        dim = kwargs.get("dim", 0)
+
+    return input_tensors, dim
+
+
+def cat_validator(node: Node, settings: Optional[CompilationSettings] = None) -> bool:
```
narendasan, line 252: I dont really understand this condition. So if we have a TRT ITensor that has a 0 in any dimension then we should break the graph? I dont think at validation time any of these ITensors will be available. Since validation is run prior to paritioning

narendasan: Should we be checking for empty PyTorch tensors?

py/torch_tensorrt/dynamo/conversion/impl/cat.py
narendasan: Also if this is caught by the validator then should this be an error? Will conversion fail or can we just ignore it?

apbose: Thanks for pointing out the error. I was earlier replacing with zeros, but later changed to `continue` since replacing with zeros is not required. I will change the warning comment.

The difference between this and the validator is that, if the empty tensor is a torch.Tensor, we can handle it in the converter.

Whereas if the empty tensor comes as an ITensor input to the converter, TensorRT complains. (I was trying to implement it earlier via replacing it with zeros, but that still leads to the error `[RemoveDeadLayers] Input Tensor y is unused or used only at compile-time, but is not being removed.` To point the difference, 

This will pass

 ```
def test_cat_with_empty_tensor(self, _, dim):
        # Handle empty tensor in concat
        class Cat(nn.Module):
            def forward(self, x):
                y = torch.empty(0, 2, 3, device="cuda")
                return torch.ops.aten.cat.default((x, y), dim)

        inputs = [
            torch.randn(1, 2, 3, device="cuda"),
        ]
        self.run_test(Cat(), inputs)
```
This will fail
```
 def test_cat_with_empty_tensor(self, _, dim):
        # Handle empty tensor in concat
        class Cat(nn.Module):
            def forward(self, x, y):
                return torch.ops.aten.cat.default((x, y), dim)

        inputs = [
            torch.randn(1, 2, 3, device="cuda"),
            y = torch.empty(0, 2, 3, device="cuda")
        ]
        self.run_test(Cat(), inputs)
```

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
apbose: Yes ideally. The validation would be based on the ITensor shape. Yes should use the meta data

apbose: But then this won't distinguish between ITensor and torch Tensor case.

apbose:
Closing in favor of #3890 and #3891

PR #3883, opened 2025-11-03
fix: torchtrtc precision setting logic

yeetypete:
# Description

Ensures that `torchtrtc` precision settings do not always contain a default `fp32` precision when the precision is explicitly passed as an argument. This is particularly important when compiling a model to run on the DLA which does not allow `fp32` precision. Currently this must not have been possible to do with the `torchtrtc` cli.

Bug example:
```
torchtrtc ssd_traced.jit.pt ssd_trt_dla.ts "(1,3,300,300)@f16%contiguous" -p fp16 --device-type=dla -v
...
INFO: Settings requested for TensorRT engine:
    Enabled Precisions: Float32 Float16
    TF32 Floating Point Computation Enabled: 1
    Truncate Long and Double: 0
    Make Refittable Engine: 0
    Debuggable Engine: 0
    GPU ID: 0
    Allow GPU Fallback (if running on DLA): 0
    Avg Timing Iterations: 1
    Max Workspace Size: 0
    DLA SRAM Size: 1048576
    DLA Local DRAM Size: 1073741824
    DLA Global DRAM Size: 536870912
    Device Type: DLA
    GPU ID: 0
    DLACore: 0
    Engine Capability: standard
    Calibrator Created: 0
```
This should only report `Float16` enabled precision.

## Type of change

# Checklist:

### DIFF
--- a/cpp/bin/torchtrtc/main.cpp
+++ b/cpp/bin/torchtrtc/main.cpp
@@ -356,6 +356,7 @@ int main(int argc, char** argv) {
   }
 
   if (enabled_precisions) {
+    compile_settings.enabled_precisions.clear();
     for (const auto& precision : args::get(enabled_precisions)) {
       auto dtype = torchtrtc::parserutil::parse_dtype(precision);
       if (dtype == torchtrt::DataType::kFloat) {


narendasan:
Thanks for the PR, it seems fine, we will test locally since there are some issues in CI. 
One question that is somewhat of an aside. 

Is there some limitation in the Dynamo / Exported Program workflow that makes it hard to use for DLA? If so this is something we would like to fix so that users can port off TorchScript since the overall PyTorch ecosystem is moving in that direction.

yeetypete:
@narendasan thanks for the quick response.

> Is there some limitation in the Dynamo / Exported Program workflow that makes it hard to use for DLA?

Currently we haven't tried using Dynamo / Exported Program workflow since we are interested in a Pythonless deployment with ahead-of-time compilation. However, I did notice that the AOT-Inductor pythonless deployment is in beta. It seems like currently you need to use Python to perform the export. This isn't a huge obstacle but it would be great to know if it is possible to perform this export step without Python just using libtorch and tensorrt c++ APIs. Maybe this could be something `torchtrtc` could support in the future?

narendasan:
AFAIK there isnt a way to `torch_tensorrt + aot_compile` in C++ currently but we can look into it / raise the usecase in the PyTorch ecosystem. But yes you should still be able to deploy in C++ through AOTInductor + Torch-TensorRT given export / compile in python today

yeetypete:
In the end we were able to do model conversion with Python via a docker container so C++ only `aot_compile` isn't a big blocker for us anymore. We will do some more testing to see if anything else comes up but so far looks good.

PR #3886, opened 2025-11-04
Cpu memory graph break

cehongwang:
# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/partitioning/_resource_partitioner.py
+++ b/py/torch_tensorrt/dynamo/partitioning/_resource_partitioner.py
@@ -59,7 +59,7 @@ from torch_tensorrt.dynamo.partitioning._atomic_subgraphs import (
 
 logger = logging.getLogger(__name__)
 
-MAX_NUM_OF_ENGINES = 40
+MAX_NUM_OF_ENGINES = 50
 ENGINE_COMPILATION_MEMORY_USAGE_MULTIPLIER = 4
 
 
@@ -126,6 +126,9 @@ class ResourcePartitioner(_SplitterBase):  # type: ignore
         subgraphs = self.break_subgraphs(
             subgraphs, subgraph_size_budget=self.calculate_size_budget()
         )
+
+        if len(subgraphs) == 1:
+            return self.module
 
         # Set the number of TRT engines to be generated
         self.num_trt_accelerated_subgraphs = len([s for s in subgraphs if s.is_acc])
@@ -315,7 +318,7 @@ class ResourcePartitioner(_SplitterBase):  # type: ignore
             )
             new_subgraphs = self.step_and_validate(new_subgraphs, step_size)
             size_0, size_1 = self.size_of_subgraphs(new_subgraphs)
-            if size_0 > size_to_break:
+            if size_0 > size_to_break or size_0 > size_1:
                 break
 
         return new_subgraphs, size_0, size_1
--- a/tests/py/dynamo/partitioning/test_001_resource_partitioning.py
+++ b/tests/py/dynamo/partitioning/test_001_resource_partitioning.py
@@ -166,7 +166,7 @@ class TestResourcePartitioning(TestCase):
         )
 
         partitioned_module = resource_partition(
-            partitioned_module, cpu_memory_budget=1.7 * 1024 * 1024 * 1024  # 1.7GB,
+            partitioned_module, cpu_memory_budget=1.4 * 1024 * 1024 * 1024  # 1.4GB,
         )
 
         assert (
@@ -177,8 +177,8 @@ class TestResourcePartitioning(TestCase):
                     if "_run_on_acc" in name
                 ]
             )
-            > 3
-        ), "The graph should have more than 3 accelerated subgraphs"
+            == 5
+        ), "The graph should have 5 accelerated subgraphs"
         assert (
             len(
                 [
@@ -298,7 +298,7 @@ class TestResourcePartitioning(TestCase):
         )
 
         partitioned_module = resource_partition(
-            partitioned_module, cpu_memory_budget=1.7 * 1024 * 1024 * 1024  # 1.7GB,
+            partitioned_module, cpu_memory_budget=1.4 * 1024 * 1024 * 1024  # 1.4GB,
         )
 
         assert (
@@ -396,7 +396,7 @@ class TestResourcePartitioning(TestCase):
         )
 
         partitioned_module = resource_partition(
-            partitioned_module, cpu_memory_budget=1.7 * 1024 * 1024 * 1024  # 1.7GB,
+            partitioned_module, cpu_memory_budget=1.4 * 1024 * 1024 * 1024  # 1.4GB,
         )
 
         assert (
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -135,7 +135,7 @@ jobs:
         pushd .
         cd tests/py/dynamo
         ../../../packaging/vc_env_helper.bat python -m pytest -ra -n 8 --junitxml=${RUNNER_TEST_RESULTS_DIR}/l0_dynamo_core_runtime_tests_results.xml runtime/test_000_*
-        ../../../packaging/vc_env_helper.bat python -m pytest -ra -n 8 --junitxml=${RUNNER_TEST_RESULTS_DIR}/l0_dynamo_core_partitioning_tests_results.xml partitioning/
+        ../../../packaging/vc_env_helper.bat python -m pytest -ra -n 8 --junitxml=${RUNNER_TEST_RESULTS_DIR}/l0_dynamo_core_partitioning_tests_results.xml partitioning/test_000_*
         ../../../packaging/vc_env_helper.bat python -m pytest -ra -n 8 --junitxml=${RUNNER_TEST_RESULTS_DIR}/l0_dynamo_core_lowering_tests_results.xml lowering/
         popd
 
@@ -219,6 +219,7 @@ jobs:
         pushd .
         cd tests/py/dynamo
         ../../../packaging/vc_env_helper.bat python -m pytest -ra -n 8 --junitxml=${RUNNER_TEST_RESULTS_DIR}/l1_dynamo_core_tests_results.xml runtime/test_001_*
+        ../../../packaging/vc_env_helper.bat python -m pytest -ra -n 8 --junitxml=${RUNNER_TEST_RESULTS_DIR}/l1_dynamo_core_partitioning_tests_results.xml partitioning/test_001_*
         popd
 
   L1-dynamo-compile-tests:


py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
```diff
@@ -26,6 +25,10 @@
 )
 
 logger = logging.getLogger(__name__)
+NON_BREAKABLE_OP_LISTS = [
+    ["addmm", "addmm"],
```
narendasan: Just a note for implementation later.

1. this should use an actual subgraph definition 
2. it should use pytorch op targets not strings
3. `addmm` should be decomposed right so the graph we want is `mm` -> `add`
4. There should be a user facing API to modify this list similar to what we have for passes

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
```diff
@@ -241,44 +245,27 @@ def partition_graph(self) -> torch.fx.GraphModule:
...
         gm = self.split()
-        self.weight_visited_nodes = set()
-        [self.size_of_subgraph(s) for s in subgraphs]
-        
 
         return gm
-    
-    def calculate_num_of_break(self, subgraphs: List[Subgraph]) -> int:
+
+    def calculate_size_budget(
```
narendasan: Should there be an API to define this manually?

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
```diff
@@ -241,44 +245,27 @@ def partition_graph(self) -> torch.fx.GraphModule:
...
+        return available_rss // engine_compilation_memory_usage_multiplier
 
     def break_subgraphs_by_node(
         self, subgraphs: List[Subgraph], num_of_break: int = 1
     ) -> List[Subgraph]:
         """
         This function breaks the subgraphs into smaller subgraphs at the specified frequency to save CPU memory.
         """
-        op_to_break = "add."
+        op_to_break = "addmm."
```
narendasan: Why is this hardcoded?

cehongwang: This function is not called. I just left there for experiment and would delete later

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
cehongwang: Yeah I think so. For now you can just hardcode and play with it

narendasan:
1. We should think about using this tech for refit vs non refit 
2. Make refit apis work across graph breaks

narendasan:
Improve usability by automating nn.Module -> atomic fx graph

py/torch_tensorrt/dynamo/partitioning/_atomic_subgraphs.py
```diff
@@ -0,0 +1,133 @@
...
+
+class ConvSilu(torch.nn.Module):  # type: ignore[misc]
+    def __init__(self) -> None:
+        super().__init__()
+
+    def forward(
+        self, x: torch.Tensor, weight: torch.Tensor, bias: torch.Tensor
+    ) -> torch.Tensor:
+        x = aten.convolution.default(
+            x, weight, bias, [1, 1], [1, 1], [1, 1], False, [0, 0], 1
```
narendasan: You probably should not encode these parameters of the conv as static, they should likely be arguments to the forward so that the graph is generic

py/torch_tensorrt/dynamo/partitioning/fusion_patterns.py
```diff
@@ -0,0 +1,133 @@
...
+
+class ConvGelu(torch.nn.Module):  # type: ignore[misc]
+    def __init__(self) -> None:
+        super().__init__()
+
+    def forward(
+        self, x: torch.Tensor, weight: torch.Tensor, bias: torch.Tensor
+    ) -> torch.Tensor:
+        x = aten.convolution.default(
+            x, weight, bias, [1, 1], [1, 1], [1, 1], False, [0, 0], 1
```
narendasan: same here

py/torch_tensorrt/dynamo/partitioning/fusion_patterns.py
```diff
@@ -0,0 +1,133 @@
...
+
+class ConvReLU(torch.nn.Module):  # type: ignore[misc]
+    def __init__(self) -> None:
+        super().__init__()
+
+    def forward(
+        self, x: torch.Tensor, weight: torch.Tensor, bias: torch.Tensor
+    ) -> torch.Tensor:
+        x = aten.convolution.default(
+            x, weight, bias, [1, 1], [1, 1], [1, 1], False, [0, 0], 1
```
narendasan: same here

py/torch_tensorrt/dynamo/partitioning/fusion_patterns.py
```diff
@@ -0,0 +1,133 @@
...
+        x: torch.Tensor,
+        weight: torch.Tensor,
+        bias: torch.Tensor,
+        bn_weight: torch.Tensor,
+        bn_bias: torch.Tensor,
+        running_mean: torch.Tensor,
+        running_var: torch.Tensor,
+    ) -> torch.Tensor:
+        x = aten.convolution.default(
+            x, weight, bias, [1, 1], [1, 1], [1, 1], False, [0, 0], 1
```
narendasan: same here

py/torch_tensorrt/dynamo/partitioning/fusion_patterns.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def forward(
+        self, x: torch.Tensor, y: torch.Tensor, z: torch.Tensor
+    ) -> torch.Tensor:
+        x = aten.mul.Tensor(x, y)
+        x = aten.mul.Tensor(x, z)
+        return x
+
+
+All_FUSION_PATTERNS = [
```
narendasan: We could cache the graphs if we do symbolic trace on register

py/torch_tensorrt/dynamo/_defaults.py
```diff
@@ -57,6 +57,7 @@
 L2_LIMIT_FOR_TILING = -1
 USE_DISTRIBUTED_MODE_TRACE = False
 OFFLOAD_MODULE_TO_CPU = False
+CPU_MEMORY_BUDGET = -1
```
narendasan: Use an optional instead since this is not a TRT api we dont need -1 to mean let us decide

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -853,3 +856,36 @@ def get_output_dtypes(output: Any, truncate_double: bool = False) -> List[dtype]
...
+    if torch.cuda.get_device_capability() in [(11, 0)]:
+        return True
+    return False
+
+
+def get_cpu_memory_usage() -> Any:
+    return psutil.Process().memory_info().rss / 1024 / 1024
+
+
+def release_memory() -> None:
```
narendasan: Did this get moved?

tests/py/dynamo/models/test_models.py
```diff
@@ -55,6 +55,52 @@ def test_resnet18(ir):
     torch._dynamo.reset()
 
 
+def compile_one(idx: int, ir: str):
```
narendasan, line 60: Why is this test here?

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
```diff
@@ -225,12 +232,304 @@ def partition_graph(self) -> torch.fx.GraphModule:
...
+        for node in nodes_in_first_subgraph:
+            for user in node.users:
+                if user not in nodes_in_first_subgraph:
+                    leaf_node.add(node)
+                    break
+        return leaf_node
+
+    def size_of_subgraphs(self, subgraphs: List[Subgraph]) -> List[int]:
+        """
+        This function calculates the size of the subgraph.
```
narendasan: Can you describe the algorithms here so we have reference for later?

py/torch_tensorrt/dynamo/partitioning/fusion_patterns.py
cehongwang: Do you mean to trace the graphs when the program starts every time? Do you think that would cause unnecessary latency when cpu memory is enough. I am thinking maybe we could use LRU cache or something so it will only be called once and it's lazy initialization

py/torch_tensorrt/dynamo/partitioning/_atomic_subgraphs.py
```diff
@@ -0,0 +1,185 @@
...
+    ConvBNReLU,
+    ConvReLU,
+    ConvGelu,
+    ConvSilu,
+    MulAdd,
+    MulMul,
+]
+
+
+@lru_cache(maxsize=None)
```
narendasan: shouldnt we lru_cache the lowered / traced graphs not the subgraph matching since the subgraph matching would change model to model

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -850,6 +854,16 @@ def preserve_module_specs(
             require_full_compilation=settings.require_full_compilation,
         )
 
+    from torch_tensorrt.dynamo.partitioning._resource_partitioner import (
```
narendasan: We should have a user setting while we are testing right?

py/torch_tensorrt/dynamo/partitioning/_atomic_subgraphs.py
```diff
@@ -0,0 +1,195 @@
...
+
+import torch
+from torch.fx.passes.utils.matcher_utils import SubgraphMatcher
+from torch.ops import aten
+
+ATOMIC_SUBGRAPHS = []
+
+
+def register_atomic_subgraph(
+    is_aten: bool = False,
```
narendasan: Add a handler for `is_aten=False` that just throws an Unimplemented Error for now

narendasan (commented):
Its looking good, just add a quick example in the examples folder and list it under contributor documentation for now

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -868,6 +885,16 @@ def preserve_module_specs(
     # Iterate over all components that can be accelerated
     # Generate the corresponding TRT Module for those
 
+    # Here we delete the frozen parameters from the graph module. Note this does not affect the submodules. We are going to delete the frozen parameters from the submodules in the convert_module function.
+    # This is done to release CPU memory.
+    for attr in dir(gm):
+        if attr.startswith("_frozen_param"):
+            delattr(gm, attr)
```
zewenli98, line 951: can we delete the entire gm after putting `parse_graph_io(gm, dryrun_tracker)` in the front?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -868,6 +885,16 @@ def preserve_module_specs(
...
+
+    from torch_tensorrt.dynamo.conversion._ConverterRegistry import DYNAMO_CONVERTERS
+
+    DYNAMO_CONVERTERS.disallowed_targets = set()
```
zewenli98: why do we need this?

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -140,6 +141,7 @@ class CompilationSettings:
     l2_limit_for_tiling: int = L2_LIMIT_FOR_TILING
     use_distributed_mode_trace: bool = USE_DISTRIBUTED_MODE_TRACE
     offload_module_to_cpu: bool = OFFLOAD_MODULE_TO_CPU
+    cpu_memory_budget: int = CPU_MEMORY_BUDGET
```
zewenli98, line 174: consider adding this arg in the `_SETTINGS_TO_BE_ENGINE_INVARIANT` below.

py/torch_tensorrt/dynamo/_compiler.py
cehongwang: Maybe. What does this do? This has to be done before the loop of conversion. `parse_graph_io(gm, dryrun_tracker)` is inside. Do you want to use a seperate loop to parse the graph before deleting the frozen attribute?

examples/dynamo/low_cpu_memory_compilation.py
```diff
@@ -0,0 +1,84 @@
...
+}
+
+settings = CompilationSettings(**compilation_options)
+with torchtrt.dynamo.Debugger(
+    log_level="debug",
+    logging_dir="/home/profile/logging/moe",
+    engine_builder_monitor=False,
+):
+
+    exp_program = torch.export.export(model, tuple(inputs))
```
narendasan, line 77: Add some comments on the expected graph you get out so that we can render this as a tutorial in the rendered docs

py/torch_tensorrt/dynamo/partitioning/_atomic_subgraphs.py
```diff
@@ -0,0 +1,210 @@
...
+
+@lru_cache(maxsize=None)
+def get_compiled_atomic_subgraphs() -> List[torch.fx.GraphModule]:
+    """
+    This function gets the compiled atomic subgraphs from the graph.
+    LRU cache the result to avoid recompiling the same pattern multiple times.
+    """
+    compiled_atomic_subgraphs = []
+    for pattern, is_aten in ATOMIC_SUBGRAPHS:
+        pattern_graph = torch.fx.symbolic_trace(pattern())
```
narendasan: @cehongwang this is what we should cache not the full list. If we cache the full list then any time the list changes we will get an invalidation. Also not sure how lru cache will work with global variables instead of arguments

narendasan: Make another function that is just 

```py
@lru_cache(maxsize=None)
def trace_atomic_graph(graph: Type[torch.nn.Module], is_core_aten: bool=True) -> torch.fx.GraphModule:
     if is_core_aten:
        return torch.fx.symbolic_trace(graph())
     else: 
         raise NotImplementedError("Resource partitioner currently does not support unlowered atomic subgraphs")
```

py/torch_tensorrt/dynamo/_compiler.py
zewenli98: I meant, I searched `gm` in the function and it's only used in parse_graph_io(gm, dryrun_tracker). I was wondering if the entire `gm` can be removed instead of only removing the attr `_frozen_param`

cehongwang: Oh, removing gm does not help because there are other references at other place. But removing the frozen attributes is effective because the only references to the weight is in gm

py/torch_tensorrt/dynamo/partitioning/_atomic_subgraphs.py
narendasan: @cehongwang this is still an issue, you need to have all parameters be inputs to the graph for this graph to be generic

narendasan: See above as an example and just copy that

narendasan: You may even want to template the conv + activation ones

cehongwang: Oh I left this one unchanged

py/torch_tensorrt/dynamo/partitioning/_resource_partitioner.py
```diff
@@ -0,0 +1,555 @@
...
+        nodes = []
+        for node in self.module.graph.nodes:
+            if hasattr(node, "tag"):
+                del node.tag
+            if node.op in CALLABLE_NODE_OPS:
+                nodes.append(node)
+        subgraphs = [Subgraph(is_acc=True, nodes=nodes)]
+        self.fusion_patterns = get_node_in_fusion_pattern(self.module.graph)
+
+        assert self.check_topological_order(
```
narendasan: When will this fail? Do we need to check this or can FX just guarantee this for us?

py/torch_tensorrt/dynamo/partitioning/_resource_partitioner.py
```diff
@@ -0,0 +1,555 @@
...
+
+        assert isinstance(module, torch.fx.GraphModule)
+
+        self.module = module
+        self.cpu_memory_budget = cpu_memory_budget
+        self.resource_split_count = 0
+        self.submodule_name = submodule_name
+        self.deps = self.find_deps()
+
+        self.non_acc_submodule_name = "_run_on_gpu_"
```
narendasan: Is this relevant? Since there will only be acc modules?

py/torch_tensorrt/dynamo/partitioning/_resource_partitioner.py
```diff
@@ -0,0 +1,555 @@
...
+
+        for i, subgraph in enumerate(subgraphs):
+            if i == 0:
+                for node in subgraph.nodes:
+                    visited_nodes[node] = i
+                # breakpoint node's users should belong to the next subgraph
+                visited_nodes[subgraph.nodes[-1]] = i + 1
+                continue
+
+            elif not subgraph.is_acc:
```
narendasan: Again will we ever see this?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -615,6 +622,10 @@ def compile(
             "'arg_inputs' and 'inputs' should not be used at the same time."
         )
 
+    assert (
```
narendasan: Why?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -854,6 +866,11 @@ def preserve_module_specs(
             require_full_compilation=settings.require_full_compilation,
         )
 
+    partitioned_module = resource_partition(
```
narendasan: Shoudnt this conditionally run based on if the user provided a budget at all?

narendasan: / should we make this opt in while testing?

tests/py/dynamo/partitioning/test_001_resource_partitioning.py
```diff
@@ -0,0 +1,265 @@
...
+                [
+                    name
+                    for name, _ in partitioned_module.named_children()
+                    if "_run_on_gpu" in name
+                ]
+            )
+            == 2
+        ), "The graph should have 2 non-accelerated subgraphs"
+
+    def test_resource_partitioning_with_global_capability_partitioning(self):
```
narendasan, line 329: We should also have a test with no fall back as well as a test which tests for making sure that the atomic subgraph system works

narendasan: Also one that tests registering a new atomic subgraph and then verifying that it has the desired effect

py/torch_tensorrt/dynamo/partitioning/_resource_partitioner.py
cehongwang: This should never fail. I put this here just in case something changes in torch and we don't know

py/torch_tensorrt/dynamo/_compiler.py
cehongwang: Torch-TensorRT will occupy some resources so it's best for users to have at least 2GBof  resources. We can also delete this because if the available memory is less than that, it will fail in a later stage anyway.

tests/py/dynamo/partitioning/test_001_resource_partitioning.py
cehongwang: `Also one that tests registering a new atomic subgraph and then verifying that it has the desired effect`
Can this prove the atomic subgraph system works?

narendasan: yeah I mean what we want is basically given a graph and an atomic subgraph, propose an invalid cut and make sure the system adjusts it s.t one of the two subgraphs created has the complete atomic subgraph and the other has no part of the atomic subgraph

narendasan: So like 
```py
graph(x):
   aten::add(x, x, 1)
   aten::mul(y, y, 1)
   aten::mul(z, z, 1)
   aten::add(a, a, 1)
   return b
```

Then set the only atomic subgraph to 
```py
graph(y):
   aten::mul(y, y, 1)
   aten::mul(z, z, 1)
   return a
```

And set the cut at `z`

The result should be either one of 

```py
graph(x):
   aten::add(x, x, 1)
   aten::mul(y, y, 1)
   aten::mul(z, z, 1)
   return a

graph(a):
   aten::add(a, a, 1)
   return b
```

or

```py
graph(x):
   aten::add(x, x, 1)
   return y

graph(y):
   aten::mul(y, y, 1)
   aten::mul(z, z, 1)
   aten::add(a, a, 1)
   return b
```

py/torch_tensorrt/dynamo/partitioning/_resource_partitioner.py
narendasan: Check if FX guarantees this. If it does then we dont need to do it

py/torch_tensorrt/dynamo/partitioning/_resource_partitioner.py
```diff
@@ -0,0 +1,547 @@
...
+
+        Returns:
+            List[Subgraph]: New list of subgraphs after resource-aware splitting.
+        """
+
+        new_subgraphs = []
+        # We throw an error if the remaining memory is almost empty compared to the model size.
+        # i.e. if the remaining memory is 4G (budget is 1G) the model size is greater than 40G, we stop the compilation.
+        sizes = self.size_of_subgraphs(subgraphs)
+        if sum(sizes) > subgraph_size_budget * 40:
```
narendasan: Why *40? just we dont want more than 40 engines? Also make this a named variable somewhere instead of just a literal

py/torch_tensorrt/dynamo/partitioning/_resource_partitioner.py
```diff
@@ -0,0 +1,547 @@
...
+                subgraph = broken_subgraphs[1]
+
+            if len(subgraph.nodes) != 0:
+                new_subgraphs.append(subgraph)
+
+        self._varify_all_fusion_nodes_in_same_subgraph(new_subgraphs)
+
+        return new_subgraphs
+
+    def _varify_all_fusion_nodes_in_same_subgraph(
```
narendasan: _varify -> _verify

py/torch_tensorrt/dynamo/partitioning/_atomic_subgraphs.py
```diff
@@ -0,0 +1,184 @@
...
+        ATOMIC_SUBGRAPHS.append((subgraph, init_args, is_core_aten))
+        return subgraph
+
+    return decorator
+
+
+@register_atomic_subgraph(init_args=(aten.silu.default,), is_core_aten=True)
+@register_atomic_subgraph(init_args=(aten.gelu.default,), is_core_aten=True)
+@register_atomic_subgraph(init_args=(aten.relu.default,), is_core_aten=True)
+@register_atomic_subgraph(init_args=(aten.sigmoid.default,), is_core_aten=True)
```
narendasan, line 27: Nice I like this design

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -512,6 +523,8 @@ def compile(
         l2_limit_for_tiling (int): The target L2 cache usage limit (in bytes) for tiling optimization (default is -1 which means no limit).
         offload_module_to_cpu (bool): Offload the module to CPU. This is useful when we need to minimize GPU memory usage.
         use_distributed_mode_trace (bool):  Using aot_autograd to trace the graph. This is enabled when DTensors or distributed tensors are present in distributed model
+        enable_resource_partitioning (bool): Enable resource-aware partitioning. This is useful when the model is large and the CPU memory is limited.
+        cpu_memory_budget (int): The maximum amount of CPU memory to use for the compilation. If the compilation requires more memory than this budget, the compilation will fail. If set to -1, the compilation will use all available CPU memory.
```
narendasan: Its not `-1` right, should be `Optional[Int]`? So, if None use all memory?

py/torch_tensorrt/dynamo/_defaults.py
```diff
@@ -57,6 +58,8 @@
 L2_LIMIT_FOR_TILING = -1
 USE_DISTRIBUTED_MODE_TRACE = False
 OFFLOAD_MODULE_TO_CPU = False
+ENABLE_RESOURCE_PARTITIONING = False
+CPU_MEMORY_BUDGET = psutil.virtual_memory().available
```
narendasan: This gets set once at the import of `_defaults`. But the available memory changes over time. We should default to `None` and call `psutil.virtual_memory().available` in `_compile`

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -140,6 +142,8 @@ class CompilationSettings:
     l2_limit_for_tiling: int = L2_LIMIT_FOR_TILING
     use_distributed_mode_trace: bool = USE_DISTRIBUTED_MODE_TRACE
     offload_module_to_cpu: bool = OFFLOAD_MODULE_TO_CPU
+    enable_resource_partitioning: bool = ENABLE_RESOURCE_PARTITIONING
+    cpu_memory_budget: int = CPU_MEMORY_BUDGET
```
narendasan, line 174: Optional[Int]

py/torch_tensorrt/dynamo/_compiler.py
cehongwang: Yes

py/torch_tensorrt/dynamo/partitioning/_atomic_subgraphs.py
```diff
@@ -0,0 +1,184 @@
...
+
+@lru_cache(maxsize=None)
+def trace_atomic_graph(
+    graph: torch.nn.Module, init_args: Any, is_core_aten: bool = True
+) -> torch.fx.GraphModule:
+    if is_core_aten:
+        return torch.fx.symbolic_trace(graph(*init_args))
+    else:
+        raise NotImplementedError(
+            "Resource partitioner currently does not support unlowered atomic subgraphs"
```
narendasan, line 183: Can you add an issue to the backlog to implement this?

narendasan: Attach it to the CPU memory story

cehongwang: ok

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -179,6 +185,8 @@ def cross_compile_for_windows(
         tiling_optimization_level (str): The optimization level of tiling strategies. A higher level allows TensorRT to spend more time searching for better tiling strategy. We currently support ["none", "fast", "moderate", "full"].
         l2_limit_for_tiling (int): The target L2 cache usage limit (in bytes) for tiling optimization (default is -1 which means no limit).
         use_distributed_mode_trace (bool):  Using aot_autograd to trace the graph. This is enabled when DTensors or distributed tensors are present in distributed model
+        enable_resource_partitioning (bool): Enable resource-aware partitioning. This is useful when the model is large and the CPU memory is limited.
+        cpu_memory_budget (Optional[int]): The maximum amount of CPU memory to use for the compilation. If the compilation requires more memory than this budget, the compilation will fail. If set to -1, the compilation will use all available CPU memory. If set to None, the compilation will use all available CPU memory.
```
narendasan: Update the docstring

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -512,6 +524,8 @@ def compile(
         l2_limit_for_tiling (int): The target L2 cache usage limit (in bytes) for tiling optimization (default is -1 which means no limit).
         offload_module_to_cpu (bool): Offload the module to CPU. This is useful when we need to minimize GPU memory usage.
         use_distributed_mode_trace (bool):  Using aot_autograd to trace the graph. This is enabled when DTensors or distributed tensors are present in distributed model
+        enable_resource_partitioning (bool): Enable resource-aware partitioning. This is useful when the model is large and the CPU memory is limited.
+        cpu_memory_budget (Optional[int]): The maximum amount of CPU memory to use for the compilation. If the compilation requires more memory than this budget, the compilation will fail. If set to -1, the compilation will use all available CPU memory. If set to None, the compilation will use all available CPU memory.
```
narendasan: Update the docstring, since -1 means nothing now

py/torch_tensorrt/dynamo/_settings.py
narendasan: Did this change happen?

narendasan: Actually we dont need this

PR #3891, opened 2025-11-06
Empty tensor handling 

apbose:
This PR addresses the case of empty tensor in torchTRT based on https://docs.nvidia.com/deeplearning/tensorrt/latest/inference-library/advanced.html#empty-tensors, and also  focuses on concat operation edge case involving empty tensor

TODO: Might have to separate the case of concat from this PR, in the case when torch.Tensor([]) and a rank greater tensor are concatenated, which is a valid case in pytorch but not TRT.

This addressed #3865. Corresponding HF transformers issue raised -  https://github.com/huggingface/transformers/issues/42027 where empty tensor should not come in the first place

Update:
This PR also handles the empty tensor input in get_trt_tensor()
Removes the obsolete FX converters from the converter registry. Two converters are ported- hardtanh and sym_numel

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/cat.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/cat.py
@@ -37,10 +37,27 @@ def unify_and_concat_trt_tensors(
         cast_dtype: Optional target dtype for casting TRT tensors.
         force_trt_output: If True, return TRT tensor even if all inputs are static ints. (True for concat operations)
     """
-    has_dynamic = any(not isinstance(x, int) for x in inputs)
+    # Normalize scalar tensors (0D) to Python values to avoid 0D vs 1D shape issues.
+    #
+    # eg case:
+    # torch.tensor(3) is a 0D tensor (shape=[])
+    # get_trt_tensor creates a 0D TRT constant for it (shape=trt.Dims())
+    # Python int 3 via get_trt_tensor creates a 1D TRT constant (shape=(1,))
+    # because to_torch(3) returns torch.tensor([3]) with shape (1,)
+    #
+    # By normalizing torch.tensor(3) -> 3, we ensure:
+    # 1. Pure static case: all ints -> returns list directly, no TRT ops needed (eg:upsample)
+    # 2. Mixed case: Python ints become 1D constants, compatible with other 1D tensors
+    normalized_inputs = []
+    for x in inputs:
+        if isinstance(x, (torch.Tensor, np.ndarray)) and x.ndim == 0:
+            normalized_inputs.append(x.item())
+        else:
+            normalized_inputs.append(x)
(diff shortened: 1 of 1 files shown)


core/runtime/execute_engine.cpp
```diff
@@ -117,6 +117,8 @@ void setup_input_tensors(
     auto shape = core::util::toVec(dims);
     LOG_DEBUG("Input Name: " << name << " Shape: " << dims);
 
+    void* tensor_addr = nullptr;
```
narendasan: I strongly want to avoid having nullptr basically anywhere, we should be looking for some sane default

narendasan (commented):
@apbose this maybe be a case where we would want TRT to properly handle this rather than us doing something hacky. Lets raise it with Yuan Yuo. Dummy inputs do not feel like the right solution

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
```diff
@@ -478,7 +481,18 @@ def __getitem__(
...
                             )
                             continue
+                    # Reached here = all dynamo validators failed for this key
+                    # Don't fall back to FX - should fall back to PyTorch instead
+                    dynamo_converter_failed_validation = True
                 else:
+                    # FX converters (legacy, stored as single function)
+                    # Skip FX if dynamo converter exists but failed validation
+                    # This ensures validator decisions are respected
+                    if dynamo_converter_failed_validation:
```
narendasan: Why is this needed for empty tensor handling? Also are we near the point we can just remove FX converters?

apbose: This is particularly for the case where x is torch.Tensor([]), and y is torch.Tensor torch.randn(3,4) where the validator comes into play. 

```
# makes use of validator
class ConcatEmptyModelEmptyConstantMisMatchDim(nn.Module):
    def __init__(self, dim=0):
        super().__init__()
        self.dim = dim

    def forward(self, x):
        y = torch.tensor([], device="cuda")
        return torch.cat([x, y], dim=self.dim)
```
This error comes -  
```
"/code/torchTRT/TensorRT/py/torch_tensorrt/fx/converters/aten_ops_converters.py", line 465, in aten_ops_cat                                                                                                                                                                                                                                  
    return acc_ops_converters.acc_ops_cat(network, target, None, kwargs_new, name) 
```
acc_ops converter is called since validator fails and it resorts to the fx converter implementation which is next in precedence. Yes we should remove the converters IMO, but till that is not done, this would be required I suppose?

apbose: Also for the test case, `class TestEmptyTensorMemoryLeak(TestCase):` should cover the case of no memory leak or memory overload in case of repeated model calls. Let me know if you think any other test case is required

PR #3897, opened 2025-11-14
Adding rank based logging for torch distributed examples

apbose:
This PR 
1. Adds rank based logging for the distributed examples
2. Corrects the fallback to pytorch case for NCCL converters
3. This with #3830 provides utilities for running distributed tensor parallel examples using torch.distributed

### DIFF
--- a/examples/distributed_inference/tensor_parallel_initialize_dist.py
+++ b/examples/distributed_inference/tensor_parallel_initialize_dist.py
@@ -3,7 +3,7 @@
 Tensor Parallel Initialize Distributed Environment
 ==================================================
 
-This module provides functions to initialize and clean up the distributed environment for tensor parallel distributed inference.
+This module provides functions to initialize and clean up the distributed environment for tensor parallel distributed inference. These utilities are useful for tensor parallel distributed inference examples using torch.distributed.
 """
 
 import logging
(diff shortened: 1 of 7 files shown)


examples/distributed_inference/tensor_parallel_initialize_dist.py
```diff
@@ -44,12 +99,49 @@ def initialize_distributed_env(rank=0, world_size=1, port=29500):
...
+    )
     device_id = (
         rank % torch.cuda.device_count()
     )  # Ensure each rank gets a unique device
     torch.cuda.set_device(device_id)
 
-    return device_mesh, world_size, rank
+    # Set C++ TensorRT runtime log level based on most verbose handler
+    # this is similar to set_log_level()
+    cpp_level = min(file_level_int, console_level_int)
```
narendasan, line 129: Dont we have an API that abstracts needing to detect if the C++ runtime is available? If not we should add one

tests/py/dynamo/distributed/test_nccl_ops.py
```diff
@@ -57,9 +74,11 @@ def forward(self, x):
 
 
 class TestNcclOpsConverter(DispatchTestCase):
+    # 1. Skip if NCCL backend is not available (e.g., Windows, Jetson) - hard requirement
+    # 2. Don't skip if TRTLLM is unavailable (e.g., CUDA 13) - falls back to PyTorch
     @unittest.skipIf(
-        not is_platform_supported_for_trtllm(),
-        "Skipped on Windows, Jetson and CUDA13: NCCL backend is not supported.",
+        not is_distributed_nccl_available(),
+        "Skipped: NCCL backend is not available (Windows/Jetson not supported).",
```
narendasan: Is it jetson or just Orin?

examples/distributed_inference/tensor_parallel_initialize_dist.py
apbose: I have added a function in _features.py for the above. And also moved all this to logging.py. Let me know if that function placment works

tests/py/dynamo/distributed/test_nccl_ops.py
apbose: yeah Orin. Changed to Jetson Orin

py/torch_tensorrt/logging.py
```diff
@@ -31,6 +34,81 @@ def log(self, severity: trt.ILogger.Severity, msg: str) -> None:
...
+        torch_tensorrt.logging.set_level(logging.DEBUG, logger=my_logger)
+    """
+    # Set the specified logger or default to root torch_tensorrt logger
+    if logger is None:
+        logging.getLogger("torch_tensorrt").setLevel(level)
+        _LOGGER.setLevel(level)
+    else:
+        logger.setLevel(level)
+
+    if has_torchscript_frontend():
```
narendasan, line 68: If we have the frontend we necessarily have the runtime, I dont think we need to use these APIs

py/torch_tensorrt/logging.py
```diff
@@ -46,13 +124,13 @@ def __enter__(self) -> None:
         self.external_lvl = _LOGGER.getEffectiveLevel()
         _LOGGER.setLevel(logging.CRITICAL)
 
-        if ENABLED_FEATURES.torchscript_frontend:
+        if has_torchscript_frontend():
```
narendasan, line 127: lets just remove the has_torchscript_frontend cases

narendasan (approved):
Just remove the TS ones since we should be able to handle both with the runtime and then LGTM

PR #3902, opened 2025-11-18
Fix Bugs

leimao:
I believe there are some bugs due to typo in programming.


cpp/src/types.cpp
narendasan, line 122: These are type constructors not direct translators https://github.com/pytorch/TensorRT/blob/98f54feede428760cfab8a000fcfaaeae65f973b/cpp/include/torch_tensorrt/torch_tensorrt.h#L144

cpp/src/types.cpp
```diff
@@ -157,11 +150,10 @@ TensorFormat::TensorFormat(at::MemoryFormat t) {
```
narendasan, line 157: Same thing here, value is held by a TensorFormat object

narendasan:
@leimao thanks for the PR. For the changes to the TS C++ API, these are not bugs. We define a set of wrapped enums that abstract the TRT and PyTorch versions. This makes it easier to write code that may mix the two APIs arbitrarily. 

Also as with FX, we are strongly pushing people to use Dynamo + AOTInductor to replace TorchScript as it is now deprecated in PyTorch and will be removed in PyTorch 2.14

cpp/src/types.cpp
leimao: I misread this one. For the purpose of clarity, I think we should use `m` prefix for object member variables. Will revert these.

leimao:
> @leimao thanks for the PR. For the changes to the TS C++ API, these are not bugs. We define a set of wrapped enums that abstract the TRT and PyTorch versions. This makes it easier to write code that may mix the two APIs arbitrarily.
> 
> Also as with FX, we are strongly pushing people to use Dynamo + AOTInductor to replace TorchScript as it is now deprecated in PyTorch and will be removed in PyTorch 2.14

@narendasan I believe there is still a bug, please review again. Thanks.

core/lowering/register_trt_placeholder_ops.cpp
```diff
@@ -20,7 +20,6 @@ RegisterOperators trt_placeholder_ops_reg({
         [](Stack& stack) {
           auto attn_mask = pop(stack).to<at::Tensor>();
           if (attn_mask.scalar_type() == at::kBool) {
-            attn_mask = attn_mask;
```
leimao, line 23: This one is unnecessary and is very confusing.

cpp/src/types.cpp
```diff
@@ -158,6 +158,7 @@ TensorFormat::TensorFormat(at::MemoryFormat t) {
   switch (t) {
     case at::MemoryFormat::ChannelsLast:
       value = TensorFormat::kChannelsLast;
+      break;
```
leimao, line 161: This one fixes an oversight of the implementation.

narendasan (approved):
LGTM, Would reiterate again that the torchscript frontend is in maintenance mode so unless there are bugs we likely will not change the implementation much. But we are highly interested in collecting reasons why Dynamo isnt working for your use case if thats the reason you are using TS so we can work to get those gaps closed prior to torchscript being removed

leimao:
The test failures seems to be irrelevant to my change. Some warnings were treated as errors by compiler flags.

leimao:
> LGTM, Would reiterate again that the torchscript frontend is in maintenance mode so unless there are bugs we likely will not change the implementation much. But we are highly interested in collecting reasons why Dynamo isnt working for your use case if thats the reason you are using TS so we can work to get those gaps closed prior to torchscript being removed

I think I mentioned Dynamo worked in my case. But for some reason, in a different environment, I only have a lower version of Torch-TensorRT which does not seem to have the Dynamo path. Is there any performance difference between the FX path and Dynamo path based on your experience?

Also could you please take a lot at the test failures?

Thanks @narendasan.

narendasan:
> Also could you please take a lot at the test failures?

Yeah if you rebase it should resolve those issues 

>  Is there any performance difference between the FX path and Dynamo path based on your experience?

They should be mostly the same but we have not been actively optimizing the FX path since Dynamo has superseded FX

leimao:
> > Also could you please take a lot at the test failures?
> 
> Yeah if you rebase it should resolve those issues
> 
> > Is there any performance difference between the FX path and Dynamo path based on your experience?
> 
> They should be mostly the same but we have not been actively optimizing the FX path since Dynamo has superseded FX

I have a noob question. Please bear with me if I did not ask it correctly. Since Dynamo will also produce FX graph as intermediate representation, which should be the same or similar to the one captured by Torch FX, what's the advantage of Torch-TensorRT-Dynamo over Torch-TensorRT-FX? Thanks.

narendasan:
> Since Dynamo will also produce FX graph as intermediate representation, which should be the same or similar to the one captured by Torch FX, what's the advantage of Torch-TensorRT-Dynamo over Torch-TensorRT-FX? Thanks.

Dynamo has better graph capture, and a fixed op set called Core ATen. This means dynamo has reproducible graphs and strong op coverage. Dynamo also fully supports current generation serialization workflows and other advanced use cases like dynamic shape. 

FX graphs can include basically any operator in PyTorch so its more likely to hit unsupported ops. Also there are more limitations in how the graphs are traced which makes it hard to use for production.

Issue #3909, opened 2025-11-19
🐛 [Bug] Engine cache failed on torch.compile backend=tensorrt


lanluo-nvidia:
##  Bug Description

engine cache test failed:

FAILED models/test_engine_cache.py::TestEngineCache::test_dynamo_compile_with_custom_engine_cache
FAILED models/test_engine_cache.py::TestEngineCache::test_torch_compile_graph_break
FAILED models/test_engine_cache.py::TestEngineCache::test_torch_compile_with_custom_engine_cache

https://gitlab-master.nvidia.com/dl/dgx/pytorch/-/jobs/234818816
https://gitlab-master.nvidia.com/dl/dgx/pytorch/-/jobs/234818814

## To Reproduce

It can be reproduced in local Linux work station

## Environment

> Build information about Torch-TensorRT can be found by turning on debug messages

 - Torch-TensorRT Version (e.g. 1.0.0):
 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source):
 - Build command you used (if compiling from source):
 - Are you using local sources or building from archives:
 - Python version:
 - CUDA version:
 - GPU models and configuration:
 - Any other relevant information:

lanluo-nvidia:
https://github.com/pytorch/TensorRT/pull/3915

zewenli98:
Hi @lanluo-nvidia, this PR #3915 can be closed because I improved and refactored engine caching in #3932

PR #3932, opened 2025-11-24
feat: improve engine caching and fix bugs

zewenli98:
# Description

As I requested, TensorRT 10.14 added an argument `trt.SerializationFlag.INCLUDE_REFIT` to allow refitted engines to keep refittable. That means engines can be refitted multiple times. Based on the capability, this PR enhances the existing engine caching and refitting features as follows:
1. To save hard disk space, engine caching will only save weight-stripped engines on disk regardless of `compilation_settings.strip_engine_weights`. Then, when users pull out the cached engine, it will be automatically refitted and kept refittable.
2. Compiled TRT modules can be refitted multiple times with `refit_module_weights()`. e.g.:
```
for _ in range(3):
    trt_gm = refit_module_weights(trt_gm, exp_program)
```
3. Due to some changes, the insertion and pulling of cached engines are located in different places, which causes #3909. This PR unified the insertion and pulling in `_conversion.py`.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/backend/backends.py
+++ b/py/torch_tensorrt/dynamo/backend/backends.py
@@ -159,8 +159,10 @@ def _pretraced_backend(
                 )
             if settings.strip_engine_weights:
                 logger.warning(
-                    "strip_engine_weights arg is not supported for torch.compile()"
+                    "strip_engine_weights=True is not supported for torch.compile(). It will be set to False automatically."
                 )
+                settings.strip_engine_weights = False
+
             trt_compiled = compile_module(
                 gm,
                 torchtrt_inputs,
--- a/tests/py/dynamo/models/test_weight_stripped_engine.py
+++ b/tests/py/dynamo/models/test_weight_stripped_engine.py
@@ -98,7 +98,9 @@ class TestWeightStrippedEngine(TestCase):
         gm1_output = gm1(*example_inputs)
 
         assertions.assertEqual(
-            gm1_output.sum(), 0, msg="gm1_output should be all zeros"
+            gm1_output.sum(),
+            0,
+            msg=f"gm1_output should be all zeros, but got {gm1_output}",
         )
 
     @unittest.skipIf(


py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -63,6 +68,128 @@ def interpret_module_to_result(
         SerializedInterpreterResult
     """
 
+    def _insert_engine_to_cache(
+        hash_val: str, interpreter_result: TRTInterpreterResult
```
narendasan: Nice, I like this

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -63,6 +68,128 @@ def interpret_module_to_result(
...
+                inputs,
+                settings,
+                interpreter_result.weight_name_map,
+                interpreter_result.requires_output_allocator,
+            ),
+        )
+        logger.info(f"Engine was successfully inserted into cache for hash: {hash_val}")
+
+    @needs_refit  # type: ignore[misc]
+    def _pull_cached_engine(hash_val: str) -> Optional[SerializedInterpreterResult]:
```
narendasan: 👍

py/torch_tensorrt/dynamo/conversion/_conversion.py
narendasan: Is there a reason that the function needs to be in the interpret functions scope?

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -63,6 +68,128 @@ def interpret_module_to_result(
...
+                interpreter_result.output_names,
+                inputs,
+                settings,
+                interpreter_result.weight_name_map,
+                interpreter_result.requires_output_allocator,
+            ),
+        )
+        logger.info(f"Engine was successfully inserted into cache for hash: {hash_val}")
+
+    @needs_refit  # type: ignore[misc]
```
narendasan: Should the insert and extract both be needs refit? 

Also shouldnt this gracefully pass through vs the typically unimplemented error?

py/torch_tensorrt/dynamo/conversion/_conversion.py
zewenli98: Not a specific reason, but I just don't know when the engine_cache will be used other than in the function `interpret_module_to_result()`. To make it safe and self-contained, I picked the smallest scope. Is there any other cases that might use engine_cache?

py/torch_tensorrt/dynamo/conversion/_conversion.py
zewenli98: > Should the insert and extract both be needs refit?

`insert` seems not to involve any refitting stuff. It supports a scenario that users insert engines on machine A that doesn't support refit but pull engines on machine B that supports refit. Please correct me if wrong.

> Also shouldnt this gracefully pass through vs the typically unimplemented error?

Not sure if I understand your question correctly. The reason why we need refit in `pull` is that we save weight-stripped engine in this implementation, which needs to be refitted to get correct weights before using.

narendasan: > Should the insert and extract both be needs refit?

If we support saving and pulling identical immutable weights engines then we dont need to guard at all, but if we are just saving immutable weights engines but never reusing them I dont see the point of supporting just insert

> Also shouldnt this gracefully pass through vs the typically unimplemented error?

The `needs_X` decorator will throw unimplemented if its hit with the feature not enabled. But in the case of refit if its not enabled then we should just bypass the cache

zewenli98: Ah I got you. fixed the two issues

py/torch_tensorrt/dynamo/conversion/_conversion.py
narendasan: Well it will get redefined each time interpret is called if its in scope, not sure its necessary unless there was some reason like it uses context from interpret module

narendasan:
@cehongwang please take a pass so we have multiple eyes on this PR

py/torch_tensorrt/dynamo/backend/backends.py
```diff
@@ -157,10 +157,6 @@ def _pretraced_backend(
                 logger.warning(
                     "require_full_compilation arg is not applicable for torch.compile with backend='torch_tensorrt"
                 )
-            if settings.strip_engine_weights:
```
narendasan, line 160: When would a torch.compile use try to use strip weights?

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -42,27 +47,168 @@ def infer_module_output_dtypes(
...
+    if not ENABLED_FEATURES.refit:
+        logger.info("Refit feature is not available, so the engine is not cached")
+        return False
+
+    # Cache the weight-stripped engine regardless of the `strip_engine_weights` setting
+    if engine_cache.check(hash_val) is not None:
+        logger.info(f"The engine already exists in cache for hash: {hash_val}")
+        return False
+
+    if not settings.strip_engine_weights:
```
narendasan: I feel like strip weights should only apply to the returned engine and not to the cache directly. So a returned cache engine with `strip weights == True` wont be refit. but you always only save stripped engine.

py/torch_tensorrt/dynamo/backend/backends.py
zewenli98: added the warning back. Not sure why strip_engine_weights arg doesn't work for torch.compile()

py/torch_tensorrt/dynamo/conversion/_conversion.py
zewenli98: The current design is to save stripped engine for using less hard disk. A returned cache engine with `strip weights == True` wont be refit as well. Only `strip weights == False` will be refit.

py/torch_tensorrt/dynamo/backend/backends.py
narendasan: It just doenst make sense. Torch compile is not serializable. So why would you ever want a callable that doesnt have the weights in it

py/torch_tensorrt/dynamo/conversion/_conversion.py
narendasan: yeah but what i mean is **every** engine we save in the cache should be a stripped weights engine right? We arent doing weight matching in the hash function for pulling from the cache so we will need to refit anyway. So the serialization config should always have `serialization_config.set_flag(trt.SerializationFlag.EXCLUDE_WEIGHTS)` regardless if the user tells us `settings.strip_engine_weights == True` or `settings.strip_engine_weights == False`

narendasan: | User Setting | Engine Saved to Cache | Engine Returned to User |
| :------- | :------: | -------: |
| settings.strip_engine_weights == True     | Weightless   | Weightless    |
| settings.strip_engine_weights == False   | Weightless   | Refit Weights  |

narendasan: Or is it that we dont need to explictly set the setting if the weight was built with stripped weights? Is there any harm in doing so? Then there is only one code path

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -42,27 +47,168 @@ def infer_module_output_dtypes(
...
+
+        for i, e in enumerate(
+            [Input.equivalent_spec(c, i) for c, i in zip(cached_engine_inputs, inputs)]
+        ):
+            assert (
+                e
+            ), f"Attempted to refit a cached engine built for a different input size (input: {i}, cached size: {cached_engine_inputs[i]}, new size: {inputs[i]}"
+
+        logger.info(
+            "Found the cached engine that corresponds to this graph. It is directly loaded."
```
narendasan: Print the hash

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -42,27 +47,168 @@ def infer_module_output_dtypes(
...
 
+    # engine_cache could be None if:
+    # 1) engine_cache is not passed in when calling this function like convert_exported_program_to_serialized_trt_engine etc., or
+    # 2) both cache_built_engines and reuse_cached_engines are False
+    if (
+        ENABLED_FEATURES.refit
+        and engine_cache is not None
+        and not settings.immutable_weights
+    ):
+        if settings.cache_built_engines or settings.reuse_cached_engines:
```
narendasan: I think this code is unclear. Would recommend something like this 

```py 

hash_val = engine_cache.get_hash(module, inputs, settings) if (settings.cache_built_engines or settings.reuse_cached_engines) else None

if settings.reuse_cached_engines:
    serialized_interpreter_result = pull_cached_engine(
        hash_val, module, engine_cache, settings, inputs
    )
    if serialized_interpreter_result is not None:  # hit the cache
        return serialized_interpreter_result

...

if (
    ENABLED_FEATURES.refit
    and not settings.immutable_weights
    and settings.cache_built_engines
    and engine_cache is not None
):
    _ = insert_engine_to_cache(
        hash_val, interpreter_result, engine_cache, settings, inputs
    )

    serialized_engine = interpreter_result.engine.serialize()
    
```

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -86,6 +232,17 @@ def interpret_module_to_result(
         f"CPU memory usage after clearing frozen parameters and building memory in conversion: {get_cpu_memory_usage()} MB"
     )
 
+    # Engine caching only for refittable engines
+    if (
+        ENABLED_FEATURES.refit
+        and not settings.immutable_weights
+        and settings.cache_built_engines
```
narendasan: We should probably throw a warning or something if `engine_cache is None and settings.cache_built_engines == True`

py/torch_tensorrt/dynamo/backend/backends.py
zewenli98: My understanding is that for torch.compile, `_pretraced_backend()` calls `compile_module()`, which is the same as the AOT to build TRT engines. If we pass in `settings.strip_engine_weights`, doesn't it tell the builder to build a weight-stripped engine?

py/torch_tensorrt/dynamo/conversion/_conversion.py
zewenli98: > Or is it that we dont need to explictly set the setting if the weight was built with stripped weights?

correct.

> Is there any harm in doing so? Then there is only one code path

At first there were multiple serialize and deserialize. For now I think there's no harm to get them in one path

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -42,27 +47,168 @@ def infer_module_output_dtypes(
...
+    engine_cache: BaseEngineCache,
+    settings: CompilationSettings,
+    inputs: Sequence[Input],
+) -> bool:
+    if not ENABLED_FEATURES.refit:
+        logger.info("Refit feature is not available, so the engine is not cached")
+        return False
+
+    # Cache the weight-stripped engine regardless of the `strip_engine_weights` setting
+    if engine_cache.check(hash_val) is not None:
```
narendasan, line 68: This should save the freshest engine

py/torch_tensorrt/dynamo/backend/backends.py
zewenli98: The reason why JIT's output is not all zeros when `strip_engine_weights=True` is that AOT and JIT generate different GM before converting to TRT engine. JIT graphs are always weightless because weights are passed by input.

AOT's weights are stored in the model:
```
graph():
    %conv1_weight : [num_users=1] = get_attr[target=conv1.weight]
    %bn1_weight : [num_users=1] = get_attr[target=bn1.weight]
    %bn1_bias : [num_users=1] = get_attr[target=bn1.bias]
    %layer1_0_conv1_weight : [num_users=1] = get_attr[target=layer1.0.conv1.weight]
    %layer1_0_bn1_weight : [num_users=1] = get_attr[target=layer1.0.bn1.weight]
    %layer1_0_bn1_bias : [num_users=1] = get_attr[target=layer1.0.bn1.bias]
    %layer1_0_conv2_weight : [num_users=1] = get_attr[target=layer1.0.conv2.weight]
    %layer1_0_bn2_weight : [num_users=1] = get_attr[target=layer1.0.bn2.weight]
    %layer1_0_bn2_bias : [num_users=1] = get_attr[target=layer1.0.bn2.bias]
    ...
    %layer4_1_bn2_running_mean : [num_users=1] = get_attr[target=layer4.1.bn2.running_mean]
    %layer4_1_bn2_running_var : [num_users=1] = get_attr[target=layer4.1.bn2.running_var]
    %convolution : [num_users=1] = call_function[target=torch.ops.aten.convolution.default](args = (%x, %conv1_weight, None, [2, 2], [3, 3], [1, 1], False, [0, 0], 1), kwargs = {})
    %_native_batch_norm_legit_no_training : [num_users=1] = call_function[target=torch.ops.aten._native_batch_norm_legit_no_training.default](args = (%convolution, %bn1_weight, %bn1_bias, %bn1_running_mean, %bn1_running_var, 0.1, 1e-05), kwargs = {})
```
but JIT uses placeholder to get the weights on the fly, so there's actually no weights to be stripped.
```
graph():
    %arg2_1 : [num_users=1] = placeholder[target=arg2_1]
    %arg3_1 : [num_users=1] = placeholder[target=arg3_1]
    %arg4_1 : [num_users=1] = placeholder[target=arg4_1]
    %arg5_1 : [num_users=1] = placeholder[target=arg5_1]
    %arg6_1 : [num_users=1] = placeholder[target=arg6_1]
    %arg7_1 : [num_users=1] = placeholder[target=arg7_1]
    %arg8_1 : [num_users=1] = placeholder[target=arg8_1]
    %arg9_1 : [num_users=1] = placeholder[target=arg9_1]
    ...
    %arg101_1 : [num_users=1] = placeholder[target=arg101_1]
    %arg102_1 : [num_users=1] = placeholder[target=arg102_1]
    %convolution : [num_users=1] = call_function[target=torch.ops.aten.convolution.default](args = (%arg1_1, %arg0_1, None, [2, 2], [3, 3], [1, 1], False, [0, 0], 1), kwargs = {})
    %_native_batch_norm_legit_no_training : [num_users=1] = call_function[target=torch.ops.aten._native_batch_norm_legit_no_training.default](args = (%convolution, %arg4_1, %arg5_1, %arg2_1, %arg3_1, 0.1, 1e-05), kwargs = {})
```

zewenli98:
The reason why JIT's output is not all zeros when `strip_engine_weights=True` is that AOT and JIT generate different GM before converting to TRT engine. JIT graphs are always weightless because weights are passed by input.

AOT's weights are stored in the model:
```
graph():
    %conv1_weight : [num_users=1] = get_attr[target=conv1.weight]
    %bn1_weight : [num_users=1] = get_attr[target=bn1.weight]
    %bn1_bias : [num_users=1] = get_attr[target=bn1.bias]
    %layer1_0_conv1_weight : [num_users=1] = get_attr[target=layer1.0.conv1.weight]
    %layer1_0_bn1_weight : [num_users=1] = get_attr[target=layer1.0.bn1.weight]
    %layer1_0_bn1_bias : [num_users=1] = get_attr[target=layer1.0.bn1.bias]
    %layer1_0_conv2_weight : [num_users=1] = get_attr[target=layer1.0.conv2.weight]
    %layer1_0_bn2_weight : [num_users=1] = get_attr[target=layer1.0.bn2.weight]
    %layer1_0_bn2_bias : [num_users=1] = get_attr[target=layer1.0.bn2.bias]
    ...
    %layer4_1_bn2_running_mean : [num_users=1] = get_attr[target=layer4.1.bn2.running_mean]
    %layer4_1_bn2_running_var : [num_users=1] = get_attr[target=layer4.1.bn2.running_var]
    %convolution : [num_users=1] = call_function[target=torch.ops.aten.convolution.default](args = (%x, %conv1_weight, None, [2, 2], [3, 3], [1, 1], False, [0, 0], 1), kwargs = {})
    %_native_batch_norm_legit_no_training : [num_users=1] = call_function[target=torch.ops.aten._native_batch_norm_legit_no_training.default](args = (%convolution, %bn1_weight, %bn1_bias, %bn1_running_mean, %bn1_running_var, 0.1, 1e-05), kwargs = {})
```
but JIT uses placeholder to get the weights on the fly, so there's actually no weights to be stripped.
```
graph():
    %arg2_1 : [num_users=1] = placeholder[target=arg2_1]
    %arg3_1 : [num_users=1] = placeholder[target=arg3_1]
    %arg4_1 : [num_users=1] = placeholder[target=arg4_1]
    %arg5_1 : [num_users=1] = placeholder[target=arg5_1]
    %arg6_1 : [num_users=1] = placeholder[target=arg6_1]
    %arg7_1 : [num_users=1] = placeholder[target=arg7_1]
    %arg8_1 : [num_users=1] = placeholder[target=arg8_1]
    %arg9_1 : [num_users=1] = placeholder[target=arg9_1]
    ...
    %arg101_1 : [num_users=1] = placeholder[target=arg101_1]
    %arg102_1 : [num_users=1] = placeholder[target=arg102_1]
    %convolution : [num_users=1] = call_function[target=torch.ops.aten.convolution.default](args = (%arg1_1, %arg0_1, None, [2, 2], [3, 3], [1, 1], False, [0, 0], 1), kwargs = {})
    %_native_batch_norm_legit_no_training : [num_users=1] = call_function[target=torch.ops.aten._native_batch_norm_legit_no_training.default](args = (%convolution, %arg4_1, %arg5_1, %arg2_1, %arg3_1, 0.1, 1e-05), kwargs = {})
```

PR #3946, opened 2025-12-03
Graph break overhead

cehongwang:
# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -1084,10 +1084,11 @@ def compile_module(
 
     output_node = list(partitioned_module.graph.nodes)[-1]
     for arg in output_node.args:
-        target = arg[0].target
-        if "_run_on_acc" not in str(target):
-            continue
-        getattr(partitioned_module, target).set_output_tensors_as_unowned(True)
+        for output in arg:
+            target = output.target
+            if "_run_on_acc" not in str(target):
+                continue
+            getattr(partitioned_module, target).set_output_tensors_as_unowned(True)
 
     # Reset settings object to user specification after fallback to global partitioning mode
     if fast_partitioner_failed:
--- a/tests/py/dynamo/runtime/test_pre_allocated_outputs.py
+++ b/tests/py/dynamo/runtime/test_pre_allocated_outputs.py
@@ -125,7 +125,7 @@ class TestPreAllocatedOutputs(TestCase):
             )
         torch._dynamo.reset()
 
-    def test_pre_allocated_outputs_unowned_outputs(self):
+    def test_pre_allocated_outputs_unowned_outputs_py_api_check_no_realloc(self):
         class SampleModel(torch.nn.Module):
             def forward(self, x):
                 return torch.softmax(x * 7 + 2, dim=0)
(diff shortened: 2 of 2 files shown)


core/runtime/TRTEngine.cpp
```diff
@@ -281,6 +289,14 @@ void TRTEngine::enable_profiling() {
   exec_ctx->setProfiler(trt_engine_profiler.get());
 }
 
+void TRTEngine::set_unowned_output_tensor(bool enable) {
+  this->unowned_output_tensor = enable;
+}
+
+bool TRTEngine::is_unowned_output_tensor() {
```
narendasan: Can we adjust these names so they are clear: 

```
TRTEngine::set_output_tensors_as_unowned(bool enable);
TRTEngine::are_output_tensors_unowned(bool enable):
```

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -218,9 +219,30 @@ def __init__(
         self.requires_output_allocator = requires_output_allocator
         self.output_allocator: Optional[DynamicOutputAllocator] = None
         self.use_output_allocator_outputs = False
-
+        self.device = torch.cuda.current_device()
+        self.cudagraphs_enabled = torch_tensorrt.runtime.get_cudagraphs_mode()
+        # If the output tensor is not owned by the engine (unowned_output_tensor=True), we need to create a new output tensor in each forward pass
+        self.unowned_output_tensor = False
```
narendasan: Same here `self.output_tensors_are_unowned`

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -218,9 +219,30 @@ def __init__(
...
         if self.serialized_engine is not None and not self.settings.lazy_engine_init:
             self.setup_engine()
+        self.is_shape_inference_io = {
+            input_name: self.engine.is_shape_inference_io(input_name)
+            for input_name in self.input_names
+        }
+
+    def set_unowned_output_tensor(self, enabled: bool) -> None:
+        """
+        Set the flag to indicate if the output tensor is unowned by the engine.
```
narendasan: To clarify the description here: 

```
Flag to set if the output tensors of this engine are wholy owned by the Torch-TensorRT Runtime or if they might be shared with a user. 

If the tensors are not owned by the runtime, then they must be recreated on every forward call which may have implications for performance.

Typically only the final engine in a graph requires output tensors to be unowned and there are performance gains to be had for intermediate engines to manage their own standing memory. 

Therefore this should only be set to True for the final module in a graph and leave false for intermediate modules.

```

core/runtime/execute_engine.cpp
```diff
@@ -145,10 +146,10 @@ void setup_input_tensors(
         // Create a new persistent input buffer
         compiled_engine->input_buffers[i] = std::move(formatted_inputs.back().clone());
       }
-
-      TORCHTRT_CHECK(
-          compiled_engine->exec_ctx->setInputShape(name.c_str(), dims), "Error while setting the input shape");
-
+      if (shape_changed) {
```
narendasan: Can't we use the same shape keys we used for cudagraphs? Why are we implementing another system?

core/runtime/TRTEngine.cpp
```diff
@@ -237,6 +237,14 @@ TRTEngine::TRTEngine(
       out_binding_names[pyt_idx] = binding_name;
     }
     num_io = std::make_pair(inputs_size, outputs);
+
+    this->current_device_id = at::cuda::current_device();
+    this->stream = c10::cuda::getCurrentCUDAStream(this->current_device_id);
```
narendasan: Is this PR supposed to implement the stream operator?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -949,6 +950,7 @@ def preserve_module_specs(
     for attr in dir(gm):
         if attr.startswith("_frozen_param"):
             delattr(gm, attr)
+    trt_module = None
```
narendasan: What is this? Why does this variable need to be outside the of scope of iterating through the module?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1070,20 +1072,28 @@ def preserve_module_specs(
...
 
     # Replace all FX Modules with TRT Modules
     for name, trt_module in trt_modules.items():
         setattr(partitioned_module, name, trt_module)
         if settings.lazy_engine_init and not settings.enable_cross_compile_for_windows:
-            getattr(partitioned_module, name).setup_engine()
+            trt_module = getattr(partitioned_module, name)
+            trt_module.setup_engine()
+
+    if trt_module:
```
narendasan: I feel like this is a pretty round about way to get the last engine

narendasan: Also the order from items is not deterministic so I dont think this is guaranteed to give you the last module anyway. You might need to work back from outputs.

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1070,20 +1072,28 @@ def preserve_module_specs(
...
+        trt_module.set_output_tensors_as_unowned(True)
 
     # Reset settings object to user specification after fallback to global partitioning mode
     if fast_partitioner_failed:
         settings.use_fast_partitioner = True
 
     dryrun_stats_display(dryrun_tracker, settings.dryrun)
+    if not settings.dryrun:
```
narendasan: Why is this conditioned on dry run? Dry run should reflect the deployment graph

setup.py
```diff
@@ -195,10 +195,10 @@ def build_libtorchtrt_cxx11_abi(
     else:
         cmd.append("//:libtorchtrt")
 
-    if develop:
-        cmd.append("--compilation_mode=dbg")
-    else:
-        cmd.append("--compilation_mode=opt")
+    # if develop:
```
narendasan: Remove

py/torch_tensorrt/dynamo/runtime/_stream_handler.py
```diff
@@ -0,0 +1,46 @@
...
+                    torch.ops.tensorrt.exit_compute_stream
+                )
+
+    partitioned_module.graph.lint()
+    partitioned_module.recompile()
+    return partitioned_module
+
+
+@torch.library.custom_op("tensorrt::enter_compute_stream", mutates_args=())
+def enter_compute_stream() -> None:
```
narendasan: IMO these should likely be dependent on the input tensors and pull their stream information from there. You arent necessarily guaranteed for the input tensors to be on the default stream. It also means you need storage for the streams. 


Also how does this work in c++?

py/torch_tensorrt/dynamo/runtime/_stream_handler.py
```diff
@@ -0,0 +1,46 @@
+import torch
+import torch.fx
+
+
+def handle_cuda_stream(
+    partitioned_module: torch.fx.GraphModule,
+) -> torch.fx.GraphModule:
+    for node in partitioned_module.graph.nodes:
+        if node.op == "placeholder":
```
narendasan: what if there are multiple placeholders or outputs?

core/runtime/execute_engine.cpp
```diff
@@ -140,15 +140,17 @@ void setup_input_tensors(
     } else {
       at::Tensor contig_input = inputs[i].view(shape).contiguous();
       formatted_inputs.emplace_back(std::move(contig_input));
-
+      bool need_cudagraphs_record = cudagraphs_enabled and
```
narendasan: Just use &&, ||, !

core/runtime/execute_engine.cpp
```diff
@@ -140,15 +140,17 @@ void setup_input_tensors(
...
+           compiled_engine->runtime_states.context_changed);
       if (need_cudagraphs_record) {
         // Create a new persistent input buffer
         compiled_engine->input_buffers[i] = std::move(formatted_inputs.back().clone());
       }
-
-      TORCHTRT_CHECK(
-          compiled_engine->exec_ctx->setInputShape(name.c_str(), dims), "Error while setting the input shape");
-
+      if (shape_changed or compiled_engine->allocated_outputs.size() == 0) {
```
narendasan: What is the allocated_outputs size test? Just that we havent allocated the outputs yet?

narendasan: Can we fold some of these conditions ahead of time?

cehongwang: So it's like for the first time you run the model or when the input shape get changed, we need to (re)set the input shape right? I don't think we can fold it

core/runtime/execute_engine.cpp
```diff
@@ -248,7 +247,11 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
       if (can_use_pre_allocated_outputs) {
         outputs = compiled_engine->pre_allocated_outputs;
       } else {
-        outputs = create_output_tensors(compiled_engine);
+        if (compiled_engine->allocated_outputs.size() == 0 or compiled_engine->output_tensors_are_unowned or
```
narendasan: Use || instead of 'or'

core/runtime/execute_engine.cpp
```diff
@@ -226,10 +226,9 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
         input_profiler_guard =
             std::make_unique<torch::autograd::profiler::RecordProfile>(compiled_engine->input_profile_path);
       }
-
-      setup_input_tensors(inputs, compiled_engine, cudagraphs_enabled, need_cudagraphs_record);
+      setup_input_tensors(inputs, compiled_engine, cudagraphs_enabled, shape_changed);
       // Check if input shapes can be inferred.
-      int32_t const io_size{compiled_engine->cuda_engine->getNbIOTensors()};
+      int32_t const io_size{compiled_engine->io_size};
```
narendasan: Are things like this actually faster? Like have we looked at the implementation to see its not cached at the TRT level?

core/runtime/execute_engine.cpp
```diff
@@ -248,7 +247,11 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
       if (can_use_pre_allocated_outputs) {
         outputs = compiled_engine->pre_allocated_outputs;
```
narendasan, line 249: What is the difference between "allocated_outputs" and "pre_allocated_outputs"?

core/runtime/TRTEngine.h
```diff
@@ -176,6 +181,7 @@ struct TRTEngine : torch::CustomClassHolder {
   std::string shape_key = "None";
   bool use_pre_allocated_outputs = false;
   std::vector<at::Tensor> pre_allocated_outputs;
+  std::vector<at::Tensor> allocated_outputs;
```
narendasan: Is there a way to reuse pre_allocated_outputs?

core/runtime/execute_engine.cpp
cehongwang: Yes they do have small difference. I am trying to minize repeating code and this repeatly call this outbound function in every iteration

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
narendasan, line 1: It seems like you are just renaming variables here? Why do we need to make that change?

core/runtime/TRTEngine.h
narendasan: @cehongwang we need to figure out how to merge these two implementations, we cant just keep adding new code paths for every situation. Either the output allocator needs to change to leverage the new system or the new system needs to leverage what the output allocator uses

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
narendasan: the `pre_allocated_outputs` -> `allocated_outputs` change

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -480,17 +502,24 @@ def run_standard_execution() -> torch.Tensor | Tuple[torch.Tensor, ...]:
                 else nullcontext()
             ):
                 if can_use_pre_allocated_outputs:
-                    outputs = self.pre_allocated_outputs
+                    outputs = self.allocated_outputs
```
narendasan: e.g. ^

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
cehongwang: I also changed `self.output_tensors = self.create_output_tensors()` to `self.allocated_outputs = self.create_output_tensors()`. Basically merged two variable together. Isn't that what you wanted me to do?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1070,14 +1070,24 @@ def preserve_module_specs(
...
 
     # Replace all FX Modules with TRT Modules
     for name, trt_module in trt_modules.items():
         setattr(partitioned_module, name, trt_module)
         if settings.lazy_engine_init and not settings.enable_cross_compile_for_windows:
-            getattr(partitioned_module, name).setup_engine()
+            trt_module = getattr(partitioned_module, name)
+            trt_module.setup_engine()
+
+    output_node = list(partitioned_module.graph.nodes)[-1]
```
narendasan, line 1085: Is this ordering gaurenteed, wouldnt it be more correct backtracing from graph outputs?

PR #3948, opened 2025-12-05
chore: Update lock file, was getting stuck and causing build issues f…

narendasan:
…or folks locally

# Description

The lock file had a stale version of torch in it and some symbols got shifted around. Caused the C++ build to use latest but in Python an older version would get pulled and prevent import. 

@lanluo-nvidia can we add that job to update the lock file regularly? 

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,8 +6,7 @@ requires = [
     "ninja>=1.11.0",
     "pyyaml>=6.0",
     "cffi>=1.15.1",
-    "torch>=2.10.0.dev,<2.11.0; platform_machine != 'aarch64' or (platform_machine == 'aarch64' and 'tegra' not in platform_release)",
-    "torch>=2.8.0,<2.9.0; platform_machine == 'aarch64' and 'tegra' in platform_release",
+    "torch>=2.10.0.dev,<2.11.0",
     "pybind11==2.6.2",
 ]
 build-backend = "setuptools.build_meta"
@@ -33,7 +32,7 @@ classifiers = [
     "Topic :: Software Development :: Libraries",
 ]
 readme = { file = "README.md", content-type = "text/markdown" }
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 keywords = [
     "pytorch",
     "torch",
@@ -100,12 +99,10 @@ index-strategy = "unsafe-best-match"
 
 [tool.uv.sources]
 torch = [
-    { index = "pytorch-nightly-cu130", marker = "platform_machine != 'aarch64' or (platform_machine == 'aarch64' and 'tegra' not in platform_release)" },
-    { index = "jetson-containers", marker = "platform_machine == 'aarch64' and 'tegra' in platform_release" },
+    { index = "pytorch-nightly-cu130" },
 ]
 torchvision = [
-    { index = "pytorch-nightly-cu130", marker = "platform_machine != 'aarch64' or (platform_machine == 'aarch64' and 'tegra' not in platform_release)" },
-    { index = "jetson-containers", marker = "platform_machine == 'aarch64' and 'tegra' in platform_release" },
+    { index = "pytorch-nightly-cu130" },
 ]
 
 [[tool.uv.index]]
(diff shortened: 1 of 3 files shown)


narendasan:
@lanluo-nvidia Also I removed the Jetson index since Thor IIRC is now on SBSAs depset, so that tegra flag will not work anyway. Makes the deps simpler but we need to revisit how we tell people how to build for Jetson Orin

setup.py
```diff
@@ -730,6 +730,7 @@ def run(self):
     "typing-extensions>=4.7.0",
     "dllist",
     "psutil",
+    "tensorrt>=10.14.1,<10.15.0",
```
lanluo-nvidia: tensorrt is intentionally not in the base requirements for two reasons:
1) DLFW build uses preinstalled tensorrt, it is tensorrt only, no tensorrt-cu12/13 related
2) gitHub build uses the tensorrt from public index, which has to be tensorrt-cu12/cu13 and tensorrt-cu12/cu13-bindings, tensorrt-cu12/cu13 libs
3) Orin build will require a different version of tensorrt
so in setup.py i dynamically add tensorrt related dependencies when it is not in DLFW build

pyproject.toml
```diff
@@ -6,8 +6,7 @@ requires = [
     "ninja>=1.11.0",
     "pyyaml>=6.0",
     "cffi>=1.15.1",
-    "torch>=2.10.0.dev,<2.11.0; platform_machine != 'aarch64' or (platform_machine == 'aarch64' and 'tegra' not in platform_release)",
```
lanluo-nvidia, line 9: Yes, build-requires can be changed to this.
We have dynamic install dependencies of torch handled in setup.py 

Latest build on Orin procedure is:
python setup.py bdist_wheel --jetpack

setup.py
```diff
@@ -744,12 +744,17 @@ def get_requirements():
...
+            requirements = requirements + [
+                "torch>=2.10.0.dev,<2.11.0"
+            ]
             if USE_TRT_RTX:
                 requirements = requirements + [
                     "tensorrt_rtx>=1.2.0.54",
                 ]
             else:
+                requirements = requirements + [
+                    "tensorrt>=10.14.1,<10.15.0",
```
narendasan: We need the meta package otherwise it will not get pulled into the env on build and users will need to manually install

narendasan: Bot now it should avoid DLFW

setup.py
narendasan: Moved it to the non DLFW section but we need it somewhere otherwise it is not installed in the env

lanluo-nvidia: Hi @narendasan,
Sorry I did not clarify the tensorrt wheel dependency clearly:
1) DLFW: no tensorrt dependency in our code due to dlfw might use a pre-installed tensorrt version which is slightly different than the one can be downloaded in public index
2) SBSA(Thor, spark etc) and Orin: tensorrt does not allow build wheel in aarch64, so it has to be installed from the tensorrt tarball, so the dependency has to be tensorrt>=10.*,*,<10.*.*  
3) Linux and Windows: have to use the wheels dependency 
   tensorrt_cu12/cu13 tensorrt_cu12/13_bindings tensorrt_cu12/13_libs 
   cannot use tensorrt>=10.*.*,<10.*.*, at that time this will automatically pull the tensorrt_cu11 on cu12 env, we end up have both tensorrt_cu11 and tensorrt_cu12 pulled in.
    so from that time, we changed to pull tensorrt_cu12/13 based on cuda version.
    
   So I think we don't need to add tensorrt dependency here

narendasan: if you dont have tensorrt here then it is missing when you import torch_tensorrt and you get an error. IMO this breaks the install ux we are going for

lanluo-nvidia:
LGTM

pyproject.toml
narendasan: ok I like pushing setup.py for orin since its more accommodating and we can be a bit more forward looking for Thor + SBSA + Windows + x86

PR #3983, opened 2025-12-25
fix: add TRT version check for engine caching feature

zewenli98:
# Description

TRT 10.14 introduces INCLUDE_REFIT in SerializationFlag

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/_conversion.py
+++ b/py/torch_tensorrt/dynamo/conversion/_conversion.py
@@ -70,16 +70,13 @@ def insert_engine_to_cache(
             f"Detected that the engine with hash: {hash_val} exists in cache. It will be refreshed"
         )
 
+    serialization_config = interpreter_result.engine.create_serialization_config()
     # for TensorRT >= 10.14, we save weight-stripped engine in cache
     if hasattr(trt.SerializationFlag, "INCLUDE_REFIT"):
-        # set EXCLUDE_WEIGHTS flag to strip weights
-        serialization_config = interpreter_result.engine.create_serialization_config()
         serialization_config.set_flag(trt.SerializationFlag.EXCLUDE_WEIGHTS)
-        serialized_engine_in_cache = interpreter_result.engine.serialize_with_config(
-            serialization_config
-        )
-    else:  # for TensorRT < 10.14, we save original engine in cache
-        serialized_engine_in_cache = interpreter_result.engine.serialize()
+    serialized_engine_in_cache = interpreter_result.engine.serialize_with_config(
+        serialization_config
+    )
 
     # Insert weight-stripped engine to cache
     engine_cache.insert(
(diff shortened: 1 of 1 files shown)


lanluo-nvidia (approved):
LGTM

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -70,18 +70,22 @@ def insert_engine_to_cache(
...
-    serialization_config.set_flag(trt.SerializationFlag.EXCLUDE_WEIGHTS)
-    weight_stripped_serialized_engine = interpreter_result.engine.serialize_with_config(
-        serialization_config
-    )
+    # for TensorRT >= 10.14, we save weight-stripped engine in cache
+    if hasattr(trt.SerializationFlag, "INCLUDE_REFIT"):
+        # set EXCLUDE_WEIGHTS flag to strip weights
+        serialization_config = interpreter_result.engine.create_serialization_config()
+        serialization_config.set_flag(trt.SerializationFlag.EXCLUDE_WEIGHTS)
+        serialized_engine_in_cache = interpreter_result.engine.serialize_with_config(
```
narendasan: Not a huge fan of taking advantage of the lack of scope conditions in python. 

Maybe something like this would be nicer 


```py

serialization_config = interpreter_result.engine.create_seralization_config()

if hasattr(trt.SeralizationFlag, "INCLUDE_REFIT"):
    serialization_config.set_flag(trt.SerializationFlag.EXCLUDE_WEIGHTS)
else:
    serialization_config.clear_flag(trt.SerializationFlag.EXCLUDE_WEIGHTS)

serialized_engine = interpreter_result.engine.serialize_with_config(serialization_config)

...
```

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -164,13 +168,17 @@ def pull_cached_engine(
...
 
-            # EXCLUDE_WEIGHTS flag must be cleared and INCLUDE_REFIT flag must be set
-            serialization_config = engine.create_serialization_config()
-            serialization_config.clear_flag(trt.SerializationFlag.EXCLUDE_WEIGHTS)
-            serialization_config.set_flag(trt.SerializationFlag.INCLUDE_REFIT)
-            serialized_engine = engine.serialize_with_config(serialization_config)
-            # Start from here, the engine is weight-included and refittable
-
+            # for TensorRT >= 10.14, we need to clear EXCLUDE_WEIGHTS flag and set INCLUDE_REFIT flag to make the engine refittable
+            if hasattr(trt.SerializationFlag, "INCLUDE_REFIT"):
```
narendasan, line 171: Same here

narendasan: We might even want to make a version safe utility function for seralization

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -214,6 +215,10 @@ def __setstate__(self, state: dict[str, Any]) -> None:
 )
 
 
+if not hasattr(trt.SerializationFlag, "INCLUDE_REFIT"):  # for TensorRT < 10.14
```
narendasan, line 218: Will this handle the case where you for some reason have different versions of TRT that have and dont have this option?

py/torch_tensorrt/dynamo/conversion/_conversion.py
zewenli98: This logic is a little different. My implementation doesn't clear EXCLUDE_WEIGHTS. Instead, it keeps the original engine, which could be weighted engine A or weightless engine B, as shown in the figure below (it‘s the design before TRT 10.14). The drawback of type B engine is that once it is refitted with weights, it would be converted to type C engine that is not refittable any more.
image

Your proposed implementation, prior to TRT 10.14, would not save B engine but save C engine so that when pulling out the engine from cache, it cannot be refitted.

My request asking TRT to add INCLUDE_REFIT has been merged into TRT 10.14, which keeps type B engine refittable after refitting with weights.

I changed the coding style to:
```python
serialization_config = interpreter_result.engine.create_serialization_config()
if hasattr(trt.SerializationFlag, "INCLUDE_REFIT"):
    serialization_config.set_flag(trt.SerializationFlag.EXCLUDE_WEIGHTS)
serialized_engine_in_cache = interpreter_result.engine.serialize_with_config(serialization_config)
```

py/torch_tensorrt/dynamo/_settings.py
zewenli98: Since the hash function also hashes engine invariant compilation_settings, the cached engine in TRT 10.12 would not hit in using TRT 10.14, and vice versa. So the cached engines cannot be shared between the different TRT versions.

narendasan: Cool

narendasan (approved):
LGTM

Issue #3972, opened 2025-12-17
🐛 [Bug] Shape mismatch when using repeat


zhaoyuanh:
##  Bug Description

```
from typing import Optional, Tuple
from contextlib import nullcontext

import torch
import torch.nn as nn
import torch_tensorrt


class CosmosLearnablePositionalEmbed(nn.Module):
    def __init__(
        self,
        hidden_size: int,
        max_size: Tuple[int, int, int],
        patch_size: Tuple[int, int, int],
        eps: float = 1e-6,
    ) -> None:
        super().__init__()

        self.max_size = [size // patch for size, patch in zip(max_size, patch_size)]
        self.patch_size = patch_size
        self.eps = eps

        self.pos_emb_t = nn.Parameter(torch.zeros(self.max_size[0], hidden_size))
        self.pos_emb_h = nn.Parameter(torch.zeros(self.max_size[1], hidden_size))
        self.pos_emb_w = nn.Parameter(torch.zeros(self.max_size[2], hidden_size))
        
    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
        batch_size, num_channels, num_frames, height, width = hidden_states.shape
        pe_size = [num_frames // self.patch_size[0], height // self.patch_size[1], width // self.patch_size[2]]

        # Use expand() instead of repeat() - torch_tensorrt compatible
        # expand() creates a view without copying data, better for dynamic shapes
        # emb_t = self.pos_emb_t[: pe_size[0]][None, :, None, None, :].expand(batch_size, -1, pe_size[1], pe_size[2], -1)
        # emb_h = self.pos_emb_h[: pe_size[1]][None, None, :, None, :].expand(batch_size, pe_size[0], -1, pe_size[2], -1)
        # emb_w = self.pos_emb_w[: pe_size[2]][None, None, None, :, :].expand(batch_size, pe_size[0], pe_size[1], -1, -1)
        emb_t = self.pos_emb_t[: pe_size[0]][None, :, None, None, :].repeat(batch_size, 1, pe_size[1], pe_size[2], 1)
        emb_h = self.pos_emb_h[: pe_size[1]][None, None, :, None, :].repeat(batch_size, pe_size[0], 1, pe_size[2], 1)
        emb_w = self.pos_emb_w[: pe_size[2]][None, None, None, :, :].repeat(batch_size, pe_size[0], pe_size[1], 1, 1)
        emb = emb_t + emb_h + emb_w
[...]

PR #4018, opened 2026-01-17
fix: expand rank mismatch on symbolic shapes; add regression test

lordaarush:
# Description

This PR fixes a failure in the Torch-TensorRT dynamo converter when lowering repeat() patterns that internally get rewritten into expand() under dynamic or symbolic shapes. In these situations, len(input_t.shape) may fail because symbolic dimensions are not standard Python sequences, leading to errors such as:

ValueError: __len__() should return >= 0

This was triggered by positional embedding patterns like:

x[..., None, None, :].repeat(batch, 1, H, W, 1)

Root cause:
After padding singleton dimensions, the converter asserted:
    assert len(input_t.shape) == shape_rank
However, calling len() on symbolic shapes can fail or return invalid results, causing the converter to raise Python-level errors during lowering.

Fix:
Replaces the unsafe assert with a robust rank check that:
- attempts to read the rank of input_t.shape safely,
- falls back to the known initial_tensor_rank if len() is not supported, and
- emits a clear RuntimeError if ranks still mismatch.

This preserves all existing expand() semantics. There are no behavior changes for static shapes.

A minimal regression test has been added at:
tests/dynamo/test_repeat_expand_repro.py
It reproduces the issue in #3972 and is skipped automatically when CUDA + Torch-TensorRT are unavailable.

Fixes: #3972

## Type of change

# Checklist:

- [x] I have commented my code where needed
- [x] New and existing unit tests pass locally where applicable (test is skipped when Torch-TensorRT runtime is unavailable)
- [x] I have added the relevant labels to my PR so that reviewers are notified

### DIFF
--- a/tests/dynamo/test_repeat_expand_repro.py
+++ b/tests/dynamo/test_repeat_expand_repro.py
@@ -1,0 +1,47 @@
+import pytest
+import torch
+import torch.nn as nn
+
+try:
+    import torch_tensorrt
+except Exception:
+    torch_tensorrt = None
+
+REQUIRES_TRT = torch.cuda.is_available() and (torch_tensorrt is not None)
+
+pytestmark = pytest.mark.skipif(not REQUIRES_TRT, reason="requires CUDA + Torch-TensorRT runtime")
+
+class CosmosLearnablePositionalEmbed(nn.Module):
+    def __init__(self, hidden_size, max_size, patch_size):
+        super().__init__()
+        self.patch_size = patch_size
+        self.pos_emb_t = nn.Parameter(torch.zeros(max_size[0] // patch_size[0], hidden_size))
+        self.pos_emb_h = nn.Parameter(torch.zeros(max_size[1] // patch_size[1], hidden_size))
+        self.pos_emb_w = nn.Parameter(torch.zeros(max_size[2] // patch_size[2], hidden_size))
+
+    def forward(self, hidden_states):
+        batch_size, _, num_frames, height, width = hidden_states.shape
+        pe_size = [num_frames // self.patch_size[0], height // self.patch_size[1], width // self.patch_size[2]]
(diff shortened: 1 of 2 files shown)


tests/dynamo/test_repeat_expand_repro.py
apbose, line 1: Thanks for the PR. I don't think the above fix addresses the issue. Since dynamic shapes are already handled in prepend_ones. Also current_rank should now be the shape it is expanded to.

```
try:
        current_rank = len(input_t.shape)
except Exception:
        current_rank = shape_rank
```

In the below test case there will be 0 computational nodes that will depend on runtime input, since the shape values will be constant. You could make them dynamic to invoke the converter.

I went through the above issue and looks like the root issue is dims being 10 here which is not permitted in TRT. It can handle till 8 max dims- https://docs.nvidia.com/deeplearning/tensorrt/latest/_static/c-api/classnvinfer1_1_1_dims64.html

Pytorch decomposes repeat to unsqueeze-> expand -> permute -> reshape

But for 5D tensor `layer.reshape_dims = new_shape` fails here, since DIMS can't be 10 here, and it fails in the dynamic case too in `layer.set_input(1, reshape_dim_layer.get_output(0))`. Hence input_tensor.shape would come invalid.

The original example would work with only 4 dimensions. Instead of `emb_t = self.pos_emb_t[: pe_size[0]][None, :, None, None, :].repeat(batch_size, 1, pe_size[1], pe_size[2], 1)` 5 dims here.

WAR would be to replace repeat with expand without broadcasting the dimension or using tile operation. I need to look into this more.

Ideally for tests below we would want the test case to be in `https://github.com/pytorch/TensorRT/tree/main/tests/py/dynamo/conversion` if it is a converter fix, so below location would not work.

lordaarush: Thanks for reviewing this, looks like I misunderstood what was actually failing and ended up fixing the wrong thing. I see the issue much more clearly now. I’ll take another pass at it with the correct context and follow up if I find something useful.

Issue #4008, opened 2026-01-13
[aten::embedding] Do we still need this check?


narendasan:
https://github.com/pytorch/TensorRT/blob/b6d49d088c323f08deaf58a2e613d9328582e1e5/py/torch_tensorrt/dynamo/conversion/impl/embedding.py#L33

Doesnt TensorRT support INT64 I/O now?

apbose:
https://github.com/pytorch/TensorRT/pull/4023 . Shows this error - IBuilder::buildEngineWithConfig: Error Code 4: API Usage Error ([SCATTER]-[aten_ops.scatter.value]-[scatter_value_scatter_layer]: Indices tensor must be Int32, input type is Int64. In validateTypes at /_src/optimizer/api/layers/scatterLayer.cpp:39)

PR #4023, opened 2026-01-21
remove downcast of int64 to int32

apbose:
Addresses https://github.com/pytorch/TensorRT/issues/4008


zewenli98 (commented):
LGTM. Is it still a draft or good to go?

Issue #4050, opened 2026-01-27
↔ [Converter] Add dynamic shape support for torch.ops.aten.scatter.src


zewenli98:
We currently only support static shape for `torch.ops.aten.scatter.src`. Dynamic shape should be supported as well.

apbose:
PR #4062 addressed this. Closing

PR #4062, opened 2026-01-30
scatter.src and scatter.value dynamic case

apbose:
https://github.com/pytorch/TensorRT/issues/4050 addresses this


apbose:
Fails-
 Build and test Linux x86, windows wheels- RTX
Build and test windows wheels
Linux x86- L2 torch compile and L2 dynamo distributed
These seem unrelated to the PR above

narendasan:
@apbose did someone review this? why do we need to change the meta op?

PR #4089, opened 2026-02-20
fix failures- cpu offloading casues device mismatch L2_dynamo_compile…

apbose:


py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1454,11 +1446,13 @@ def convert_exported_program_to_serialized_trt_engine(
             f"Conversion of module {gm} not currently fully supported or convertible!",
             exc_info=True,
         )
+        raise
```
narendasan: If we are throwing exceptions, provide an actual error type and explanation

tests/py/dynamo/models/test_models_export.py
```diff
@@ -386,6 +386,7 @@ def calibrate_loop(model):
                 exp_program,
                 inputs=[input_tensor],
                 enabled_precisions={torch.float8_e4m3fn},
+                use_explicit_typing=False,
```
narendasan: This is deprecated, this test should be ported to explicit typing

py/torch_tensorrt/dynamo/_compiler.py
apbose: The bare raise should be able to catch the same exception object or the UnsupportedOperatorException. Do we need another exception message. That would be code repeat. But can add.

apbose:
@narendasan had discussion with @zewenli98. Should we remove this https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/_compiler.py#L641-648 from the code, since explicit typing means that we should not provide precision now.
As for porting the test to explicit typing, that means is it meant that we should remove the `enabled_precisions={torch.float8_e4m3fn}` from test?

lanluo-nvidia (approved):
LGTM

PR #4104, opened 2026-02-28
converter: add sdpa, flash-sdpa, efficient-sdpa, and cudnn-sdpa converters

zewenli98:
# Description

This PR adds sdpa, flash-sdpa, efficient-sdpa, and cudnn-sdpa converters, which are implemented with TRT IAttention Layer. These 4 sdpa ops will not be decomposed. 
The test cases only cover flash-sdpa and efficient-sdpa because I didn't find scenarios that Pytorch transforms the op to sdpa and cudnn-sdpa.

Replace the PR #3933

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -3794,9 +3794,38 @@ def aten_ops_linear(
     )
 
 
+def scaled_dot_product_attention_validator(
+    node: Node, settings: Optional[CompilationSettings] = None
+) -> bool:
+    if node.kwargs.get("enable_gqa", False):
+        _LOGGER.debug(
+            "enable_gqa is not yet supported by the converter. Please try setting decompose_attention=True in the compilation settings."
+        )
+        return False
+
+    query_shape, key_shape, value_shape = None, None, None
+    if "val" in node.args[0].meta:
+        query_shape = node.args[0].meta["val"].size()
+    if "val" in node.args[1].meta:
+        key_shape = node.args[1].meta["val"].size()
+    if "val" in node.args[2].meta:
+        value_shape = node.args[2].meta["val"].size()
+    if (
+        query_shape != key_shape
+        or query_shape != value_shape
+        or key_shape != value_shape
+    ):
+        _LOGGER.debug(
+            "query, key, and value have different shapes. Please try setting decompose_attention=True in the compilation settings."
+        )
+        return False
+    return True
+
+
 @dynamo_tensorrt_converter(
     torch.ops.aten.scaled_dot_product_attention.default,
     supports_dynamic_shapes=True,
+    capability_validator=scaled_dot_product_attention_validator,
 )
 def aten_ops_scaled_dot_product_attention(
     ctx: ConversionContext,
@@ -3822,9 +3851,36 @@ def aten_ops_scaled_dot_product_attention(
     )
 
 
+def scaled_dot_product_flash_attention_validator(
+    node: Node, settings: Optional[CompilationSettings] = None
+) -> bool:
+    if args_bounds_check(node.args, 5, False):
+        _LOGGER.debug("return_debug_mask is not yet supported.")
+        return False
+
+    query_shape, key_shape, value_shape = None, None, None
+    if "val" in node.args[0].meta:
+        query_shape = node.args[0].meta["val"].size()
+    if "val" in node.args[1].meta:
+        key_shape = node.args[1].meta["val"].size()
+    if "val" in node.args[2].meta:
+        value_shape = node.args[2].meta["val"].size()
+    if (
+        query_shape != key_shape
+        or query_shape != value_shape
+        or key_shape != value_shape
+    ):
+        _LOGGER.debug(
+            "query, key, and value have different shapes. Please try setting decompose_attention=True in the compilation settings."
+        )
+        return False
+    return True
+
+
 @dynamo_tensorrt_converter(
     torch.ops.aten._scaled_dot_product_flash_attention.default,
     supports_dynamic_shapes=True,
+    capability_validator=scaled_dot_product_flash_attention_validator,
 )
 def aten_ops_scaled_dot_product_flash_attention(
     ctx: ConversionContext,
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -447,139 +447,6 @@ def view_decomposition(x: torch.Tensor, size: List[torch.SymInt]) -> torch.Tenso
     return aten._reshape_copy.default(x, size)
 
 
-@register_torch_trt_decomposition(
```
narendasan, line 450: Can we keep the decomp code and gate it behind a flag? like decompose_attention?

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
```diff
@@ -146,6 +146,13 @@ def _is_subnode_dynamic(subnode: torch.fx.Node) -> bool:
             meta_val = subnode.meta["val"]
             if isinstance(meta_val, (list, tuple)):
                 for val in meta_val:
+                    if val is None:
+                        continue
+                    if isinstance(val, (SymFloat, SymInt, SymBool)):
```
narendasan, line 151: What is the distinction between this case and the one below?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3792,3 +3792,115 @@ def aten_ops_linear(
...
+    supports_dynamic_shapes=True,
+)
+def aten_ops_scaled_dot_product_flash_attention(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    # _scaled_dot_product_flash_attention(Tensor query, Tensor key, Tensor value, float dropout_p=0.0, bool is_causal=False, bool return_debug_mask=False, *, float? scale=None) -> (Tensor output, Tensor logsumexp, Tensor cum_seq_q, Tensor cum_seq_k, SymInt max_q, SymInt max_k, Tensor rng_state, Tensor unused, Tensor debug_attn_mask)
```
narendasan, line 3892: Do we support all flags here? is causal? return debug mask scale?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3792,3 +3792,115 @@ def aten_ops_linear(
...
+    return impl.attention.scaled_dot_product_flash_attention(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        query=args[0],
+        key=args[1],
+        value=args[2],
+        dropout_p=args_bounds_check(args, 3, 0.0),
```
narendasan, line 3901: Should these arg bounds check be in a validator?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3792,3 +3792,115 @@ def aten_ops_linear(
...
+    return impl.attention.scaled_dot_product_attention(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        query=args[0],
+        key=args[1],
+        value=args[2],
+        attn_mask=args_bounds_check(args, 3, None),
+        dropout_p=args_bounds_check(args, 4, 0.0),
```
narendasan, line 3847: Same thing here

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
narendasan: Are there any cases where we might want to fall back to decomp? This can be a 2.12 feature but I am thinking basically if we cant run via IAttn layer then we could decompose and run as lower level ops

py/torch_tensorrt/dynamo/conversion/impl/attention.py
```diff
@@ -0,0 +1,425 @@
...
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+]:
+    assert return_debug_mask is False, "return_debug_mask is not supported yet"
```
narendasan: Validator for this!

py/torch_tensorrt/dynamo/conversion/impl/attention.py
```diff
@@ -0,0 +1,425 @@
...
+    query: TRTTensor,
+    key: TRTTensor,
+    value: TRTTensor,
+    attn_bias: Optional[TRTTensor] = None,
+    compute_log_sumexp: bool = False,
+    dropout_p: float = 0.0,
+    is_causal: bool = False,
+    scale: Optional[float] = None,
+) -> Tuple[TRTTensor, Optional[TRTTensor], Optional[TRTTensor], Optional[TRTTensor]]:
+    assert compute_log_sumexp is False, "compute_log_sumexp is not supported yet"
```
narendasan: Validator!

narendasan: The validator should tell users to use decompose_attention

py/torch_tensorrt/dynamo/conversion/impl/attention.py
```diff
@@ -0,0 +1,425 @@
...
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+    Optional[TRTTensor],
+]:
+    assert return_debug_mask is False, "return_debug_mask is not supported yet"
```
narendasan: Validator

py/torch_tensorrt/dynamo/conversion/impl/attention.py
```diff
@@ -0,0 +1,425 @@
...
+        attn_mask (Optional[TRTTensor], optional): Attention mask tensor. The shape must be broadcastable to the shape of attention weights. Two types of masks are supported. A boolean mask where a value of True indicates that the element should take part in attention. A float mask of the same type as query, key, value that is added to the attention score. If is_causal is set to True, attn_mask should be None.
+        dropout_p (float, optional): Ignored in inference stage
+        is_causal (bool, optional): Whether to apply causal masking. If True, the attention mask will be a lower triangular mask. If is_causal is set to True, the attn_mask argument should be None.
+        scale (Optional[float], optional): Scaling factor for the attention layer
+        enable_gqa (bool, optional): Whether to enable Group Query Attention (GQA). If True, the query tensor will be split into groups and the attention will be computed for each group.
+
+    Returns:
+        TRTTensor: Attention output tensor with shape [batch, heads, seq_len, head_dim]
+    """
+    assert enable_gqa is False, "enable_gqa is not supported yet"
```
narendasan: Validator + tell users to use decompose_attn until we can auto decompose

py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
zewenli98: This one checks if `val` itself is a SymVal, while the below one checks `dim` of `val`

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: 1. Thanks for the reminder. I'll put all unsupported args into validator.
2. Since many returns are something like intermediate results which are not returned by IAttn Layer, we return None as placeholders.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zewenli98: Yeah I'm also thinking about it. We can have a mechanism to go through converters first (assumably faster than a collection of decomposed low-level ops), and then if any op is not supported in converters, we can check if it can be decomposed

narendasan: Yeah we can do like an early call the to validator library (like in the pre_aten_lowering phase) that only runs the attn validator maybe and then triggers the decomp

zewenli98:
> I think this is mostly looking good, was there anything you wanted me to specifically look at otherwise LGTM

nope. I think it's good to go

narendasan:
Ok please ping @lanluo-nvidia to merge and cherry pick

PR #4108, opened 2026-03-03
feat: Add TensorRT Edge-LLM AttentionPlugin backend support

chohk88:
Add plugin backend as an alternative to the default SDPA lowering for LLM inference, providing ~1.7x-3.3x speedup over SDPA and ~8x-11x over PyTorch eager execution.

Supported Models: Llama 3.x (3.1/3.2), Qwen 2.5, Qwen 3

Changes:
- examples/dynamo/attention_plugin_example.py: Standalone plugin demo with correctness validation against PyTorch SDPA
- examples/dynamo/end_to_end_llm_generation_example.py: End-to-end LLM generation example with plugin integration and benchmarks
- tools/llm/plugin_utils.py: Model-agnostic plugin utilities including op registration (tensorrt_edge_llm::xqa_attn), TensorRT converter, PluginAttention module, LLMPluginWrapper, compilation and generation
- tools/llm/run_llm.py: Add --backend plugin/sdpa selection with plugin workflow integration
- tools/llm/README.md: Plugin backend documentation with build guide, usage examples, and performance summary

Plugin library built from TensorRT-Edge-LLM 0.4.0: https://github.com/chohk88/TensorRT-Edge-LLM/tree/feature/torch-tensorrt-python-runtime

# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/tools/llm/README.md
+++ b/tools/llm/README.md
@@ -8,11 +8,12 @@ This directory provides utilities and scripts for compiling, optimizing, and ben
 - **VLM Support:** Supports Visual Language Models like Qwen2.5-VL and Eagle2.
 - **Precision Modes:** Supports FP16, BF16, and FP32.
 - **Multiple Backends:**
-  - **SDPA Backend** (default): Registers custom lowering pass for SDPA operations, enabling TensorRT conversion with optional static KV cache support
-  - **Plugin Backend**: Uses TensorRT Edge-LLM attention plugin for optimized inference with built-in KV cache management
-- **KV Cache:** Supports static and dynamic KV cache for efficient autoregressive decoding.
+  - **SDPA Backend** (`--backend sdpa`, default): Registers custom lowering pass for SDPA operations, converting attention to matmul+softmax+matmul with optional static KV cache support (`--cache static_v1`/`static_v2`)
+  - **IAttention Backend** (`--backend iattention`): Uses TensorRT's native `IAttention` layer for attention conversion. Single-pass inference only (KV cache not yet supported with this backend)
+  - **Plugin Backend** (`--backend plugin`): Uses TensorRT Edge-LLM attention plugin for optimized inference with built-in KV cache management
+- **KV Cache:** Supports static KV cache for efficient autoregressive decoding (SDPA and Plugin backends).
+- **Quantization:** Supports FP8 and NVFP4 quantization formats for reduced memory usage and improved inference speed.
 - **Benchmarking:** Measures and compares throughput and latency for PyTorch and TensorRT backends.
-- **Custom Attention:** Registers and converts custom scaled dot-product attention (SDPA) for compatibility with TensorRT.
 
 
 ### Supported Models
@@ -62,6 +63,12 @@ TensorRT model generated text:  What is parallel programming? Parallel programmi
 PyTorch and TensorRT outputs match: True
 ```
 </details>
+
+*IAttention Backend (native TRT IAttention layer, no KV cache):*
+```bash
+python run_llm.py --model meta-llama/Llama-3.2-1B-Instruct --backend iattention \
+  --prompt "What is parallel programming?" --model_precision FP16 --num_tokens 30 --enable_pytorch_run
+```
 
 *Plugin Backend:*
 ```bash
@@ -115,7 +122,10 @@ python run_vlm.py --model nvidia/Eagle2-2B --precision FP16 --cache static_v1 --
 **Model Configuration:**
 - `--model`: Name or path of the HuggingFace LLM/VLM.
 - `--tokenizer`: (Optional) Tokenizer name; defaults to model.
-- `--backend`: Backend to use (`sdpa` or `plugin`). Default is `sdpa`. Only applicable for LLM models.
+- `--backend`: Backend to use (`sdpa`, `iattention`, or `plugin`). Default is `sdpa`. Only applicable for LLM models.
+  - `sdpa`: Custom SDPA lowering pass + converter. Supports `--cache static_v1`/`static_v2` for KV caching.
+  - `iattention`: TensorRT native IAttention layer. No KV cache support yet (single-pass inference only).
+  - `plugin`: TensorRT Edge-LLM attention plugin. KV cache managed internally by the plugin.
 
 **Generation Settings:**
 - `--prompt`: Input prompt for generation (generation mode only, ignored in benchmark mode).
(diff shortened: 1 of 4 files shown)


tools/llm/README.md
```diff
@@ -129,4 +265,4 @@ This codebase can be extended to
   - `pip install qwen-vl-utils` (for Qwen2.5-VL-3B-Instruct model)
   - **Flash Attention**: For models using flash attention operations (e.g., Eagle2-2B), install one of the following:
     - **Fast installation (recommended)**: `pip install flash-attn==2.8.1` (pre-built wheel, should work)
-    - **Source build (slow)**: `pip install flash-attn --no-build-isolation -v` (fallback if pre-built wheels fail)
\ No newline at end of file
+    - **Source build (slow)**: `pip install flash-attn --no-build-isolation -v` (fallback if pre-built wheels fail)
```
narendasan: Make sure to add the `MAX_JOBS=8` otherwise you might take peoples systems down

tools/llm/run_llm.py
```diff
@@ -322,54 +365,116 @@ def measure_perf(trt_model, input_signature, backend_name):
...
+                        input_ids.shape[1],
+                        args.num_tokens,
+                        max_seq_len,
+                        DEVICE,
+                        dtype,
+                    )
+                    trt_timings.append(elapsed_ms / 1000.0)
+        else:
+            # SDPA backend (default)
+            if args.cache == "static_v1":
```
narendasan, line 457: We have a few threads, backend and cache and with @zewenli98's PR in core Attention. can we merge these settings so its easy to understand when you will get TRT-Edge-LLM, when you get native IAttention and when you get Static KV Cache?

tools/llm/run_llm.py
```diff
@@ -35,6 +41,24 @@
     time_generate,
 )
 
+# Import plugin utilities (optional)
+try:
+    from plugin_utils import (
```
narendasan: Where does plugin_utils come from?

narendasan (commented):
Overall, I think its close. @zewenli98 should take a pass but we can merge near as is. but I want to think about next how we might create lowering passes that insert the placeholder ops programmatically. Evan is about to disable decomposition by default for sdpa so we can basically dynamically insert a pass that keys on those ops

tools/llm/plugin_utils.py
```diff
@@ -0,0 +1,952 @@
...
+):
+    _register_plugin_op_impl()
+
+
+# -----------------------------------------------------------------------------
+# TensorRT Converter
+# -----------------------------------------------------------------------------
+
+
+@dynamo_tensorrt_converter(
```
narendasan: Lets put all the converters for our edgellm ops in their own file

narendasan (commented):
Do we have lowering passes to insert the tensorrt edge llm ops in place of pytorch ops?

tools/llm/plugin_utils.py
```diff
@@ -0,0 +1,952 @@
...
+        dynamic_shapes=dynamic_shapes,
+        strict=False,
+    )
+
+    # Compile
+    trt_model = torch_tensorrt.dynamo.compile(
+        ep,
+        inputs=[dummy_input_ids, dummy_pos_ids, dummy_kvs, dummy_ctx_len],
+        enabled_precisions={torch.float32},
+        use_explicit_typing=True,
```
zewenli98: when `use_explicit_typing` is true, `enabled_precisions` should be removed.

tools/llm/plugin_utils.py
```diff
@@ -0,0 +1,952 @@
...
+        use_fp32_acc=True,
+        device=device,
+        disable_tf32=True,
+        min_block_size=1,
+        debug=debug,
```
zewenli98: `debug` is deprecated. Please use `with torch_tensorrt.dynamo.Debugger(...)`

tools/llm/README.md
```diff
@@ -7,7 +7,9 @@ This directory provides utilities and scripts for compiling, optimizing, and ben
 - **Model Support:** Works with popular LLMs such as Llama-3, Qwen2.5, etc.
 - **VLM Support:** Supports Visual Language Models like Qwen2.5-VL and Eagle2.
 - **Precision Modes:** Supports FP16, BF16, and FP32.
-- **Quantization:** Supports FP8 and NVFP4 quantization formats for reduced memory usage and improved inference speed.
```
zewenli98, line 10: should we keep quant?

tools/llm/run_llm.py
zewenli98: @chohk88 I implemented the converters for some attention variants in https://github.com/pytorch/TensorRT/pull/4104. Can you take a look how to integrate?

tools/llm/README.md
chohk88: Updated. Thanks for the suggestion!

tools/llm/run_llm.py
chohk88: @narendasan @zewenli98 Thanks for the reviews! 

Now, `--backend` supports three options — `sdpa`, `iattention`, and `plugin` and update it in README on what each provides:
- `sdpa`: SDPA lowering pass + optional `--cache static_v1/v2`
- `iattention`: TRT native IAttention converters from #4104 (no KV cache yet)
- `plugin`: Edge-LLM attention plugin with built-in KV cache


Additionally, I've added `--backend iattention` and it works correctly (outputs match PyTorch for all tested models). However, I found that the TRT `IAttention` layer is never actually used with HF models — HF always passes `attn_bias` (causal mask tensor) to `_scaled_dot_product_efficient_attention`, which causes the converter to take the manual matmul+softmax decomposition path instead of `ctx.net.add_attention()`. Filed #4129 to track this.

tools/llm/plugin_utils.py
chohk88: Updated. Thanks for the suggestion!

tools/llm/plugin_utils.py
chohk88: Updated. Thanks for the suggestion!

tools/llm/plugin_utils.py
chohk88: Updated. Thanks for the suggestion!

tools/llm/README.md
chohk88: That was my mistake! Thanks for the suggestion!

chohk88:
> Do we have lowering passes to insert the tensorrt edge llm ops in place of pytorch ops?

Not yet. Currently the plugin backend does module-level replacement before `torch.export`: it iterates over the HF model's decoder layers (`model.model.layers[i].self_attn`) and swaps each attention module with a `PluginAttention` that calls the Edge-LLM custom op.

Implementing this as an FX lowering pass instead would require pattern-matching the full `Q/K/V projection → reshape → RoPE → SDPA` op sequence in the exported graph, which is non-trivial since these patterns differ across architectures (Qwen, Llama, Gemma, etc.).

chohk88:
@narendasan @zewenli98 I ran correctness and benchmark tests across all backend × model combinations to verify the changes work as expected. All backends produce outputs identical to PyTorch, and performance numbers are in a reasonable range. Note that the specific performance numbers in the README have been removed since these are not from an official benchmarking setup — the README now describes approximate speedup ranges instead.

### Correctness (PyTorch vs TensorRT output match)

| Model | sdpa (no cache) | sdpa (static_v1) | plugin | iattention |
|-------|:-:|:-:|:-:|:-:|
| Qwen2.5-0.5B | ✅ | ✅ | ✅ | ✅ |
| Qwen3-0.6B | ✅ | ✅ | ✅ | ✅ |
| Llama-3.2-1B | ✅ | ✅ | ✅ | ✅ |

### Benchmark (Median Latency ms, A100 80GB, FP16, ISL=2048, OSL=128, Batch=1)

| Model | PyTorch | sdpa (no cache) | sdpa (static_v1) | plugin | iattention |
|-------|-------:|-------:|-------:|-------:|-------:|
| Qwen2.5-0.5B | 4751 | 3271 | 1238 | **421** | 5421 |
| Qwen3-0.6B | 6875 | 4031 | 1708 | **569** | 6792 |
| Llama-3.2-1B | 7053 | 5466 | 1379 | **465** | 8283 |

> **Note on iattention**: Currently on par with or slightly slower than PyTorch for autoregressive generation because the TRT `IAttention` layer is not actually used — HF passes `attn_bias` (causal mask) to `efficient_attention`, causing the converter to take the manual decomposition path. Tracked in #4129

PR #4123, opened 2026-03-06
split index.Tensor converter for bool vs int indexing

wenbingl:
# Description

Add `index_has_bool_indices` validator and register two separate converters for `torch.ops.aten.index.Tensor`:
- Integer indexing (HIGH priority, no output allocator): output shape is deterministic based on index tensor shape.
- Boolean indexing (requires output allocator): uses nonzero() internally, producing data-dependent output shapes.

Fixes # (issue)
support cuda-graph feature for some beam search based model which failed due to output-allocator limit

## Type of change

# Checklist:

### DIFF
--- a/tests/py/dynamo/conversion/test_index_bool_split_aten.py
+++ b/tests/py/dynamo/conversion/test_index_bool_split_aten.py
@@ -1,0 +1,165 @@
+"""Tests for the bool/int index.Tensor converter split (commit b91168b).
+
+Verifies that:
+1. `index_has_bool_indices` validator correctly distinguishes bool vs int indices.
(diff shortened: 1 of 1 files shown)


wenbingl:
@narendasan , please take a look whether this PR is needed, for cuda-graph.

narendasan (commented):
Cool, I think this is a totally reasonable approach to WAR DDS cases. One thing that came to mind is if there is a way to make this validator generic because I think it makes sense to have "different" converters which use a the same internal defensive implementation but may have different restrictions with different codepaths like dynamic shape, dds, etc. to register multiple times and use validators to distinguish the restrictions. So a validator could take an arg index and a type allow list and deny list and pass or fail the converter @zewenli98 @apbose what do you think 

Maybe a API similar to enforce_tensors

PR #4124, opened 2026-03-07
handle symbolic shape for non tensor inputs in symbolic shape extraction

apbose:
This PR handles-
1. SymInt/scalar inputs in `_symbolic_shape_capture.py` 
2. `inputShapeTensorValues` was a local variable in `setup_input_tensors()`. When the function returned, the CPU buffer pointers registered with TRT via `setTensorAddress()` became dangling. `inferShapes()` and `enqueueV3() `then read garbage from freed memory, producing nonsensical reshape dimensions.

### DIFF
--- a/core/runtime/execute_engine.cpp
+++ b/core/runtime/execute_engine.cpp
@@ -96,9 +96,8 @@ void setup_input_tensors(
     std::vector<at::Tensor> inputs,
     c10::intrusive_ptr<TRTEngine> compiled_engine,
     bool cudagraphs_enabled,
-    bool need_cudagraphs_record) {
-  // this is a buffer to store shape tensor input addresses throughout the runtime scope
-  std::list<std::vector<int64_t>> inputShapeTensorValues;
+    bool need_cudagraphs_record,
+    std::list<std::vector<int64_t>>& inputShapeTensorValues) {
   std::list<at::Tensor> formatted_inputs(compiled_engine->num_io.first);
 
   for (size_t i = 0; i < inputs.size(); i++) {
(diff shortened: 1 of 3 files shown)


narendasan (commented):
@apbose can you add a test case for this?

narendasan:
@lanluo-nvidia can you merge and cherry-pick this for 2.11?

core/runtime/execute_engine.cpp
```diff
@@ -233,6 +230,9 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
 
     std::vector<at::Tensor> outputs(compiled_engine->num_io.second);
 
+    // Shape tensor CPU buffers must outlive inferShapes() and enqueueV3()
```
narendasan, line 236: Can you explain why? Would help others understand (basically afaict, its just that these inputs are not provided by torch and so we need to produce them ourselves and ensure they are available through enqueue)

narendasan (approved):
Think this looks mostly good, we can merge if its urgent otherwise documenting the reason for the C++ change would be good

core/runtime/execute_engine.cpp
apbose: Yeah the jist is that. Shape tensor values require CPU memory buffers whose addresses are registered with TRT via `setTensorAddress()`. Unlike regular tensor inputs (whose GPU memory is reference-counted by PyTorch and stays alive via the caller's tensor references), shape tensor values are copied into `std::vector<int64_t>` buffers that we allocate ourselves (I should restore back the comment, will do that). TRT holds raw pointers to these buffers and reads from them during` inferShapes()` and `enqueueV3()`. Previously, these buffers were local to `setup_input_tensors()` and were freed on return. Moving the declaration here ensures the buffers outlive both calls.

PR #4131, opened 2026-03-17
feat: support `attn_bias` for efficient SDPA

### LINKED ISSUE #4129: ↔ [Converter] IAttention converter bypasses TRT native IAttention layer due to HuggingFace causal mask (attn_bias)
### Summary

When using the `--backend iattention` option in `tools/llm/run_llm.py`, the TensorRT native `IAttention` layer (`ctx.net.add_attention()`) is **never used**. Instead, the `scaled_dot_product_efficient_attention` converter falls back to manual matmul+softmax+matmul decomposition for every attention layer. This results in **no speedup** (or even **slight slowdown**) compared to PyTorch eager for autoregressive generation.

### Root Cause

HuggingFace's SDPA attention implementation always generates a causal mask tensor and passes it as `attn_bias` to `aten._scaled_dot_product_efficient_attention`. The converter in `py/torch_tensorrt/dynamo/conversion/impl/attention.py` has two code paths:

```python
def scaled_dot_product_efficient_attention(..., attn_bias=None, ...):
    if attn_bias is not None:
        # SLOW PATH: manual matmul → add bias → softmax → matmul
        attn_weight = matmul(scaled_query, key^T)
        attn_weight = add(attn_weight, attn_bias)
        attn_weight = softmax(attn_weight)
        out = matmul(attn_weight, value)
        return out, None, None, None
    else:
        # FAST PATH: TRT native IAttention layer
        attention_layer = ctx.net.add_attention(scaled_query, key, value, SOFTMAX, is_causal)
        return attention_layer.get_output(0), None, None, None
```

Since HuggingFace **always** passes `attn_bias` (a causal mask tensor), the fast path is never taken.

### Evidence (debug log analysis)

Running with `--debug` on Qwen2.5-0.5B-Instruct (24 layers):

```
$ grep -c "attn_bias_add" debug_stderr.log
24                    # ← manual decomposition used for ALL 24 layers

$ grep -c "add_attention\|IAttention" debug_stderr.log
0                     # ← TRT IAttention layer NEVER used
```

### Benchmark Results

Tested on NVIDIA A100 80GB PCIe, FP16, ISL=2048, OSL=128, Batch=1:

| Model | PyTorch (ms) | sdpa no cache (ms) | iattention (ms) | sdpa static_v1 (ms) | plugin (ms) |
|-------|---:|---:|---:|---:|---:|
| Qwen2.5-0.5B | 4751 | 3271 | 5421 | **1238** | **421** |
| Qwen3-0.6B | 6875 | 4031 | 6792 | **1708** | **569** |
| Llama-3.2-1B | 7053 | 5466 | 8283 | **1379** | **465** |

- `iattention` is **slower** than `sdpa no cache` despite both lacking KV cache
- `sdpa no cache` is faster because the SDPA lowering pass replaces `attn_bias` with `is_causal=True` and dynamically generates the mask, which is more efficient
- `iattention` keeps the HF-provided `attn_bias` tensor, taking the slow manual decomposition path

### Why sdpa backend doesn't have this problem

The SDPA lowering pass in `tools/llm/torchtrt_ext/register_sdpa.py` transforms the graph:

```
BEFORE (HF output):
  aten._scaled_dot_product_efficient_attention(Q, K, V, attn_bias=<mask_tensor>, is_causal=False)

AFTER (lowering pass):
  F.scaled_dot_product_attention(Q, K, V, attn_mask=None, is_causal=True)
```

The lowering pass discards `attn_bias` and sets `is_causal=True`, so the custom SDPA converter generates the causal mask dynamically inside TensorRT — avoiding the large mask tensor overhead entirely.

### Suggested Fix

Option A: In the `efficient_attention` converter, detect when `attn_bias` is a standard causal mask and convert it to `is_causal=True` + `add_attention()`:

```python
if attn_bias is not None and _is_causal_mask(attn_bias):
    # Use IAttention with is_causal=True
    attention_layer = ctx.net.add_attention(scaled_query, key, value, SOFTMAX, is_causal=True)
else:
    # Keep manual decomposition for non-standard masks
    ...
```

Option B: Use the `IAttention` layer's mask input directly:

```python
attention_layer = ctx.net.add_attention(scaled_query, key, value, SOFTMAX, is_causal=False)
attention_layer.mask = attn_bias  # pass mask to IAttention layer
```

Option C: Add a lowering pass (similar to `register_sdpa`) that normalizes HF's `attn_bias` to `is_causal=True` before the converter runs.

### Environment

- Torch-TensorRT: built from `attn-plugin-workflow` branch (includes PR #4104)
- Transformers: v4.52.3
- TensorRT: 10.15.1
- GPU: NVIDIA A100 80GB PCIe

### Related

- PR #4104: Added `IAttention` converters
- Branch `attn-plugin-workflow`: Added `--backend iattention` option
```

### PR DESCRIPTION
**Pull request by zewenli98**
# Description

Support `attn_bias` for efficient SDPA

Fixes #4129  #4145

## Type of change

# Checklist:

### DIFF
--- a/docsrc/user_guide/compilation/compilation_settings.rst
+++ b/docsrc/user_guide/compilation/compilation_settings.rst
@@ -123,6 +123,11 @@ Optimization Tuning
      - Disable TensorFloat-32 (TF32) accumulation. TF32 is enabled by default on
        Ampere and newer GPUs and provides FP32-range with FP16-speed for matmul/conv.
        Disable only when you need strict IEEE FP32 semantics.
+   * - ``attn_bias_is_causal``
+     - ``True``
+     - Whether the attn_bias in efficient SDPA is causal. Default is True. This can 
+       accelerate models from HF because attn_bias is always a causal mask in HF. 
+       If you want to use non-causal attn_bias, you can set this to False.
 
 ----
 
@@ -448,7 +453,7 @@ rebuilt from scratch:
 ``l2_limit_for_tiling``, ``enable_autocast``, ``autocast_low_precision_type``,
 ``autocast_excluded_nodes``, ``autocast_excluded_ops``,
 ``autocast_max_output_threshold``, ``autocast_max_depth_of_reduction``,
-``autocast_calibration_dataloader``.
+``autocast_calibration_dataloader``, ``decompose_attention``, ``attn_bias_is_causal``.
 
 Settings not in this list (e.g., ``debug``, ``dryrun``, ``pass_through_build_failures``)
 can be changed without invalidating the cache.
--- a/docsrc/user_guide/compilation/compilation_settings.rst
+++ b/docsrc/user_guide/compilation/compilation_settings.rst
@@ -350,6 +350,10 @@ Graph Partitioning
      - ``False``
      - Use ``aot_autograd`` for tracing instead of the default path. Required when the
        model contains ``DTensor`` or other distributed tensors.
+   * - ``decompose_attention``
+     - ``False``
+     - Decompose attention layers into smaller ops. We have converters for handling attention ops,
+       but if you want to decompose them into smaller ops, you can set this to True.
 
 ----
 
@@ -372,7 +376,7 @@ Compilation Workflow
      - ``False``
      - Defer TRT engine deserialization until all engines have been built.
        Works around resource contraints and builder overhad but engines
-       may be less well tuned to their deployment resource availablity
+       may be less well tuned to their deployment resource availability
    * - ``debug``
      - ``False``
      - Enable verbose TRT builder logs at ``DEBUG`` level.
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -112,6 +112,7 @@ def cross_compile_for_windows(
     cpu_memory_budget: Optional[int] = _defaults.CPU_MEMORY_BUDGET,
     dynamically_allocate_resources: bool = _defaults.DYNAMICALLY_ALLOCATE_RESOURCES,
     decompose_attention: bool = _defaults.DECOMPOSE_ATTENTION,
+    attn_bias_is_causal: bool = _defaults.ATTN_BIAS_IS_CAUSAL,
     **kwargs: Any,
 ) -> torch.fx.GraphModule:
     """Compile an ExportedProgram module using TensorRT in Linux for Inference in Windows
@@ -190,6 +191,7 @@ def cross_compile_for_windows(
         cpu_memory_budget (Optional[int]): The maximum amount of CPU memory to use for the compilation. If the compilation requires more memory than this budget, the compilation will fail.
         dynamically_allocate_resources (bool): Dynamically allocate resources during engine execution.
         decompose_attention (bool): Whether to decompose attention layers. We have converters for handling attention ops, but if you want to decompose them into smaller ops, you can set this to True.
+        attn_bias_is_causal (bool): Whether the attn_bias in efficient SDPA is causal. Default is True. This can accelerate models from HF because attn_bias is always a causal mask in HF. If you want to use non-causal attn_bias, you can set this to False.
         **kwargs: Any,
     Returns:
         torch.fx.GraphModule: Compiled FX Module, when run it will execute via TensorRT
@@ -349,6 +351,7 @@ def cross_compile_for_windows(
         "cpu_memory_budget": cpu_memory_budget,
         "dynamically_allocate_resources": dynamically_allocate_resources,
         "decompose_attention": decompose_attention,
+        "attn_bias_is_causal": attn_bias_is_causal,
     }
 
     # disable the following settings is not supported for cross compilation for windows feature
@@ -471,6 +474,7 @@ def compile(
     enable_resource_partitioning: bool = _defaults.ENABLE_RESOURCE_PARTITIONING,
     dynamically_allocate_resources: bool = _defaults.DYNAMICALLY_ALLOCATE_RESOURCES,
     decompose_attention: bool = _defaults.DECOMPOSE_ATTENTION,
+    attn_bias_is_causal: bool = _defaults.ATTN_BIAS_IS_CAUSAL,
     **kwargs: Any,
 ) -> torch.fx.GraphModule:
     """Compile an ExportedProgram module for NVIDIA GPUs using TensorRT
@@ -559,6 +563,7 @@ def compile(
         cpu_memory_budget (Optional[int]): The maximum amount of CPU memory to use for the compilation. If the compilation requires more memory than this budget, the compilation will fail.
         dynamically_allocate_resources (bool): Dynamically allocate resources during engine execution.
         decompose_attention (bool): Whether to decompose attention layers. We have converters for handling attention ops, but if you want to decompose them into smaller ops, you can set this to True.
+        attn_bias_is_causal (bool): Whether the attn_bias in efficient SDPA is causal. Default is True. This can accelerate models from HF because attn_bias is always a causal mask in HF. If you want to use non-causal attn_bias, you can set this to False.
         **kwargs: Any,
     Returns:
         torch.fx.GraphModule: Compiled FX Module, when run it will execute via TensorRT
@@ -763,6 +768,7 @@ def compile(
         "cpu_memory_budget": cpu_memory_budget,
         "dynamically_allocate_resources": dynamically_allocate_resources,
         "decompose_attention": decompose_attention,
+        "attn_bias_is_causal": attn_bias_is_causal,
     }
     logger.debug(f"CPU memory usage before lowering: {get_cpu_memory_usage()} MB")
     settings = CompilationSettings(**compilation_options)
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

**Comment by chohk88** [comment]
Hi Evan, I tested this PR on three HuggingFace models (Qwen2.5-0.5B, Qwen3-0.6B, Llama-3.2-1B) with the `iattention` backend. Here's what I found.

### Background: How HuggingFace uses `_scaled_dot_product_efficient_attention`

HuggingFace always passes the causal mask as `attn_bias` with `is_causal=False`, rather than setting `is_causal=True` directly:

```python
# HuggingFace's actual call pattern:
aten._scaled_dot_product_efficient_attention(
    Q, K, V,
    attn_bias=[[[ 0,   -inf, -inf, -inf],
                [ 0,    0,   -inf, -inf],
                [ 0,    0,    0,   -inf],
                [ 0,    0,    0,    0  ]]],
    is_causal=False
)
```

This additive causal mask (0 for attend, -inf for block) is semantically equivalent to `is_causal=True`, but it forces the converter to handle a large mask tensor.

### Issue 1: Dynamic head dimension causes `add_attention` failure

When running with HuggingFace models, `torch.export` marks the query tensor's head dimension as dynamic (`-1`), even though it should be static. This is because `query` goes through `.view().transpose()` during tracing, while `key`/`value` go through an additional `repeat_kv` (`.expand().reshape()`) that preserves the static head count.

```
Observed shapes at converter:
  query:        (1, -1, -1, 64)    # head dim is dynamic
  key:          (1, 14, -1, 64)    # head dim is static (14)
  scaled_query: (1, -1, -1, 64)    # still dynamic after scaling
```

TRT's `add_attention` requires a static head dimension for validation, so it fails with:
```
Error Code 3: API Usage Error ((Unnamed Layer*) [AttentionInput]:
query number of heads must be divisible by key/value number of heads)
```

I worked around this by reshaping `scaled_query` to match `key`'s shape (they should be identical after HF's GQA expansion via `repeat_kv`):

```python
if (
    isinstance(scaled_query.shape[1], int)
    and scaled_query.shape[1] < 0
    and isinstance(key.shape[1], int)
    and key.shape[1] > 0
):
    shape_layer = ctx.net.add_shape(key)
    shape_layer.name = name + "_key_shape"
    shuffle = ctx.net.add_shuffle(scaled_query)
    shuffle.set_input(1, shape_layer.get_output(0))
    shuffle.name = name + "_fix_head_dim"
    scaled_query = shuffle.get_output(0)
```

### Issue 2: IAttention with mask is slower than the previous manual decomposition

With the dynamic head dim fix applied, the PR's approach (IAttention + mask) does run, but the large causal mask tensor (`[1, 14, 2048, 2048]` per attention layer) introduces overhead that offsets the IAttention kernel's benefit. Benchmark on A100 80GB, FP16, ISL=2048, OSL=128, Batch=1:

- **PyTorch**: 4743 ms
- **iattention (main branch, manual matmul decomposition)**: 5421 ms
- **iattention (this PR, IAttention + mask)**: 5633 ms ← **slower than main**

### Observation: Discarding the mask and using `is_causal=True` gives dramatic improvement

Since HuggingFace's `attn_bias` is always a causal mask, discarding it and setting `is_causal=True` is semantically equivalent. I tested this by adding the following before `add_attention`:

```python
if mask_tensor is not None:
    mask_tensor = None
    use_causal = True
```

Full benchmark results (Median Latency in ms):

| Model | PyTorch | iattention (main) | iattention (PR+mask) | iattention (is_causal) | sdpa no cache | sdpa static_v1 | plugin |
|-------|--------:|------------------:|---------------------:|-----------------------:|--------------:|---------------:|-------:|
| Qwen2.5-0.5B | 4743 | 5421 (0.88x) | 5633 (0.84x) | **2303 (2.06x)** | 3271 (1.5x) | 1238 (3.8x) | 421 (11.3x) |
| Qwen3-0.6B | 6891 | 6792 (1.01x) | — | **3097 (2.22x)** | 4031 (1.7x) | 1708 (4.0x) | 569 (12.1x) |
| Llama-3.2-1B | 7064 | 8283 (0.85x) | — | **4426 (1.60x)** | 5466 (1.3x) | 1379 (5.1x) | 465 (15.2x) |

*Parentheses show speedup vs PyTorch.*

The `is_causal=True` approach makes `iattention` **1.6–2.2x faster than PyTorch** and even **faster than `sdpa no cache`** (which goes through a separate graph-level lowering pass to achieve the same mask-to-causal conversion).

### Trade-off discussion

However, unconditionally discarding the mask means the converter can no longer support arbitrary (non-causal) masks. The existing tests in `test_attention_aten.py` pass random float tensors as `attn_bias`, which would produce incorrect results if the mask were discarded.

If the `iattention` backend is intended primarily for HuggingFace LLM inference (where `attn_bias` is always a causal mask), then discarding the mask is the right optimization. But if the converter needs to remain general-purpose for arbitrary masks, we need a different approach — perhaps detecting the causal mask pattern at the graph level, or providing a flag (e.g., `decompose_attention`) to let users control this behavior.

What are your thoughts on how to proceed?

chohk88:
After further investigation, I think adding an `iattention-specific lowering pass` would be a cleaner solution than unconditionally discarding the mask inside the converter like above:
```python
if mask_tensor is not None:
    mask_tensor = None
    use_causal = True
```

The `sdpa` backend avoids the mask overhead by using a lowering pass (`enable_sdpa_converter()`) that sets `attn_bias=None` and `is_causal=True` at the graph level before conversion. We could take a similar approach for the `iattention` backend, but with one difference: the lowering pass should modify the args of `_efficient_attention` **without changing the target op**, so that the built-in converter with `add_attention()` (IAttention kernel) is still used.

```
# sdpa lowering pass (existing):
_efficient_attention(Q, K, V, attn_bias=<mask>, is_causal=False)
  → F.sdpa(Q, K, V, attn_mask=None, is_causal=True)     ← replaces the op
  → sdpa_converter.py (matmul decomposition, no IAttention)

# iattention lowering pass (proposed):
_efficient_attention(Q, K, V, attn_bias=<mask>, is_causal=False)
  → _efficient_attention(Q, K, V, attn_bias=None, is_causal=True)  ← same op, args only
  → attention.py built-in converter (add_attention → IAttention kernel)
```

This keeps the converter general-purpose (arbitrary masks still work when no lowering pass is registered, e.g., in unit tests), while getting the optimal IAttention performance for LLM inference. It also feels cleaner than unconditionally discarding the mask inside the converter itself.

Note: the dynamic head dimension fix from Issue 1 is still needed in the converter regardless, since `torch.export` can mark Q's head dim as dynamic even when `attn_bias=None`.

What do you think?

zewenli98:
> This keeps the converter general-purpose (arbitrary masks still work when no lowering pass is registered, e.g., in unit tests), while getting the optimal IAttention performance for LLM inference. It also feels cleaner than unconditionally discarding the mask inside the converter itself.

Thanks for the detailed experiments! For this part, did you mean users need to manually register the lowering pass if they are setting `attn_bias=<mask>, is_causal=False`? like `register_default_sdpa_pass`? I'm thinking if we can internally do this so that users don't need to register. Since there are only two types of attn_bias which equals to mask: 1) float mask and 2) bool mask, we can check attn_bias in the converter. If attn_bias matches the mask pattern, we can overwrite `attn_bias=None, is_causal=True`

zewenli98:
@chohk88 I updated the PR with an additional compilation arg `attn_bias_is_causal` and a lowering pass to control the behavior. I tried a few ways and finally I feel this way is fast and also convenient for users. What do you think?
Besides, I couldn't repro the dynamic shape error. can you share how to repro? thanks!

zewenli98:
Observed that `tests/py/dynamo/models/test_models.py::test_bert_base_uncased` failed in TRT 10.14.1 with the error:
```
ERROR    torch_tensorrt [TensorRT Conversion Context]:logging.py:25 INetworkDefinition::addAttention: Error Code 3: API Usage Error (Parameter check failed, condition: mScopedOps.size() == mGraph.scopedOps.size().  In addAttention at /_src/optimizer/api/network.cpp:2125)
```
but works in TRT 10.15.1

chohk88:
@zewenli98 

The `attn_bias_is_causal` lowering pass completely fixes the slow iattention backend issue with HF models. Previously iattention was on par or slower than PyTorch (0.85–1.01x), now it's **faster than sdpa no cache** across all tested models.

Tested by cherry-picking onto the `attn-plugin-workflow` branch. 
A100 80GB, FP16, ISL=2048, OSL=128, Batch=1:

| Model | PyTorch | iattention (main) | iattention (prev PR + mask) | **iattention (current PR)** | sdpa no cache | sdpa static_v1 | plugin |
|-------|--------:|------------------:|-------------------------:|-------------------------------:|--------------:|---------------:|-------:|
| Qwen2.5-0.5B | 4730 | 5421 (0.88x) | 5633 (0.84x) | **2112 (2.24x)** | 3271 (1.5x) | 1238 (3.8x) | 421 (11.3x) |
| Qwen3-0.6B | 6838 | 6792 (1.01x) | — | **2892 (2.36x)** | 4031 (1.7x) | 1708 (4.0x) | 569 (12.1x) |
| Llama-3.2-1B | 7051 | 8283 (0.85x) | — | **4278 (1.65x)** | 5466 (1.3x) | 1379 (5.1x) | 465 (15.2x) |

*Parentheses show speedup vs PyTorch.*

chohk88 (approved):
One note: I still hit the dynamic head dimension error from Issue 1 in my previous comment. The workaround (reshaping `scaled_query` to match `key`'s shape) is still needed on my setup.

Filed as [4145](https://github.com/pytorch/TensorRT/issues/4145)

zewenli98:
@chohk88 Thanks a lot for the detailed feedback! good to see the perf boost!

cehongwang (commented):
I think this is fixed in another PR: https://github.com/pytorch/TensorRT/pull/4156/changes

Which one should we keep? @narendasan

narendasan:
I think rebase and it should be fine

narendasan:
What is this PR waiting on?

zewenli98:
@narendasan The fp32 works but fp16 still gets some mismatched values, even for decompose_attention=True. Do you think we can directly merge this PR? or make fp32 go IAttention path and fp16 go decompose path since the decompose path has fewer mismatched values?

zewenli98:
After discussion with TRT team, they can repro the errors in fp16 and they have filed a bug to TRT. This PR skips the fp16 tests for now.

PR #4147, opened 2026-03-30
feat(plugins): add QDP AOT callbacks to custom_op

BowenFu:
## Summary

- Extend the existing automatic-plugin path instead of creating a parallel annotation/plugin stack.
- Let `custom_op` and `generate_plugin` accept user-supplied TensorRT QDP `aot_impl` and `autotune` callbacks.
- Preserve the generated JIT implementation, converter, dynamic-shape controls, and JIT fallback.
- Keep kernel compilation and the QDP launch contract with the caller; Torch-TensorRT does not rewrite PTX or CUBIN.

## Tests

- Focused unit coverage verifies callback forwarding and unchanged JIT defaults.
- An end-to-end automatic-plugin test registers a valid fixed PTX kernel through the new AOT path and compares TensorRT output with PyTorch.
- Repository-pinned pre-commit hooks pass: mypy, Ruff, Black, isort, typos, and NVIDIA internal checks.

Runtime GitHub workflows require maintainer approval because the branch is from a fork.

### DIFF
--- a/py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
+++ b/py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
@@ -11,9 +11,12 @@ AOT compilation pipeline for cuTILE kernels
    *pointer binding indices* and *scalar SymInt32 expressions* that TRT evaluates at
    runtime.
 
-3. **CUBIN compilation** — ``compile_tile(pyfunc, example_tensors, CompilerOptions())``
-   from ``cuda.tile._compile`` produces a CUBIN file containing an embedded PTX
-   debug section (``.nv_debug_ptx_txt``).
+3. **CUBIN compilation** — ``cuda.tile.compilation.export_kernel(kernel,
+   [KernelSignature(...)], buf, gpu_code=..., output_format="cubin")``
+   produces a CUBIN with an embedded PTX debug section (``.nv_debug_ptx_txt``).
+   Signatures are built from the TensorDesc + scalar attrs of the registered
+   plugin so the cubin's symbol carries the right rank/dtype/constant mangling
+   (e.g. ``_kernel_Kt1_A1f32_1l0_I128``).
 
 4. **PTX extraction and parameter reordering** — ``_extract_ptx_from_cubin`` recovers
    the PTX from the CUBIN.  The cuTILE kernel ABI uses per-tensor groups of
@@ -89,6 +92,7 @@ from .._qdp_utils import (
 from ..._recorders import CuTileLaunchRecorder
 from ..._specs import CuTileSpec
 from .._symbolic import SymbolicTensor, TensorRole
+from . import downgrade_ptx_version
 
 
 def _params_per_tensor_for_rank(rank: int) -> int:
@@ -142,9 +146,7 @@ _PTX_ENTRY_PARAM_RE = re.compile(
     re.DOTALL,
 )
 
-_PTX_VERSION_RE = re.compile(r"\.version\s+(\d+)\.(\d+)")
 _PTX_REQNTID_RE = re.compile(r"\.reqntid\s+(\d+)")
-_PTX_MAX_SUPPORTED_VERSION = (9, 0)
 
 
 def _extract_ptx_from_cubin(cubin_bytes: bytes) -> Optional[str]:
@@ -188,31 +190,14 @@ def _extract_ptx_from_cubin(cubin_bytes: bytes) -> Optional[str]:
     return "\n".join(lines) + "\n"
 
 
-# LIMITATION (short-term workaround): PTX version downgrade is a best-effort
-# text substitution.  CUDA 13.1 tileiras emits .version 9.1 but the driver in
-# our container (CUDA 13.0) only JIT-compiles up to PTX 9.0.  Lowering the
-# declared version by editing the .version line is safe only if the PTX body
-# uses no ISA features introduced after the target version.  This will silently
-# produce incorrect PTX if a future tileiras version emits instructions that
-# require a higher PTX version than the declared one after downgrade.
-# Long-term fix: align the driver and CUDA toolkit versions so no downgrade is
-# needed.
-def _downgrade_ptx_version(ptx: str) -> str:
-    """Downgrade .version directive if it exceeds what the runtime driver supports.
-
-    CUDA 13.1 tileiras emits .version 9.1 but a CUDA 13.0 driver only JIT-compiles
-    up to 9.0.  Lowering the declared version is safe as long as the PTX body doesn't
-    use features introduced after the target version (CuTile-generated PTX doesn't).
-    """
-    m = _PTX_VERSION_RE.search(ptx)
-    if not m:
-        return ptx
-    major, minor = int(m.group(1)), int(m.group(2))
-    max_major, max_minor = _PTX_MAX_SUPPORTED_VERSION
-    if (major, minor) <= (max_major, max_minor):
-        return ptx
-    replacement = f".version {max_major}.{max_minor}"
-    return ptx[: m.start()] + replacement + ptx[m.end() :]
+# PTX header version downgrade lives in ``_aot.downgrade_ptx_version`` —
+# both Triton and cuTile backends share that helper.  LIMITATION (short-term
+# workaround): the downgrade is a best-effort text substitution; safe only if
+# the PTX body uses no ISA features introduced after the target version.  This
+# will silently produce incorrect PTX if a future tileiras version emits
+# instructions that require a higher PTX version than the declared one after
+# downgrade.  Long-term fix: align the driver and CUDA toolkit versions so no
+# downgrade is needed.
 
 
 def _parse_reqntid(ptx: str) -> Optional[int]:
@@ -295,7 +280,7 @@ def _reorder_cutile_ptx_for_trt(
     reordered_ptx = _reorder_ptx_params(ptx, runtime_order)
     if reordered_ptx == ptx:
         return cubin_bytes
-    reordered_ptx = _downgrade_ptx_version(reordered_ptx)
+    reordered_ptx = downgrade_ptx_version(reordered_ptx)
     return reordered_ptx.encode("utf-8")
 
 
@@ -768,7 +753,7 @@ def aot_impl_cutile(
         # If code_out is still a cubin but we have ptx_raw available, convert to PTX
         # so we can do a safe string-replace for the kernel name.
         if not is_ptx and ptx_raw is not None:
-            downgraded = _downgrade_ptx_version(ptx_raw)
+            downgraded = downgrade_ptx_version(ptx_raw)
             code_out = downgraded.encode("utf-8")
             is_ptx = True
         if is_ptx:
--- a/py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
+++ b/py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
@@ -305,15 +305,19 @@ def _infer_cutile_extra_args(
     block_size: int,
     scalar_symints: Optional[List[Any]] = None,
     rank: int = 1,
+    num_scalars: Optional[int] = None,
 ) -> Any:
     """Build SymIntExprs: (extent per dim, stride per dim) per tensor then scalar(s).
 
     Uses symbolic SymInt32 expressions for dynamic dimensions so that TRT can
     evaluate them at runtime with actual input shapes.
+
+    ``num_scalars=0`` skips the trailing scalar slot entirely; this matches the
+    cuda-tile >=1.4 ABI where ``ct.Constant``-annotated parameters are baked
+    into the kernel symbol and not exposed at runtime.
     """
-    num_scalars = 1
-    if scalar_symints:
-        num_scalars = len(scalar_symints)
+    if num_scalars is None:
+        num_scalars = len(scalar_symints) if scalar_symints else 1
     n_tensors = len(inp_descs) + len(out_descs)
     r = max(1, rank)
     n = n_tensors * 2 * r + num_scalars
@@ -541,15 +545,18 @@ def aot_impl_cutile(
 
     def _build_signature(flatten_all: bool) -> KernelSignature:
         # Match the kernel ABI: arrays in (inputs..., outputs...) order,
-        # followed by scalar attributes as compile-time constants.
+        # followed by scalar attributes as compile-time constants.  The
+        # convention used by ``_ct_launch_*`` helpers (reshape(-1) before
+        # ``ct.launch``) compiles every kernel against rank-1 views; that
+        # is what the AOT path bakes in.  ``flatten_all`` is retained for
+        # symmetry with the legacy retry path and currently has no effect.
+        del flatten_all  # noqa: F841 — kept for caller signature parity
         params: List[Any] = []
         for td in all_descs:
-            natural_ndim = len(td.shape_expr) or 1
-            ndim = 1 if (flatten_all or natural_ndim > 2) else natural_ndim
             params.append(
                 ArrayConstraint(
                     dtype=_trt_dtype_to_ct(td.dtype),
-                    ndim=ndim,
+                    ndim=1,
                     index_dtype=ct.int32,
                     # cuTile rejects negative strides by default; the TRT
                     # tensors we receive always have non-negative strides.
@@ -614,7 +621,11 @@ def aot_impl_cutile(
         or "cutile_kernel"
     )
 
-    num_scalars = len(scalar_symints) if scalar_symints else 1
+    # Under cuda-tile >=1.4, scalar attributes are encoded as ConstantConstraint
+    # and baked into the kernel symbol; no runtime scalar PTX params survive.
+    # The legacy ABI exposed one runtime scalar per ``scalar_symints`` entry
+    # (or the default block size), so the reorder helpers had a non-zero count.
+    num_scalars = 0
     rank = max(1, max(len(td.shape_expr) for td in all_descs))
     ptx_raw = _extract_ptx_from_cubin(cubin_bytes)
     num_ptx_params = _count_ptx_params(ptx_raw) if ptx_raw else 0
@@ -719,8 +730,15 @@ def aot_impl_cutile(
     launch.block_x = actual_block
     launch.shared_mem = 0
 
+    # num_scalars=0: cuda-tile >=1.4 bakes ``ct.Constant`` scalars into the
+    # kernel symbol; no runtime scalar args follow the (extent, stride) tuples.
     extra_args = _infer_cutile_extra_args(
-        inp_descs, out_descs, block_size, scalar_symints, rank=effective_rank
+        inp_descs,
+        out_descs,
+        block_size,
+        scalar_symints,
+        rank=effective_rank,
+        num_scalars=0,
     )
 
     ext = ".ptx" if is_ptx else ".cubin"
--- a/py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
+++ b/py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
@@ -1,0 +1,754 @@
+"""cuTILE AOT backend: source → sandbox recording → CUBIN/PTX → KernelLaunchParams.
+
+AOT compilation pipeline for cuTILE kernels
+============================================
+1. **Sandbox execution** — A patched ``cuda.tile.launch`` shim is injected alongside
+   ``CuTileLaunchRecorder`` proxies for each cuTILE program object found in the
+   user's module globals.  ``launch_fn`` is called with ``SymbolicTensor`` arguments
+   so that the recorded grid and tensor bindings are captured symbolically.
+
+2. **Argument analysis** — ``analyze_launch_args`` splits recorded call arguments into
+   *pointer binding indices* and *scalar SymInt32 expressions* that TRT evaluates at
+   runtime.
+
+3. **CUBIN compilation** — ``compile_tile(pyfunc, example_tensors, CompilerOptions())``
+   from ``cuda.tile._compile`` produces a CUBIN file containing an embedded PTX
+   debug section (``.nv_debug_ptx_txt``).
+
+4. **PTX extraction and parameter reordering** — ``_extract_ptx_from_cubin`` recovers
+   the PTX from the CUBIN.  The cuTILE kernel ABI uses per-tensor groups of
+   ``(ptr, extents..., strides...)`` parameters in manual (kernel-source) order.
+   The QDP runtime passes ``(input_ptrs, extra_args, output_ptrs)`` in a different
+   order.  ``_reorder_cutile_ptx_for_trt`` reorders the ``.param`` declarations to
+   match the runtime expectation and downgrades the ``.version`` directive if the
+   CUDA driver is older than the PTX version emitted by ``tileiras``.
+
+5. **Tactic uniquification** — Identical to the Triton backend: a config-derived
+   suffix is appended to the kernel name so TRT registers separate PTX/cubin entries
+   per tactic.
+
+The public entry-point is ``aot_impl_cutile``, which returns the QDP AOT 4-tuple
+``(kernel_name_bytes, code_bytes, KernelLaunchParams, SymIntExprs)``.
+``compile_cutile_program`` is a thin tactic-manager wrapper.
+"""
+from __future__ import annotations
+
+import logging
+import os
+import re
+from typing import Any, Dict, List, Mapping, Optional, Tuple
+
+logger = logging.getLogger(__name__)
+
+try:
+    import tensorrt as trt
+    import tensorrt.plugin as trtp
+except ImportError as e:
+    raise ImportError(
+        "TensorRT with plugin support is required for cuTILE AOT compilation."
+    ) from e
+
+from .._qdp_utils import (
+    AOTMetadata,
+    TTAPluginError,
+    _as_symint32,
+    _assign_recorded_grid,
+    _launch_params_from_trt,
+    _safe_dim,
+    _shape_expr_to_ints,
+    _sym_dim,
+    analyze_launch_args,
+    dump_code_artifact,
+    is_cutile_program,
+    run_kernel_sandbox,
+    td_dtype_to_torch,
+)
+from ..._recorders import CuTileLaunchRecorder
+from ..._specs import CuTileSpec
+from .._symbolic import SymbolicTensor, TensorRole
+
+
+def _params_per_tensor_for_rank(rank: int) -> int:
+    """CuTile kernel ABI: ptr + extent per dim + stride per dim = 1 + 2*rank."""
+    return 1 + 2 * max(1, rank)
+
+
+def _cutile_trt_order_for_io(
+    num_inputs: int,
+    num_outputs: int,
+    num_scalars: int = 1,
+    rank: int = 1,
+) -> Tuple[int, Tuple[int, ...]]:
+    """Compute (num_params, runtime_order) for CuTile kernel.
+
+    Per-tensor params: ptr (1) + extent per dim (rank) + stride per dim (rank) = 1 + 2*rank.
+    Manual (kernel) order: for each input [ptr, ext..., str...]; for each output [ptr, ext..., str...]; scalars.
+    QDP runtime passes: input_ptrs, extra_args (extent/stride + scalars), output_ptrs.
+    Returns permutation so that reordered[physical_idx] = manual_param_index.
+    """
+    n_in = num_inputs
+    n_out = num_outputs
+    ppt = _params_per_tensor_for_rank(rank)
+    num_params = (n_in + n_out) * ppt + num_scalars
+    manual_ptr_in = [i * ppt for i in range(n_in)]
+    manual_ptr_out = [n_in * ppt + j * ppt for j in range(n_out)]
+    extent_stride_indices = list(range(1, ppt))
+    manual_extent_stride_in = [
+        i * ppt + k for i in range(n_in) for k in extent_stride_indices
+    ]
+    manual_extent_stride_out = [
+        n_in * ppt + j * ppt + k
+        for j in range(n_out)
+        for k in extent_stride_indices
+    ]
+    manual_scalars = [(n_in + n_out) * ppt + s for s in range(num_scalars)]
+    trt_order = (
+        list(manual_ptr_in)
+        + manual_extent_stride_in
+        + manual_extent_stride_out
+        + manual_scalars
+        + list(manual_ptr_out)
+    )
+    return num_params, tuple(trt_order)
+
+
+_ELF_MIN_SIZE = 64
+
+_PTX_ENTRY_PARAM_RE = re.compile(
+    r"(\.visible\s+\.entry\s+(\w+)\s*\()([^)]*)\)",
+    re.DOTALL,
+)
+
+_PTX_VERSION_RE = re.compile(r"\.version\s+(\d+)\.(\d+)")
+_PTX_REQNTID_RE = re.compile(r"\.reqntid\s+(\d+)")
+_PTX_MAX_SUPPORTED_VERSION = (9, 0)
+
+
+def _extract_ptx_from_cubin(cubin_bytes: bytes) -> Optional[str]:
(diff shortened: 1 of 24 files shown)


py/torch_tensorrt/annotation/_custom_plugin/_lowering.py
```diff
@@ -0,0 +1,227 @@
...
+        f"spec object from tta.custom_plugin() or tta.export_as(impl=...)."
+    )
+
+
+# ---------------------------------------------------------------------------
+# Public lowering API
+# ---------------------------------------------------------------------------
+
+
+def lower_custom_plugin(
```
narendasan: Can we merge this stuff with already existing plugin autogeneration in torch_tensorrt.dynamo.conversion.plugin? Like we already automate converter generation key'ed on operator name

py/torch_tensorrt/annotation/_custom_plugin/_lowering.py
```diff
@@ -0,0 +1,227 @@
...
+            inconsistent descriptor (propagated from
+            ``lower_custom_plugin_descriptor``).
+        trt.tensorrt.IBuilderError: If TRT rejects the plugin layer (e.g.
+            unsupported dtype/format combination).
+    """
+    descriptor: CustomPluginSpec = _spec_to_descriptor(spec)
+    return lower_custom_plugin_descriptor(ctx, descriptor, trt_inputs, name)
+
+
+def register_custom_plugin_qdp(
```
narendasan: Same here, these facilities already exist, they should be extended not duplicated

narendasan (commented):
@BowenFu can we split this out in to a PR stack? lets put tta.custom_plugin at the bottom. What I want to focus on is lets say a user already has implemented a custom operator in PyTorch backed by one of these kernels. We want to enable the AOT QDP launch of that kernel without a bunch of boilerplate: Basically this example https://docs.pytorch.org/TensorRT/tutorials/_rendered_examples/dynamo/auto_generate_plugins.html with AOT QDP. There is already facilities for the converter generation and plugin registration from PyTorch Meta Kernel. Once we have that we have a solid base to look at region labeling / manual fusion and other advance usecases.

narendasan:
Also @BowenFu please follow the instructions I sent for how to get added to the CLA

BowenFu:
> @BowenFu can we split this out in to a PR stack? lets put tta.custom_plugin at the bottom. What I want to focus on is lets say a user already has implemented a custom operator in PyTorch backed by one of these kernels. We want to enable the AOT QDP launch of that kernel without a bunch of boilerplate: Basically this example https://docs.pytorch.org/TensorRT/tutorials/_rendered_examples/dynamo/auto_generate_plugins.html with AOT QDP. There is already facilities for the converter generation and plugin registration from PyTorch Meta Kernel. Once we have that we have a solid base to look at region labeling / manual fusion and other advance usecases.

Sure. Will include only tta.custom_plugin in this PR.

py/torch_tensorrt/annotation/_custom_plugin/_lowering.py
BowenFu: I just cleaned up all tta.lower_as/tta.export_as related codes from this PR.

tests/py/annotation/unit/test_specs.py
```diff
@@ -0,0 +1,303 @@
...
+
+        descriptor = tta.custom_plugin(tta.triton(kernel))
+        self.assertIs(tta.normalize_impl_to_spec(descriptor), descriptor)
+
+    def test_rejects_invalid_type(self):
+        with self.assertRaisesRegex(TypeError, "normalize_impl_to_spec"):
+            tta.normalize_impl_to_spec(123)
+
+
+if __name__ == "__main__":
```
bowang007: This test spec seems a little opaque to me. I have no idea what it tests without diving into the code and understand that the spec includes difference specification for each type of kernel. We could make it more straightforward.

tests/py/annotation/unit/test_specs.py
```diff
@@ -0,0 +1,303 @@
...
+        normalized = tta.normalize_impl_to_spec(tta.cutile(kernel))
+        self.assertIsInstance(normalized, tta.CustomPluginSpec)
+        self.assertIsInstance(normalized.specs[0], tta.CuTileSpec)
+
+    def test_bare_cutedsl_spec_wrapped(self):
+        def kernel(x, out):
+            pass
+
+        normalized = tta.normalize_impl_to_spec(tta.cutedsl(kernel))
+        self.assertIsInstance(normalized, tta.CustomPluginSpec)
```
bowang007: If I understand correctly, this won't include the runtime test right? 
How can we make sure that the kernel produces correct output with different config?

py/torch_tensorrt/dynamo/conversion/plugins/_custom_op.py
```diff
@@ -20,20 +26,29 @@ def custom_op(
...
-    generate_plugin(op_name)
-    generate_plugin_converter(
-        op_name,
-        capability_validator,
-        priority,
-        supports_dynamic_shapes,
-        requires_output_allocator,
-    )
+    if impl is not None:
+        impl.register_dynamo_plugin(
```
narendasan: We dont need a complete second code path for this, lets reuse what we have. For example creating the converter is likely the only user of `capability_validator`, `priority`, `supports_dynamic_shapes`, `requires_output_allocator`.  `generate_plugin_converter` already creates this converter key'ed on name,

Really I would expect the code to look like: 


```py
generate_plugin(op_name) # Generates JIT QDP Plugin 
generate_plugin_converter(op_name, capability_validator, priority, supports_dynamic_shapes, requires_output_allocator) # Generates the converter that inserts the QDP plugin 

if impl: #this should be kernel_impl probably 
    impl.generate_plugin_aot_impl() 
```

py/torch_tensorrt/annotation/_specs.py
```diff
@@ -0,0 +1,372 @@
...
+    """Attach :class:`AnnotationMetadata` to a callable."""
+    setattr(fn, _ANNOTATION_METADATA_ATTR, metadata)
+
+
+def get_annotation_metadata(fn: Callable) -> Optional[AnnotationMetadata]:
+    """Retrieve :class:`AnnotationMetadata` from a callable, if present."""
+    return getattr(fn, _ANNOTATION_METADATA_ATTR, None)
+
+
+# ── Custom kernel specs (Triton / CuTile / CuTeDSL) ──────────────────────────
```
narendasan: Do we have one for NVRTC?

py/torch_tensorrt/annotation/_specs.py
```diff
@@ -0,0 +1,372 @@
...
+        raise TypeError(
+            f"{spec_name}: configs must be a list or None, got {type(configs).__name__!r}"
+        )
+
+
+# ── Metadata helpers ──────────────────────────────────────────────────────────
+
+
+@dataclass
+class AnnotationMetadata:
```
narendasan: Is this needed in this PR?

py/torch_tensorrt/annotation/_layer_metadata.py
```diff
@@ -0,0 +1,546 @@
...
+            best, best_pri = parsed, pri
+    return best
+
+
+# ---------------------------------------------------------------------------
+# Layer metadata writer
+# ---------------------------------------------------------------------------
+
+
+def set_tta_layer_metadata(
```
narendasan: Why do we need this?

py/torch_tensorrt/annotation/_layer_metadata.py
```diff
@@ -0,0 +1,546 @@
+"""TTA metadata stored on TensorRT ``ILayer`` objects as a plain string.
```
narendasan: Is this needed in this PR or one of the higher level ones in the PR stack?

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
```diff
@@ -0,0 +1,512 @@
...
+       ``prevGrid``) beyond the user-declared arguments.  TRT passes exactly
+       ``num_used_params`` args; strip the extras and fix the trailing comma.
+    """
+    # Rewrite 1: PTX version downgrade (9.x → 9.0).
+    # Triton on CUDA 13.x emits `.version 9.1`; TRT's QDP PTX JIT caps at 9.0.
+    # Only touch lines where the major version is 9 and the minor is > 0.
+    lines = ptx.split("\n")
+    result_pre: List[str] = []
+    for line in lines:
+        if line.startswith(".version "):
```
narendasan: Is this really the best we can do? just spoof the PTX version?

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
```diff
@@ -0,0 +1,512 @@
...
+            line = re.sub(
+                r"^(\.version\s+)9\.([1-9]\d*)",
+                r"\g<1>9.0",
+                line,
+            )
+        result_pre.append(line)
+    ptx = "\n".join(result_pre)
+
+    # Rewrites 2 & 3: param reorder + trailing-param strip.
+    num_ptrs = len(param_binding_indices)
```
narendasan: I am a bit unsure about how much string manipulation is happening here

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
```diff
@@ -0,0 +1,512 @@
...
+                f"but kernel has {expected_num_scalars} scalar args (depends on descriptor ranks for this invocation)"
+            ),
+        )
+
+    # triton.compile raises a mix of Exception subclasses (CompilationError,
+    # subprocess.CalledProcessError, RuntimeError) depending on the failure mode;
+    # a broad catch is necessary here to give a structured diagnostic in all cases.
+    try:
+        compiled = triton.compile(
+            triton.compiler.ASTSource(
```
narendasan: What does the recorder do if we just call ASTSource in the end anyway?

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
```diff
@@ -0,0 +1,512 @@
...
+            op=qdp_symbol,
+            stage="aot_impl",
+            backend=backend,
+            msg=f"triton.compile failed for kernel '{recorder.real_kernel.__name__}': {exc}",
+        ) from exc
+
+    # 6. Build KernelLaunchParams (grid can be symbolic; QDP evaluates at runtime).
+    # Pattern: grid from shape-derived expressions, block/shared from compiled metadata,
+    # extra_args from shape_expr-derived scalars; only tile sizes in constexprs.
+    launch = trtp.KernelLaunchParams()
```
narendasan: Cant we centralize the calculation of the trtp.KernelLaunchParams instead of doing per kernel source?

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
```diff
@@ -0,0 +1,512 @@
...
+                    op=qdp_symbol,
+                    stage="aot_impl",
+                    backend="triton",
+                    msg=(
+                        f"Triton config values must be integers for AOT compilation; "
+                        f"got {k}={v!r} ({type(v).__name__})"
+                    ),
+                )
+            parts.append(f"{k}{v}")
+        suffix = "_".join(parts)
```
narendasan: Necessary mutations of the ptx should be done in discrete atomic functions so its easy to understand whats happening instead of inline in a 300 line function. So for example here, we can call this `specialize_ptx_name` or something

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
```diff
@@ -0,0 +1,512 @@
...
+#      as a plain string — any change to Triton's param naming convention silently
+#      produces incorrect PTX.
+#   2. It scans for ``.entry {kernel_name}(`` as a literal string; multi-line or
+#      differently-formatted entry declarations will not be recognised.
+#   3. Unused trailing params (constexprs not referenced in the body) are stripped
+#      by counting references — this is correct only if Triton doesn't reuse param
+#      indices in non-obvious ways.
+# The proper fix is for TRT's QDP AOT API to expose a parameter-order remapping
+# mechanism so that post-compilation PTX rewriting is not needed.
+def _fix_triton_ptx_for_trt(
```
narendasan: To be honest, I would much rather reject unsupportable ptx and provide strong error messages to the user on how they can adjust their kernel implementation to work with TRT rather than doing this patching.

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
```diff
@@ -0,0 +1,1295 @@
...
+    num_outputs: int = 1
+    attrs: Dict[str, Any] = field(default_factory=dict)
+    # Tensor weights are excluded from __hash__ and __eq__ because torch.Tensor
+    # is not hashable.  Identity / equality of a descriptor is captured by
+    # op_name (which already encodes num_weights via derive_impl_id).
+    weights: Dict[str, "torch.Tensor"] = field(
+        default_factory=dict, hash=False, compare=False
+    )
+
+    def register_dynamo_plugin(
```
narendasan: This is mostly redundant, all we need is `kernel PTX` -> `AOT IMPL`

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
```diff
@@ -0,0 +1,1295 @@
...
+        meta_sig = inspect.signature(self.meta_impl)
+        param_names = list(meta_sig.parameters.keys())
+
+        first_spec = self.specs[0]
+        first_config = first_spec.configs[0] if getattr(first_spec, "configs", None) else {}
+        _launch = first_spec.launch_fn
+        _meta = self.meta_impl
+        _num_outputs = self.num_outputs
+
+        # CuTeDSL @cute.jit functions expect cute.Tensor, not torch.Tensor.
```
narendasan: I think we need this only if users are using the kernel in TensorRT and not in PyTorch as custom op, I dont expect this to be very common. Even in eager execution of manually fused regions in future PRs, I would think the expectation is that there should be some custom op that we generate a fx pass to insert

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
```diff
@@ -0,0 +1,1295 @@
...
+
+# LIMITATION: _infer_num_outputs probes meta_impl with rank-1 dummy tensors to count
+# outputs.  It returns 1 (silently) when:
+#   (a) a dispatch mode (e.g. FakeTensorMode during torch.export) is active, or
+#   (b) meta_impl raises with rank-1 dummies (shape-sensitive implementations).
+# In either case a multi-output plugin will be silently treated as single-output,
+# causing a downstream shape mismatch at TRT engine build time.  Users with
+# shape-sensitive or rank-sensitive meta_impl functions must pass num_outputs
+# explicitly to custom_plugin() to avoid this.
+def _infer_num_outputs(meta_impl: Callable[..., Any]) -> int:
```
narendasan: We have this already for custom ops

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
```diff
@@ -0,0 +1,1295 @@
...
+    for d in shape_expr:
+        # Check for TRT IDimensionExpr which are symbolic
+        if hasattr(d, 'is_constant') or hasattr(d, 'get_constant_value'):
+            return True  # It's a TRT dimension expression
+        if not isinstance(d, (int, float)):
+            return True
+    return False
+
+
+def _build_meta_impl_desc_fn(
```
narendasan: same here, see: https://github.com/pytorch/TensorRT/blob/2e26bfa13ddd565396f0455591bb2c275c225416/py/torch_tensorrt/dynamo/conversion/plugins/_generate_plugin.py#L130

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
```diff
@@ -0,0 +1,1295 @@
...
+    return _desc
+
+
+# ---------------------------------------------------------------------------
+# Autotune function builder (format / tactic combinations via @trtp.autotune)
+# ---------------------------------------------------------------------------
+
+# Mapping from TRT dtype enum values to the string tokens accepted by
+# AutoTuneCombination.  Populated lazily only when TRT is available.
+_TRT_DTYPE_TOKEN: Dict[Any, str] = (
```
narendasan: Just add autotune support to the existing plugin system

narendasan (commented):
Marked a bunch of immediate stuff that stood out. the TL;DR is theres a ton of re-implementation here and some stuff (particularly the triton stuff) seems hacky. 

My general recommendation is focus on adding `aot_impl` and perhaps `autotune` (@bowang007 did you look at autotune at all?) to the **existing** plugin system which should handle the rest, rather than essentially making a whole second version. If there are limitations you are running into with what is there then I believe that is where the technical discussion should be centered. Like perhaps the locking system (cc: @bowang007)

I would also recommend trying to make the systems for defining launch parameters more generically applicable, then we dont need to do as much work to say add support for pallas or nvrtc kernels. 

Also I would recommend that all the sort of kernel encapsulation stuff would nicely fit in a namespace called `torch_tensorrt.kernels` or `torch_tensorrt.dynamo.kernels` It would be immediately obvious what you would use the namespace for then.

py/torch_tensorrt/annotation/_specs.py
BowenFu: @narendasan I would expect one TVM path later after TensorRT actually supports that. NVRTC is too flexible which we need to add lots of constraints on it or bridge work to make it work with the existing QDP path.

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
BowenFu: @narendasan actually this is a limitation of TRT QDP AOT path. So if we do not adjust it, most triton kernels will fail. All the workarounds can be cleaned up if we use the TVM FFI for all these kernel providers and plugin runtime.

py/torch_tensorrt/annotation/_layer_metadata.py
BowenFu: layer metadata is not necessary for the functionality. Will defer it to a separate PR. It is used to map back TRT engine layers back to torch codes, which would be used by some other high level annotations (which will come later).

py/torch_tensorrt/dynamo/conversion/plugins/_custom_op.py
BowenFu: Updated as suggested.

py/torch_tensorrt/annotation/_specs.py
BowenFu: Pushed the Metadata related changes to a future PR.

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
BowenFu: This is TRT's PTX loader's limitation. We can remove this after TRT supports PTX version > 9.0.

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
BowenFu: I've rewritten the functions to split "version downgrade" and "signature fix".

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
BowenFu: The recorder is to capture how to map from the launch fn's parameters (which we require to minic the op's sig, especially the tensor order) to the kernel parameters and launch parameters, including some symbolic computation from shape to strides/gridsize/blocksize and so on.

tests/py/annotation/unit/test_specs.py
BowenFu: unit tests does not test accuracy. Please check e2e tests in integration folder.

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
BowenFu: Done

py/torch_tensorrt/annotation/_specs.py
narendasan: There are users of the plugin system who are explicitly using NVRTC which is why I ask

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
BowenFu: Added inline documentation to explain.

tests/py/annotation/unit/test_specs.py
BowenFu: Update the file to make it more straightfoward.

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
BowenFu: Updated to reuse and enhance the existing version.

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
BowenFu: Updated to reuse and enhance the existing version.

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
BowenFu: Hi Naren, cutedsl kernels only accept cute.Tensor, so we need to convert from torch.Tensor to it even for capturing the launch params.

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
BowenFu: Done

py/torch_tensorrt/annotation/_specs.py
BowenFu: We can support that in the future. For now, they can register to trt plugin registery manually, and use the incoming tta.plugin annotation (on par with tta.custom_plugin) to refer to it.

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
BowenFu: fixed

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
BowenFu: fixed

py/torch_tensorrt/annotation/_layer_metadata.py
BowenFu: Removed.

py/torch_tensorrt/annotation/_custom_plugin/_lowering.py
BowenFu: Fixed

py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
```diff
@@ -0,0 +1,754 @@
...
+
+# LIMITATION (short-term workaround): PTX version downgrade is a best-effort
+# text substitution.  CUDA 13.1 tileiras emits .version 9.1 but the driver in
+# our container (CUDA 13.0) only JIT-compiles up to PTX 9.0.  Lowering the
+# declared version by editing the .version line is safe only if the PTX body
+# uses no ISA features introduced after the target version.  This will silently
+# produce incorrect PTX if a future tileiras version emits instructions that
+# require a higher PTX version than the declared one after downgrade.
+# Long-term fix: align the driver and CUDA toolkit versions so no downgrade is
+# needed.
```
narendasan, line 177: I still dont think this is a good idea. It would be better to give people clear instructions on exactly how to generate the right PTX version

py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
```diff
@@ -0,0 +1,754 @@
...
+    params_raw = [p.strip().rstrip(",") for p in param_block.split(",") if p.strip()]
+    if len(params_raw) != len(runtime_order):
+        return ptx
+    reordered = [params_raw[runtime_order[i]] for i in range(len(runtime_order))]
+    new_param_block = ",\n    ".join(reordered)
+    new_entry = prefix + "\n    " + new_param_block + "\n)"
+    return ptx[: m.start()] + new_entry + ptx[m.end() :]
+
+
+# LIMITATION (fragile PTX rewrite): _reorder_cutile_ptx_for_trt extracts PTX
```
narendasan, line 236: Same thing here, this is a very extreme solution for something we could have a convention for that we inform users of

py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
```diff
@@ -0,0 +1,754 @@
...
+            op=qdp_symbol,
+            stage="aot_impl",
+            backend=backend,
+            msg=f"failed to import cuTILE internals (compile_tile/CompilerOptions) for op '{qdp_symbol}': {exc}",
+        ) from exc
+
+    # Ensure tileiras is findable: add the Python-packaged nvidia/cu13/bin if needed.
+    import shutil as _shutil
+    import sysconfig as _sysconfig
+    if not _shutil.which("tileiras"):
```
narendasan, line 492: Have the user set their path correctly, dont do this for them

py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
```diff
@@ -0,0 +1,754 @@
...
+            effective_rank = 1 if rank > 2 else rank
+    else:
+        effective_rank = 1 if rank > 2 else rank
+    if rank > 1 and ptx_raw:
+        expected_n, _ = _cutile_trt_order_for_io(
+            num_inputs, num_outputs, num_scalars, rank=effective_rank
+        )
+        # Dump diagnostic PTX for rank>1 kernels and log the path.
+        diag_filename = f"cutile_rank_gt1_{kernel_name_str}_nparam_{num_ptx_params}_expected_{expected_n}.ptx"
+        dump_code_artifact("CUTILE_DUMP_DIR", diag_filename, ptx_raw, default_dir="/tmp/cutile_dump")
```
narendasan: We have a torch_tensorrt temp directory convention already for things like caching, lets reuse that so multi user systems wont break

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
```diff
@@ -0,0 +1,552 @@
...
+    dump_code_artifact,
+    is_triton_kernel,
+    run_kernel_sandbox,
+)
+from ..._recorders import TritonLaunchRecorder
+from ..._specs import TritonSpec
+from .._symbolic import SymbolicTensor, TensorRole
+
+
+def _trt_dtype_to_triton_ptr(trt_dtype: Any, qdp_symbol: str) -> str:
```
narendasan, line 81: We could add this to the dtype enum system so we dont have LUTs all over the library. we already to do the same for numpy and pytorch

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
```diff
@@ -0,0 +1,552 @@
...
+        return "i32"
+    raise TTAPluginError(
+        op=qdp_symbol,
+        stage="aot_impl",
+        backend="triton",
+        msg=f"unsupported tensor dtype {trt_dtype} for Triton kernel signature",
+    )
+
+
+# LIMITATION (fragile PTX rewrite): the three _ptx_* helpers below rewrite .param
```
narendasan, line 104: Same here, give people a validated toolchain and calling convention vs patching their ptx under the hood

py/torch_tensorrt/annotation/_custom_plugin/_aot/_triton.py
narendasan: Honestly Id rather wait for TRT to support PTX > 9.0 if this is the only option. We cant give people a validated toolchain?

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
```diff
@@ -0,0 +1,951 @@
...
+    return _tls.aot_fn_cache
+
+
+# ---------------------------------------------------------------------------
+# Public descriptor
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class CustomPluginSpec:
```
narendasan, line 97: Do we really need this class? I would assume at least for this stage that users will already register their own meta kernel and custom operator in PyTorch. The existing system already does trtp.register

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
```diff
@@ -0,0 +1,951 @@
...
+                return list(result)
+            return result
+
+        _fake_body.__signature__ = sig
+        torch.library.register_fake(op_name)(_fake_body)
+
+
+
+
+def custom_plugin(
```
narendasan, line 309: Similar here, where will this entry point be used if we already have the existing `custom_op` class

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
```diff
@@ -0,0 +1,951 @@
...
+        for i in range(num_inputs)
+    ]
+
+
+# ---------------------------------------------------------------------------
+# Descriptor function builder (shape / dtype via @trtp.register)
+# ---------------------------------------------------------------------------
+
+
+def _build_desc_fn(
```
narendasan, line 422: Also is redundant, fold missing functionality into the existing register generator

py/torch_tensorrt/dynamo/conversion/plugins/_generate_plugin.py
```diff
@@ -233,3 +358,46 @@ def generate_plugin(plugin_name: str) -> None:
             There should be existing kernels and pytorch custom operation for this plugin name.
     """
     _generate_plugin(plugin_name)
+
+
+def register_plugin_with_aot(
```
narendasan: This is unnecessary indirection, just use the one API endpoint (`register_plugin`)

narendasan (commented):
My high level feedback is I am concerned about the PTX hacks and I think we can radically simplify this PR by de-duplicating. I would say we should centralize both the existing plugin system and the new one in a `torch_tensorrt.plugin` and/or `torch_tensorrt.kernels` for the kernel wrappers. It would make it easier for people to understand

py/torch_tensorrt/annotation/_custom_plugin/_aot/_cutile.py
BowenFu: @narendasan I think we can mark the entire AOT path as experimental. Does this sound good to you? The entire annotation layer can be marked as experimental, just like https://github.com/NVIDIA/TensorRT-Incubator.

BowenFu:
Rebased onto current `main` (148 commits) and fixed all current repository hooks; the exact pinned pre-commit suite now passes. Fresh runtime/GPU CI is still required.

The remaining design decisions are:

1. Reject unsupported PTX/toolchain combinations, or retain PTX version/parameter rewrites.
2. Split kernel AOT support from `tta.custom_plugin`/weighted registration.

@narendasan @bowang007, please confirm the preferred direction. I’ll make that architectural change rather than treating metadata-only checks as validation.

BowenFu:
Rewritten around the existing automatic-plugin path as requested. The branch is now one 205-line commit: `custom_op`/`generate_plugin` accept QDP `aot_impl` and `autotune` callbacks, while the existing descriptor, JIT implementation, converter, dynamic-shape controls, and fallback remain unchanged. All annotation descriptors, backend wrappers, weighted registration, and PTX rewriting were removed. Added focused forwarding/JIT-default tests plus an end-to-end fixed-PTX AOT plugin test. Pinned repository hooks pass. The fork workflows still need maintainer approval to execute. Please re-review the reduced API and approve the workflows when ready.

BowenFu:
@bowang007 @zewenli98 — friendly ping when you have time: this PR is ready for another review after the requested rewrite and rebase. Thank you!

bowang007 (approved):
LGTM, thanks!

PR #4157, opened 2026-04-01
Multi-Device TensorRT Runtime with Native NCCL Collectives

apbose:
- C++ runtime: NCCL communicator init via c10d, rank/world_size serialization
- Python runtime: distributed support in PythonTorchTensorRTModule and TorchTensorRTModule, NCCL library auto-detection
- Conversion: native TRT DistCollective API (AllGather, ReduceScatter, AllReduce) with TRT-LLM plugin fallback
- Graph lowering: fuse c10d_functional collectives + wait_tensor into single ops
- Feature detection: native_trt_collectives flag, platform validation, graceful fallback chain
- Build: conditional NCCL compilation via torch_nccl toolchain
- Examples: tensor_parallel_simple_example.py, tensor_parallel_llama_llm.py


core/runtime/runtime.h
```diff
@@ -39,6 +39,8 @@ typedef enum {
   TARGET_PLATFORM_IDX,
   REQUIRES_OUTPUT_ALLOCATOR_IDX,
   RESOURCE_ALLOCATION_STRATEGY_IDX,
+  RANK_IDX,
+  WORLD_SIZE_IDX,
```
narendasan: Make sure to bump the ABI version

narendasan: Fields that are only sometimes applicable should be prefixed with OPTIONAL_ and need guards

core/runtime/TRTEngine.cpp
```diff
@@ -88,7 +95,15 @@ TRTEngine::TRTEngine(std::vector<std::string> serialized_info)
           serialized_info[SERIALIZED_METADATA_IDX],
           (static_cast<bool>(std::stoi(serialized_info[RESOURCE_ALLOCATION_STRATEGY_IDX]))
                ? ResourceAllocationStrategy::kDynamic
-               : ResourceAllocationStrategy::kStatic)) {}
+               : ResourceAllocationStrategy::kStatic)) {
+  // Load distributed info if available (backward compatible with older ABI versions)
```
narendasan: We dont need backwards compat unless some semantic definition changed, just bump the version

core/runtime/TRTEngine.cpp
```diff
@@ -519,6 +534,149 @@ void TRTEngine::set_resource_allocation_strategy(TRTEngine::ResourceAllocationSt
...
+bool TRTEngine::set_nccl_communicator_to_trt_context() {
+  // Set NCCL communicator on TensorRT execution context
+  // The communicator should be set from Python via set_nccl_comm() or set_process_group()
+
+  if (!exec_ctx) {
+    LOG_ERROR("Cannot set NCCL communicator: execution context is null");
+    return false;
+  }
+
+  if (this->nccl_comm == nullptr) {
```
narendasan: if it can be avoided, do not let anything be null. Should be a sentinel value or use a smart pointer

core/runtime/TRTEngine.cpp
```diff
@@ -519,6 +534,149 @@ void TRTEngine::set_resource_allocation_strategy(TRTEngine::ResourceAllocationSt
...
+    LOG_WARNING(
+        "Distributed inference enabled but no NCCL communicator set. "
+        "Call set_process_group() or set_nccl_comm() from Python first.");
+    return false;
+  }
+
+  // Set NCCL communicator on TensorRT execution context
+  try {
```
narendasan: We need like a real state machine or some true semantics here. Nothing else in the runtime just try catches

core/runtime/TRTEngine.cpp
```diff
@@ -519,6 +534,149 @@ void TRTEngine::set_resource_allocation_strategy(TRTEngine::ResourceAllocationSt
...
+
+bool TRTEngine::set_process_group_from_registry(const std::string& group_name) {
+  // Get ProcessGroup from C++ registry and extract NCCL communicator
+  // This avoids the need to pass the ProcessGroup from Python
+  LOG_INFO("TRTEngine::set_process_group_from_registry() called with group_name: " << group_name);
+  LOG_INFO("   Current rank: " << this->rank);
+  LOG_INFO("   Current world_size: " << this->world_size);
+  LOG_INFO("   Current device_id: " << this->device_info.id);
+
+  try {
```
narendasan: same here, no nested try catches, you need to provide gaurentees about the state of the system.

core/runtime/TRTEngine.h
```diff
@@ -9,12 +9,29 @@
...
+#include "NvInferVersion.h"
 #include "c10/cuda/CUDAStream.h"
 #include "torch/custom_class.h"
 
 #include "core/runtime/TRTEngineProfiler.h"
 #include "core/util/prelude.h"
 
+// TensorRT 10.16+ has native NCCL collective support via IExecutionContext::setCommunicator()
+#if NV_TENSORRT_MAJOR > 10 || (NV_TENSORRT_MAJOR == 10 && NV_TENSORRT_MINOR >= 16)
+#define TRT_HAS_NATIVE_NCCL 1
```
narendasan, line 21: Does this work for say jetson or windows?

core/runtime/TRTEngine.h
```diff
@@ -9,12 +9,29 @@
...
+
+// Full TRT NCCL collectives support requires both:
+// 1. PyTorch built with NCCL (USE_C10D_NCCL defined via Bazel)
+// 2. TensorRT 10.16+ (TRT_HAS_NATIVE_NCCL defined above)
+#if defined(USE_C10D_NCCL) && defined(TRT_HAS_NATIVE_NCCL)
+#define ENABLE_TRT_NCCL_COLLECTIVES 1
+#endif
+
+#ifdef ENABLE_TRT_NCCL_COLLECTIVES
+#include <nccl.h>
```
narendasan: You are pulling nccl.h but this is not a bazel dep, where are you getting it from?

core/runtime/TRTEngine.h
```diff
@@ -196,6 +213,22 @@ struct TRTEngine : torch::CustomClassHolder {
   bool use_output_allocator_outputs = false; // users specify to use output allocator
   std::shared_ptr<DynamicOutputAllocator> output_allocator;
 
+  // Member variables for distributed inference (-1 indicates non-distributed mode)
+  int64_t rank = -1;
+  int64_t world_size = -1;
+
+  // Set rank and world_size for distributed inference
+  void set_rank(int64_t rank_val);
```
narendasan: Why are these apis always available but not the nccl ones, is there a way to use MD-TRT without the below apis?

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -280,6 +288,222 @@ def set_default_device_memory_budget(self) -> int:
...
         return self._set_device_memory_budget(budget_bytes)
 
+    # Distributed functions
+    @property
+    def is_distributed(self) -> bool:
+        """Check if this module is configured for distributed execution."""
+        return self.rank >= 0 and self.world_size > 1
+
+    @property
+    def has_native_trt_collectives(self) -> bool:
```
narendasan: why does this matter as a property of a runtime module?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -146,6 +151,11 @@ def __init__(
         self.requires_output_allocator = requires_output_allocator
         self.dynamically_allocate_resources = settings.dynamically_allocate_resources
         self.symbolic_shape_expressions = symbolic_shape_expressions
+        self.rank = rank
+        self.world_size = world_size
```
narendasan: We write these values down twice both in TRTEngine and in this class which raises a syncing problem, choose an owner, if its TRTEngine, use that as source of truth and dont make a second copy in this class just use the engine or do atomic updates

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -146,6 +151,11 @@ def __init__(
...
+        self._nccl_setup_done = (
```
narendasan: I want this module as stateless as possible, we cant be checking this stuff in every module in the graph.

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -329,6 +341,159 @@ def set_pre_allocated_outputs(self, enable: bool) -> None:
     def set_use_output_allocator(self, enable: bool) -> None:
         self.engine.use_output_allocator_outputs = enable
 
+    def _auto_init_distributed(self) -> None:
```
narendasan: I thought we talked about how its on the user to setup nccl properly, the runtimes should not care

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -329,6 +341,159 @@ def set_pre_allocated_outputs(self, enable: bool) -> None:
...
+                if world_size is None:
+                    world_size = dist.get_world_size()
+            else:
+                raise RuntimeError(
+                    "torch.distributed is not initialized and rank/world_size not provided. "
+                    "Call dist.init_process_group() first or provide rank/world_size explicitly."
+                )
+
+        # Set on C++ TRTEngine
+        self.engine.set_rank(rank)
```
narendasan: right here is a desync, the TRTEngine gets the updated value and you dont change the modules value

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
narendasan, line 1: Honestly i want very little of this nccl stuff in this class. Things related to setup of the nccl env is on the user and things related to passing information to the TRT engine are in C++, this class just transparently passes information

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -329,6 +341,159 @@ def set_pre_allocated_outputs(self, enable: bool) -> None:
...
+            raise RuntimeError(
+                "Native TRT NCCL collectives are not available. "
+                "Requires TensorRT 10.16+ and PyTorch built with NCCL support."
+            )
+        if self.engine is None:
+            raise RuntimeError(
+                "Engine has not been setup yet. Call setup_engine() first."
+            )
+
+        setup_nccl_library()
```
narendasan: Give something for users to call that takes their torch distributed and links it to TRT that they do once instead of stuffing it in here.

py/torch_tensorrt/_utils.py
```diff
@@ -299,6 +299,39 @@ def load_and_initialize_trtllm_plugin(plugin_lib_path: str) -> bool:
...
+        torchtrt_runtime_lib_path: Path to libtorchtrt_runtime.so (runtime-only library)
+
+    Returns:
+        bool: True if native TRT collectives are available, False otherwise.
+    """
+    try:
+        # Load the runtime library to register torch ops
+        # Prefer full library if available, otherwise runtime-only
+        if torchtrt_lib_path and os.path.isfile(torchtrt_lib_path):
+            torch.ops.load_library(torchtrt_lib_path)
```
narendasan, line 324: You could even do the nccl linking here when we check if its available

tools/llm/tensor_parallel_llama_llm.py
```diff
@@ -0,0 +1,340 @@
...
+    if args.model_precision == "FP16":
+        enabled_precisions = {torch.float32}
+        use_fp32_acc = True
+        use_explicit_typing = True
+    elif args.model_precision == "BF16":
+        enabled_precisions = {torch.bfloat16}
+    else:
+        enabled_precisions = {torch.float32}
+
+    # torch.export does not support DTensor-parallelized models (sharding propagation
```
narendasan, line 171: Wasnt the problem that we were not decomposing linear? add that decomposition if we are using `distributed_trace_mode`

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
narendasan, line 1: pull the nccl stuff out, make an api for people to link to TRT once and then the runtimes can transparently handle it

core/runtime/runtime.h
```diff
@@ -39,6 +39,8 @@ typedef enum {
   TARGET_PLATFORM_IDX,
   REQUIRES_OUTPUT_ALLOCATOR_IDX,
   RESOURCE_ALLOCATION_STRATEGY_IDX,
+  RANK_IDX,
```
narendasan: You need a field that is like `IS_MD_ENGINE` that controls these optional values

py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -358,6 +357,41 @@ def convert_module(
...
+            world_size = dist.get_world_size()
+        else:
+            # Fallback to environment variables
+            rank = int(os.environ.get("RANK", -1))
+            world_size = int(os.environ.get("WORLD_SIZE", -1))
+
+        if rank >= 0 and world_size > 0:
+            logger.info(
+                f"Creating TRT module for distributed execution: rank={rank}, world_size={world_size}"
+            )
```
narendasan: Are engines tied to a particular rank and world size or is there ever a case where I might change these at runtime?

narendasan: Basically we are changing the function signature of the runtime modules to fetch information that we pull from the environment. To me this code can just be in module constructors (in C++ for TRTEngine) given that we are just taking information from torch distributed

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -134,6 +135,8 @@ def __init__(
         requires_output_allocator: bool = False,
         symbolic_shape_expressions: Optional[Dict[str, List[Dict[str, Any]]]] = None,
         _debugger_config: Optional[DebuggerConfig] = None,
+        rank: int = -1,
```
narendasan: The one property we need is if the engine needs md-trt, the rank and world_size can be fetched internally

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -280,6 +288,222 @@ def set_default_device_memory_budget(self) -> int:
...
+                    logger.warning(
+                        f"set_communicator returned False (rank={self.rank})"
+                    )
+            except AttributeError:
+                logger.warning("TensorRT context does not support set_communicator")
+            except TypeError as e:
+                logger.error(f"Failed to set NCCL communicator: {e}")
+                raise
+
+    def get_nccl_communicator(self) -> Optional[Any]:
```
narendasan: we dont need getters and setters if you want people to use this field just make it public

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -280,6 +288,222 @@ def set_default_device_memory_budget(self) -> int:
...
+        """Get the NCCL communicator if set."""
+        return self._nccl_comm
+
+    def setup_nccl(self, use_pytorch_comm: bool = True) -> None:
```
narendasan: This should be on the user

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -280,6 +288,222 @@ def set_default_device_memory_budget(self) -> int:
...
+            logger.debug(f"Rank {rank} created NCCL unique ID ({len(uid_bytes)} bytes)")
+        else:
+            uid_tensor = torch.zeros(128, dtype=torch.uint8, device="cuda")
+
+        dist.broadcast(uid_tensor, src=0)
+        logger.debug(f"Rank {rank} received NCCL unique ID")
+
+        uid = nccl.UniqueId.from_bytes(bytes(uid_tensor.cpu().numpy()))
+
+        comm = nccl.Communicator.init(world_size, rank, uid)
```
narendasan: Is this a singleton owned by NCCL or can you have multiple, can a user just give this to us?

core/runtime/TRTEngine.cpp
```diff
@@ -519,6 +534,149 @@ void TRTEngine::set_resource_allocation_strategy(TRTEngine::ResourceAllocationSt
...
+        "(rank="
+        << this->rank << ", device=" << this->device_info.id << ")");
+    return true;
+  } catch (const std::exception& e) {
+    LOG_ERROR("Failed to set NCCL communicator on execution context: " << e.what());
+    return false;
+  }
+}
+
+void TRTEngine::init_nccl_comm(const std::string& group_name) {
```
narendasan: why do we need a method that transparently calls another method?

core/runtime/TRTEngine.cpp
```diff
@@ -519,6 +534,149 @@ void TRTEngine::set_resource_allocation_strategy(TRTEngine::ResourceAllocationSt
...
+    }
+
+    if (!backend) {
+      LOG_ERROR("ProcessGroup '" << group_name << "' does not have an NCCL backend");
+      return false;
+    }
+    LOG_INFO("   Got NCCL backend from ProcessGroup");
+
+    // Cast the backend to ProcessGroupNCCL
+    auto* nccl_pg = dynamic_cast<c10d::ProcessGroupNCCL*>(backend.get());
```
narendasan: Try to use smart pointers if possible

narendasan (commented):
The high order bits are there is too much NCCL management happening all over the place and state is not managed tightly enough. Fundamentally setting up NCCL is not our job. The user tells us the information we need (the communicator and what rank the process is on) and we trust them to do the rest. 

The runtime modules do not need to care about if nccl is set up as long as the information needed to deserialize and setup the engine is available. 

Also seems like the C++ runtime only uses NCCL from c10d but then theres other code which falls back to nccl from the nccl python bindings. What do we actually want to support? just c10d? or both?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -88,6 +91,8 @@ def __init__(
         weight_name_map: Optional[dict[Any, Any]] = None,
         requires_output_allocator: bool = False,
         symbolic_shape_expressions: Optional[Dict[str, List[Dict[str, Any]]]] = None,
+        rank: int = -1,
```
narendasan: Think all we need is that this is an md_engine and we can fetch the info from torch.distributed or the env var internally

core/runtime/TRTEngine.cpp
apbose: The rank and world_size fields are optional — they are only populated for distributed engines. The size/empty guards ensure backward compatibility with non-distributed engines and older ABI versions where these fields don't exist. Since these fields are not always present, bumping the ABI version would unnecessarily break non-distributed engines that don't use them.
Do you want something like we bump the ABI version and introduce a field to handle the rank and world_size fields?  Something like IS_MD_ENGINE as you said? Would that not just be a different approach?

PR #4164, opened 2026-04-04
Merged two runtimes

cehongwang:
# Description

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/runtime/meta_ops/register_meta_ops.py
```diff
@@ -226,6 +260,31 @@ def fake_tensorrt_execute_engine(
         )
 
 
+@torch.library.register_fake("tensorrt::execute_engine_python")  # type: ignore
```
narendasan: Why do we need a seperate operator for this arent we just changing the implementation of TRTEngine to either be python or C++?

narendasan: The problem is that if its a separate op then you can't interchange between C++ and Python only builds

py/torch_tensorrt/dynamo/runtime/_PythonTRTEngine.py
```diff
@@ -0,0 +1,668 @@
...
+            need_cudagraphs_reset,
+        )
+
+
+# ---------------------------------------------------------------------------
+# PythonTRTEngine
+# ---------------------------------------------------------------------------
+
+
+class PythonTRTEngine:
```
narendasan: I think this class should be "TRTEngine" and only "registered" if the C++ runtime is unavailable. It should also be a valid script object so that the same operator works with the Python and C++ versions of the objects and it should uses the exact same APIs as the ones we expose in the JIT_hooks file

py/torch_tensorrt/dynamo/runtime/_PythonTRTEngine.py
```diff
@@ -0,0 +1,668 @@
...
+        logger.debug(
+            f"Using the standard execution runtime mode with cudagraphs={torch_tensorrt.runtime.get_cudagraphs_mode()}."
+        )
+        return self._execute_standard(contiguous_inputs)
+
+
+register_opaque_type(PythonTRTEngine, typ="reference")
+
+
+@torch.library.custom_op(  # type: ignore[misc]
```
narendasan, line 819: Same thing here. this operator should only get registered if the C++ library is not available and it should take the name of the C++ op

py/torch_tensorrt/dynamo/runtime/_PythonTRTEngine.py
```diff
@@ -0,0 +1,668 @@
...
+    "tensorrt::execute_engine_python", mutates_args=()
+)
+def execute_engine_python(
+    input_tensors: List[torch.Tensor], engine: PythonTRTEngine
+) -> List[torch.Tensor]:
+    outputs = engine.execute(input_tensors)
```
narendasan: Would rather use a struct + function design rather than some masked call to a method similar to the c++ structure

py/torch_tensorrt/dynamo/runtime/_RuntimeBackendSelection.py
narendasan, line 1: Its cool that we have this but we should look into if there is a way to drop / mask registrations to change the runtime implementation rather than relying on distinct graph constructions

py/torch_tensorrt/dynamo/runtime/_serialized_engine_layout.py
```diff
@@ -0,0 +1,70 @@
...
+
+from __future__ import annotations
+
+from typing import Any, Dict, List, Union
+
+import tensorrt as trt
+import torch
+import torch_tensorrt
+
+ABI_VERSION = "8"
```
narendasan, line 18: We need some sort of consistency check with the C++ runtime to make sure the format is the same

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -4,77 +4,71 @@
...
-    Union[str, bytes]
-]  # Aligned with  //core/runtime/register_jit_hooks.cpp
 SerializedTorchTensorRTModuleFmt = Tuple[
+    str,
+    Optional[SerializedTensorRTEngineFmt],
+    List[str],
+    List[str],
+    Optional[str],
+]
+# Checkpoints written before the trailing ``runtime_backend`` slot used four elements.
```
narendasan: What is this for?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -142,10 +108,17 @@ def __init__(
         self.settings = copy.deepcopy(settings)
         self.weight_name_map = weight_name_map
         self.serialized_engine = serialized_engine
-        self.engine = None
+        self.engine: Optional[Any] = None
```
narendasan: Shouldnt this be bytes?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -154,6 +127,21 @@ def __init__(
         ):
             self.setup_engine()
 
+    def _require_engine(self) -> Any:
```
narendasan: Im not too big a fan of having a loose state we need to check all the time. We should not have to defensively program if we are clear about when things can happen. For instance, setup engine is called by either lazy_engine init or constructor, so why would there be a case where we have no engine at inference time

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -250,6 +242,16 @@ def setup_engine(self) -> None:
         """
         if self.engine is not None:
             return
+
+        if self._is_python_runtime:
+            self.engine = PythonTRTEngine(
```
narendasan: Yeah we should be trying to monitor torch bind registration and register a class if there is no C++ api rather than two code paths

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -259,167 +261,175 @@ def encode_metadata(self, metadata: Any) -> str:
         return encoded_metadata
 
     @staticmethod
-    def decode_metadata(encoded_metadata: bytes) -> Any:
-        dumped_metadata = base64.b64decode(encoded_metadata.encode("utf-8"))
-        metadata = pickle.loads(dumped_metadata)
-        return metadata
+    def decode_metadata(encoded_metadata: bytes | str) -> Any:
+        if isinstance(encoded_metadata, str):
```
narendasan: Why was this rewritten?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -259,167 +261,175 @@ def encode_metadata(self, metadata: Any) -> str:
...
+        return (
+            self.name,
+            None,
+            self.input_binding_names,
+            self.output_binding_names,
+            None,
+        )
 
-    def set_extra_state(self, state: SerializedTorchTensorRTModuleFmt) -> None:
+    def set_extra_state(self, state: TorchTensorRTModuleExtraState) -> None:
```
narendasan: Why are we changing any of this it should be the same

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -259,167 +261,175 @@ def encode_metadata(self, metadata: Any) -> str:
...
+        else:
+            if not ENABLED_FEATURES.torch_tensorrt_runtime:
+                raise NotImplementedError("Torch-TensorRT Runtime is not available")
+            self.engine = torch.classes.tensorrt.Engine(serialized_engine_info)
+
+        self.engine.set_output_tensors_as_unowned(
+            metadata["output_tensors_are_unowned"]
+        )
+
+    def __del__(self) -> None:
```
narendasan: Is this necessary, cant we just use __del__ in the actual TRTEngine class?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
narendasan, line 1: Generally theres a lot of changes in this file I dont really understand why we are changing. Isnt the entire point of this feature to detect when the C++ runtime is not available and drop in compatible Python implementations of C++ runtime APIs? rather than just fold two separate implementations into one class.

narendasan (commented):
Just reviewed the core stuff for now. I think this mostly is not really solving the issue. The core idea is that we want to have a Python implementation of Torchbind endpoints (execute_engine / TRTEngine) that lets us run the same programs with either standard torch-trt or python only rather than just having two implementations that are kind of mixed together

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -4,77 +4,52 @@
...
+    serialize_binding_names,
+    serialize_device_info,
+)
 
 logger = logging.getLogger(__name__)
 
-SerializedTensorRTEngineFmt = List[
-    Union[str, bytes]
-]  # Aligned with  //core/runtime/register_jit_hooks.cpp
 SerializedTorchTensorRTModuleFmt = Tuple[
```
narendasan, line 35: This should be with the other serialization info

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -170,30 +123,36 @@ def _pack_engine_info(self) -> List[str | bytes]:
...
         }
-        target_platform = (
-            Platform.current_platform()
-            if not self.settings.enable_cross_compile_for_windows
-            else Platform.WIN_X86_64
-        )  # Change to match target for engine
 
         engine_info: List[str | bytes] = [""] * SERIALIZATION_LEN
-        engine_info[ABI_TARGET_IDX] = torch.ops.tensorrt.ABI_VERSION()
+        engine_info[ABI_TARGET_IDX] = (
```
narendasan, line 135: We should centralize this so its like `generate_engine_info` and will choose the right source internally

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -266,16 +231,19 @@ def decode_metadata(encoded_metadata: bytes) -> Any:
 
     def get_extra_state(self) -> SerializedTorchTensorRTModuleFmt:
         if self.engine:
+            engine_info = self._pack_engine_info()
+            assert isinstance(engine_info[ENGINE_IDX], (bytes, bytearray))
+            engine_info[ENGINE_IDX] = base64.b64encode(engine_info[ENGINE_IDX])
             return (
                 self.name,
-                self.engine.__getstate__(),
+                engine_info,
```
narendasan, line 246: Does python self.engine not have __getstate__()?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -330,73 +305,52 @@ def set_use_output_allocator(self, enable: bool) -> None:
...
-            for i in inputs
+            (value if isinstance(value, torch.Tensor) else torch.tensor(value).cuda())
+            for value in inputs
         ]
-
-        outputs: List[torch.Tensor] = torch.ops.tensorrt.execute_engine(
-            list(input_tensors), self.engine
-        )
-
+        outputs = list(torch.ops.tensorrt.execute_engine(input_tensors, self.engine))
```
narendasan: Why do we need this casting?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -330,73 +305,52 @@ def set_use_output_allocator(self, enable: bool) -> None:
...
 
-        # If the inputs are not Torch Tensors, which can occur in scenarios such as shape tensors
-        # which are outputs of a preceding Torch subgraph (where the Dynamic input may be an integer)
-        # directly cast the input to a Torch Tensor.
-        #
-        # This also avoids the need for type-checking inputs, since they are now explicitly casted to Torch tensors
         input_tensors: List[torch.Tensor] = [
-            (i if isinstance(i, torch.Tensor) else torch.tensor(i).cuda())
-            for i in inputs
+            (value if isinstance(value, torch.Tensor) else torch.tensor(value).cuda())
```
narendasan: Is this just renaming a variable?

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -483,36 +484,18 @@ def inline_trt_modules(
...
-            # set trt_node.meta with trt_module_node.meta
+            # Always embed engine data as primitive string args via no_op_placeholder
+            # so torch.export does not pickle torch.classes.tensorrt.Engine (which
+            # requires the C++ TorchBind class at load time).
+            # torch_tensorrt.load() lowers placeholders → execute_engine.
+            engine_info = trt_module._pack_engine_info()
+            engine_bytes = engine_info[ENGINE_IDX]
+            engine_info[ENGINE_IDX] = base64.b64encode(engine_bytes).decode("utf-8")
+            trt_node = gm.graph.call_function(
+                torch.ops.tensorrt.no_op_placeholder_for_execute_engine.default,
```
narendasan: Why do we need to use this? This means that we break compatibility with torch.export.load

py/torch_tensorrt/dynamo/runtime/_PythonTRTEngine.py
```diff
@@ -0,0 +1,731 @@
...
+            need_cudagraphs_reset,
+        )
+
+
+# ---------------------------------------------------------------------------
+# TRTEngine (Python implementation)
+# ---------------------------------------------------------------------------
+
+
+class TRTEngine:
```
narendasan: I really think the right way to solve this is to formally register this as a torchbind class or modern equivalent pytorch object class. Using the placeholder op is a hack to get around this

py/torch_tensorrt/dynamo/_exporter.py
narendasan: Also does this even work if you try to load and run with a python only build?

narendasan:
Like why can we use the c++ version of this https://github.com/pytorch/pytorch/blob/e2584b2554d11fda4998a8d2be6145b0eded5049/torch/_library/opaque_object.py#L73 like you do in python and then align the serialization so we dont need the no op

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
cehongwang: Yes I feel like this is better than "i" but if you think it is unnecessary I can change it back

py/torch_tensorrt/dynamo/_exporter.py
cehongwang: I use this to decouple the dependencies on the C++ build. When serializing a C++ engine and loading it in the Python runtime.

This works with python only build, python built engine and load in C++ runtime, C++ build engine and load in Python runtime.

cehongwang: Let me look into the opaque object

core/runtime/TRTEngine.h
```diff
@@ -167,7 +167,6 @@ struct TRTEngine : torch::CustomClassHolder {
   int64_t get_streamable_device_memory_budget();
   int64_t get_automatic_device_memory_budget();
   std::vector<at::Tensor> infer_outputs(std::vector<std::vector<int64_t>> input_shapes);
-  void set_pre_allocated_outputs(bool enable);
```
narendasan, line 184: Why was this removed,

py/torch_tensorrt/dynamo/runtime/meta_ops/register_meta_ops.py
```diff
@@ -199,24 +245,9 @@ def fake_tensorrt_execute_engine(
...
-    metadata = None
-    if hasattr(fake_trt_engine, "real_obj"):
-        # Wrapped C++ engine with real_obj
-        trt_engine = fake_trt_engine.real_obj
-        metadata = TorchTensorRTModule.decode_metadata(
-            trt_engine.get_serialized_metadata()
-        )
-    else:
-        metadata = TorchTensorRTModule.decode_metadata(
-            fake_trt_engine.get_serialized_metadata()
```
narendasan, line 212: Do we not have this method anymore?

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
narendasan, line 1: Can we just call this file _TRTEngine.py

py/torch_tensorrt/dynamo/runtime/_PythonTRTEngine.py
```diff
@@ -0,0 +1,831 @@
...
+            output_shapes = [list(s) for s in engine.output_shapes]
+
+        return [
+            torch.empty(
+                shape, dtype=engine.output_dtypes[i], device=input_tensors[0].device
+            )
+            for i, shape in enumerate(output_shapes)
+        ]
+
+    @torch.library.custom_op(  # type: ignore[misc]
```
narendasan: Is this still needed here?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -86,7 +86,7 @@ def cross_compile_for_windows(
     max_aux_streams: Optional[int] = _defaults.MAX_AUX_STREAMS,
     version_compatible: bool = _defaults.VERSION_COMPATIBLE,
     optimization_level: Optional[int] = _defaults.OPTIMIZATION_LEVEL,
-    use_python_runtime: bool = _defaults.USE_PYTHON_RUNTIME,
+    use_python_runtime: bool = False,  # Does nothing. Kept for backward compatibility.
```
narendasan: When we detect this is set, throw a deprecation warning

core/runtime/TRTEngine.h
cehongwang: This was never used. The setter is here: https://github.com/pytorch/TensorRT/blob/17550014e85960c510860ceb4608ecc1474e9a53/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py#L343

cehongwang:
Use the other PR

PR #4185, opened 2026-04-14
Add support for SymInt start values in slice_scatter_decomposition

shahe-nv:
# Description

Add support for SymInt start values in slice_scatter_decomposition.

This fixes a crash issue when I run with ZoomASR.

Fixes # (issue): None

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -197,8 +197,13 @@ def slice_scatter_decomposition(
     dim_size = input_tensor.shape[dim]
     device_input_tensor = input_tensor.device
 
-    start = 0 if start is None else start  # Ensure start is int
-    start = get_positive_dim(start, input_tensor.shape[dim])
+    if start is None:
+        start = 0
+    elif isinstance(start, int):
+        start = get_positive_dim(start, dim_size)
+    elif isinstance(start, torch.SymInt):
+        if start < 0:
+            start = start + dim_size
     if end is None:  # Ensure end is int
         end = dim_size
     end = (


py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -197,8 +197,13 @@ def slice_scatter_decomposition(
...
-    start = 0 if start is None else start  # Ensure start is int
-    start = get_positive_dim(start, input_tensor.shape[dim])
+    if start is None:
+        start = 0
+    elif isinstance(start, int):
+        start = get_positive_dim(start, dim_size)
+    elif isinstance(start, torch.SymInt):
+        if start < 0:
+            start = start + dim_size
     if end is None:  # Ensure end is int
```
apbose, line 207: Hi can you have a testcase for the above?

shahe-nv: Hi, I tested with my ZoomASR code. Without this, it will crash. After adding this it will run.

apbose: Thanks for the info! Generally for any change in the codebase, we recommend adding a corresponding test case to capture the corner case for future CI runs.

shahe-nv: Hi, do you know where I need to add the test? Do you have examples?

py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -197,8 +197,13 @@ def slice_scatter_decomposition(
...
 
-    start = 0 if start is None else start  # Ensure start is int
-    start = get_positive_dim(start, input_tensor.shape[dim])
+    if start is None:
+        start = 0
+    elif isinstance(start, int):
+        start = get_positive_dim(start, dim_size)
+    elif isinstance(start, torch.SymInt):
+        if start < 0:
+            start = start + dim_size
```
zewenli98, line 206: According to [the aten schema](https://github.com/pytorch/pytorch/blob/5bcf6257308559061d638a615da7c47790a8f8e4/aten/src/ATen/native/native_functions.yaml#L5717), all of `start`, `end`, and `step` could be SymInt, not just `start`. Can you fix all of them btw? I think a more general way is to support SymInt in `get_positive_dim()` like:
```
def get_positive_dim(
    dim: Union[int, SymInt, Sequence[int], Sequence[SymInt]], dim_size: int
):
```

py/torch_tensorrt/dynamo/lowering/_decompositions.py
zewenli98, line 195: the types should be changed accordingly, and step=1 by default according to the schema

py/torch_tensorrt/dynamo/lowering/_decompositions.py
zewenli98: @shahe-nv I think @apbose meant you could add unit tests in `tests/py/dynamo/lowering/test_decompositions.py`

py/torch_tensorrt/dynamo/lowering/_decompositions.py
zewenli98, line 213: not needed if step=1 by default

zewenli98, line 216: what if they are symint?

zewenli98, line 222: shouldn't they be put in the front of the function?

zewenli98:
close in favor of https://github.com/pytorch/TensorRT/pull/4236
@shahe-nv Can you try if the new PR works for ZoomASR?

PR #4184, opened 2026-04-12
feat: add dynamic shapes kernel specialization strategy for TRT-RTX

tp5uiuc:
# Description

Expose `IRuntimeConfig.setDynamicShapesKernelSpecializationStrategy()` through the Torch-TensorRT Python API for TensorRT-RTX builds.

Users can now control how shape-specialized kernels are compiled at runtime for dynamic shapes via the new `dynamic_shapes_kernel_specialization_strategy` compilation setting:
- `"lazy"` (default): Compile shape-specialized kernels in the background, use fallback until ready
- `"eager"`: Compile immediately (blocking)
- `"none"`: Always use fallback kernels, never specialize

**Depends on:** #4180 (runtime cache API — provides the `IRuntimeConfig` infrastructure)

## Type of change

- New feature (non-breaking change which adds functionality)

# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests to verify my fix or my feature
- [x] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #4180, opened 2026-04-10
feat: add runtime cache API for TensorRT-RTX

tp5uiuc:
# Description

Add runtime cache support for TensorRT-RTX JIT compilation results, replacing the timing cache which is not used by RTX (no autotuning).

TensorRT-RTX uses JIT compilation at inference time. The **runtime cache** (`IRuntimeCache`) stores these compilation results so that kernels and execution graphs are not recompiled on subsequent runs. This is analogous to the timing cache but operates at inference time rather than build time.

Fixes pytorch/TensorRT#3817

## Changes

- **Skip timing cache for RTX**: Early return in `_create_timing_cache()` and `_save_timing_cache()` when `ENABLED_FEATURES.tensorrt_rtx` is True (timing cache is a no-op in TRT-RTX)
- **Add `runtime_cache_path` setting**: New `RUNTIME_CACHE_PATH` default and `runtime_cache_path` field in `CompilationSettings`, threaded through all compile functions
- **Wire up `IRuntimeCache` in `PythonTorchTensorRTModule`**: Create `RuntimeConfig` with runtime cache on engine setup, load from disk if available, save on module destruction
- **File locking**: Uses `filelock` for concurrent access safety when multiple processes share the same cache file
- **Documentation**: Updated docstrings, compilation settings RST, and engine cache tutorial with new "Runtime Cache (TensorRT-RTX)" section

## Type of change

- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas and hacks
- [x] I have made corresponding changes to the documentation
- [x] I have added tests to verify my fix or my feature
- [x] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #4187, opened 2026-04-14
feat: add TRT-RTX native CUDA graph support

tp5uiuc:
# Description

Add `cuda_graph_strategy` compilation setting and automatic RTX-native CUDA graph integration for the Python runtime path (`PythonTorchTensorRTModule`).

TensorRT-RTX has native CUDA graph support via `IRuntimeConfig.cuda_graph_strategy`, where the JIT compiler handles capture/replay/invalidation internally. This is superior to manual `torch.cuda.CUDAGraph()` capture on RTX because:
- Manual capture freezes fallback kernels; lazy-compiled specialized kernels can never replace them
- Runtime allocation or data-dependent shapes can cause `cudaStreamBeginCapture` to fail
- The JIT compiler automatically manages graph staleness (shape changes, pointer changes, kernel readiness)

## Key changes

- New `cuda_graph_strategy` setting on `CompilationSettings` (`"disabled"` / `"whole_graph_capture"`)
- Mapped to `trt.CudaGraphStrategy` on `IRuntimeConfig` (same pattern as `dynamic_shapes_kernel_specialization_strategy`)
- **SUBGRAPH mode** (`set_cudagraphs_mode(True)`): On RTX, always use RTX-native CUDA graphs — manual capture is bypassed. If `cuda_graph_strategy` was not explicitly set, the runtime overrides to `whole_graph_capture` and warns.
- **WHOLE_GRAPH mode** (`enable_cudagraphs()` with mixed TRT + PyTorch ops): Validates all TRT engines are monolithically capturable via `context.is_stream_capturable(stream)` and `strategy != "lazy"`. If capturable, proceeds with outer monolithic capture (RTX-native disabled per-engine). If not capturable, raises `RuntimeError`.
- `_is_monolithic_capturable()` — runtime check combining stream capturability and kernel specialization strategy
- `_enable_rtx_native_cudagraphs()` — recreates execution context with `WHOLE_GRAPH_CAPTURE`
- `_check_monolithic_capturability()` in `CudaGraphsTorchTensorRTModule` for mixed graph validation

## Behavior matrix

| Graph type | cudagraph mode | RTX? | Behavior |
|---|---|---|---|
| TRT-only | SUBGRAPH | Yes | RTX-native always (override if needed) |
| TRT-only | SUBGRAPH | No | Manual capture (existing) |
| Mixed | WHOLE_GRAPH | Yes + capturable | Monolithic capture; RTX-native disabled per-engine |
| Mixed | WHOLE_GRAPH | Yes + NOT capturable | RuntimeError |
| Mixed | WHOLE_GRAPH | No | Monolithic capture (existing) |
| Any | No cudagraphs + strategy set | Yes | RTX-native runs transparently |

Depends on #4180 (runtime cache) and #4184 (dynamic shapes strategy).

## Type of change

# Checklist:


narendasan:
@tp5uiuc The Python runtime is about to get re-implemented with #4164. Make sure to align your implementation with that. Also there are existing CUDA graph facilities you might want to integrate into.

tp5uiuc:
> @tp5uiuc The Python runtime is about to get re-implemented with #4164. Make sure to align your implementation with that. Also there are existing CUDA graph facilities you might want to integrate into.

Thanks for calling this out Naren, I was unaware. I will align my implementation as you suggest, thanks!

py/torch_tensorrt/dynamo/runtime/_CudaGraphsTorchTensorRTModule.py
```diff
@@ -114,6 +114,53 @@ def __del__(self) -> None:
     def set_use_output_allocator(self, enable: bool) -> None:
         self.use_output_allocator_outputs = enable
 
+    def _check_monolithic_capturability(self, stream: torch.cuda.Stream) -> None:
```
tp5uiuc, line 117: TRT-RTX would need to avoid the 
"If your input shapes change between requests, the graph is re-recorded for each new shape. "
behavior from torch-TRT here in subgraphs mode. TRT-RTX takes care of re-capturing graphs internally if shapes have changed.

https://docs.pytorch.org/TensorRT/tutorials/runtime_opt/cuda_graphs.html

We should add an explicit test to verify this.

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -466,6 +486,32 @@ def _reset_captured_graph(self) -> None:
...
+        """
+        if not ENABLED_FEATURES.tensorrt_rtx:
+            return True  # non-RTX: assume capturable (existing behavior)
+        # Check 1: TRT-RTX stream capturability (runtime allocation, DDS, etc.)
+        if not self.context.is_stream_capturable(stream.cuda_stream):
+            return False
+        # Check 2: Lazy kernel specialization would invalidate captured graph
+        if self.settings.dynamic_shapes_kernel_specialization_strategy == "lazy":
+            return False
+        return True
```
tp5uiuc, line 503: Refactor to use any(conditions) rather than individual checks.

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -309,6 +318,10 @@ def setup_engine(self) -> None:
 
         if ENABLED_FEATURES.tensorrt_rtx:
             self._setup_runtime_config()
+            self._rtx_native_cudagraphs = (
+                ENABLED_FEATURES.tensorrt_rtx
```
tp5uiuc, line 322: ENABLED_FEATURES.tensorrt_rtx is already true, don't need to double check.

py/torch_tensorrt/dynamo/runtime/_PythonTorchTensorRTModule.py
```diff
@@ -587,7 +652,7 @@ def run_standard_execution() -> torch.Tensor | Tuple[torch.Tensor, ...]:
                 ), f"Wrong number of inputs, expect {len(self.input_names)} get {len(contiguous_inputs)}."
 
                 self.setup_input_tensors(
```
tp5uiuc, line 654: Add a comment here that is a linear version of

When runtime_config.cuda_graph_strategy = WHOLE_GRAPH_CAPTURE is in effect, TRT-RTX's JIT compiler handles graph
  capture/replay inside execute_async_v3(). It tracks the bound pointer + shape and re-captures internally when either changes. So the manual _input_buffers static-pointer dance is unnecessary in the base implementation — and
  skipping it actually avoids an extra clone() + copy_() on every call. However, this is a potential optimization opportunity.

tp5uiuc:
Closing the stale MR.

PR #4181, opened 2026-04-10
Rather than passing a power value of 1 to the scale op, send None.

SandSnip3r:
# Description

Rather than passing 1 for the exponent for the `scale` op, pass None. This will save a wasteful elementwise `pow` op being run.
This also prevents us from hitting a bug in TensorRT-RTX where `pow(x,y)` where `x` is negative returns NaN.

Fixes #3699 

## Type of change

- Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas and hacks
- [x] I have made corresponding changes to the documentation
- [x] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #4192, opened 2026-04-16
fix(test): enable TRT-RTX refit and engine cache tests

### LINKED ISSUE #3752: 🐛 [Bug] Refitter test failed when constant fold is disabled
##  Bug Description

Since constant fold is giving the nan issue in RTX,
https://github.com/pytorch/TensorRT/issues/3699
I have disabled constant fold which caused the refit test failed due to missing eps weights

From the code all the eps weights became unknown because we only capture weight/bias/running_mean/running_var:
https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py#L540

(Pdb) refitter.get_missing_weights()
[‘layer3.0.downsample.1/_native_batch_norm_legit_no_training_12_eps CONSTANT’, ‘layer4.1.bn2/_native_batch_norm_legit_no_training_19_eps CONSTANT’, ‘layer1.1.bn1/_native_batch_norm_legit_no_training_3_eps CONSTANT’, ‘layer3.0.bn1/_native_batch_norm_legit_no_training_10_eps CONSTANT’, ‘layer1.0.bn2/_native_batch_norm_legit_no_training_2_eps CONSTANT’, ‘layer2.0.bn1/_native_batch_norm_legit_no_training_5_eps CONSTANT’, ‘layer4.0.bn1/_native_batch_norm_legit_no_training_15_eps CONSTANT’, ‘layer4.0.bn2/_native_batch_norm_legit_no_training_16_eps CONSTANT’, ‘layer1.0.bn1/_native_batch_norm_legit_no_training_1_eps CONSTANT’, ‘layer2.0.bn2/_native_batch_norm_legit_no_training_6_eps CONSTANT’, ‘layer3.1.bn1/_native_batch_norm_legit_no_training_13_eps CONSTANT’, ‘layer3.0.bn2/_native_batch_norm_legit_no_training_11_eps CONSTANT’, ‘layer2.1.bn1/_native_batch_norm_legit_no_training_8_eps CONSTANT’, ‘layer1.1.bn2/_native_batch_norm_legit_no_training_4_eps CONSTANT’, ‘layer2.1.bn2/_native_batch_norm_legit_no_training_9_eps CONSTANT’, ‘layer4.0.downsample.1/_native_batch_norm_legit_no_training_17_eps CONSTANT’, ‘layer4.1.bn1/_native_batch_norm_legit_no_training_18_eps CONSTANT’, ‘layer3.1.bn2/_native_batch_norm_legit_no_training_14_eps CONSTANT’, ‘bn1/_native_batch_norm_legit_no_training_eps CONSTANT’, ‘layer2.0.downsample.1/_native_batch_norm_legit_no_training_7_eps CONSTANT’]

## To Reproduce

1.
2.
3.

## Environment

 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - Python version:
 - CUDA version:

tp5uiuc:
# Description

Two related cleanups that together let the previously-skipped RTX refit and engine cache tests run.

**Test changes**: #4181 removed the RTX-specific batch-norm workaround that bypassed constant folding, so the original refit bug (#3752) signature — "eps as a separate non-refittable layer" — is gone. Removing the `@unittest.skipIf(tensorrt_rtx, ...)` decorators on those tests, however, exposed a *different* latent bug along the fast-refit path that prevented the tests from actually exercising refit (they fell back to GraphModule.forward).

**Library fix**: in #3573 (June 2025) `ctx.weight_refit_map` switched from `np.ndarray` to `torch.Tensor`, but two consumer call sites kept the old `np.ndarray` API. Both went unnoticed until refit on RTX surfaced them:

1. `_TRTInterpreter._construct_refit_mapping` filtered scalar constants with `v.size == 1`. `Tensor.size` is a *method*, so the comparison was always False and `constant_mapping` was always empty — scalar constants like batch-norm `eps` never reached the cached `weight_name_map["constant_mapping"]`. Standard TRT happened to mask this because `refitter.get_missing_weights()` does not list these constants; on TRT-RTX, the stricter `unset_weights` check (#4198) flagged all of them.
2. `_refit_single_trt_engine_with_gm` rehydrated those values via `torch.from_numpy(val).cuda()`, which raises `TypeError: expected np.ndarray (got Tensor)`. This was hidden behind (1) — once `constant_mapping` actually had entries, the TypeError surfaced.

Both fixes are minimal and consistent with the post-#3573 `torch.Tensor` storage contract.

Fixes #3752

### Library
- `_TRTInterpreter.py`: `if v.size == 1` → `if v.numel() == 1` so scalar constants are kept in `constant_mapping`.
- `_refit.py`: scalar `constant_mapping` rehydration uses `val.cuda()` directly instead of `torch.from_numpy(val).cuda()`. Local renamed `np_weight_type` → `weight_dtype` to reflect the actual type.

### Tests
- **Removed RTX skip** from `test_dynamo_compile_with_refittable_weight_stripped_engine` (test_weight_stripped_engine.py)
- **Removed RTX skip** from `test_dynamo_compile_with_custom_engine_cache` and `test_dynamo_compile_change_input_shape` (test_engine_cache.py)
- **Kept RTX skip** on `test_caching_small_model` — this test fails a timing assertion (cached compilation is slower than uncached on TRT-RTX). Updated the skip message to reflect the actual reason rather than referencing #3752.
- **Fixed import ordering** in `test_weight_stripped_engine.py`: `import tensorrt as trt` must come after `import torch_tensorrt` so the `tensorrt_rtx` module alias is resolved. Added `# isort: skip` to prevent automated reordering.

## Verification (TRT-RTX, A100, nightly torch_tensorrt_rtx 2.13.0.dev20260505+cu130)

| Test | Before fix | After fix |
|---|---|---|
| `test_torch_compile_with_default_disk_engine_cache` (xfail) | XPASS, but captured log shows `AssertionError: Fast refit failed on TensorRT-RTX: 20 of 20 engine weight(s) had no entry in weight_name_map` | XPASS, no fast-refit assertion, no "is not found in weight mapping" warnings |
| `test_dynamo_compile_with_custom_engine_cache` | FAILED — `TypeError: expected np.ndarray (got Tensor)` at `_refit.py:181` | PASSED |
| `test_dynamo_compile_change_input_shape` | PASSED | PASSED |
| `test_caching_small_model` | SKIPPED (timing assertion) | SKIPPED (timing assertion) |

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_refit.py
+++ b/py/torch_tensorrt/dynamo/_refit.py
@@ -175,10 +175,10 @@ def _refit_single_trt_engine_with_gm(
             constant_mapping_with_type = {}
 
             for constant_name, val in constant_mapping.items():
-                np_weight_type = val.dtype
-                val_tensor = torch.from_numpy(val).cuda()
-                trt_dtype = dtype._from(np_weight_type).to(trt.DataType)
-                torch_dtype = dtype._from(np_weight_type).to(torch.dtype)
+                weight_dtype = val.dtype
+                val_tensor = val.cuda()
+                trt_dtype = dtype._from(weight_dtype).to(trt.DataType)
+                torch_dtype = dtype._from(weight_dtype).to(torch.dtype)
                 constant_mapping_with_type[constant_name] = (
                     val_tensor.clone().reshape(-1).contiguous().to(torch_dtype),
                     trt_dtype,
--- a/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
+++ b/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
@@ -486,7 +486,7 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
         sd = {k: v.to(torch_device) for k, v in self.module.state_dict().items()}
         weight_name_map: dict[str, Any] = {}
         weight_refit_map = self.ctx.weight_refit_map
-        constant_mapping = {k: v for k, v in weight_refit_map.items() if v.size == 1}
+        constant_mapping = {k: v for k, v in weight_refit_map.items() if v.numel() == 1}
         net = self.ctx.net
         for i in range(net.num_layers):
             layer = net[i]
--- a/tests/py/dynamo/models/test_engine_cache.py
+++ b/tests/py/dynamo/models/test_engine_cache.py
@@ -267,11 +267,6 @@ class TestEngineCache(TestCase):
     @unittest.skipIf(
         not importlib.util.find_spec("torchvision"), "torchvision not installed"
     )
-    @unittest.skipIf(
-        torch_trt.ENABLED_FEATURES.tensorrt_rtx,
-        # TODO: need to fix this https://github.com/pytorch/TensorRT/issues/3752
-        "There is bug in refit, so we skip the test for now",
-    )
     def test_dynamo_compile_with_custom_engine_cache(self):
         model = models.resnet18(pretrained=True).eval().to("cuda")
 
@@ -339,11 +334,6 @@ class TestEngineCache(TestCase):
     )
     @unittest.skipIf(
         not importlib.util.find_spec("torchvision"), "torchvision not installed"
-    )
-    @unittest.skipIf(
-        torch_trt.ENABLED_FEATURES.tensorrt_rtx,
-        # TODO: need to fix this https://github.com/pytorch/TensorRT/issues/3752
-        "There is bug in refit, so we skip the test for now",
     )
     def test_dynamo_compile_change_input_shape(self):
         """Runs compilation 3 times, the cache should miss each time"""
@@ -652,8 +642,7 @@ class TestEngineCache(TestCase):
     )
     @unittest.skipIf(
         torch_trt.ENABLED_FEATURES.tensorrt_rtx,
-        # TODO: need to fix this https://github.com/pytorch/TensorRT/issues/3752
-        "There is bug in refit, so we skip the test for now",
+        "Engine caching compilation time assertion is unreliable with TensorRT-RTX",
     )
     def test_caching_small_model(self):
         from torch_tensorrt.dynamo._refit import refit_module_weights
--- a/tests/py/dynamo/models/test_weight_stripped_engine.py
+++ b/tests/py/dynamo/models/test_weight_stripped_engine.py
@@ -269,8 +269,8 @@ class TestWeightStrippedEngine(TestCase):
         "torchvision is not installed",
     )
     @unittest.skipIf(
-            torch_trt.ENABLED_FEATURES.tensorrt_rtx,
-            "Engine caching compilation time assertion is unreliable with TensorRT-RTX",
+        torch_trt.ENABLED_FEATURES.tensorrt_rtx,
+        "Engine caching compilation time assertion is unreliable with TensorRT-RTX",
     )
     def test_dynamo_compile_with_refittable_weight_stripped_engine(self):
         pyt_model = models.resnet18(pretrained=True).eval().to("cuda")
--- a/tests/py/dynamo/models/test_weight_stripped_engine.py
+++ b/tests/py/dynamo/models/test_weight_stripped_engine.py
@@ -268,6 +268,10 @@ class TestWeightStrippedEngine(TestCase):
         not importlib.util.find_spec("torchvision"),
         "torchvision is not installed",
     )
+    @unittest.skipIf(
+            torch_trt.ENABLED_FEATURES.tensorrt_rtx,
+            "Engine caching compilation time assertion is unreliable with TensorRT-RTX",
+    )
     def test_dynamo_compile_with_refittable_weight_stripped_engine(self):
         pyt_model = models.resnet18(pretrained=True).eval().to("cuda")
         example_inputs = (torch.randn((100, 3, 224, 224)).to("cuda"),)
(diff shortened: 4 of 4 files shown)


tests/py/dynamo/models/test_engine_cache.py
```diff
@@ -673,8 +663,7 @@ def forward(self, c, d):
     )
     @unittest.skipIf(
         torch_trt.ENABLED_FEATURES.tensorrt_rtx,
-        # TODO: need to fix this https://github.com/pytorch/TensorRT/issues/3752
-        "There is bug in refit, so we skip the test for now",
+        "Engine caching compilation time assertion is unreliable with TensorRT-RTX",
```
tp5uiuc, line 666: ~I don't know about if we should skip this. engine caching should work I think.~

tp5uiuc: This is most definitely another bug, will investigate and patch in another MR.

tests/py/dynamo/models/test_weight_stripped_engine.py
```diff
@@ -278,9 +279,8 @@ def test_engine_caching_saves_weight_stripped_engine(self):
         "torchvision is not installed",
     )
     @unittest.skipIf(
-        torch_trt.ENABLED_FEATURES.tensorrt_rtx,
-        # TODO: need to fix this https://github.com/pytorch/TensorRT/issues/3752
-        "There is bug in refit, so we skip the test for now",
+            torch_trt.ENABLED_FEATURES.tensorrt_rtx,
```
cehongwang: Why extra indentation here?

cehongwang (approved):
LGTM

tests/py/dynamo/models/test_engine_cache.py
```diff
@@ -659,8 +649,7 @@ def forward(self, c, d):
...
```
tp5uiuc, line 645: Confirmed this is fixed in 1.5. TRT-RTX doesn't cache refit graphs/kernels, so upon refitting, we are essentially recompiling kernels. Refit kernels are now generated AoT with v1.5, so caching behavior can be restored.

zewenli98 (approved):
LGTM. waiting for CI

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -486,7 +486,7 @@ def _save_weight_mapping(self) -> None:
         sd = {k: v.to(torch_device) for k, v in self.module.state_dict().items()}
         weight_name_map: dict[str, Any] = {}
         weight_refit_map = self.ctx.weight_refit_map
-        constant_mapping = {k: v for k, v in weight_refit_map.items() if v.size == 1}
+        constant_mapping = {k: v for k, v in weight_refit_map.items() if v.numel() == 1}
```
tp5uiuc, line 489: This path was exercised with 
- `tests/py/dynamo/models/test_engine_cache.py::TestEngineCache::test_dynamo_compile_with_custom_engine_cache`
- `tests/py/dynamo/models/test_engine_cache.py::TestEngineCache::test_torch_compile_with_default_disk_engine_cache` (XFAILED)
in the RTX path only. In the earlier diff `Tensor.size` is a method (and it was not called), so the comparison was always False and `constant_mapping` is always empty. The current PR fixes this to allow CI to pass.

py/torch_tensorrt/dynamo/_refit.py
```diff
@@ -177,10 +177,10 @@ def _refit_single_trt_engine_with_gm(
...
 
             for constant_name, val in constant_mapping.items():
-                np_weight_type = val.dtype
-                val_tensor = torch.from_numpy(val).cuda()
-                trt_dtype = dtype._from(np_weight_type).to(trt.DataType)
-                torch_dtype = dtype._from(np_weight_type).to(torch.dtype)
+                weight_dtype = val.dtype
+                val_tensor = val.cuda()
+                trt_dtype = dtype._from(weight_dtype).to(trt.DataType)
+                torch_dtype = dtype._from(weight_dtype).to(torch.dtype)
```
tp5uiuc, line 181: As a result of fix #1 in https://github.com/pytorch/TensorRT/pull/4192/changes#r3192971184  and from previous commit (https://github.com/pytorch/TensorRT/commit/02737260d2335ce70b4b7120a997c6f84c192235) all constant_mapping items are torch tensors and not np arrays.

tp5uiuc:
[by Claude Code]
The failing checks on this PR are not regressions from this branch — they're either pre-existing main breakage that #4294 fixes, or unrelated infra issues:

- **RTX L0 dynamo core (9 fails) + non-RTX L0 dynamo core (1 fail)** — all in `tests/py/dynamo/runtime/test_000_runtime_cache.py` (`TestRuntimeCacheSetup`, `TestRuntimeCachePersistence`, `TestRuntimeCacheConcurrency`, `TestNonRTXUnchanged::test_no_runtime_config_for_standard_trt`). They reference `TRTEngine.runtime_cache_path` / `runtime_config` / `_runtime_cache`, which don't exist on main yet — that machinery is reintroduced by #4294. This PR's diff doesn't touch `test_000_runtime_cache.py` or `_TRTEngine.py`, so these are not regressions from this branch and will go green once #4294 lands (or once this branch is rebased on top of it).
- **Build Windows wheel + RTX Build Windows wheel** — `pip install torch --pre` fails with `[WinError 206] The filename or extension is too long` while extracting `torch-2.13.0.dev20260525+cu130.dist-info/licenses/third_party/kineto/.../civetweb/.../duktape-1.5.2`. Windows PATH_MAX vs torch nightly's deeply nested third_party tree — a self-hosted-runner infra issue (needs `core.longpaths` / a shorter `_temp`), not anything from this PR.

PR #4198, opened 2026-04-21
fix: detect incomplete fast refit on TRT-RTX via unset weights check

tp5uiuc:
## Summary
- Add complementary completeness check in fast refit path that compares engine weights against the mapping to detect unset weights directly
- On TRT-RTX, `get_missing_weights()` may not report weights in independent wtsEngines (TRT-RTX native refit closures have no cross-engine dependencies), causing the fast refit to silently produce wrong results with stale weights
- The new `unset_weights` check catches this regardless of the wtsEngine dependency structure
- Unwaive `test_refit_one_engine_with_wrong_weightmap` on TRT-RTX

## Test plan
- [x] `test_refit_one_engine_with_wrong_weightmap` passes on standard TRT (torch 2.13.0.dev20260416, TRT 10.16.1.11)
- [x] `test_refit_one_engine_with_wrong_weightmap` passes on TRT-RTX (torch 2.13.0.dev20260416, TRT-RTX 1.4.0.76)
- [ ] CI L0/L1 refit tests

### DIFF
--- a/py/torch_tensorrt/dynamo/_refit.py
+++ b/py/torch_tensorrt/dynamo/_refit.py
@@ -187,9 +187,18 @@ def _refit_single_trt_engine_with_gm(
                     weight_dtype, weight.data_ptr(), torch.numel(weight)
                 )
                 refitter.set_named_weights(layer_name, trt_wt_tensor, trt_wt_location)
-            assert (
-                len(refitter.get_missing_weights()) == 0
-            ), "Fast refitting failed due to incomplete mapping"
+            # Check completeness via two methods:
+            # 1. get_missing_weights(): reports weights in connected engines
+            #    that were not set.
+            # 2. Compare weights actually set vs all engine weights: catches
+            #    weights in independent engines that get_missing_weights() may not report.
+            missing_weights = refitter.get_missing_weights()
+            unset_weights = {w for w in weight_list if w not in mapping}
+            assert len(missing_weights) == 0 and len(unset_weights) == 0, (
+                f"Fast refitting failed due to incomplete mapping"
+                f" ({len(missing_weights)} missing,"
+                f" {len(unset_weights)} unset)"
+            )
 
         else:
             mapping = construct_refit_mapping(new_gm, input_list, settings)
--- a/tests/py/dynamo/models/test_model_refit.py
+++ b/tests/py/dynamo/models/test_model_refit.py
@@ -21,7 +21,8 @@ from torch_tensorrt.dynamo.lowering import (
 )
 from torch_tensorrt.logging import TRT_LOGGER
 
-import tensorrt as trt
+# must import after torch_tensorrt to resolve tensorrt_rtx alias
+import tensorrt as trt  # isort: skip
 
 assertions = unittest.TestCase()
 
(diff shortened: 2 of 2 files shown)


tests/py/dynamo/models/test_model_refit.py
```diff
@@ -4,6 +4,7 @@
 import unittest
 
 import pytest
+import tensorrt as trt
```
tp5uiuc: This is a bug with isort.

tp5uiuc: Fixd in [bcc25ae](https://github.com/pytorch/TensorRT/pull/4198/commits/bcc25ae3b19be021b381b7ea031ab1866d6b1ebe)

py/torch_tensorrt/dynamo/_refit.py
```diff
@@ -187,9 +187,18 @@ def _refit_single_trt_engine_with_gm(
...
-                len(refitter.get_missing_weights()) == 0
-            ), "Fast refitting failed due to incomplete mapping"
+            # Check completeness via two methods:
+            # 1. get_missing_weights(): reports weights in connected engines
+            #    that were not set.
+            # 2. Compare weights actually set vs all engine weights: catches
+            #    weights in independent engines that get_missing_weights() may not report.
+            missing_weights = refitter.get_missing_weights()
+            unset_weights = {w for w in weight_list if w not in mapping}
+            assert len(missing_weights) == 0 and len(unset_weights) == 0, (
```
tp5uiuc, line 197: On TRT-RTX, get_missing_weights() may return 0 even when some engine weights were not set during fast refit. This happens because TRT-RTX engines can have independent weight groups with no cross-dependencies. Setting weights in one group doesn't make other groups visible to the missing-weights check. TRT has this behavior because global optimizations like batchnorm folding occurs at a user-visible level.

Note that `get_missing_weights` will NOT report all missing weights, it only reports weights missing from that group/combination ([C++ API](https://docs.nvidia.com/deeplearning/tensorrt/latest/_static/python-api/infer/Core/Refitter.html#tensorrt.Refitter.get_missing_weights), [Python API](https://docs.nvidia.com/deeplearning/tensorrt/latest/_static/c-api/classnvinfer1_1_1_i_refitter.html#a7accd1b3424a659a8999bf47d667e60c))
```
For example, if some Weights have been set, but the engine was optimized in a way that combines weights, any unsupplied Weights in the combination are considered missing.
```
For TRT-RTX, all weights are treated independently (i.e. no combination exists).

The existing fast refit path relies solely on get_missing_weights() == 0 to verify completeness. When this check passes incorrectly, the fallback to slow refit never triggers, and the engine runs with stale weights producing wrong output. The existing check is insufficient for TRT-RTX.


Propsed fix: After the fast refit loop, also compare the set of engine weights (get_all_weights()) against the weights that were actually present in the mapping. If any engine weight was skipped (not in the mapping), the fast refit is considered incomplete and falls back to slow refit — regardless of what `get_missing_weights()` reports.

This is a one-line addition (unset_weights = {w for w in weight_list if w not in mapping}) plus the extended assert. Works correctly on both standard TensorRT and TensorRT-RTX.

cehongwang (approved):
LGTM

PR #4199, opened 2026-04-21
feat: TorchTRT Cuda generated kernels plugin support

bowang007:
# Description

This PR introduces torch_tensorrt.annotation, an experimental module for registering hand-written CUDA C++ kernels as both PyTorch custom ops (for eager execution) and TensorRT Quick Deployable Plugins with AOT support (for torch_tensorrt.compile). 

## Usage

```
  import torch, torch_tensorrt
  import torch_tensorrt.kernels as ttk                                                                                                                           
   
  CU = """                                                                                                                                                          
  extern "C" __global__ void my_sigmoid(const float* x, int n, float* y) {
      int i = blockIdx.x * blockDim.x + threadIdx.x;
      if (i < n) y[i] = 1.0f / (1.0f + __expf(-x[i]));                                                                                                              
  }
  """                                                                                                                                                               
                  
ttk.cuda_kernel_op(
    "kern_ex::sigmoid",
    ttk.KernelSpec(
        kernel_source=CU_SIGMOID,
        kernel_name="my_sigmoid",
        inputs=[ttk.InputDecl("x")],
        outputs=[ttk.OutputDecl("y", shape=ttk.SameAs("x"))],
        extras=[ttk.Numel("x")],
        geometry=ttk.Elementwise(block=(256,), layout="flat"),
    ),
    supports_dynamic_shapes=True,
)
```

 After this call, torch.ops.ann_ex.sigmoid is available in eager and is embedded as a TensorRT plugin during torch_tensorrt.compile. The meta function, eager      
  launch, AOT implementation, and PyTorch schema are all derived from the KernelSpec.
                                                                                                                                                                    
## API Surface                                                                                                                                                       
   
  The module exposes two primary entry points, layered by declarativeness:                                                                                          
                  
  auto_cuda_kernel_plugin is the recommended default. The caller supplies a KernelSpec dataclass describing the kernel's inputs, outputs (with a shape relation such
   as SameAs or ReduceDims), scalar extras (Numel, DimSize), and launch geometry (Elementwise or Reduction). The framework derives the meta function, eager CUDA
  launch, TensorRT AOT implementation, and PyTorch schema. This path covers pointwise kernels (1-D flat or N-D grid launches), reductions (with optional keepdim),  
  multi-input kernels, and scalar (non-tensor) kernel arguments via ScalarInput.

  manual_cuda_kernel_plugin is the lower-level alternative for kernels outside the declarative DSL — shape-changing outputs, multi-output kernels, or non-standard  
  launch geometries. The caller provides eager_fn and aot_fn directly; the decorator still registers the PyTorch op, TRT plugin, AOT implementation, and converter
  in a single call.                                                                                                                                                 
                  
  A Custom(fn=...) geometry is also available for callers who want the declarative path's schema/meta derivation but need to hand-write the TRT KernelLaunchParams. 

## Type of change

# Checklist:


examples/dynamo/custom_cuda_kernel_op.py
```diff
@@ -0,0 +1,165 @@
...
+_module = _program.compile("ptx", name_expressions=("repeat2_kernel",))
+_kernel = _module.get_kernel("repeat2_kernel")
+
+
+class _PTStream:
+    def __cuda_stream__(self):
+        return (0, torch.cuda.current_stream().cuda_stream)
+
+
+def _eager_repeat2(x: torch.Tensor) -> torch.Tensor:
```
BowenFu: This is like "launch_fn" we planned for other kernel providers. Is it possible to keep only this one and infer _aot_repeat2 from this?

examples/dynamo/manual_cuda_kernel_plugin_annotation.py
```diff
@@ -0,0 +1,165 @@
...
+
+@tta.manual_cuda_kernel_plugin(
+    op_name="ann_ex::repeat2",
+    kernel_source=CU_REPEAT2,
+    kernel_name="repeat2_kernel",
+    eager_fn=_eager_repeat2,
+    aot_fn=_aot_repeat2,
+    supports_dynamic_shapes=True,
+)
+def _repeat2_meta(x: torch.Tensor) -> torch.Tensor:
```
BowenFu: Is this a dead fn?

examples/dynamo/custom_cuda_kernel_op.py
bowang007: @BowenFu  No, the aot_repeat func is required since it provides TensorRT symbolic launch metadata

examples/dynamo/manual_cuda_kernel_plugin_annotation.py
bowang007: @BowenFu This is fakeTensor functions used for shape inference. PyTorch relies on to infer the output shape

BowenFu: Marking as resolved.

examples/dynamo/custom_cuda_kernel_op.py
BowenFu: Marking as resolved.

docsrc/py_api/annotation.rst
```diff
@@ -0,0 +1,156 @@
+.. _torch_tensorrt_annotation_py:
+
+torch_tensorrt.annotation
```
narendasan: I think this should be called torch_tensorrt.kernels

examples/dynamo/auto_cuda_kernel_plugin_annotation.py
```diff
@@ -0,0 +1,120 @@
...
+    print(
+        "[auto_cuda_kernel_plugin_annotation] Skipping example: "
+        "tensorrt.plugin unavailable."
+    )
+    sys.exit(0)
+
+try:
+    import cuda.core  # noqa: F401
+except ImportError:
+    try:
```
narendasan: For example code just assume all these libraries are there. makes it easier to read

examples/dynamo/auto_cuda_kernel_plugin_annotation.py
```diff
@@ -0,0 +1,120 @@
...
+# Numel("x")         pass x.numel() to the kernel as an int extra.
+# Elementwise(flat)  1-D launch over the flattened output; any input rank works.
+
+tta.auto_cuda_kernel_plugin(
+    "ann_ex::sigmoid",
+    tta.KernelSpec(
+        kernel_source=CU_SIGMOID,
+        kernel_name="my_sigmoid",
+        inputs=[tta.InputDecl("x")],
+        outputs=[tta.OutputDecl("y", shape=tta.SameAs(0))],
```
narendasan: You should be able to use names of other known symbols (like this could be tta.SameAs("x"))

examples/dynamo/manual_cuda_kernel_plugin_annotation.py
```diff
@@ -0,0 +1,165 @@
...
+    params = trtp.KernelLaunchParams()
+    params.grid_x = trtp.cdiv(n, 256)
+    params.block_x = 256
+    params.shared_mem = 0
+    extra = trtp.SymIntExprs(1)
+    extra[0] = trtp.SymInt32(n)
+    return params, extra
+
+
+@tta.manual_cuda_kernel_plugin(
```
narendasan: What is the distinction between manual and auto from a user perspective?

py/torch_tensorrt/annotation/_custom_plugin/_descriptor.py
```diff
@@ -0,0 +1,240 @@
...
+    else:
+        ret_str = _TORCH_TYPE_TO_SCHEMA.get(ret, "Tensor")
+
+    return f"({args_str}) -> {ret_str}"
+
+
+def _torch_op_already_registered(op_name: str) -> bool:
+    """Return True if ``op_name`` is already known to the torch dispatcher."""
+    try:
+        return bool(torch._C._jit_get_schemas_for_operator(op_name))
```
narendasan: Theres probably a more modern approach here (like some api in torch.library)

py/torch_tensorrt/kernels/_custom_plugin/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+            include_paths
+            if include_paths is not None
+            else _default_cuda_include_paths()
+        ),
+        compile_std=compile_std,
+        arch_override=arch_override,
+    )
+
+
+def custom_plugin(
```
narendasan: there is now a `torch_tensorrt.annotations(kernels).custom_plugin` and a `torch_tensorrt.dynamo.conversion.plugins.custom_op`. Why cant we just centralize on one?

narendasan: Or make it clear what is used for what by disambiguating the names

examples/dynamo/auto_cuda_kernel_plugin_annotation.py
```diff
@@ -0,0 +1,120 @@
...
+    if (i < n) y[i] = 1.0f / (1.0f + __expf(-x[i]));
+}
+"""
+
+
+# SameAs(0)          output has the same shape and dtype as input 0.
+# Numel("x")         pass x.numel() to the kernel as an int extra.
+# Elementwise(flat)  1-D launch over the flattened output; any input rank works.
+
+tta.auto_cuda_kernel_plugin(
```
narendasan: maybe we can call this something like `torch_tensorrt.kernels.cuda_kernel_op`

examples/dynamo/manual_cuda_kernel_plugin_annotation.py
narendasan: To me it just seems like the function bodies are just manually configured here, why not just support additional kwargs in the same api?

examples/dynamo/manual_cuda_kernel_plugin_annotation.py
narendasan: Why is the meta kernel the one that we decorate? to me the obvious thing to decorate is the jit_impl_fn

narendasan: I think the manual api as a decorator is somewhat confusing, imo we either do the workflow that we already have with multiple decorators (https://docs.pytorch.org/TensorRT/tutorials/_rendered_examples/dynamo/nvrtc_aot_plugin.html) or we dont decorate anything and just have a function that takes all of kernel source, meta, jit and aot as argument

py/torch_tensorrt/kernels/_custom_plugin/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+        include_paths if include_paths is not None else _default_cuda_include_paths()
+    )
+    runtime_cache: dict[str, Any] = {}
+
+    def _ensure_compiled() -> None:
+        if runtime_cache:
+            return
+        from torch_tensorrt.annotation._custom_plugin._nvrtc import compile_to_ptx
+
+        _ptx, device, kernel = compile_to_ptx(
```
narendasan: Could we also expose a `torch_tensorrt.kernels.ptx_op`, that just takes externally created valid ptx?

narendasan (commented):
Cool I think this is getting really close. I think we just have a few naming things to make this more user friendly and I think we should let users provide PTX directly in addition to the cuda apis. Also did you add nvrtc as an optional dependency in the pyproject.toml) (maybe under an a extras called kernels)?

docsrc/py_api/annotation.rst
bowang007: Hey @narendasan should we just deprecate the naming related to `annotation`? I'm a little bit confused here since this `annotation` is really opaque to the users and there is no `annotation` module right now

py/torch_tensorrt/kernels/_custom_plugin/__init__.py
bowang007: @narendasan  we could do that by just removing the `annotation` related stuff for now. I'm just putting that in the annotation in case the TensorRT team wants to develop some `annotation` related stuff in the future. Do we need to keep that stub for them now?

examples/dynamo/manual_cuda_kernel_plugin_annotation.py
bowang007: I’m planning to present how to manually write a CUDA kernel plugin. We have the option to completely remove this manual approach and rely entirely on automatic generation so users don't have to worry about it.

However, doing so means users lose control over launch functions and configurations. In this specific case, automatic generation doubles the number of launching threads, which could lead to memory inefficiencies and limit our ability to support dynamic cases.

@narendasan – We can drop the manual options for the sake of simplicity, but it comes at the cost of efficiency and flexibility. Which approach do you prefer?

docsrc/py_api/annotation.rst
narendasan: yeah I think it makes sense in the context of manual fusions that @BowenFu proposed but we can add the .annotation namespace when those land

py/torch_tensorrt/kernels/_custom_plugin/__init__.py
narendasan: no need for the stub until we use it

examples/dynamo/manual_cuda_kernel_plugin_annotation.py
narendasan: I think I pointed to something below on how we might be able to frame this for users

examples/dynamo/custom_cuda_kernel_op.py
```diff
@@ -0,0 +1,136 @@
...
+    extra = trtp.SymIntExprs(1)
+    extra[0] = trtp.SymInt32(n)
+    return params, extra
+
+
+def _repeat2_meta(x: torch.Tensor) -> torch.Tensor:
+    return torch.empty((x.numel() * 2,), device=x.device, dtype=x.dtype)
+
+
+ttk.custom_cuda_kernel_op(
```
narendasan: custom_cuda_kernel_op vs cuda_kernel_op is not clear. Is there a reason we cant just make the manual fields optional kwargs for the same api?

py/torch_tensorrt/kernels/_custom_plugin/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+        supports_dynamic_shapes=supports_dynamic_shapes,
+        requires_output_allocator=requires_output_allocator,
+        priority=priority,
+        capability_validator=capability_validator,
+        register_torch_op=True,
+        schema=schema,
+    )
+
+
+def ptx_op(
```
narendasan: Can we add an example for this?

narendasan (commented):
Cool almost there, last thing is this `cuda_kernel_op` vs `custom_cuda_kernel_op`. Its not clear what the difference is. I think we should just take optional kwargs for the manual fields for `cuda_kernel_op`.

Also can we get an example for ptx_op?

examples/dynamo/ptx_op.py
```diff
@@ -0,0 +1,177 @@
...
+    return params, extra
+
+
+# ---------------------------------------------------------------------------
+# Step 3: register via ptx_op
+# ---------------------------------------------------------------------------
+# After this call, ``torch.ops.ptx_ex.gelu`` exists and works in eager,
+# torch.compile, and torch_tensorrt.compile — same as ``cuda_kernel_op``.
+
+ttk.ptx_op(
```
narendasan, line 136: We dont have a kernel spec for ptx right?

py/torch_tensorrt/dynamo/conversion/plugins/_generate_plugin.py
```diff
@@ -133,50 +181,87 @@ def _generic_plugin_desc(*args: Any, **kwargs: Any) -> Tuple[trtp.TensorDesc]:
...
+        # for ``LocalSource('0')``), which are not valid Python identifiers.
+        # ``sympy.lambdify`` pastes arg names verbatim into the generated
+        # ``def`` body, so without this substitution we'd emit
+        # ``lambda L['0']: ...`` — a SyntaxError. Renaming to ``_a0, _a1, ...``
+        # before lambdifying sidesteps the issue.
+        #
+        # File against pytorch/pytorch (torch.fx.experimental.symbolic_shapes).
+        # Once ShapeEnv produces identifier-safe symbol names, the substitution
+        # can be removed and ``input_node_expr`` passed straight to lambdify.
+        # Tracking issue: <ADD PYTORCH ISSUE URL>.
```
narendasan: Is there an issue here?

py/torch_tensorrt/kernels/_ops.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Two paths share one entry point:
+
+    * **Declarative** — pass a fully-populated :class:`KernelSpec` and the
+      meta fn, eager fn, AOT fn, and PyTorch schema are all derived for you.
+      Covers Elementwise / Reduction kernels out of the box.
+    * **Override** — pass any of ``meta_fn`` / ``eager_fn`` / ``aot_fn`` /
+      ``schema`` as keyword arguments and the corresponding ``KernelSpec``
+      fields become optional. Use this for shape-changing kernels,
+      multi-output kernels, or anything outside the declarative DSL.
```
narendasan, line 52: What does declarative DSL mean?

narendasan (approved):
I think this looks good, I think my main remaining questions is why the KernelSpec doesnt work for PTX and we can probably do some work on the documentation. We can do that in another PR but right now the kernel docs/examples will not be rendered, make sure you add it into the toc tree in docsrc

examples/dynamo/ptx_op.py
bowang007: no, but we could extend the kernel spec for ptx as well. 
However, PTX is so low level that the user might have some issues like signature mismatch between the ptx and the kernelSpec even without noticing it.

py/torch_tensorrt/kernels/_ops.py
bowang007: it's to describe what the desired outcome is, rather than how to achieve it through step-by-step instructions. In our case, the user only specify 
```
  ttk.KernelSpec(
      kernel_source=CU_SIGMOID,
      kernel_name="my_sigmoid",
      inputs=[ttk.InputDecl("x")],
      outputs=[ttk.OutputDecl("y", shape=ttk.SameAs("x"))],
      extras=[ttk.Numel("x")],
      geometry=ttk.Elementwise(block=(256,), layout="flat"),
  )
```
Then TorchTensorRT derives the rest of required functions for users.

examples/dynamo/ptx_op.py
narendasan: We should think about it but doesnt need to be in this PR

PR #4202, opened 2026-04-21
feat(runtime): add TensorRT-RTX runtime cache, dynamic shapes strategy, and native CUDA graph support to C++ runtime

tp5uiuc:
# Description

Extends three TensorRT-RTX runtime features that landed on the Python runtime (`PythonTorchTensorRTModule`) to the **C++ runtime path** (`TorchTensorRTModule` → `core/runtime/TRTEngine`). All three features center on `nvinfer1::IRuntimeConfig`, which the C++ runtime previously did not use — it called `createExecutionContext(...)` directly at four sites.

Features ported in this stack:

| Feature | Python PR | This PR adds C++ runtime support |
|---|---|---|
| Runtime cache API | #4180 (merged) | yes |
| Dynamic shapes kernel specialization strategy | #4184 (merged) | yes |
| TensorRT-RTX native CUDA graph strategy | #4187 (draft) | yes |

**Without this PR**, users on the C++ runtime path (TorchScript deployments, `use_python_runtime=False`) cannot access any of these TRT-RTX features, and runtime-cache warm-start savings (~8× measured in #4180) are unavailable on that path.

## Commits

Original stack (reviewable independently):

1. **`feat(runtime): introduce IRuntimeConfig scaffolding and bump ABI to v9`** — shared infra. Adds the `IRuntimeConfig`/`IRuntimeCache` members (RTX-only), a private `recreate_execution_context()` helper replacing 4 direct `createExecutionContext` call sites, three new `SerializedInfoIndex` entries (`RUNTIME_CACHE_PATH_IDX`, `DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX`, `CUDA_GRAPH_STRATEGY_IDX`) with a single ABI bump `8 → 9`. Python settings + `compile()` parameter threading. No behavior change — the per-feature `apply_*` appliers are empty stubs filled in by subsequent commits.

2. **`feat(runtime): add runtime cache to C++ runtime for TensorRT-RTX`** — mirror of #4180. Load on engine setup, atomic save on destructor (`tmp + rename`).

3. **`feat(runtime): add dynamic shapes kernel specialization strategy to C++ runtime`** — mirror of #4184. Wires `IRuntimeConfig::setDynamicShapesKernelSpecializationStrategy`.

4. **`feat(runtime): add TensorRT-RTX native CUDA graph strategy to C++ runtime`** — mirror of #4187. Wires `IRuntimeConfig::setCudaGraphStrategy` and makes `execute_engine.cpp` TensorRT-RTX-aware: bypasses manual `at::cuda::CUDAGraph` capture on RTX (TRT-RTX handles it internally) and uses `cudaStreamIsCapturing` to detect outer whole-graph capture.

5. **`test: consolidate C++ runtime tests`** — folds C++ runtime cases into the existing `test_000_runtime_cache.py` and `test_001_dynamic_shapes_kernel_strategy.py`; adds model-level coverage in `test_runtime_cache_models.py` and `test_dynamic_shapes_kernel_strategy_models.py`.

Review-response commits (on top of the feature stack):

6. **`refactor(runtime): extract TRTRuntimeConfig`** — moves all TensorRT-RTX-specific IRuntimeConfig state into `core/runtime/TRTRuntimeConfig.{h,cpp}`, collapses three separate constructor args into a single `TRTRuntimeConfig` sink, uses enums (`DynamicShapesKernelStrategy`, `CudaGraphStrategyOption`), makes the destructor exception-safe, and contains the `#ifdef TRT_MAJOR_RTX` scatter to the new TU.

7. **`refactor(runtime): third-round review polish + cross-backend verification`** — inlines the FlattenedState macro, moves enum helpers + cache I/O into an anonymous namespace in the cpp, replaces `(void)param;` with `TORCHTRT_UNUSED`, uses `std::underlying_type_t` throughout, adds `[[nodiscard]]`, adds `ENABLE_FEATURE_DISABLE_RUNTIME_ALLOCATION` as a `local_define` for the RTX Bazel configs so the runtime-allocation feature gate matches the RTX header's expectation.

## Why bundle three features in one PR

All three features require an `IRuntimeConfig` on the engine, a single ABI bump, and extensions to the same serialization/deserialization code paths. Splitting into three independent PRs would trigger three consecutive ABI bumps and triple the surface area for backward-compat fallout. Keeping them in one stack keeps ABI changes atomic while still giving reviewers clean per-feature diffs.

## Type of change

- [x] New feature (non-breaking change which adds functionality — on the Python API surface)
- [x] Breaking change (serialized-engine ABI bump from `"8"` to `"9"` — old `.pt`/`.ep` files targeting the C++ runtime will fail `verify_serialization_fmt` with a clear error, as with every prior ABI bump)
- [ ] This change requires a documentation update — default values (`"lazy"`, `"disabled"`) keep existing behavior; existing docs for the Python-runtime runtime cache already cover the concept

## Checklist

- [x] My code follows the style guidelines of this project (clang-format, black, isort, ruff, mypy, typos, buildifier — all pre-commit checks pass)
- [x] I have commented my code, particularly in hard-to-understand areas (outer-capture detection, cache save-on-destructor contract, ABI bump rationale)
- [ ] I have made corresponding changes to the documentation (docstrings updated on `CompilationSettings`; full user-guide updates pending the feature PRs on the Python path)
- [x] I have added tests to verify my fix or my feature (C++ runtime cases folded into `test_000_runtime_cache.py`, `test_001_dynamic_shapes_kernel_strategy.py`, `test_001_cuda_graph_strategy.py`; model-level coverage in `test_runtime_cache_models.py`, `test_dynamic_shapes_kernel_strategy_models.py`)
- [x] New and existing unit tests pass locally on **both** backends (see test plan below)
- [ ] I have added the relevant labels to my PR

## Test plan

Cross-backend verification on an A100 (ipp1 node) inside the `main-native-x86_64-ubuntu24.04-cuda13.0` dev container, PyTorch nightly 2.13.0.dev20260420, CUDA 13.0.

### TensorRT-RTX build

Wheel: `torch_tensorrt_rtx-2.12.0.dev0+612556ba0` (built with `python3 setup.py bdist_wheel --use-rtx`, TensorRT-RTX 1.4.0.76).

Import smoke:
- `ENABLED_FEATURES.tensorrt_rtx = True`
- `ABI_VERSION = 9`
- `SERIALIZATION_LEN = 14`
- `RUNTIME_CACHE_PATH_IDX = 11`, `DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX = 12`, `CUDA_GRAPH_STRATEGY_IDX = 13`

```
$ pytest runtime/test_000_runtime_cache.py \
         runtime/test_001_dynamic_shapes_kernel_strategy.py \
         runtime/test_001_cuda_graph_strategy.py
================= 35 passed, 3 skipped, 119 warnings, 3 subtests passed =================
```

Breakdown of the 35 RTX-side passes:

- **Runtime cache** — Python-runtime setup (`TestRuntimeCacheSetup`), persistence and warm-cache roundtrip on both Python and C++ runtimes (`TestRuntimeCachePersistence`, `TestRuntimeCacheCppPersistence`), concurrency/filelock (`TestRuntimeCacheConcurrency`), timing-cache skip (`TestTimingCacheSkipped`), and serialization-index registration (`TestCppSerializationIndices`).
- **Dynamic shapes kernel strategy** — setup / enum plumbing (`TestDynamicShapesKernelStrategySetup`), full `{lazy, eager, none}` end-to-end through both runtimes (`TestDynamicShapesKernelStrategyCpp`), dynamic-shape traversal, and invalid-value rejection (`TestDynamicShapesKernelStrategyCppInvalidValue`).
- **CUDA graph strategy** — settings defaults (`TestCudaGraphStrategySettings`), `{disabled, whole_graph_capture}` via C++ runtime (`TestCudaGraphStrategyCpp`), RTX-native override when `set_cudagraphs_mode(True)` is combined with a strategy, repeated inference, and invalid-value rejection (`TestCudaGraphStrategyInvalidValue`).

Regression:
```
$ pytest runtime/test_005_dynamic_allocation.py runtime/test_002_cudagraphs_cpp.py
================= 9 passed =================
```

Model-level: e2e ResNet18 compilation + inference via the C++ runtime path with each `{lazy, eager, none}` strategy and with runtime cache warm-roundtrip (added in `tests/py/dynamo/models/`).

### Standard TensorRT build

Wheel: `torch_tensorrt-2.12.0.dev0+612556ba0` (built with plain `python3 setup.py bdist_wheel`, TensorRT 10.16.0).

Import smoke:
- `ENABLED_FEATURES.tensorrt_rtx = False`
- `ABI_VERSION = 9`
- `SERIALIZATION_LEN = 11` (no RTX-only slots in the FlattenedState)
- RTX-only index accessors are not registered by `register_jit_hooks.cpp`

```
$ pytest runtime/test_000_runtime_cache.py \
         runtime/test_001_dynamic_shapes_kernel_strategy.py \
         runtime/test_001_cuda_graph_strategy.py
================= 7 passed, 31 skipped =================
```

The 7 passes on standard TRT are:

- `TestNonRTXUnchanged` — confirms the existing Python-runtime paths are unaffected (runtime cache / timing cache behavior, no `runtime_config` member) — 2 tests.
- `TestDynamicShapesKernelStrategyNonRTX::test_setting_ignored_on_non_rtx` — confirms the new `dynamic_shapes_kernel_specialization_strategy` setting is silently ignored on non-RTX.
- `TestCudaGraphStrategySettings::{test_default_value, test_settable_values}` — `CompilationSettings` accepts the new fields on any backend.
- `TestDynamicShapesKernelStrategyCppInvalidValue::test_invalid_strategy_raises` + `TestCudaGraphStrategyInvalidValue::test_invalid_strategy_raises` — unknown strategy names are rejected at `_pack_engine_info` time even on non-RTX (validation is cross-platform; the engine-info serialization slots themselves are RTX-only).

The 31 skips all carry clean messages (sample):
```
SKIPPED [1] runtime/test_000_runtime_cache.py:68: Runtime cache is only available with TensorRT-RTX
SKIPPED [1] runtime/test_001_dynamic_shapes_kernel_strategy.py:58: Dynamic shapes kernel specialization strategy requires TensorRT-RTX
SKIPPED [1] runtime/test_001_cuda_graph_strategy.py:63:  CUDA graph strategy is a TensorRT-RTX feature
```

Regression on standard TRT:
```
$ pytest runtime/test_005_dynamic_allocation.py runtime/test_002_cudagraphs_cpp.py
================= 9 passed =================
```

### Summary

| Backend | Wheel | New / merged suites | Regression |
|---|---|---|---|
| TensorRT-RTX 1.4.0.76 | `torch_tensorrt_rtx-2.12.0.dev0+612556ba0` | **35 passed, 3 skipped** | **9 passed** |
| Standard TensorRT 10.16.0 | `torch_tensorrt-2.12.0.dev0+612556ba0` | **7 passed, 31 skipped** | **9 passed** |

0 failures on either backend; all RTX-gated suites skip cleanly with descriptive messages on the standard TRT build.


core/runtime/runtime.h
```diff
@@ -39,6 +39,9 @@ typedef enum {
   TARGET_PLATFORM_IDX,
   REQUIRES_OUTPUT_ALLOCATOR_IDX,
   RESOURCE_ALLOCATION_STRATEGY_IDX,
+  RUNTIME_CACHE_PATH_IDX,
+  DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX,
+  CUDA_GRAPH_STRATEGY_IDX,
```
tp5uiuc, line 46: Can we guard this with `#ifdef TRT_MAJOR_RTX`?

core/runtime/register_jit_hooks.cpp
```diff
@@ -150,6 +150,9 @@ TORCH_LIBRARY(tensorrt, m) {
   m.def("REQUIRES_OUTPUT_ALLOCATOR_IDX", []() -> int64_t { return REQUIRES_OUTPUT_ALLOCATOR_IDX; });
   m.def("SERIALIZATION_LEN", []() -> int64_t { return SERIALIZATION_LEN; });
   m.def("RESOURCE_ALLOCATION_STRATEGY_IDX", []() -> int64_t { return RESOURCE_ALLOCATION_STRATEGY_IDX; });
+  m.def("RUNTIME_CACHE_PATH_IDX", []() -> int64_t { return RUNTIME_CACHE_PATH_IDX; });
+  m.def("DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX", []() -> int64_t { return DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX; });
+  m.def("CUDA_GRAPH_STRATEGY_IDX", []() -> int64_t { return CUDA_GRAPH_STRATEGY_IDX; });
```
tp5uiuc, line 205: Guard with `#ifdef TRT_MAJOR_RTX`

core/runtime/TRTEngine.cpp
```diff
@@ -73,7 +84,10 @@ TRTEngine::TRTEngine(
           hardware_compatible,
           requires_output_allocator,
           serialized_metadata,
-          resource_allocation_strategy) {}
+          resource_allocation_strategy,
+          runtime_cache_path,
+          dynamic_shapes_kernel_strategy,
+          cuda_graph_strategy) {}
```
tp5uiuc: It maybe better to put TRT-RTX specific runtime options in a separate struct, which can then be transparently passed onto `runtimeConfig`.

core/runtime/TRTEngine.cpp
```diff
@@ -100,7 +117,20 @@ TRTEngine::TRTEngine(
...
+  this->runtime_cache_path = runtime_cache_path;
+  TORCHTRT_CHECK(
+      dynamic_shapes_kernel_strategy >= 0 && dynamic_shapes_kernel_strategy <= 2,
+      "Invalid dynamic_shapes_kernel_strategy: " << dynamic_shapes_kernel_strategy
+                                                 << ". Expected 0 (lazy), 1 (eager), or 2 (none).");
+  this->dynamic_shapes_kernel_strategy = dynamic_shapes_kernel_strategy;
+  TORCHTRT_CHECK(
+      cuda_graph_strategy >= 0 && cuda_graph_strategy <= 1,
+      "Invalid cuda_graph_strategy: " << cuda_graph_strategy << ". Expected 0 (disabled) or 1 (whole_graph_capture).");
+  this->cuda_graph_strategy = cuda_graph_strategy;
```
tp5uiuc: See https://github.com/pytorch/TensorRT/pull/4202/changes#r3120561924, push to a new `struct`

core/runtime/TRTEngine.cpp
```diff
@@ -278,8 +307,7 @@ void TRTEngine::disable_profiling() {
   torch::cuda::synchronize(device_info.id);
   profile_execution = false;
   trt_engine_profiler.reset();
-  exec_ctx = make_trt(cuda_engine->createExecutionContext());
-  TORCHTRT_CHECK((exec_ctx.get() != nullptr), "Unable to recreate TensorRT execution context");
+  recreate_execution_context();
```
tp5uiuc, line 307: Disabling profiling doesn't seem to respect `nvinfer1::ExecutionContextAllocationStrategy::kUSER_MANAGED` if `this->resource_allocation_strategy == ResourceAllocationStrategy::kDynamic`. Is this expected?

core/runtime/TRTEngine.cpp
```diff
@@ -376,10 +404,7 @@ bool TRTEngine::set_device_memory_budget(int64_t budget) {
     trt_engine_profiler.reset();
   }
   bool result = cuda_engine->setWeightStreamingBudgetV2(budget);
-  exec_ctx = make_trt(cuda_engine->createExecutionContext());
-  TORCHTRT_CHECK(
-      (exec_ctx.get() != nullptr),
-      "Unable to recreate TensorRT execution context after setting new device memory budget");
+  recreate_execution_context();
```
tp5uiuc, line 404: Same as https://github.com/pytorch/TensorRT/pull/4202/changes#r3120574611, confirm

core/runtime/TRTEngine.cpp
```diff
@@ -428,6 +453,11 @@ std::string TRTEngine::to_str() const {
   ss << "  Hardware Compatibility: " << (hardware_compatible ? "Enabled" : "Disabled") << std::endl;
   ss << "  Target Platform: " << target_platform << std::endl;
   ss << "  Resource Allocation Strategy: " << (resource_allocation_strategy == ResourceAllocationStrategy::kDynamic ? "Dynamic" : "Static") << std::endl;
+  ss << "  Runtime Cache Path: " << (runtime_cache_path.empty() ? "<disabled>" : runtime_cache_path) << std::endl;
+  ss << "  Dynamic Shapes Kernel Strategy: " << dynamic_shapes_kernel_strategy
+     << " (0=lazy, 1=eager, 2=none)" << std::endl;
+  ss << "  CUDA Graph Strategy: " << cuda_graph_strategy
+     << " (0=disabled, 1=whole_graph_capture)" << std::endl;
```
tp5uiuc: Similar to   
```
ss << "  Resource Allocation Strategy: " << (resource_allocation_strategy == ResourceAllocationStrategy::kDynamic ? "Dynamic" : "Static") << std::endl;
```

Do the conversion to a string and emit something that's human readable instead of an index.

core/runtime/TRTEngine.h
```diff
@@ -217,6 +226,49 @@ struct TRTEngine : torch::CustomClassHolder {
   ResourceAllocationStrategy resource_allocation_strategy = kStatic;
   void set_resource_allocation_strategy(ResourceAllocationStrategy new_strategy);
   ResourceAllocationStrategy get_resource_allocation_strategy();
+
+  // TRT-RTX runtime config state. The plain fields are stored unconditionally so that
+  // serialization remains ABI-stable on non-RTX builds; the IRuntimeConfig / IRuntimeCache
+  // handles themselves only exist on RTX.
+  std::string runtime_cache_path = "";
+  int dynamic_shapes_kernel_strategy = 0; // 0=lazy, 1=eager, 2=none
+  int cuda_graph_strategy = 0; // 0=disabled, 1=whole_graph_capture
```
tp5uiuc: Have an enumeration here rather than an integer.

core/runtime/TRTEngine.cpp
```diff
@@ -508,16 +544,191 @@ void TRTEngine::reset_captured_graph() {
...
+    LOG_DEBUG(
+        "Setting resource allocation strategy to "
+        << (this->resource_allocation_strategy == TRTEngine::ResourceAllocationStrategy::kDynamic ? "dynamic"
+                                                                                                  : "static"));
+    recreate_execution_context();
+  }
+}
+
+bool TRTEngine::is_monolithic_capturable(cudaStream_t stream) const {
+#if defined(TRT_MAJOR_RTX) && defined(ENABLE_FEATURE_DISABLE_RUNTIME_ALLOCATION)
```
tp5uiuc: We can remove the check for defined `ENABLE_FEATURE_DISABLE_RUNTIME_ALLOCATION`. It is on for TRT-RTX always.

core/runtime/TRTEngine.h
```diff
@@ -217,6 +226,49 @@ struct TRTEngine : torch::CustomClassHolder {
...
+  // One-shot flag: set the first time execute_engine detects an outer stream capture around
+  // this engine, at which point its TRT-RTX native CUDA graph capture is turned off so the
+  // two do not fight. The flag stays set for the remainder of the engine's lifetime.
+  bool rtx_native_cudagraphs_disabled = false;
+
+#ifdef TRT_MAJOR_RTX
+  std::shared_ptr<nvinfer1::IRuntimeConfig> runtime_config;
+  std::shared_ptr<nvinfer1::IRuntimeCache> runtime_cache;
+#endif
```
tp5uiuc: We should encapsulate these in another struct/class instead. Couple of reasons:
1. Non-RTX vs RTX separation should be made transparent
2. Destructor has non-trivial semantics in this case. If runtime_cache save fails, then stack unwinding fails. We should catch exceptions properly in TRTEngine's destructor instead.

All of the `apply_runtime_cache` etc. functions can be pushed to that struct (as a public member/free functions, no need for information hiding in this case) and called in from `TRTEngine`. Free-functions are overall easier to test.

core/runtime/TRTEngine.cpp
```diff
@@ -508,16 +544,191 @@ void TRTEngine::reset_captured_graph() {
...
+  apply_cuda_graph_strategy();
+  recreate_execution_context();
+  rtx_native_cudagraphs_disabled = true;
+#endif
+}
+
+void TRTEngine::recreate_execution_context() {
+#ifdef TRT_MAJOR_RTX
+  if (!runtime_config) {
+    runtime_config = make_trt(cuda_engine->createRuntimeConfig());
```
tp5uiuc: `createRuntimeConfig()`, `createExecutionContext(IRuntimeConfig*)` API is available in both standard TensorRT and TensorRT-RTX : so it can be used to unify code-paths.

`IRuntimeConfig` then has the following APIs for setting different members and can be used for both TRT and TRT-RTX
- `setExecutionContextAllocationStrategy`
- `getExecutionContextAllocationStrategy`

The following are available only in TRT-RTX
- `setRuntimeCache`
- `setDynamicShapesKernelSpecializationStrategy`
- `setCudaGraphStrategy`

tp5uiuc: Use this API to reduce divergence.

core/runtime/TRTEngine.h
tp5uiuc: (1) means that any #ifdef TRT_MAJOR_RTX can be pushed to these functions/classes instead so that call sites become clean.

core/runtime/TRTEngine.cpp
```diff
@@ -508,16 +544,191 @@ void TRTEngine::reset_captured_graph() {
...
+    TORCHTRT_CHECK(runtime_config.get() != nullptr, "Unable to create TensorRT IRuntimeConfig");
+    apply_runtime_cache();
+    apply_dynamic_shapes_kernel_strategy();
+    apply_cuda_graph_strategy();
+  }
+  runtime_config->setExecutionContextAllocationStrategy(
```
tp5uiuc: If `runtime_config` is not none, I wonder if we should `save_runtime_cache()` here so that the next reload can reuse any compiled kernels from the present run.

core/runtime/TRTEngine.cpp
```diff
@@ -508,16 +544,191 @@ void TRTEngine::reset_captured_graph() {
...
+  } else {
+    exec_ctx = make_trt(cuda_engine->createExecutionContext());
+  }
+#endif
+  TORCHTRT_CHECK(exec_ctx.get() != nullptr, "Unable to (re)create TensorRT execution context");
+}
+
+#ifdef TRT_MAJOR_RTX
+void TRTEngine::apply_runtime_cache() {
+  if (runtime_cache_path.empty()) {
```
tp5uiuc: Maybe a debug assert for existence of runtime_config would be nice to have before attempting to dereference it. By program construction I think this is not a possible state, but never hurts to be defensive.

core/runtime/TRTEngine.cpp
```diff
@@ -508,16 +544,191 @@ void TRTEngine::reset_captured_graph() {
...
+  if (!ok) {
+    LOG_WARNING("Failed to attach runtime cache to IRuntimeConfig; cache will be unused.");
+    runtime_cache.reset();
+    return;
+  }
+  LOG_DEBUG("TensorRT-RTX runtime cache configured at " << runtime_cache_path);
+}
+
+void TRTEngine::apply_dynamic_shapes_kernel_strategy() {
+  runtime_config->setDynamicShapesKernelSpecializationStrategy(
```
tp5uiuc: Maybe a debug assert for existence of runtime_config would be nice to have before attempting to dereference it. By program construction I think this is not a possible state, but never hurts to be defensive.

core/runtime/TRTEngine.cpp
```diff
@@ -508,16 +544,191 @@ void TRTEngine::reset_captured_graph() {
...
+      static_cast<nvinfer1::DynamicShapesKernelSpecializationStrategy>(dynamic_shapes_kernel_strategy));
+  LOG_DEBUG("Dynamic shapes kernel specialization strategy set to " << dynamic_shapes_kernel_strategy);
+}
+
+void TRTEngine::apply_cuda_graph_strategy() {
+  bool ok = runtime_config->setCudaGraphStrategy(
```
tp5uiuc: Maybe a debug assert for existence of runtime_config would be nice to have before attempting to dereference it. By program construction I think this is not a possible state, but never hurts to be defensive.

core/runtime/TRTEngine.cpp
```diff
@@ -508,16 +544,191 @@ void TRTEngine::reset_captured_graph() {
...
+void TRTEngine::load_runtime_cache() {
+  if (runtime_cache == nullptr || runtime_cache_path.empty()) {
+    return;
+  }
+  if (!std::filesystem::exists(runtime_cache_path)) {
+    LOG_DEBUG("No existing runtime cache at " << runtime_cache_path);
+    return;
+  }
+#ifndef _WIN32
+  int fd = ::open(runtime_cache_path.c_str(), O_RDONLY);
```
tp5uiuc: Let's remove the need for file-locking and just use `std::filesystem` for now. We will have a followup task to imbibe a platform independent file-locking system and then introduce file locking support for runtime-caches.

core/runtime/TRTEngine.cpp
```diff
@@ -508,16 +544,191 @@ void TRTEngine::reset_captured_graph() {
...
+void TRTEngine::save_runtime_cache() {
+  if (runtime_cache == nullptr || runtime_cache_path.empty()) {
+    return;
+  }
+  auto host_mem = make_trt(runtime_cache->serialize());
+  if (host_mem.get() == nullptr || host_mem->size() == 0) {
+    return;
+  }
+  try {
+    std::filesystem::path path(runtime_cache_path);
```
tp5uiuc: Same comment. Let's remove the need for file-locking and just use `std::filesystem` for now. We will have a followup task to imbibe a platform independent file-locking system and then introduce file locking support for runtime-caches.

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -53,7 +57,22 @@
...
         torch.ops.tensorrt.RESOURCE_ALLOCATION_STRATEGY_IDX()
     )  # 10
-    SERIALIZATION_LEN = torch.ops.tensorrt.SERIALIZATION_LEN()  # 11
+    RUNTIME_CACHE_PATH_IDX = torch.ops.tensorrt.RUNTIME_CACHE_PATH_IDX()  # 11
+    DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX = (
+        torch.ops.tensorrt.DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX()
+    )  # 12
+    CUDA_GRAPH_STRATEGY_IDX = torch.ops.tensorrt.CUDA_GRAPH_STRATEGY_IDX()  # 13
+    SERIALIZATION_LEN = torch.ops.tensorrt.SERIALIZATION_LEN()  # 14
+
```
tp5uiuc, line 73: Wrap in a `if ENABLED_FEATURES.tensorrt_rtx:`

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -203,6 +227,29 @@ def _pack_engine_info(self) -> List[str | bytes]:
         engine_info[RESOURCE_ALLOCATION_STRATEGY_IDX] = str(
             int(self.dynamically_allocate_resources)
         )
+        engine_info[RUNTIME_CACHE_PATH_IDX] = self.runtime_cache_path or ""
```
tp5uiuc: Wrap in a `if ENABLED_FEATURES.tensorrt_rtx:`

tests/py/dynamo/runtime/test_000_dynamic_shapes_kernel_strategy.py
```diff
@@ -0,0 +1,133 @@
+import unittest
```
tp5uiuc: Can me merge this file and [test_001_dynamic_shapes_kernel_strategy.py](https://github.com/pytorch/TensorRT/commit/8903707b90ff0a8d6dc04eb7757a9eb8772fb76f#diff-bd73cd48732b052410e03f5a102d62fa147a6748d000373ffeb4f05e34171f2d)

tests/py/dynamo/runtime/test_000_runtime_cache_cpp.py
```diff
@@ -0,0 +1,146 @@
+import gc
```
tp5uiuc: Can we merge this file and tests/py/dynamo/runtime/test_000_runtime_cache.py?

tests/py/dynamo/runtime/test_000_runtime_cache_cpp.py
```diff
@@ -0,0 +1,146 @@
...
+import os
```
tp5uiuc: Also have a models level test, you can reuse `tests/py/dynamo/models/test_runtime_cache_models.py`

tests/py/dynamo/runtime/test_000_dynamic_shapes_kernel_strategy.py
```diff
@@ -0,0 +1,133 @@
...
+
```
tp5uiuc: Also have a models level test, we can reuse `tests/py/dynamo/models/test_dynamic_shapes_kernel_strategy_models.py`

py/torch_tensorrt/dynamo/_settings.py
tp5uiuc, line 157: `dynamic_shapes_kernel_specialization_strategy` is repeated.

tp5uiuc, line 101: Docstring for dynamic shapes kernels specialization strategy is repeated

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -124,6 +125,8 @@ class CompilationSettings:
         autocast_calibration_dataloader (Optional[torch.utils.data.DataLoader]): The dataloader to use for autocast calibration. Default is None.
         offload_module_to_cpu (bool): Offload the model to CPU to reduce memory footprint during compilation
         dynamically_allocate_resources (bool): Dynamically allocate resources for TensorRT engines
+        dynamic_shapes_kernel_specialization_strategy (str): TensorRT-RTX dynamic shapes kernel specialization strategy: "lazy" (default, compile specialized kernels in background and use fallbacks until ready), "eager" (compile specialized kernels synchronously, blocking first inference), or "none" (always use fallback kernels). Not used for standard TensorRT.
```
tp5uiuc: Remove this line

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -189,6 +192,10 @@ class CompilationSettings:
     enable_resource_partitioning: bool = ENABLE_RESOURCE_PARTITIONING
     cpu_memory_budget: Optional[int] = CPU_MEMORY_BUDGET
     dynamically_allocate_resources: bool = DYNAMICALLY_ALLOCATE_RESOURCES
+    dynamic_shapes_kernel_specialization_strategy: str = (
+        DYNAMIC_SHAPES_KERNEL_SPECIALIZATION_STRATEGY
+    )
```
tp5uiuc: Remove.

core/runtime/TRTEngine.cpp
```diff
@@ -62,7 +63,10 @@ TRTEngine::TRTEngine(
     bool hardware_compatible,
     bool requires_output_allocator,
     const std::string& serialized_metadata,
-    const ResourceAllocationStrategy resource_allocation_strategy)
+    const ResourceAllocationStrategy resource_allocation_strategy,
+    const std::string& runtime_cache_path,
```
tp5uiuc: Capture by value and `std::move()` instead since it is a sink parameter.

core/runtime/TRTEngine.cpp
```diff
@@ -73,7 +77,10 @@ TRTEngine::TRTEngine(
           hardware_compatible,
           requires_output_allocator,
           serialized_metadata,
-          resource_allocation_strategy) {}
+          resource_allocation_strategy,
+          runtime_cache_path,
```
tp5uiuc: move here.

core/runtime/TRTEngine.cpp
```diff
@@ -88,7 +95,15 @@ TRTEngine::TRTEngine(std::vector<std::string> serialized_info)
           serialized_info[SERIALIZED_METADATA_IDX],
           (static_cast<bool>(std::stoi(serialized_info[RESOURCE_ALLOCATION_STRATEGY_IDX]))
                ? ResourceAllocationStrategy::kDynamic
-               : ResourceAllocationStrategy::kStatic)) {}
+               : ResourceAllocationStrategy::kStatic)
+#ifdef TRT_MAJOR_RTX
+              ,
+          serialized_info[RUNTIME_CACHE_PATH_IDX],
```
tp5uiuc: can move here

core/runtime/TRTEngine.h
```diff
@@ -13,12 +13,30 @@
...
+    std::tuple<std::string, std::string>, // output binding names
+    std::tuple<std::string, std::string>, // HW compatibility
+    std::tuple<std::string, std::string>, // requires_output_allocator
+    std::tuple<std::string, std::string>, // serialized metadata
+    std::tuple<std::string, std::string>, // Platform
+    std::tuple<std::string, std::string>, // Resource Allocation Strategy
+    std::tuple<std::string, std::string>, // Runtime Cache Path (TRT-RTX)
+    std::tuple<std::string, std::string>, // Dynamic Shapes Kernel Specialization Strategy (TRT-RTX)
+    std::tuple<std::string, std::string>>; // CUDA Graph Strategy (TRT-RTX)
+#else
```
tp5uiuc: Don't repeat information, put the macros within the FlattenedState

core/runtime/TRTEngine.h
```diff
@@ -134,7 +153,10 @@ struct TRTEngine : torch::CustomClassHolder {
       bool requires_output_allocator = false,
       const std::string& serialized_metadata = "",
       const TRTEngine::ResourceAllocationStrategy resource_allocation_strategy =
-          TRTEngine::ResourceAllocationStrategy::kStatic);
+          TRTEngine::ResourceAllocationStrategy::kStatic,
+      const std::string& runtime_cache_path = "",
+      int dynamic_shapes_kernel_strategy = 0,
+      int cuda_graph_strategy = 0);
```
tp5uiuc: Does taking a `TRTRuntimeConfig` not work here? Why bother to repeat all the nested options across all the constructors?

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+std::string to_string(DynamicShapesKernelStrategy s) {
+  switch (s) {
+    case DynamicShapesKernelStrategy::kLazy:
+      return "lazy";
+    case DynamicShapesKernelStrategy::kEager:
+      return "eager";
+    case DynamicShapesKernelStrategy::kNone:
+      return "none";
+  }
+  return "unknown";
```
tp5uiuc: TORCHTRT_CHECK(false), we cannot use `std::unreachable` yet.

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+}
+
+std::string to_string(CudaGraphStrategyOption s) {
+  switch (s) {
+    case CudaGraphStrategyOption::kDisabled:
+      return "disabled";
+    case CudaGraphStrategyOption::kWholeGraphCapture:
+      return "whole_graph_capture";
+  }
+  return "unknown";
```
tp5uiuc: Same comment as before TORCHTRT_CHECK(false),

core/runtime/TRTEngine.cpp
tp5uiuc: Why aren't there `#ifdef TRT_RTX` guards here? This wont even compile in standard TRT path

core/runtime/TRTEngine.cpp
```diff
@@ -100,7 +115,13 @@ TRTEngine::TRTEngine(
...
     bool requires_output_allocator,
     const std::string& serialized_metadata,
-    const ResourceAllocationStrategy resource_allocation_strategy) {
+    const ResourceAllocationStrategy resource_allocation_strategy,
+    const std::string& runtime_cache_path,
+    int dynamic_shapes_kernel_strategy,
+    int cuda_graph_strategy) {
+  runtime_cfg.runtime_cache_path = runtime_cache_path;
+  runtime_cfg.dynamic_shapes_kernel_strategy = to_dynamic_shapes_kernel_strategy(dynamic_shapes_kernel_strategy);
+  runtime_cfg.cuda_graph_strategy = to_cuda_graph_strategy_option(cuda_graph_strategy);
```
tp5uiuc: Ideally we pass in a TRTRuntimeCfg and move

core/runtime/TRTEngine.cpp
```diff
@@ -264,7 +279,13 @@ TRTEngine::TRTEngine(
 }
 
 TRTEngine::~TRTEngine() {
-  torch::cuda::synchronize(device_info.id);
+  // Destructors must not throw; `save_runtime_cache_nothrow` is itself no-throw but we
+  // wrap it defensively to keep stack unwinding safe in all circumstances.
+  try {
+    torch::cuda::synchronize(device_info.id);
```
tp5uiuc: Restore old behavior for this.

core/runtime/TRTEngine.cpp
```diff
@@ -497,6 +522,12 @@ std::vector<std::string> TRTEngine::serialize() {
   serialized_info[TARGET_PLATFORM_IDX] = this->target_platform.serialize();
   serialized_info[RESOURCE_ALLOCATION_STRATEGY_IDX] =
       this->resource_allocation_strategy == ResourceAllocationStrategy::kDynamic ? "1" : "0";
+#ifdef TRT_MAJOR_RTX
+  serialized_info[RUNTIME_CACHE_PATH_IDX] = runtime_cfg.runtime_cache_path;
+  serialized_info[DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX] =
+      std::to_string(static_cast<int>(runtime_cfg.dynamic_shapes_kernel_strategy));
+  serialized_info[CUDA_GRAPH_STRATEGY_IDX] = std::to_string(static_cast<int>(runtime_cfg.cuda_graph_strategy));
```
tp5uiuc: static_cast but always use underlying_t instead of raw int.

core/runtime/TRTEngine.cpp
```diff
@@ -62,7 +63,10 @@ TRTEngine::TRTEngine(
     bool hardware_compatible,
     bool requires_output_allocator,
     const std::string& serialized_metadata,
-    const ResourceAllocationStrategy resource_allocation_strategy)
+    const ResourceAllocationStrategy resource_allocation_strategy,
+    const std::string& runtime_cache_path,
+    int dynamic_shapes_kernel_strategy,
+    int cuda_graph_strategy)
```
tp5uiuc: DO NOT TAKE int. Take the underlying type of the enum and always use that. For example since the enum is `int32_t`, use `int32_t`. The std::vector<std::string> is the only method that gets called to instantiate the objects, so we can take care of casting to the right type in that constructor.

core/runtime/TRTRuntimeConfig.h
```diff
@@ -0,0 +1,93 @@
...
+// TensorRT-RTX-only configuration for how CUDA graph capture/replay is handled.
+enum class CudaGraphStrategyOption : int32_t {
+  kDisabled = 0,
+  kWholeGraphCapture = 1,
+};
+
+std::string to_string(DynamicShapesKernelStrategy s);
+std::string to_string(CudaGraphStrategyOption s);
+DynamicShapesKernelStrategy to_dynamic_shapes_kernel_strategy(int v);
+CudaGraphStrategyOption to_cuda_graph_strategy_option(int v);
```
tp5uiuc: Use underlying types in signatures always

core/runtime/TRTRuntimeConfig.h
```diff
@@ -0,0 +1,93 @@
...
+  // Load the runtime cache from disk using std::filesystem. No-throw: errors log and
+  // return. Invoked internally from `ensure_initialized` when a cache path is set.
+  void load_runtime_cache_nothrow() noexcept;
+
+  // Save the runtime cache to disk using std::filesystem (tmp + rename). No-throw:
+  // errors log and return, so it is safe to call from a destructor.
+  void save_runtime_cache_nothrow() noexcept;
+
+  // Append a human-readable summary to a TRTEngine::to_str stream.
+  void write_to_str(std::ostream& os) const;
```
tp5uiuc: We should have two functions:
1. `to_str` that generates a `str` representation from the runtime cfg, similar to `write_to_str` but returning a string without taking in any user argument.
2. std::ostream& operator<<(std::ostream& os, const TRTRuntimeCfg& cfg) {
  os << " Runtime cfg { " << endl
  os << cfg.to_str();
  return os;
}

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+#endif
+}
+
+void TRTRuntimeConfig::set_execution_context_allocation_strategy(
+    nvinfer1::ExecutionContextAllocationStrategy strategy) {
+  TORCHTRT_ASSERT(config, "TRTRuntimeConfig::config must be initialized before setting allocation strategy");
+  config->setExecutionContextAllocationStrategy(strategy);
+}
+
+bool TRTRuntimeConfig::uses_internal_capture(bool cudagraphs_enabled) const {
```
tp5uiuc: nodiscard

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+}
+
+DynamicShapesKernelStrategy to_dynamic_shapes_kernel_strategy(int v) {
+  TORCHTRT_CHECK(
+      v >= 0 && v <= 2,
+      "Invalid dynamic shapes kernel strategy value: " << v << ". Expected 0 (lazy), 1 (eager), or 2 (none).");
+  return static_cast<DynamicShapesKernelStrategy>(v);
+}
+
+CudaGraphStrategyOption to_cuda_graph_strategy_option(int v) {
```
tp5uiuc: nodsicard

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+  switch (s) {
+    case CudaGraphStrategyOption::kDisabled:
+      return "disabled";
+    case CudaGraphStrategyOption::kWholeGraphCapture:
+      return "whole_graph_capture";
+  }
+  return "unknown";
+}
+
+DynamicShapesKernelStrategy to_dynamic_shapes_kernel_strategy(int v) {
```
tp5uiuc: nodsicard

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+#include <stdexcept>
+#include <vector>
+
+#include "core/util/prelude.h"
+
+namespace torch_tensorrt {
+namespace core {
+namespace runtime {
+
+std::string to_string(DynamicShapesKernelStrategy s) {
```
tp5uiuc: nodiscard

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+    LOG_DEBUG("Runtime cache disabled (no path configured).");
+  }
+
+  // Dynamic shapes kernel specialization strategy -- TRT-RTX only.
+  config->setDynamicShapesKernelSpecializationStrategy(
+      static_cast<nvinfer1::DynamicShapesKernelSpecializationStrategy>(dynamic_shapes_kernel_strategy));
+  LOG_DEBUG("Dynamic shapes kernel specialization strategy set to " << to_string(dynamic_shapes_kernel_strategy));
+
+  // CUDA graph strategy -- TRT-RTX only.
+  bool ok = config->setCudaGraphStrategy(
```
tp5uiuc: Can we enforce const correctness, this should be `bool const ok` or better yet, don't make the variable in the first place, but use directly in the site

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+  // isStreamCapturable is declared inside `#if ENABLE_FEATURE_DISABLE_RUNTIME_ALLOCATION`
+  // in the TensorRT-RTX header; conservatively assume the engine is capturable when that
+  // feature flag is not enabled at compile time.
+  (void)exec_ctx;
+  (void)stream;
+  return true;
+#endif
+}
+
+void TRTRuntimeConfig::load_runtime_cache_nothrow() noexcept {
```
tp5uiuc: Can this be refactored into two functions:
1. A free function which takes in the path and the runtime_cache pointer and deserializes it. No error checking in this path, so non-noexcept
2. A member function (so that TRTEngine consumes it), that handles exceptions and is non-throwing and hence noexcept

Easier to debug/test this way.

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+  } catch (const std::exception& e) {
+    LOG_WARNING("Failed to load runtime cache: " << e.what());
+  } catch (...) {
+    LOG_WARNING("Failed to load runtime cache (unknown exception).");
+  }
+#endif
+}
+
+void TRTRuntimeConfig::save_runtime_cache_nothrow() noexcept {
```
tp5uiuc: Same comment as load_runtime_cache(). You don't need to explicitly say nothrow the function signature will guarantee that

core/runtime/TRTEngine.cpp
```diff
@@ -54,26 +55,28 @@ void DynamicOutputAllocator::notifyShape(char const* tensorName, nvinfer1::Dims
 }
 
 TRTEngine::TRTEngine(
-    const std::string& serialized_engine,
+    std::string serialized_engine,
```
tp5uiuc, line 65: This is technically not needed, but also doesn't hurt.

core/runtime/TRTEngine.cpp
```diff
@@ -428,6 +427,7 @@ std::string TRTEngine::to_str() const {
   ss << "  Hardware Compatibility: " << (hardware_compatible ? "Enabled" : "Disabled") << std::endl;
   ss << "  Target Platform: " << target_platform << std::endl;
   ss << "  Resource Allocation Strategy: " << (resource_allocation_strategy == ResourceAllocationStrategy::kDynamic ? "Dynamic" : "Static") << std::endl;
+  ss << runtime_cfg;
```
tp5uiuc: Use ss << runtime_cfg.to_str() here.

core/runtime/TRTEngine.h
```diff
@@ -30,7 +43,8 @@ using FlattenedState = std::tuple<
     std::tuple<std::string, std::string>, // requires_output_allocator
     std::tuple<std::string, std::string>, // serialized metadata
     std::tuple<std::string, std::string>, // Platform
-    std::tuple<std::string, std::string>>; // Resource Allocation Strategy
+    std::tuple<std::string, std::string> /* Resource Allocation Strategy */
+        TRTRTX_FLATTENED_STATE_EXTRAS>;
```
tp5uiuc: TODO : Inline and fix

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+TRTRuntimeConfig make_runtime_config_from_serialized(const std::vector<std::string>& info) {
+  TRTRuntimeConfig cfg;
+#ifdef TRT_MAJOR_RTX
+  cfg.runtime_cache_path = info[RUNTIME_CACHE_PATH_IDX];
+  cfg.dynamic_shapes_kernel_strategy =
+      to_dynamic_shapes_kernel_strategy(std::stoi(info[DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX]));
+  cfg.cuda_graph_strategy = to_cuda_graph_strategy_option(std::stoi(info[CUDA_GRAPH_STRATEGY_IDX]));
+#else
+  (void)info;
```
tp5uiuc: Use `TORCHTRT_UNUSED` instead

core/runtime/TRTRuntimeConfig.h
```diff
@@ -0,0 +1,113 @@
...
+};
+
+// Conversion helpers. Signatures use the enum's underlying type (int32_t) rather than
+// raw `int` so call sites pass validated strategy codes directly without implicit
+// narrowing.
+[[nodiscard]] std::string to_string(DynamicShapesKernelStrategy s);
+[[nodiscard]] std::string to_string(CudaGraphStrategyOption s);
+[[nodiscard]] DynamicShapesKernelStrategy to_dynamic_shapes_kernel_strategy(
+    std::underlying_type_t<DynamicShapesKernelStrategy> v);
+[[nodiscard]] CudaGraphStrategyOption to_cuda_graph_strategy_option(std::underlying_type_t<CudaGraphStrategyOption> v);
```
tp5uiuc: We don't need to expose these functions in the header, perfectly fine to use helper functions in C++ files, in an anonymous namespace.

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,256 @@
...
+    }
+  }
+  rtx_native_cudagraphs_disabled = true;
+#else
+  (void)engine_name;
+#endif
+}
+
+bool TRTRuntimeConfig::is_monolithic_capturable(nvinfer1::IExecutionContext* exec_ctx, cudaStream_t stream) const {
+#if defined(TRT_MAJOR_RTX) && defined(ENABLE_FEATURE_DISABLE_RUNTIME_ALLOCATION)
```
tp5uiuc: Remove the defined(ENABLE_FEATURE_DISABLE_RUNTIME_ALLOCATION) everywhere. It is always true for TRT_MAJOR_RTX

core/runtime/BUILD
```diff
@@ -75,11 +77,19 @@ cc_library(
         "RTDevice.h",
         "TRTEngine.h",
         "TRTEngineProfiler.h",
+        "TRTRuntimeConfig.h",
         "runtime.h",
     ],
     linkopts = [
         "-lstdc++fs",
     ],
+    local_defines = select({
```
tp5uiuc, line 102: This is an unfortunate side-effect of having feature-gating in public headers, this should have been caught earlier. I will raise it with the broader team. No action needed for this MR.

core/runtime/TRTEngine.cpp
```diff
@@ -508,17 +522,38 @@ void TRTEngine::reset_captured_graph() {
...
+  runtime_cfg.disable_rtx_native_cudagraphs(name);
+  if (!was_disabled && runtime_cfg.rtx_native_cudagraphs_disabled) {
+    // The CUDA graph strategy on the IRuntimeConfig has been flipped; rebuild exec_ctx
+    // so the new strategy takes effect for subsequent enqueueV3 calls.
+    recreate_execution_context();
+  }
+}
+
+void TRTEngine::recreate_execution_context() {
+  runtime_cfg.ensure_initialized(cuda_engine.get());
```
tp5uiuc: TODO : Should we serialize the cache here to avoid multiple kernels being compiled again and again?

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,251 @@
...
+    if (!ok) {
+      LOG_WARNING("Failed to update CUDA graph strategy on IRuntimeConfig after disable.");
+    }
+  }
+  rtx_native_cudagraphs_disabled = true;
+#endif
+}
+
+bool TRTRuntimeConfig::is_monolithic_capturable(
+    TORCHTRT_UNUSED nvinfer1::IExecutionContext* exec_ctx,
```
tp5uiuc, line 198: I could have used `[[maybe_unused]]` but that goes against project conventions (at least I don't see it being used)

tests/py/dynamo/models/test_dynamic_shapes_kernel_strategy_models.py
```diff
@@ -129,5 +129,69 @@ def test_dynamic_batch_none(self):
...
+    def tearDown(self):
+        torch._dynamo.reset()
+
+    def _compile_and_verify_cpp(self, model, strategy):
+        input_tensor = torch.randn(4, 3, 224, 224).cuda()
+        compiled = torchtrt.compile(
+            model,
+            ir="dynamo",
+            inputs=[
+                torchtrt.Input(
```
tp5uiuc: Can't we reuse all the code that we wrote for python instead of writing the same thing again? Refactor to not repeat yourself. Do it for all cases, across runtime cache, dynamic shapes and cuda-gaphs.

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,251 @@
...
+    LOG_DEBUG("No existing runtime cache at " << path);
+    return;
+  }
+  std::ifstream f(path, std::ios::binary);
+  std::vector<char> buf((std::istreambuf_iterator<char>(f)), std::istreambuf_iterator<char>());
+  if (buf.empty()) {
+    return;
+  }
+  bool ok = cache->deserialize(buf.data(), buf.size());
+  TORCHTRT_CHECK(ok, "IRuntimeCache::deserialize returned false for " << path);
```
tp5uiuc:
```suggestion
  TORCHTRT_CHECK(cache->deserialize(buf.data(), buf.size()), "IRuntimeCache::deserialize returned false for " << path);
```
Brevity is the soul of wit

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,251 @@
...
+      } catch (const std::exception& e) {
+        LOG_WARNING("Failed to load runtime cache from " << runtime_cache_path << ": " << e.what());
+      }
+      bool ok = config->setRuntimeCache(*runtime_cache);
+      if (!ok) {
+        LOG_WARNING("Failed to attach runtime cache to IRuntimeConfig; cache will be unused.");
+        runtime_cache.reset();
+      } else {
+        LOG_DEBUG("TensorRT-RTX runtime cache configured at " << runtime_cache_path);
+      }
```
tp5uiuc:
```suggestion
      if (config->setRuntimeCache(*runtime_cache)) {
              LOG_DEBUG("TensorRT-RTX runtime cache configured at " << runtime_cache_path);
      } else {
        LOG_WARNING("Failed to attach runtime cache to IRuntimeConfig; cache will be unused.");
        runtime_cache.reset();
       }
```

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -0,0 +1,251 @@
...
+      << engine_name << " for the remainder of its lifetime.");
+  // Persist any kernels the engine-internal capture has compiled so far; the outer
+  // capture will run without them otherwise, and we want future reloads to reuse them.
+  save_runtime_cache();
+  cuda_graph_strategy = CudaGraphStrategyOption::kDisabled;
+  if (config) {
+    bool ok = config->setCudaGraphStrategy(nvinfer1::CudaGraphStrategy::kDISABLED);
+    if (!ok) {
+      LOG_WARNING("Failed to update CUDA graph strategy on IRuntimeConfig after disable.");
+    }
```
tp5uiuc:
```suggestion
    if (!config->setCudaGraphStrategy(nvinfer1::CudaGraphStrategy::kDISABLED)) {
      LOG_WARNING("Failed to update CUDA graph strategy on IRuntimeConfig after disable.");
    }
```

core/runtime/TRTEngine.cpp
tp5uiuc: [by Claude Code] Handled by the extraction of `recreate_execution_context()` in the refactor round (commit a4989c760 and later). That helper now derives the allocation strategy from `resource_allocation_strategy`, so `disable_profiling` preserves `kUSER_MANAGED` when the engine was created with the dynamic resource allocation strategy. Resolving.

core/runtime/TRTEngine.cpp
tp5uiuc: [by Claude Code] Same resolution as the `disable_profiling` thread — `set_device_memory_budget` now funnels through `recreate_execution_context()`, which re-derives the allocation strategy from `resource_allocation_strategy`. Resolving.

core/runtime/TRTEngine.cpp
tp5uiuc: [by Claude Code] Addressed in e8521233c — `recreate_execution_context` now calls `runtime_cfg.save_runtime_cache()` before rebuilding the `IExecutionContext`. The destructor still saves on teardown, but flushing at every recreate provides a checkpoint so a process kill between an allocator flip and teardown no longer loses kernels compiled during the previous context. No-op on standard TRT / when no cache path is configured.

core/runtime/TRTRuntimeConfig.cpp
tp5uiuc: [by Claude Code] Addressed in e8521233c — the intermediate `bool ok` is gone, the `setCudaGraphStrategy` call is inlined into the `if` condition.

core/runtime/TRTEngine.cpp
tp5uiuc: [by Claude Code] Addressed in e8521233c — `recreate_execution_context` now flushes the runtime cache before rebuilding the execution context. Comment above the call explains the motivation.

core/runtime/TRTRuntimeConfig.cpp
tp5uiuc: [by Claude Code] Acknowledged — `TORCHTRT_UNUSED` matches existing project convention, keeping it. Resolving.

tests/py/dynamo/models/test_dynamic_shapes_kernel_strategy_models.py
tp5uiuc: [by Claude Code] Addressed in e8521233c — converged on `parameterized.expand(_RUNTIMES)` for all three features. test_000_runtime_cache.py / test_runtime_cache_models.py / test_dynamic_shapes_kernel_strategy_models.py each have a single parameterized body that runs on both runtimes; the dedicated Cpp\* sibling classes and helpers (`_compile_cpp`, `CppSimpleModel`, `_fresh_cpp_model_and_inputs`) are gone. The Cpp variant skips itself inside the test body when `torch_tensorrt_runtime` is off. cuda_graph_strategy is C++-only by design (no Python runtime path exposed yet) so nothing to dedup there. Net: ~200 lines removed, same coverage, symmetric test execution across runtimes.

core/runtime/TRTRuntimeConfig.cpp
tp5uiuc: [by Claude Code] Applied in e8521233c.

core/runtime/TRTRuntimeConfig.cpp
tp5uiuc: [by Claude Code] Applied in e8521233c.

core/runtime/TRTRuntimeConfig.cpp
tp5uiuc: [by Claude Code] Applied in e8521233c.

tp5uiuc:
Gated by https://github.com/pytorch/TensorRT/pull/4164 which needs to merge first

core/runtime/TRTRuntimeConfig.h
```diff
@@ -10,6 +10,13 @@
 
 #include "NvInfer.h"
 
+// `TRT_HAS_IRUNTIME_CONFIG` is injected by `core/runtime/BUILD` and reflects whether
+// the linked TensorRT exports `nvinfer1::IRuntimeConfig` (introduced in TensorRT
+// 10.11). It is defined on every build configuration except `:jetpack`, whose
+// l4t-r36.4 TensorRT bundle predates 10.11. On the Jetpack path the runtime falls
+// back to the legacy `ICudaEngine::createExecutionContext()` no-arg overload and
+// treats the rest of this struct's IRuntimeConfig state as inert.
+
```
tp5uiuc: Remove this comment.

core/runtime/TRTEngine.cpp
```diff
@@ -637,19 +648,50 @@ void TRTEngine::release_nccl_comm() {
...
+          ? nvinfer1::ExecutionContextAllocationStrategy::kUSER_MANAGED
+          : nvinfer1::ExecutionContextAllocationStrategy::kSTATIC);
+#ifdef TRT_HAS_IRUNTIME_CONFIG
+  exec_ctx = make_trt(cuda_engine->createExecutionContext(runtime_cfg.config.get()));
+#else
+  // Older standard TensorRT (e.g. Jetpack) doesn't expose IRuntimeConfig; fall back
+  // to the legacy createExecutionContext() overload, which uses TensorRT's built-in
+  // default allocation strategy. The set_execution_context_allocation_strategy call
+  // above is a no-op on this path.
+  exec_ctx = make_trt(cuda_engine->createExecutionContext());
```
tp5uiuc: You might want to set the allocation strategy here too.

      this->exec_ctx =
          make_trt(cuda_engine->createExecutionContext(
this->resource_allocation_strategy == TRTEngine::ResourceAllocationStrategy::kDynamic ? nvinfer1::ExecutionContextAllocationStrategy::kUSER_MANAGED : nvinfer1::ExecutionContextAllocationStrategy::kSTATIC));

See if 
```
resource_allocation_strategy == ResourceAllocationStrategy::kDynamic
          ? nvinfer1::ExecutionContextAllocationStrategy::kUSER_MANAGED
          : nvinfer1::ExecutionContextAllocationStrategy::kSTATIC
```
can be extracted out as a common const variable and set both `TRT_HAS_IRUNTIME_CONFIG` and non `TRT_HAS_IRUNTIME_CONFIG` with it.

core/runtime/TRTRuntimeConfig.cpp
```diff
@@ -146,12 +147,23 @@ void TRTRuntimeConfig::ensure_initialized(nvinfer1::ICudaEngine* cuda_engine) {
...
 }
 
-void TRTRuntimeConfig::set_execution_context_allocation_strategy(
-    nvinfer1::ExecutionContextAllocationStrategy strategy) const {
-  TORCHTRT_ASSERT(config, "TRTRuntimeConfig::config must be initialized before setting allocation strategy");
-  config->setExecutionContextAllocationStrategy(strategy);
+std::shared_ptr<nvinfer1::IExecutionContext> TRTRuntimeConfig::create_execution_context(
+    nvinfer1::ICudaEngine* cuda_engine,
+    nvinfer1::ExecutionContextAllocationStrategy allocation_strategy) {
+  TORCHTRT_CHECK(cuda_engine != nullptr, "create_execution_context requires a live ICudaEngine");
```
tp5uiuc: Redundant check, remove.

core/runtime/TRTRuntimeConfig.h
tp5uiuc: [by Claude Code] Done in abeca04b6 (header comment block removed when the BUILD-side define landed).

core/runtime/TRTEngine.cpp
tp5uiuc: [by Claude Code] Done in abeca04b6 — extracted as const allocation_strategy and used the legacy createExecutionContext(strategy) overload on the non-IRuntimeConfig path.

core/runtime/TRTRuntimeConfig.cpp
tp5uiuc: [by Claude Code] Done in abeca04b6 — removed; ensure_initialized already does the same null check.

tp5uiuc:
Closing the stale MR.

PR #4211, opened 2026-04-24
engine caching for distributed

apbose:


py/torch_tensorrt/dynamo/conversion/_conversion.py
```diff
@@ -262,6 +262,41 @@ def interpret_module_to_result(
...
+    if _distributed_caching:
+        import os as _os
+
+        from filelock import FileLock
+
+        # is_distributed_caching_enabled guarantees engine_cache and hash_val are set.
+        assert engine_cache is not None
+        assert hash_val is not None
+
+        _lock_path = _os.path.join(engine_cache.engine_cache_dir, f".{hash_val}.lock")
```
narendasan, line 283: You should build a locking facility into the core BaseEngineCache class so that if someone uses some other method other than disk cache it still works the same. For DiskCache FileLocks might be the right choice but say you use S3 buckets, it might be something else

zewenli98: yeah I think it would be better to make the distributed feature self-contained. Specifically, we need an abstract template defined in BaseEngineCache (so that other people can easily implement their own caching strategy) and we write a concrete implementation for DiskEngineCache

apbose:
Closing this for now. Will reopen when I work on this

PR #4213, opened 2026-04-27
user provided bound for torchtrt compile when export dimension is unb…

apbose:
This PR  validates user bounds against any finite exporter bounds, with three branches:                                                                                                                                                                                                                                                                                                                                     
  - Upper overflow (user_max > exp_max) → ValueError. TRT engine profile follows the exporter; shapes above exp_max would be rejected at runtime anyway.
  - Lower underflow (user_min < exp_min) → ValueError, except the (user_min=1, exp_min=2) case which is just PyTorch's 0/1-specialization artifact (Dim(min=1) silently becomes 2 in ShapeEnv), this particular case emits a warning.                                                                                                                                                                                                       
  -  Subset (user range strictly inside exporter range) → engine profile is narrowed to the user's bounds (logged at INFO). The user's Input(min_shape, max_shape) is the contract for the TRT profile; a subset of the exporter's valid range is still correct, and a tighter profile lets TRT specialize more aggressively.                                                                                             
  - Dim.DYNAMIC (unbounded exporter upper) → no check, user fills the gap (the intended use).

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -895,7 +895,7 @@ def _build_user_symbol_bounds(
     gm: torch.fx.GraphModule,
     sample_arg_inputs: Sequence[Input],
     sample_kwarg_inputs: dict[Any, Any],
-    graph_signature: Optional[torch.export.graph_signature.ExportGraphSignature] = None,
+    graph_signature: torch.export.graph_signature.ExportGraphSignature,
 ) -> Dict[sympy.Symbol, Tuple[int, int]]:
     """Map ``sympy.Symbol -> (min, max)`` from dynamic ``Input``s, used to
     fill ``Dim.DYNAMIC`` upper bounds without mutating ``ShapeEnv``.
@@ -910,30 +910,34 @@ def _build_user_symbol_bounds(
     # only contains USER_INPUT nodes, but graph_signature.input_specs still lists
     # all specs (PARAMETER, BUFFER, USER_INPUT). A positional zip misaligns.
     # Use name-based lookup keyed on graph_signature USER_INPUT names instead.
-    # When graph_signature is absent (e.g. direct unit-test calls), fall back to
-    # all placeholder nodes in graph order (which are already USER_INPUT only
-    # after ep.module()).
     placeholder_by_name = {n.name: n for n in gm.graph.nodes if n.op == "placeholder"}
-    if graph_signature is not None:
-        user_input_names = [
-            spec.arg.name
-            for spec in graph_signature.input_specs
-            if spec.kind == InputKind.USER_INPUT and hasattr(spec.arg, "name")
-        ]
-        placeholders = [
-            placeholder_by_name[name]
-            for name in user_input_names
-            if name in placeholder_by_name
-        ]
-    else:
-        placeholders = list(placeholder_by_name.values())
+    user_input_names = [
+        spec.arg.name
+        for spec in graph_signature.input_specs
+        if spec.kind == InputKind.USER_INPUT and hasattr(spec.arg, "name")
+    ]
+    placeholders = [
+        placeholder_by_name[name]
+        for name in user_input_names
+        if name in placeholder_by_name
+    ]
 
-    # Avoid flatten_up_to: the user may compile with a different arg/kwarg split
-    # than was used at export time, causing a pytree arity mismatch. Match the
-    # existing pattern in compile_module: positional args first, then kwargs.
-    flat_inputs: List[Any] = list(sample_arg_inputs)
-    if isinstance(sample_kwarg_inputs, dict):
-        flat_inputs.extend(sample_kwarg_inputs.values())
+    # Use _in_spec.flatten_up_to to preserve export-time kwarg ordering.
+    # Fall back to name-based kwarg lookup when the arg/kwarg split differs
+    # from export time (which would cause flatten_up_to to raise an arity error).
+    in_spec = getattr(gm, "_in_spec", None)
+    try:
+        if in_spec is None:
+            raise AttributeError("_in_spec not found on graph module")
+        flat_inputs = in_spec.flatten_up_to(
+            (tuple(sample_arg_inputs), sample_kwarg_inputs)
+        )
+    except (ValueError, AttributeError):
+        flat_inputs = list(sample_arg_inputs)
+        if isinstance(sample_kwarg_inputs, dict):
+            for name in user_input_names[len(sample_arg_inputs) :]:
+                if name in sample_kwarg_inputs:
+                    flat_inputs.append(sample_kwarg_inputs[name])
 
     user_symbol_bounds: Dict[sympy.Symbol, Tuple[int, int]] = {}
 
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -11,6 +11,7 @@ import sympy
 import torch
 from torch.export import ExportedProgram
 from torch.fx.node import Target
+from torch.utils._sympy.numbers import int_oo
 from torch_tensorrt._Device import Device
 from torch_tensorrt._enums import EngineCapability, dtype
 from torch_tensorrt._features import ENABLED_FEATURES, needs_cross_compile
@@ -895,20 +896,21 @@ def _build_user_symbol_bounds(
     sample_kwarg_inputs: dict[Any, Any],
     graph_signature: torch.export.graph_signature.ExportGraphSignature,
 ) -> Dict[sympy.Symbol, Tuple[int, int]]:
-    """Collect ``{sympy.Symbol: (min, max)}`` from user-supplied ``Input``s.
+    """Build a read-only ``{sympy.Symbol: (min, max)}`` map from dynamic ``Input``s.
 
-    This is a *read-only* bridge between ``torch_tensorrt.Input`` (where the
-    user declares ``min_shape``/``max_shape``) and the partitioner's shape
-    reader (``extract_var_range_info``). Each sympy symbol that appears in a
-    top-level placeholder's ``meta["val"].shape`` is recorded once with the
-    corresponding ``(min_shape[d], max_shape[d])`` from the user-provided
-    dynamic ``Input``.
+    Symbols are taken from top-level placeholder ``meta["val"].shape``; the
+    map is threaded to ``extract_var_range_info`` to fill upper bounds left
+    unbounded by ``Dim.DYNAMIC`` (``int_oo``). ``ShapeEnv`` is never mutated.
 
-    The map is consulted only to *fill* missing upper bounds (``int_oo`` /
-    unbounded) left by ``Dim.DYNAMIC``; the parent ``ShapeEnv`` is never
-    mutated. As a result, downstream consumers such as
-    :func:`torch_tensorrt.save(..., output_format="exported_program")` see
-    the original ``range_constraints`` from the exporter.
+    Validation against the exporter's finite bounds (when present):
+
+    - **Outside** (``user_min < exp_min`` or ``user_max > exp_max``): raises
+      ``ValueError`` -- shapes the user listed in ``Input`` would be rejected
+      by TRT at runtime since the engine profile follows the exporter.
+    - **Subset** (user fully inside exporter's range): emits a warning that
+      the engine profile will be widened to the exporter's bounds.
+    - **``Dim.DYNAMIC``** (unbounded exporter upper): no check; user's
+      ``Input`` fills the gap.
     """
     name_to_node = {n.name: n for n in gm.graph.nodes if n.op == "placeholder"}
     placeholders = [name_to_node[name] for name in graph_signature.user_inputs]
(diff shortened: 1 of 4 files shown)


apbose:
CI index put cases failing. Looking at it.

narendasan:
This PR is only for externally exported programs?

apbose:
@narendasan yeah it addresses (torch_tensorrt.dynamo.compile(exported_program, ...)). The flow for torch_tensorrt.compile(model, inputs=..) should be unaffected, since in that case export bounds and input bounds should be same

narendasan (commented):
You should use the user bounds in the narrow case

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -874,6 +876,139 @@ def _insert_complex_io_adapters(
...
+            if exp_max_unbounded:
+                # Dim.DYNAMIC: user fills the gap (intended use).
+                continue
+            try:
+                exp_min = int(exp_lower)
+                exp_max = int(exp_upper)
+            except (TypeError, ValueError):
+                continue
+            if user_min == exp_min and user_max == exp_max:
+                continue
```
narendasan, line 1024: Can you explain what the logic here is?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -874,6 +876,139 @@ def _insert_complex_io_adapters(
...
+                f"exporter({exp_min}, {exp_max})."
+            )
+            hint = (
+                f" Re-export with Dim('{expr}', min={user_min}, "
+                f"max={user_max}) or adjust Input to match."
+            )
+
+            if user_max > exp_max:
+                raise ValueError(
+                    f"{mismatch} Input.max_shape exceeds the exporter's max "
```
narendasan: the exported program right?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -874,6 +876,139 @@ def _insert_complex_io_adapters(
...
+                    f"({user_max} > {exp_max}); TRT will reject shapes above "
```
narendasan: TRT will reject or pytorch will?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -874,6 +876,139 @@ def _insert_complex_io_adapters(
...
+                    f"{exp_max} at runtime.{hint}"
+                )
+
+            if user_min < exp_min:
+                # 1->2 is the 0/1 specialization artifact, not a user error.
+                if user_min == 1 and exp_min == 2:
+                    logger.warning(
+                        "%s Input.min_shape=1 vs exporter min=2 is the "
```
narendasan: I think we need to make sure these error messages are clear to follow from a user perspective

py/torch_tensorrt/dynamo/_compiler.py
apbose: This block runs only when the exported program already bounds the symbol to a finite range (e.g. exported with Dim("batch", min=10, max=20)). The compiled engine's TRT optimization profile follows that exported range, so we compare the user's Input range against it here, at compile time, and handle four cases:
 - exported upper is unbounded (Dim.DYNAMIC) → user's Input fills the gap, nothing to validate (the continue just above);
- Input range == exported range → nothing to do;
- Input extends beyond the exported range (user_max > exp_max or user_min < exp_min) → those shapes can never run, so we raise now;
- Input strictly inside the exported range → safe to honor, we narrow the engine profile to the Input.
 The user_min == exp_min and user_max == exp_max check on this line is just the "exact match → no-op" early-out.

py/torch_tensorrt/dynamo/_compiler.py
apbose: yeah the exported program. will change the wording

py/torch_tensorrt/dynamo/_compiler.py
apbose: TRT will.  [10,20] + Input min_shape=2: shows
IExecutionContext::setInputShape: Error Code 3: API Usage Error
(Set dimension [2,8] for tensor x does not satisfy any optimization profiles.
 Valid range for profile 0: [10,8]..[20,8])
 through core/runtime/execute_engine.cpp:149.

apbose:
failing tests- 
L0 dynamo core tests- RTX Linux- runtime cache failure
L0 dynamo core tests- Linux- same
L0 dynamo converter tests- RTX  Linux- cumsum operator
L0 torchscript tests- windows- test_e2e_behavior.py worker crashes,  test_get_layer_info Key Bindings missing
Python only runtime tests- wrong bounda

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -898,6 +900,149 @@ def _insert_complex_io_adapters(
...
+        sample_input = sample_by_name[node.target]
+        fake_val = node.meta.get("val")
+        if not isinstance(fake_val, torch.Tensor):
+            continue
+
+        min_shape = sample_input.shape["min_shape"]
+        max_shape = sample_input.shape["max_shape"]
+
+        for d, dim in enumerate(fake_val.size()):
+            if not isinstance(dim, torch.SymInt) or d >= len(min_shape):
```
narendasan: in the case that its not SymInt shouldnt we check if they match the user dim?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -929,6 +1074,12 @@ def compile_module(
     if sample_kwarg_inputs is None:
         sample_kwarg_inputs = {}
 
+    # Forwarded to the partitioner to fill Dim.DYNAMIC upper bounds.
+    # Read-only w.r.t. ShapeEnv so range_constraints survive save/re-export.
+    user_symbol_bounds = _build_user_symbol_bounds(
```
narendasan: instead of trying to handle args, kwargs, why not operate on the flattened pytree inputs? If you really do need to do this work this early then you can always call in_spec to flatten the args, kwargs. Then the order is deterministic and you arent doing name matching

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -898,6 +900,149 @@ def _insert_complex_io_adapters(
...
+
+    Validates against finite exporter bounds: ``user_max > exp_max`` and
+    ``user_min < exp_min`` raise (TRT would reject those shapes at runtime);
+    a strict subset narrows the engine profile to the user's bounds (info
+    log only); the ``user_min=1, exp_min=2`` case warns -- it's PyTorch's
+    0/1 specialization artifact, not a user error.
+    """
+    placeholders = [n for n in gm.graph.nodes if n.op == "placeholder"]
+
+    sample_by_name: dict[str, Input] = {}
```
narendasan: Like this code should go away with pytree flattening and you can just zip the placeholders

py/torch_tensorrt/dynamo/_compiler.py
narendasan: `d >= len(min_shape)`: in what case would we want this to be let through? the rank of the tensor cant change

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -898,6 +900,149 @@ def _insert_complex_io_adapters(
...
+        min_shape = sample_input.shape["min_shape"]
+        max_shape = sample_input.shape["max_shape"]
+
+        for d, dim in enumerate(fake_val.size()):
+            if not isinstance(dim, torch.SymInt) or d >= len(min_shape):
+                continue
+            expr = dim.node.expr
+            # Composite exprs (e.g. ``2*s0``) are recomputed by
+            # ``ShapeEnv.bound_sympy``; overriding them directly would lie.
+            if not isinstance(expr, sympy.Symbol):
```
narendasan, line 976: You should log that this dynamic dimension is bounded by some other dynamic dimension

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -898,6 +901,164 @@ def _insert_complex_io_adapters(
...
+    ``user_min < exp_min`` raise (TRT would reject those shapes at runtime);
+    a strict subset narrows the engine profile to the user's bounds (info
+    log only); the ``user_min=1, exp_min=2`` case warns -- it's PyTorch's
+    0/1 specialization artifact, not a user error.
+    """
+    placeholders = [n for n in gm.graph.nodes if n.op == "placeholder"]
+
+    # Flatten args+kwargs in pytree order — guaranteed to match placeholder
+    # order by torch.export, so we can zip directly without name matching.
+    flat_inputs, _ = pytree.tree_flatten((list(sample_arg_inputs), sample_kwarg_inputs))
```
narendasan: This should come from the exported program (i believe its called in_spec)

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -898,6 +901,164 @@ def _insert_complex_io_adapters(
...
+    """Map ``sympy.Symbol -> (min, max)`` from dynamic ``Input``s, used to
+    fill ``Dim.DYNAMIC`` upper bounds without mutating ``ShapeEnv``.
+
+    Validates against finite exporter bounds: ``user_max > exp_max`` and
+    ``user_min < exp_min`` raise (TRT would reject those shapes at runtime);
+    a strict subset narrows the engine profile to the user's bounds (info
+    log only); the ``user_min=1, exp_min=2`` case warns -- it's PyTorch's
+    0/1 specialization artifact, not a user error.
+    """
+    placeholders = [n for n in gm.graph.nodes if n.op == "placeholder"]
```
narendasan: Same thing here, there should be a flat list of inputs to the graph

apbose: we would need the placeholders for the faketensor values right?

apbose: Pushed in the change to use in_spec. But I believe we still need  the placeholders, since in_spec would just give the ordering, but not the symbolic shape info?

narendasan: `exported_program.graph_signature.user_inputs` I think this is the API to use because in theory with functionalized graphs (like the torch.compile graphs) you might have additional inputs that users have not provided right? We wouldnt have specs for those

apbose: Updated the PR to use graph_signature.user_inputs for filtering placeholder nodes instead of iterating all placeholders. Need to rebase though

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -884,13 +888,174 @@ def _insert_complex_io_adapters(
...
+    fill ``Dim.DYNAMIC`` upper bounds without mutating ``ShapeEnv``.
+
+    Validates against finite exporter bounds: ``user_max > exp_max`` and
+    ``user_min < exp_min`` raise (TRT would reject those shapes at runtime);
+    a strict subset narrows the engine profile to the user's bounds (info
+    log only); the ``user_min=1, exp_min=2`` case warns -- it's PyTorch's
+    0/1 specialization artifact, not a user error.
+    """
+    name_to_node = {n.name: n for n in gm.graph.nodes if n.op == "placeholder"}
+    placeholders = [name_to_node[name] for name in graph_signature.user_inputs]
```
narendasan: what user_inputs are not placeholders?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -922,6 +1087,16 @@ def compile_module(
...
         )
+
+    # Forwarded to the partitioner to fill Dim.DYNAMIC upper bounds.
+    # Read-only w.r.t. ShapeEnv so range_constraints survive save/re-export.
+    if graph_signature is not None:
+        user_symbol_bounds = _build_user_symbol_bounds(
+            gm, sample_arg_inputs, sample_kwarg_inputs, graph_signature
+        )
+    else:
+        user_symbol_bounds = {}
```
narendasan: Why would this case ever be triggered? What happens if it does?

py/torch_tensorrt/dynamo/_compiler.py
apbose: all user_inputs would be placeholders, but not all placeholders would be user_inputs. placeholders would also have the lifted params and buffers

py/torch_tensorrt/dynamo/_compiler.py
apbose: its a safety fallback if the compile_module is passed without the graph_signature. In that case we revert back to the behavior earlier. I can actually remove it and keep the graph_signature as not Optional. Both call sites in compile() calls it with the graph signature

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -884,6 +889,170 @@ def _insert_complex_io_adapters(
...
+    a strict subset narrows the engine profile to the user's bounds (info
+    log only); the ``user_min=1, exp_min=2`` case warns -- it's PyTorch's
+    0/1 specialization artifact, not a user error.
+    """
+    all_placeholders = [n for n in gm.graph.nodes if n.op == "placeholder"]
+    placeholders = [
+        p
+        for p, s in zip(all_placeholders, graph_signature.input_specs)
+        if s.kind == InputKind.USER_INPUT
+    ]
```
zewenli98: Is the order of all_placeholders and graph_signature.input_specs guaranteed? 
If input_specs contains more entries (e.g., inserted lifted inputs), zip() will silently truncate to the shorter list and can misalign placeholders. 
If the graph contains placeholders in a different order, some user inputs will be skipped or matched to the wrong spec.

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -884,6 +889,170 @@ def _insert_complex_io_adapters(
...
+
+    in_spec = getattr(gm, "_in_spec", None)
+    assert in_spec is not None, "Exported graph module missing _in_spec"
+    flat_inputs = in_spec.flatten_up_to((list(sample_arg_inputs), sample_kwarg_inputs))
+
+    user_symbol_bounds: Dict[sympy.Symbol, Tuple[int, int]] = {}
+
+    for node, inp in zip(placeholders, flat_inputs):
```
zewenli98, line 944: Are placeholders and flat_inputs matched as well?

py/torch_tensorrt/dynamo/_compiler.py
apbose: The ordering is guaranteed — input_specs and placeholder nodes are built together during torch.export(). Pytorch uses the same pattern in here https://github.com/pytorch/pytorch/blob/main/torch/export/exported_program.py#L1683-L1687

py/torch_tensorrt/dynamo/_compiler.py
apbose: In torch/export/_trace.py https://github.com/pytorch/pytorch/blob/main/torch/export/_trace.py#L1509
 _, original_in_spec = pytree.tree_flatten((args, kwargs))
This original_in_spec is what becomes gm._in_spec on the graph module (stored via ExportArtifact.in_spec). The placeholder nodes are created from the same pytree flattening of the same (args, kwargs) — so their order is determined by the same spec.

When we call in_spec.flatten_up_to((sample_arg_inputs, sample_kwarg_inputs)), we're reproducing that exact flattening order on the user's inputs. And USER_INPUT entries in input_specs are in the same placeholder order. Both come from the same source, so the zip should be correctly aligned.

PR #4216, opened 2026-04-28
ASR PR changes for arange and expand, adding testcases

apbose:


apbose:
the fix seems correct but the test cases of expand need to be worked on, to catch the fix scenario

apbose:
closing this for now

PR #4228, opened 2026-05-01
Fix tensor lifetime issue

SandSnip3r:
# Description

This change fixes a correctness issue that I and others were seeing when running the FLUX2 diffusion model. The model, when compiled with either TensorRT or TensorRT-RTX was producing garbage images.

The issue was that the input tensor's lifetime was incorrect. The input tensor's ref count dropped to 0 before the engine ran with `enqueueV3()`. In this specific case, it was a bit of a perfect storm with an output having the same size and shape and also there being a fp32->bf16 cast. Another tensor was being allocated (the output tensor) and that was given the address of the input tensor.

## Type of change

# Checklist:

### DIFF
--- a/core/runtime/execute_engine.cpp
+++ b/core/runtime/execute_engine.cpp
@@ -92,7 +92,13 @@ bool _validate_shapes(std::vector<at::Tensor> inputs, c10::intrusive_ptr<TRTEngi
   return false;
 }
 
-void setup_input_tensors(
+// Returns the contiguified input tensors. The caller must keep the returned
+// list alive until enqueueV3 completes: the engine is bound to .data_ptr()s
+// inside it, and freshly-allocated contig copies (from .contiguous() on
+// non-contig inputs) would otherwise be freed and their CUDA addresses
+// recycled by the caching allocator for output tensors, aliasing inputs
+// onto outputs and corrupting reads after the first output write.
+std::list<at::Tensor> setup_input_tensors(
     std::vector<at::Tensor> inputs,
     c10::intrusive_ptr<TRTEngine> compiled_engine,
     bool cudagraphs_enabled,
@@ -169,6 +175,7 @@ void setup_input_tensors(
           "Failed to bind tensor address for " << name);
     }
   }
+  return formatted_inputs;
 }
 
 std::vector<at::Tensor> create_output_tensors(c10::intrusive_ptr<TRTEngine> compiled_engine) {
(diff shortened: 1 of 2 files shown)


core/runtime/execute_engine.cpp
SandSnip3r, line 193: By the way, a separate cleanup should be done where this line is instead
```
outputs[pyt_idx] = at::empty(dims, at::TensorOptions().device(at::kCUDA).dtype(type));
```
This would improve from two allocations & a dtype-conversion kernel to just a single allocation.

narendasan: I think this is the same line Shane identified as well.

narendasan (approved):
This looks good to me

core/runtime/execute_engine.cpp
```diff
@@ -92,7 +92,13 @@ bool _validate_shapes(std::vector<at::Tensor> inputs, c10::intrusive_ptr<TRTEngi
...
 
-void setup_input_tensors(
+// Returns the contiguified input tensors. The caller must keep the returned
+// list alive until enqueueV3 completes: the engine is bound to .data_ptr()s
+// inside it, and freshly-allocated contig copies (from .contiguous() on
+// non-contig inputs) would otherwise be freed and their CUDA addresses
+// recycled by the caching allocator for output tensors, aliasing inputs
+// onto outputs and corrupting reads after the first output write.
+std::list<at::Tensor> setup_input_tensors(
     std::vector<at::Tensor> inputs,
```
narendasan, line 102: What if we mutated inputs inplace so something like 

```c++
void setup_input_tensors( 
    std::vector<at::Tensor>& inputs,
    ...

```

Similar to what is happening for the inputShapeTensors

narendasan (approved):
I think the only alternative design I can think of is that we mutate the vector of input tensors inplace with the new contiguous versions.

core/runtime/execute_engine.cpp
SandSnip3r: Yeah, it kinda seems like the same thing but wrapped in a different package

narendasan: I think fundamentally we need to hold the tensors through the whole execution. I guess the question is if we need to keep the original user copies as well or just the contig copies

SandSnip3r:
closed in favor of a different design: https://github.com/pytorch/TensorRT/pull/4247

PR #4232, opened 2026-05-01
fix: respect caller's CUDA stream in TRT runtime (Green Context support)

shoumikhin:
### What this fixes

Callers commonly pin work to a specific CUDA stream via `torch.cuda.stream(...)` (Python) or `c10::cuda::CUDAStreamGuard(...)` (C++) for NCCL, AMP overlap, custom scheduling, and most recently CUDA Green Contexts (CUDA 12.4+) for SM partitioning. A green-context-derived stream is the only way to confine a TensorRT engine to a partition.

Today the Torch-TensorRT runtime ignores the caller's stream. It reads `getCurrentCUDAStream()`, then substitutes one from PyTorch's pool. Pool streams are not bound to any green context, so the partition is silently bypassed. The same pattern lives in both Python runtime modules. Net effect: Torch-TensorRT cannot be used for GPU partitioning at all today.

### What this PR does

The runtime now picks the engine's stream like this:

- If the caller is on the regular default stream, behavior is unchanged. Grab a stream from the pool so the device does not get accidentally serialized on the default stream.
- If the caller is on any other stream (custom user stream, Green Context stream, NCCL stream, etc.), run the engine directly on that stream.

Two small pieces of bookkeeping fall out of this change:

- When the chosen stream changes between calls, force any captured CUDA Graph to re-record. Otherwise replay would launch on a stale stream and silently produce wrong results.
- When the engine stream and the caller's stream are the same, skip the pre/post sync events between them. They become no-ops in that case.

The same logic is mirrored in `_PythonTorchTensorRTModule.py` (hoisted to the outer `forward` so both the standard and output-allocator paths see the resolved stream) and `_CudaGraphsTorchTensorRTModule.py`.

### What stays the same for existing users

Anyone who calls `model(x)` without setting a stream sees the exact same behavior as before. No new public API. No new arguments. No `.pt2` file format change. No new fields on engine state.

### Behavior change for existing non-default-stream callers

Users who already wrap inference in `torch.cuda.stream(my_stream)` (or `c10::cuda::CUDAStreamGuard(my_stream)`) for non-green reasons (custom scheduling, NCCL, AMP) will see the engine run on their stream instead of a separate pool stream. Results are identical; the timing characteristics may differ, since the previous separate-stream + event-fence pattern is replaced by direct execution on the caller's stream. For CUDA Graphs users who alternate between several non-default streams, the captured graph re-records on each stream change.

### What a Green Context user does

```cpp
auto green_stream = c10::cuda::getStreamFromExternal(my_green_ctx_stream, /*device=*/0);
{
  c10::cuda::CUDAStreamGuard guard(green_stream);
  auto outputs = loader.run(inputs);   // runs on green_stream, inside the SM partition
}
```

### Test plan

- Existing test suite passes (no API surface changed).
- Manual: build a small `.pt2`, run inside a `CUDAStreamGuard` over a Green-Context-derived stream, capture an `nsys` trace, confirm kernels run on the green stream and stay inside the partition's SMs.
- CUDA Graphs path: switch streams between calls, confirm the captured graph re-records on stream change.

### DIFF
--- a/core/runtime/execute_engine.cpp
+++ b/core/runtime/execute_engine.cpp
@@ -243,6 +243,23 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
     bool cudagraphs_enabled = (CUDAGRAPHS_MODE == SUBGRAPH_CUDAGRAPHS);
     bool shape_changed = _validate_shapes(inputs, compiled_engine);
 
+    auto current_device_id = inputs.size() > 0 ? inputs[0].device().index() : at::cuda::current_device();
+    auto default_stream = compiled_engine->default_stream;
+    auto previous_engine_stream = compiled_engine->engine_stream;
+    compiled_engine->caller_stream = c10::cuda::getCurrentCUDAStream(current_device_id);
+    bool caller_on_default = (compiled_engine->caller_stream == default_stream);
+    if (caller_on_default) {
+      compiled_engine->engine_stream = compiled_engine->owned_pool_stream;
+    } else {
+      // Honor caller's non-default stream so its scheduling choice (e.g. SM
+      // partitioning via a CUDA Green Context) is preserved end to end.
+      compiled_engine->engine_stream = compiled_engine->caller_stream;
+    }
+    if (cudagraphs_enabled && compiled_engine->engine_stream != previous_engine_stream) {
+      // Captured CUDA graph was recorded against the old stream; force re-record.
+      compiled_engine->runtime_states.context_changed = true;
+    }
+
     // Whether cudagraphs needs to record the graph on this pass
     auto result = compiled_engine->runtime_states.set_runtime_states(
         cudagraphs_enabled, compiled_engine->use_pre_allocated_outputs, shape_changed);
@@ -310,19 +327,6 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
       }
     }
 
-    auto current_device_id = -1;
-    if (inputs.size() > 0) {
-      current_device_id = inputs[0].device().index(); // Done this way to avoid a call to cudart
-    } else if (outputs.size() > 0) {
-      current_device_id = outputs[0].device().index(); // Done this way to avoid a call to cudart
-    }
-
-    compiled_engine->caller_stream = c10::cuda::getCurrentCUDAStream(current_device_id);
-    if (compiled_engine->engine_stream == c10::cuda::getDefaultCUDAStream(current_device_id)) {
-      // Create a new stream if the engine stream is the default stream
-      compiled_engine->engine_stream = c10::cuda::getStreamFromPool(false, current_device_id);
-    }
-
     compiled_engine->record_active_input_tensor_stream_usage(
         cudagraphs_enabled ? compiled_engine->caller_stream : compiled_engine->engine_stream);
 
@@ -335,10 +339,12 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
             std::make_unique<torch::autograd::profiler::RecordProfile>(compiled_engine->enqueue_profile_path);
       }
 
-      // Block engine stream until results are available on caller stream
-      at::cuda::CUDAEvent caller_exec_complete;
-      caller_exec_complete.record(compiled_engine->caller_stream);
-      caller_exec_complete.block(compiled_engine->engine_stream);
+      if (caller_on_default) {
+        // Block engine stream until results are available on caller stream
+        at::cuda::CUDAEvent caller_exec_complete;
+        caller_exec_complete.record(compiled_engine->caller_stream);
+        caller_exec_complete.block(compiled_engine->engine_stream);
+      }
 
       if (!cudagraphs_enabled) {
         // Direct execution uses the caller buffers directly
@@ -371,10 +377,12 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
       compiled_engine->pre_allocated_outputs = create_output_tensors(compiled_engine);
     }
 
-    // Block caller stream until engine execution is complete
-    at::cuda::CUDAEvent trt_exec_complete;
-    trt_exec_complete.record(compiled_engine->engine_stream);
-    trt_exec_complete.block(compiled_engine->caller_stream);
+    if (caller_on_default) {
+      // Block caller stream until engine execution is complete
+      at::cuda::CUDAEvent trt_exec_complete;
+      trt_exec_complete.record(compiled_engine->engine_stream);
+      trt_exec_complete.block(compiled_engine->caller_stream);
+    }
 
     if (cudagraphs_enabled) {
       // If in CUDAGraph mode, copy persistent staging outputs to returned tensors on the caller stream.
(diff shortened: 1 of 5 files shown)


shoumikhin:
## Long-term plan: upstream PyTorch PR

Opened https://github.com/pytorch/pytorch/pull/182149 to add `AOTIModelPackageLoader::get_custom_objs()`. Once it lands and reaches a tagged PyTorch release, AOTI / `.pt2` C++ users can reach the live `TRTEngine` torchbind instances inside the loaded `.so` and use the per-engine `set_external_stream` API directly (no need for the process-wide `set_engine_stream_passthrough` flag).

```cpp
torch::inductor::AOTIModelPackageLoader loader("model.pt2");
for (auto& [name, ivalue] : loader.get_custom_objs()) {
  if (auto e = ivalue.toCustomClass<torch_tensorrt::TRTEngine>()) {
    e->set_external_stream(reinterpret_cast<int64_t>(my_green_stream));
  }
}
loader.run(inputs);
```

This PR (#4232) ships `set_engine_stream_passthrough` as the interim mechanism so edge / on-device users are unblocked today. A follow-up `torch_tensorrt::aoti::TRTAOTILoader` wrapper (gated on a CMake `TORCH_TRT_HAVE_AOTI_CUSTOM_OBJS` probe) will provide the clean per-engine API once upstream lands.

shoumikhin:
Friendly ping @narendasan @cehongwang.

This has been open for a week. The upstream PyTorch dependency (pytorch/pytorch#182149) is landing today, which unblocks the full end-to-end path.

Could one of you take a first pass when you get a chance? Happy to address feedback or split it up if that helps review.

narendasan:
@shoumikhin I will leave the more detailed feedback to @cehongwang but one thing I wanted to raise was would it make sense for your usecase to just pull stream management out of the execute engine op all together. So the lazy stream from the stream pool would be a bail out in the case the user is not managing this externally.  


My thought here is basically we operationalize stream management in the fx graph with a set of operators that serves as a guard / interface to prevent computation on the default stream. The concern I have is if you are mixing pytorch and tensorrt operations and using say green contexts would you be able to express everything correctly 

Something to this effect in the green context case: 
smolvla_stream_plan

I can upload a prototype if this makes sense. I think we would still use the AOTI hooks you created, just in these stream ops rather than engine by engine

shoumikhin:
@narendasan many thanks for the feedback and prototype! I rewrote this e2e on top of latest main. Could you take another look please?

What I dropped: the process-wide flag, the per-engine setter, the Python multi-stream facade. No new public API, no .pt2 format change, no torchbind additions, no FX graph changes, no header changes.

What is in: stream selection moved to the top of the lambda. If the caller is on the default stream, behavior is unchanged (grab a pool stream). If the caller is on any other stream, the engine runs on that stream directly. When the chosen stream changes between calls, set runtime_states.context_changed = true so any captured CUDA Graph re-records. The pre/post sync events are skipped when the engine stream and the caller's stream are the same. Mirrored in _PythonTorchTensorRTModule.py and _CudaGraphsTorchTensorRTModule.py.

This matches the shape of execute_engine.cpp on your stream_gaurds branch. I borrowed the caller_on_default local and the fence-skip pattern from there.

On your "pull stream management out entirely" suggestion: with this change that is what happens for any caller who installs a non-default stream. The pool-stream path stays only as the fallback for the default-stream case, which matches your wording: the lazy stream from the pool would be a fallback when the user is not managing this externally.

On mixed PyTorch + TensorRT in a green context: this PR does not push stream awareness into the FX graph. For the common case (one model per .pt2, one green context, caller wraps the whole forward in a CUDAStreamGuard), every CUDA op inside that guard inherits the green stream from the calling thread, both PyTorch ops and the TensorRT engine. So the slim change covers the "uniform green context for the whole graph" case without graph-level plumbing.

Where I think the two approaches fit together: this PR is the runtime contract (the engine runs on whatever stream is current at enqueueV3 time). stream_gaurds is the layer above for cases where the caller wants intra-graph parallelism across multiple streams (different branches on different green contexts, fan-out, etc.). Single-stream callers do not need the FX rewrite to get green contexts working, and graph-level ops still need the runtime to honor the chosen stream, so the slim runtime fix is a prerequisite for stream_gaurds rather than a competing design.

One thought I want to share, since it shaped the design: stream identity is imo a runtime scheduling property, not a model property. A cudaStream_t does not survive serialization, and the same .pt2 may run in different deployment topologies (single-tenant, multi-tenant green contexts, NCCL, etc.). Even with apply_stream_plan taking actual streams at runtime, baking the plan structure into the graph means re-exporting whenever the topology changes shape (different number of streams, different fan-out). For the multi-tenant on-device case where the caller picks the partition per process, host-level stream selection feels like the right granularity. I think stream_gaurds is the right design for the intra-graph multi-stream case but I would be careful about making it the default path for users who just want "engine respects my stream guard."

Happy to discuss any of this. Mind taking another pass?

cehongwang (approved):
Looks good. Can you rebase this PR to the main branch? You might need to move changes in PythonTorchTensorRTmodule to the new file TRTEngine.py

shoumikhin:
Rebased on `main`. @narendasan @cehongwang please merge when ready.

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -675,9 +675,65 @@ def _profile_section(self, label: str) -> ContextManager[None]:
...
+        current invocation re-records the graph against the new stream.
+        Call sites MUST invoke this before ``runtime_states.set_runtime_states``
+        because that call consumes and resets ``context_changed``.
+        """
+        current_device = (
+            contiguous_inputs[0].device
+            if contiguous_inputs and contiguous_inputs[0].is_cuda
+            else torch.device("cuda", torch.cuda.current_device())
+        )
+        default_stream = torch.cuda.default_stream(current_device)
```
cehongwang: Maybe move this to the __init__ function to avoid running it in every forward pass. Ideally, we want to keep minimal code in the forward to reduce runtime overhead. Based on my profiling experience, calling torch.cuda is time consuming

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -675,9 +675,65 @@ def _profile_section(self, label: str) -> ContextManager[None]:
...
+        needed.
+
+        Returns ``caller_on_default`` so call sites can gate the wait_stream
+        pair on it.  Also flips ``runtime_states.context_changed`` whenever
+        the engine stream changes while CUDA graphs are enabled so the
+        current invocation re-records the graph against the new stream.
+        Call sites MUST invoke this before ``runtime_states.set_runtime_states``
+        because that call consumes and resets ``context_changed``.
+        """
+        current_device = (
```
cehongwang, line 700: Same here

py/torch_tensorrt/dynamo/runtime/_CudaGraphsTorchTensorRTModule.py
```diff
@@ -121,7 +121,33 @@ def forward(
...
         if cudagraphs_enabled:
             shape_changed = self.validate_input_shapes(inputs)
-            need_cudagraphs_record = shape_changed or self.is_weight_streaming_set
+
+            current_device = (
+                inputs[0].device
+                if inputs and inputs[0].is_cuda
+                else torch.device("cuda", torch.cuda.current_device())
+            )
+            default_stream = torch.cuda.default_stream(current_device)
```
cehongwang: Same here

core/runtime/execute_engine.cpp
```diff
@@ -243,6 +243,28 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
     bool cudagraphs_enabled = (CUDAGRAPHS_MODE == SUBGRAPH_CUDAGRAPHS);
     bool shape_changed = _validate_shapes(inputs, compiled_engine);
 
+    auto current_device_id = inputs.size() > 0 ? inputs[0].device().index() : at::cuda::current_device();
+    auto default_stream = c10::cuda::getDefaultCUDAStream(current_device_id);
```
cehongwang: Same here

cehongwang (approved):
Just fix the minor changes and rebase again to main for CI tests.

shoumikhin:
@cehongwang Addressed all four inline comments by moving the `default_stream` and pool stream allocations to `__init__` / the constructor; please take another look when you have a chance.

cehongwang:
LGTM

PR #4235, opened 2026-05-06
fix: remove unnecessary elementwise kernel

narendasan:
# Description

Removes the unnecessary elementwise kernel that gets invoked when output tensors are allocated then casted. Also fixes the search path for finding torch from the local venv to prioritize in repo venvs. 

Fixes # (issue)

## Type of change

cc: @shahe-nv 

# Checklist:


lanluo-nvidia:
@narendasan 
Here are the codex review comments:

• Findings

  1. core/runtime/execute_engine.cpp:186-190 breaks multi-device safe mode for hardware-compatible engines.
     The PR now allocates outputs on compiled_engine->device_info.id, but MULTI_DEVICE_SAFE_MODE can select a different runtime device at lines 532-539 via select_rt_device(...) and set_rt_device(device). In that case inputs are moved to
     device.id, while outputs are still allocated on the original device_info.id, so TensorRT may bind output memory from the wrong GPU. The previous at::empty(..., {at::kCUDA}) allocated on the current CUDA device after set_rt_device, which
     matched the selected runtime device.
     Suggested fix: allocate on the active/selected CUDA device, or update/use the selected device id rather than compiled_engine->device_info.id.

  Source: PR head execute_engine.cpp (https://github.com/pytorch/TensorRT/blob/f3cb7118ea8c70107d4e9ad2b3fc63152a30dd36/core/runtime/execute_engine.cpp#L186-L190), device switch context execute_engine.cpp
  (https://github.com/pytorch/TensorRT/blob/f3cb7118ea8c70107d4e9ad2b3fc63152a30dd36/core/runtime/execute_engine.cpp#L532-L539).

zewenli98 (approved):
LGTM

PR #4237, opened 2026-05-06
feat(runtime): file-lock the TRT-RTX runtime cache

tp5uiuc:
## Summary

Adds a cross-platform RAII file-lock primitive (`core/util/file_lock.{h,cpp}`) wired into `load_runtime_cache` (shared) and `save_runtime_cache_impl` (exclusive) so the Python and C++ TRT-RTX runtimes sharing a `runtime_cache_path` do not race the rename and silently drop compiled kernels. Follow-up to a reviewer ask on #4202 to land locking as a separate pass.

## Backend

Matches the `filelock` Python library so the two interop on the same `<cache>.lock` file:
- **Linux/macOS:** `flock(2)` — not POSIX `fcntl(F_SETLK)`, which lives in an independent namespace and would silently fail to interop on Linux.
- **Windows:** `LockFileEx` on byte range `(0, 1)` — matches `msvcrt.locking(..., 1)` on the Python side.

`flock(2)` has no native timeout, so `try_lock_for` is a 50ms-cadence poll loop with a 10s default matching `filelock`'s `acquire(timeout=10)`. Errors propagate via `TORCHTRT_CHECK`; the existing `try/catch` in `ensure_initialized` and the `noexcept save_runtime_cache` wrapper preserve behavior on contention.

## Tests

- **C++** (`tests/cpp/test_file_lock.cpp`): 12 gtest cases — exclusive/shared/mixed contention, timeout edges, RAII release, move semantics, no-unlink-on-release, and a same-namespace \`flock(2)\` interop check.
- **Python** (`tests/py/dynamo/runtime/test_000_runtime_cache.py`): parameterizes \`test_filelock_works\` and \`test_sequential_save_load\` over both runtimes; adds \`test_python_lock_blocks_cpp_save\` (Python \`filelock\` blocks C++ save → timeout, cache unchanged, post-release save succeeds) and \`test_filelock_cross_runtime_parallel\` (two subprocesses, one per runtime, on a shared \`cache_path\`).

Local (A100, RTX): \`bazel test //tests/cpp:test_file_lock\` → 12/12 pass; full \`test_000_runtime_cache.py\` → all pass + RTX-gated skips.

## Notes

- **Depends on #4202** — should merge after that.
- More backend / API rationale and review history in tp5uiuc/TensorRT#1.

## Test plan

- [x] C++ unit tests pass (\`bazel test //tests/cpp:test_file_lock\`)
- [x] Python e2e pass (\`pytest tests/py/dynamo/runtime/test_000_runtime_cache.py\`)
- [x] Pre-commit clean (clang-format, black, isort, ruff, buildifier, typos)


core/runtime/runtime.h
```diff
@@ -40,6 +40,11 @@ typedef enum {
   REQUIRES_OUTPUT_ALLOCATOR_IDX,
   RESOURCE_ALLOCATION_STRATEGY_IDX,
   REQUIRES_NATIVE_MULTIDEVICE_IDX,
+#ifdef TRT_MAJOR_RTX
```
narendasan, line 43: I would rather not ifdef the serialization format because users might accidentally cross packages here. We can make optional slots with a sentinel value. But TRT produced programs and TRT-RTX programs should share the format

core/runtime/TRTEngine.cpp
```diff
@@ -61,26 +62,28 @@ void DynamicOutputAllocator::notifyShape(char const* tensorName, nvinfer1::Dims
 }
 
 TRTEngine::TRTEngine(
-    const std::string& serialized_engine,
+    std::string serialized_engine,
```
narendasan, line 65: why does this need to be a deep copy?

core/runtime/TRTEngine.cpp
```diff
@@ -495,7 +495,14 @@ FlattenedState TRTEngine::__obj_flatten__() {
       std::tuple("requires_output_allocator", serialized_info[REQUIRES_OUTPUT_ALLOCATOR_IDX]),
       std::tuple("target_platform", serialized_info[TARGET_PLATFORM_IDX]),
       std::tuple("resource_allocation_strategy", serialized_info[RESOURCE_ALLOCATION_STRATEGY_IDX]),
-      std::tuple("requires_native_multidevice", serialized_info[REQUIRES_NATIVE_MULTIDEVICE_IDX]));
+      std::tuple("requires_native_multidevice", serialized_info[REQUIRES_NATIVE_MULTIDEVICE_IDX])
+#ifdef TRT_MAJOR_RTX
```
narendasan, line 499: See above comment

core/runtime/TRTEngine.cpp
```diff
@@ -522,6 +529,13 @@ std::vector<std::string> TRTEngine::serialize() {
       this->resource_allocation_strategy == ResourceAllocationStrategy::kDynamic ? "1" : "0";
   serialized_info[REQUIRES_NATIVE_MULTIDEVICE_IDX] = this->requires_native_multidevice ? "1" : "0";
   // rank/world_size are runtime facts (may differ at load time); not serialized.
+#ifdef TRT_MAJOR_RTX
```
narendasan, line 532: Same here

core/runtime/runtime.h
narendasan: Also are these properties of the engine or are they runtime mode configurations? The point of this interface is the bare minimum information to reconstruct the program from disk

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -57,7 +61,25 @@
...
+        RUNTIME_CACHE_PATH_IDX = torch.ops.tensorrt.RUNTIME_CACHE_PATH_IDX()  # 12
+        DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX = (
+            torch.ops.tensorrt.DYNAMIC_SHAPES_KERNEL_STRATEGY_IDX()
+        )  # 13
+        CUDA_GRAPH_STRATEGY_IDX = torch.ops.tensorrt.CUDA_GRAPH_STRATEGY_IDX()  # 14
+    SERIALIZATION_LEN = (
+        torch.ops.tensorrt.SERIALIZATION_LEN()
+    )  # 15 (RTX) / 12 (standard)
+
+_DYNAMIC_SHAPES_KERNEL_STRATEGY_MAP: Dict[str, int] = {
```
narendasan, line 74: Are these things that need user apis?

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -57,7 +61,25 @@
...
+    "lazy": 0,
+    "eager": 1,
+    "none": 2,
+}
+_CUDA_GRAPH_STRATEGY_MAP: Dict[str, int] = {
+    "disabled": 0,
+    "whole_graph_capture": 1,
+}
```
narendasan, line 82: See above?

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -119,6 +120,7 @@ class CompilationSettings:
         autocast_calibration_dataloader (Optional[torch.utils.data.DataLoader]): The dataloader to use for autocast calibration. Default is None.
         offload_module_to_cpu (bool): Offload the model to CPU to reduce memory footprint during compilation
         dynamically_allocate_resources (bool): Dynamically allocate resources for TensorRT engines
+        cuda_graph_strategy (str): TensorRT-RTX CUDA graph strategy: "disabled" (default) or "whole_graph_capture" (let TensorRT-RTX manage CUDA graph capture/replay internally). When set and combined with `torch_tensorrt.runtime.set_cudagraphs_mode(True)` on RTX, overrides manual capture. Not used for standard TensorRT.
```
narendasan, line 123: Runtime mode controls should be controlled via context managers rather than passed in at compile time. Only information that is fixed at runtime needs to be here

tests/cpp/test_file_lock.cpp
```diff
@@ -0,0 +1,187 @@
+#include <atomic>
```
narendasan, line 1: move into //tests/core/runtime or //tests/core/util

tp5uiuc:
Closing the stale MR.

PR #4244, opened 2026-05-08
reduce_scatter bug fix native

apbose:
Checking CI

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
@@ -55,11 +55,11 @@ def _get_distributed_rank_and_world_size() -> Tuple[int, int]:
         dist_rank = dist.get_rank()
         dist_world_size = dist.get_world_size()
 
-        env_world_size = os.environ.get("WORLD_SIZE")
-        env_rank = os.environ.get("RANK")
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -341,13 +341,12 @@ def nccl_reduce_scatter_native(
...
     input_tensor = plug_inputs[0]
 
-    # Map string reduction op to TensorRT ReduceOperation enum
     reduce_op_map = {
         "sum": trt.ReduceOperation.SUM,
-        "prod": trt.ReduceOperation.PROD,
         "min": trt.ReduceOperation.MIN,
         "max": trt.ReduceOperation.MAX,
         "avg": trt.ReduceOperation.AVG,
+        "product": trt.ReduceOperation.PROD,
```
zewenli98, line 349: Just curious, is it a must to call "product" rather than "prod"?

zewenli98 (approved):
LGTM

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
apbose: yup product is the arg which is used in torch.ops._c10d_functional.reduce_scatter instead of prod

PR #4247, opened 2026-05-11
Fix TensorRT runtime input buffer lifetimes

SandSnip3r:
# Description

This fixes a C++ runtime lifetime issue where temporary formatted input buffers, especially .contiguous() copies of non-contiguous inputs, could be destroyed before TensorRT finished using the bound input addresses. In the failing case, the CUDA caching allocator could then reuse that freed input storage for an output buffer with the same shape and dtype, causing input/output aliasing and large numerical corruption in monolithic TRT engines such as FLUX.2-klein-9B.

The fix moves per-execution input and shape-tensor storage onto the `TRTEngine` so the buffers remain alive through `enqueueV3`, records CUDA stream usage for active input tensors, and clears the retained references only after execution has been launched. This also makes the runtime behavior more robust for non-contiguous inputs across the standard C++ runtime path, CUDA graphs, and output allocator mode.

Added regression coverage exercises a non-contiguous bf16 input whose output has the same shape, matching the allocator/lifetime pattern that exposed the original bug. Verification included the new runtime test plus the FLUX transformer numerical repro and full image-generation path.

## Type of change

Bug fix

# Checklist:

### DIFF
--- a/core/runtime/TRTEngine.cpp
+++ b/core/runtime/TRTEngine.cpp
@@ -2,6 +2,7 @@
 
 #include <cuda_runtime.h>
 #include "NvInfer.h"
+#include "c10/cuda/CUDACachingAllocator.h"
 #include "c10/cuda/CUDAStream.h"
 #include "torch/csrc/jit/frontend/function_schema_parser.h"
 #include "torch/cuda.h"
@@ -58,6 +59,24 @@ void* DynamicOutputAllocator::reallocateOutputAsync(
 
 void DynamicOutputAllocator::notifyShape(char const* tensorName, nvinfer1::Dims const& dims) noexcept {
   shapes[tensorName] = dims;
+}
+
+void TRTEngine::clear_active_input_tensors() {
+  active_input_tensors.clear();
+  active_shape_tensor_values.clear();
+}
+
+void TRTEngine::reset_active_input_tensors() {
+  clear_active_input_tensors();
+  active_input_tensors.resize(num_io.first);
+}
+
+void TRTEngine::record_active_input_tensor_stream_usage(const c10::cuda::CUDAStream& stream) {
+  for (const auto& input : active_input_tensors) {
+    if (input.defined() && input.is_cuda() && input.has_storage() && input.numel() > 0) {
+      c10::cuda::CUDACachingAllocator::recordStream(input.storage().data_ptr(), stream);
+    }
+  }
 }
 
 TRTEngine::TRTEngine(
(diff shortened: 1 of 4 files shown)


core/runtime/TRTEngine.cpp
```diff
@@ -237,7 +237,7 @@ TRTEngine::TRTEngine(
 
     uint64_t outputs = _out_binding_names.size();
     out_binding_names.resize(outputs);
-    output_buffers.resize(outputs);
+    cudagraph_output_staging_buffers.resize(outputs);
```
narendasan, line 259: Does this system also get used for the pre_allocated outputs / output allocator?

core/runtime/execute_engine.cpp
```diff
@@ -92,13 +93,35 @@ bool _validate_shapes(std::vector<at::Tensor> inputs, c10::intrusive_ptr<TRTEngi
   return false;
 }
 
+void clear_active_input_tensors(c10::intrusive_ptr<TRTEngine> compiled_engine) {
+  compiled_engine->active_input_tensors.clear();
+  compiled_engine->active_shape_tensor_values.clear();
+}
+
+void reset_active_input_tensors(c10::intrusive_ptr<TRTEngine> compiled_engine) {
```
narendasan: Do you want to just make this a method of the engine object. I dont think it needs to get lifted into python or anything

core/runtime/execute_engine.cpp
```diff
@@ -92,13 +93,35 @@ bool _validate_shapes(std::vector<at::Tensor> inputs, c10::intrusive_ptr<TRTEngi
   return false;
 }
 
+void clear_active_input_tensors(c10::intrusive_ptr<TRTEngine> compiled_engine) {
```
narendasan: same here

narendasan (commented):
I think this makes sense to me to use the engine object to manage lifetimes. We might want to make the active buffer management functions methods of the engine object. 

Also it would be good to make sure the rename works correctly with IOutputAllocator / pre_allocated_outputs which I think reuse some of the cudagraphs system

core/runtime/execute_engine.cpp
SandSnip3r: Done

core/runtime/TRTEngine.cpp
SandSnip3r: No

SandSnip3r:
> Also it would be good to make sure the rename works correctly with IOutputAllocator / pre_allocated_outputs which I think reuse some of the cudagraphs system

I think that path uses `compiled_engine->output_allocator->getBuffers()` which is not going to use these cuda graph staging buffers.

Issue #4240, opened 2026-05-07
✨[Feature] Support in-place custom plugins (and multiple outputs)


peterbjorgensen:
**Is your feature request related to a problem? Please describe.**
I am trying to use this package to put models into production and I am not sure if I am using it wrongly.
The usual route is pytorch->onnx->tensorrt, but now that I have custom triton kernels it becomes a pain.

The first problem is that the `generate_plugin_converter` only supports single outputs, but I have managed to hack this by adding a `number_of_outputs` argument to `generate_plugin_converter` and do `return tuple(layer.get_output(i) for i in range(number_of_outputs))` inside the `custom_kernel_converter` function. Having multiple outputs seems like a basic feature.

The second problem is that I am trying to make in-place custom plugins work. I don't think it is supported at all, because I had to dig into the library and set `builder_config.set_preview_feature(trt.PreviewFeature.ALIASED_PLUGIN_IO_10_03, True)`.
The following code does run after setting the flag, but it does not use aliased in/out:
```
from pathlib import Path
import triton
import triton.language as tl
import tensorrt.plugin as trtp
from typing import override
import torch
from torch import Tensor
from torch import nn
import torch_tensorrt
import tensorrt as trt

@triton.jit
def add_one_kernel(x_ptr, n_elements, y_ptr, BLOCK_SIZE: tl.constexpr):
    pid = tl.program_id(0)
    block_start = pid * BLOCK_SIZE
    offsets = block_start + tl.arange(0, BLOCK_SIZE)
    mask = offsets < n_elements
    x = tl.load(x_ptr + offsets, mask=mask)
    output = x + 1
    tl.store(y_ptr + offsets, output, mask=mask)


@torch.library.custom_op("my::add_one", mutates_args=())  # type: ignore[misc]
def add_one(X: torch.Tensor) -> torch.Tensor:
    assert X.is_cuda
    BLOCK_SIZE = 256
    grid = lambda meta: (triton.cdiv(X.numel(), meta["BLOCK_SIZE"]),)
    Y = torch.empty_like(X)
    add_one_kernel[grid](X, X.numel(), Y, BLOCK_SIZE=BLOCK_SIZE)
    return Y


@torch.library.register_fake("my::add_one")
def _(X: torch.Tensor) -> torch.Tensor:
    return X

@trtp.register("my::add_one")
def add_plugin_desc(X: trtp.TensorDesc) -> tuple[trtp.TensorDesc]:
[...]

PR #4251, opened 2026-05-12
feat: Allow for users / kv cache to add aliased I/O for inplace operations

narendasan:
Description

Adds support for in-place ATen operators by extending the Torch-TensorRT compile pipeline and C++ runtime with aliased input/output bindings. The motivating case is streaming inference with a key/value cache (e.g. autoregressive decoders, ZoomASR): each step writes a single timestep into the cache, and without aliasing every step pays a full cache-size copy at the engine boundary. With aliased I/O the TensorRT engine writes directly into the user's (or module-held) cache storage; no fresh allocation, no post-engine copy.

Two `aliased_io` "kinds" are tracked so the runtime can reason about provenance:

* `kv_cache_update` — TensorRT-enforced via `IKVCacheUpdateLayer`; reported through `ICudaEngine::getAliasedInputTensor`.
* `user` — Torch-TensorRT-declared; reserved for future expansion if TRT exposes a public non-KV aliasing API.

## What this PR does

**Pipeline (Python)**

* New `slice_scatter` and `index_copy` converters that recognize KV-cache-update patterns (4-D static cache, `dim=2`, batch=1) and emit `IKVCacheUpdateLayer` with the output aliased to the cache input. Non-eligible cases fall back to scatter in TRT — no graph break.
* For `index_copy`, two disjoint converters (validator-gated KV fast path at `ConverterPriority.HIGH` + scatter fallback at standard priority) cleanly split the cases.
* `aliased_io` plumbed through `TRTInterpreter` → `TRTInterpreterResult` → `SerializedInterpreterResult` → `TorchTensorRTModule`. The `output()` step automatically promotes layer outputs that need to be network outputs (KVCacheUpdate requires it) and appends them after user outputs. The user/side-effect boundary is *derived* at runtime, not stored.

**Buffer-style support**

* `lift_mutated_buffers` lowering pass detects `BUFFER_MUTATION` patterns (the trailing `aten.copy_(get_attr_buffer, ...)` that `ep.module()` emits) and lifts each mutated buffer from `get_attr` to `placeholder` so the engine sees it as an input binding.
* `inline_lifted_buffers_into_gm` post-compile transform registers the buffers as `nn.Module` state on the compiled `GraphModule` and rewrites the lifted placeholders to `get_attr` reads. The result is a plain `fx.GraphModule` (no custom wrapper class) that serializes cleanly through `torch_tensorrt.save` / `torch.export`.
* Low-level entry point `convert_exported_program_to_serialized_trt_engine` gains `lift_mutable_buffers: bool = False` for power users who want to manage the resulting bindings themselves.

**C++ runtime (ABI v9 → v10)**

* Bumped `ABI_VERSION` to `"10"`; added `ALIASED_IO_IDX` to `SerializedInfoIndex`.
* `serialize_aliased_io` / `deserialize_aliased_io` helpers (wire format: `output@input@kind` records joined by `BINDING_DELIM`). Helpers live in `runtime_utils.cpp` alongside `serialize_bindings`.
* `TRTEngine` constructor reconciles the build-time map against `ICudaEngine::getAliasedInputTensor` — the engine API is the source of truth for KV-style aliasing.
* `execute_engine` records bound input tensors by binding name; for each output binding in `aliased_io`, binds the same `data_ptr` as the source input and skips fresh allocation. Pre-allocated outputs are disabled when aliased I/O is present.
* CUDA Graphs integration: aliased *input* bindings bypass the persistent-clone path so the engine writes through to user storage; aliased outputs are skipped in the post-exec copy-back loop. Capture + replay both correctly mutate the user's tensor.

**Docs + examples**

* `docsrc/contributors/inplace_operations.rst` — full design doc covering motivation, primitives, pipeline, runtime, serialization format, and known limitations.
* Three examples under `examples/dynamo/`:
  * `aliased_io_user_inputs.py` — caller-owned cache (simplest case)
  * `aliased_io_buffers.py` — module-owned cache via `register_buffer`
  * `aliased_io_kv_attention.py` — realistic single-layer transformer attention block with static KV cache

Fixes partially #4240 (in-place custom plugins / multiple outputs — addresses the in-place-operator side; plugin-side aliased I/O is explicitly out of scope here).

## Type of change

- New feature (non-breaking for callers who don't opt in to aliased I/O; **ABI-breaking for existing engine binaries** — older serialized engines fail the version check and need to be rebuilt, consistent with prior ABI bumps).
- This change requires a documentation update (included).

# Checklist

- [ ] I have added the relevant labels to my PR

## Test summary

38 new tests across 8 files, all passing:

| File | Cases | Covers |
|---|---|---|
| `tests/py/dynamo/conversion/test_slice_scatter_aten.py` | 8 | scatter-fallback path (numerical correctness via the standard converter harness) |
| `tests/py/dynamo/runtime/test_aliased_io.py` | 8 | end-to-end aliased I/O (user-input single/paired/streaming + buffer-style + regressions) |
| `tests/py/dynamo/runtime/test_index_copy_kv.py` | 4 | KV fast path + 3 fallback shapes for `aten.index_copy` |
| `tests/py/dynamo/runtime/test_lift_mutable_buffers_api.py` | 4 | low-level `lift_mutable_buffers=True` flag round-trip (introspect engine, construct module, run) |
| `tests/py/dynamo/runtime/test_aliased_io_serialization.py` | 3 | `torch_tensorrt.save` / `load` round-trip for user-input + buffer-backed + streaming buffer |
| `tests/py/dynamo/runtime/test_aliased_io_cudagraphs.py` | 3 | CUDA Graph capture + replay for both user-input and buffer-backed KV; cudagraphs vs non-cudagraphs parity |
| `tests/py/dynamo/runtime/test_hf_static_cache_xfail.py` | 1 | xfail documenting the current HF + StaticCache gap (asserts known failure mode) |
| `tests/py/dynamo/lowering/test_buffer_lifting.py` | 9 | `lift_mutated_buffers` + `inline_lifted_buffers_into_gm` unit tests |

## Known gaps (documented)

* Stock HuggingFace decoder LMs with `StaticCache` don't compile end-to-end yet: torch.export's `run_decompositions` raises internally on the EP that `convert_and_export_with_cache` produces. The xfail test asserts the known failure so a future upstream fix surfaces as a test failure. Path forward documented in the design doc.
* `IKVCacheUpdateLayer` requires static `s_max`. Dynamic-sequence-length cache shapes fall through to the scatter path (still correct, no aliasing).


cehongwang:
Anything we could do to avoid manually resetting the KV cache before every run?

core/runtime/execute_engine.cpp
```diff
@@ -139,20 +140,35 @@ void setup_input_tensors(
...
 
-      if (need_cudagraphs_record) {
+      // An aliased input is one whose data_ptr will also be the address of an
+      // aliased output binding. Cudagraphs normally clone inputs into a
+      // persistent buffer so addresses are stable across replays; for an
+      // aliased input we deliberately bind to the user's tensor instead, so
+      // the engine writes through to the user's storage. The user is already
+      // required to pass stable input addresses under cudagraphs, so the
+      // aliasing contract is compatible.
+      bool is_aliased_input = false;
```
cehongwang: I think we should have this info in the compiled time and not recompute this in the execution phase every iteration

core/runtime/TRTEngine.h
cehongwang, line 118: Should add a condition of no aliased I/O

core/runtime/execute_engine.cpp
```diff
@@ -391,11 +463,19 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
...
     // When the pre-allocated output mode is turned on, for intermediate modules, we only create the output in the first
-    // execution or when shape is changed.
-    if (compiled_engine->use_pre_allocated_outputs &&
+    // execution or when shape is changed. If the engine has aliased outputs we
+    // disable pre-allocation entirely: aliased outputs share storage with
+    // user-supplied inputs that may change between calls, so caching the
+    // tensor reference would lead to writes against stale storage.
+    if (compiled_engine->use_pre_allocated_outputs && !compiled_engine->aliased_io.empty()) {
+      LOG_DEBUG(
+          "Skipping pre_allocated_outputs cache because engine has aliased I/O; "
```
cehongwang: Do you think we should just skip aliased I/O and only pre-allocate other I/O?

cehongwang: Looks like it is already implemented, right? We can call create_output tensors here

core/runtime/TRTEngine.cpp
```diff
@@ -281,6 +285,40 @@ TRTEngine::TRTEngine(
...
+  for (const auto& out_name : this->out_binding_names) {
+    // TRT returns nullptr / empty string for non-aliased outputs; any thrown
+    // exception is a real error in the engine state and propagates.
+    const char* aliased_in = cuda_engine->getAliasedInputTensor(out_name.c_str());
+    if (aliased_in == nullptr || aliased_in[0] == '\0') {
+      continue;
+    }
+    auto it = this->aliased_io.find(out_name);
+    if (it == this->aliased_io.end()) {
+      this->aliased_io[out_name] = AliasedIOSpec{std::string(aliased_in), AliasKind::kKVCacheUpdate};
```
cehongwang, line 304: What about kUser?

docsrc/contributors/inplace_operations.rst
```diff
@@ -0,0 +1,505 @@
...
+Background: TensorRT Primitives
+--------------------------------
+
+KVCacheUpdate Operator
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``IKVCacheUpdateLayer`` performs ``output[i, :, writeIndices[i] + s, :] =
+update[i, :, s, :]`` and aliases ``output`` to ``cache``. Inputs:
+
+* ``cache`` — shape ``[b, d, s_max, h]``, network input, static ``s_max``.
```
cehongwang: It's better to make the symbol clearer: [b, num_heads, s_max, dim]

docsrc/contributors/inplace_operations.rst
```diff
@@ -0,0 +1,505 @@
...
+
+       std::unordered_map<std::string, std::string> aliased_io;  // out → in
+
+   Populated from two sources:
+
+   1. The serialized engine metadata (see :ref:`serialization_format`).
+   2. **Source-of-truth reconciliation at deserialize time:** for every output
+      binding, query ``cuda_engine->getAliasedInputTensor(out_name)`` and merge
+      the result into ``aliased_io``. The TRT API is authoritative; the
+      serialized map is a cache that allows the runtime to avoid a per-call
```
cehongwang: Why per call? We only need to call it once and store it right? Should we consider dropping the metadata for simplicity?

docsrc/contributors/inplace_operations.rst
```diff
@@ -0,0 +1,505 @@
...
+``aten.copy_(get_attr_buffer, slice_scatter_result)`` node. Left alone,
+Torch-TensorRT's constant-folding pass would bake the buffer into the
+engine as a weight; the slice-scatter would scatter against that constant
+and produce a result the engine discards.
+
+To preserve buffer state across calls and let the KV-cache aliasing fast
+path fire, a pre-compile pass ``lift_mutated_buffers`` runs after
+``ep.module()`` and before ``post_lowering``:
+
+1. Scans for ``aten.copy_.default(get_attr, _)`` patterns — the marker for
```
cehongwang, line 480: Does this have a side effect on .copy() operators themselves? i.e. not KV cached ones

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -892,6 +948,9 @@ def output(self, target: str, args: Any, kwargs: Any) -> List[Any]:
             else:
                 output_dtype = dtype.unknown
 
+            # Capture identity before any cast — we use it to find aliasing.
+            original_id = id(output)
+
             if output_dtype is not dtype.unknown:
                 output = self._cast_output_dtype(
```
cehongwang, line 818: Should we skip casting if it is aliased?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1062,6 +1063,133 @@ def aten_ops_gather(
...
+) -> bool:
+    """Validator for the KV-cache fast path of ``aten.index_copy.default``.
+
+    Returns True only for the narrow case our ``IKVCacheUpdateLayer``
+    emitter can handle without a graph break:
+
+    * Input is an FX placeholder (i.e. a network input — required for
+      aliasing).
+    * Input has rank 4 and fully static shape ``[b, d, s_max, h]``.
+    * ``dim`` argument is exactly ``2``.
```
cehongwang, line 1077: why does it have to be 2?

core/runtime/execute_engine.cpp
narendasan: Its per input, maybe we can include it in the binding data structure you added

core/runtime/runtime.h
```diff
@@ -64,6 +68,12 @@ std::string base64_encode(const std::string& in);
 std::string base64_decode(const std::string& in);
 std::string serialize_bindings(const std::vector<std::string>& bindings);
 
+// Encode/decode the aliased_io map. Records are separated by BINDING_DELIM
+// ('%') and each record is "output_name@input_name@kind" (the '@' avoids
+// collision with TRT binding names which are alphanumeric + underscore).
+std::string serialize_aliased_io(const std::unordered_map<std::string, AliasedIOSpec>& aliased_io);
```
narendasan, line 74: We should fold the binding structure and this together @cehongwang

core/runtime/TRTEngine.cpp
narendasan: I think this code is only to detect IKVCacheUpdate bindings basically the internal under the hood loop back case. If its in the map then its a user binding which needs to be returned

core/runtime/TRTEngine.h
narendasan: I think its just that the aliased binding cannot be pre allocated, not that we cant have any pre allocated bindings

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
narendasan: Should be the input type right?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
narendasan: Entries in the cache, that get updated by index_put should be rank 2 right?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
cehongwang: Is there a case it is not input type if it is aliased?

docsrc/contributors/inplace_operations.rst
narendasan: sure

docsrc/contributors/inplace_operations.rst
narendasan: Yeah lets fold it into your binding metadata structure

docsrc/contributors/inplace_operations.rst
narendasan: `copy_` should only be for inplace ops. Without functionalizing these, `copy_` and `copy` are distinct operations

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
narendasan: No harm in checking i guess

core/runtime/TRTEngine.cpp
```diff
@@ -308,6 +303,17 @@ TRTEngine::TRTEngine(
     if (it == this->aliased_io.end()) {
       this->aliased_io[out_name] = AliasedIOSpec{std::string(aliased_in), AliasKind::kKVCacheUpdate};
       LOG_DEBUG("aliased_io reconciliation: discovered " << out_name << " -> " << aliased_in << " (kv_cache_update)");
+    } else if (it->second.kind == AliasKind::kUser) {
+      // A kUser alias is declared by Torch-TensorRT, not TRT. If the engine
+      // also reports an alias for this output the two views should agree on the
+      // source input; keep the kUser kind either way so the provenance recorded
+      // at build time is not silently rewritten to kv_cache_update.
+      if (it->second.input_binding_name != std::string(aliased_in)) {
```
narendasan: How would the user and engine declare differently?

PR #4256, opened 2026-05-12
Move all Python test dependencies to pyproject

SandSnip3r:
Install Python test dependencies from pyproject dependency groups in CI to avoid duplicated requirements.

# Description

### Summary

This PR ports Python test dependency installation in CI from `tests/py/requirements.txt` to `pyproject.toml` dependency groups, so local development and CI use the same dependency source of truth.

### Problem

The test dependencies were defined in more than one place and had drifted. In particular, `tests/py/requirements.txt` pinned `transformers==4.53.1`, while `pyproject.toml` specified `transformers>=5.0.0`.

That was risky because GitHub CI was installing from `tests/py/requirements.txt`, while local development with `uv` uses the dependency groups in `pyproject.toml`. As a result, CI and local dev could resolve meaningfully different test environments.

### What Changed

- Moved the shared Python test dependencies into the `test`, `test-ext`, and `quantization` dependency groups in `pyproject.toml`.
- Updated `.github/scripts/install-torch-tensorrt.sh` to install test dependencies with:

  ```bash
  pip install --group test --group test-ext --group quantization

- Updated noxfile.py to install the same dependency groups instead of tests/py/requirements.txt.
- Reduced tests/py/requirements.txt to a compatibility stub for tooling that still expects the file to exist.
- Updated the release checklist to point maintainers at pyproject.toml dependency groups instead of the old test requirements file.

### Design Decisions

The main design choice was to make pyproject.toml the source of truth for shared test dependencies. This matches the local uv workflow and avoids maintaining duplicate package constraints in both dependency groups and a requirements file.

CI now uses native pip dependency group support instead of a custom parser or generated requirements file. This keeps the CI path simple and close to the intended interface for PEP 735 dependency groups.

install-torch-tensorrt.sh upgrades to pip>=25.1 before installing groups because pip --group support requires a recent pip. The script also installs tomli only for Python versions before 3.11 because it parses pyproject.toml to recover the torchvision requirement used later in the script.

The old tests/py/requirements.txt file is intentionally left in place as an empty compatibility stub because there is still non-CI tooling that references it by path. It no longer contains dependency constraints, so it cannot drift from pyproject.toml.

## Type of change

- CI Refactor

# Checklist:

### DIFF
--- a/.github/scripts/install-torch-tensorrt.sh
+++ b/.github/scripts/install-torch-tensorrt.sh
@@ -12,8 +12,8 @@ if [[ $(uname -m) == "aarch64" ]]; then
 fi
 
 # Install all the dependencies required for Torch-TensorRT
-pip install --upgrade "pip>=25.1" "tomli>=1.1.0; python_version < '3.11'"
-pip install \
+python -m pip install --upgrade "pip>=25.1" "tomli>=1.1.0; python_version < '3.11'"
+python -m pip install \
     --pre \
     --extra-index-url https://pypi.nvidia.com \
     --extra-index-url https://download.pytorch.org/whl/nightly/cu130 \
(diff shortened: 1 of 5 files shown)


lanluo-nvidia (approved):
LGTM
noxfile.py is deprecated anyway, we no longer use nox for pytest.

PR #4274, opened 2026-05-16
[ExecuTorch] Remove libtorch dependency from .pte runtime backend

shoumikhin:
## ExecuTorch backend currently needs libtorch

But ExecuTorch's `.pte` runtime is meant to be a standalone runtime with no libtorch dependency. That's the primary reason a user picks the ExecuTorch path over the JIT/AOTI runtime path in order to ship models that run on edge / embedded / mobile devices where bundling libtorch (~200+ MB on Linux) is impractical.

The current `core/runtime/executorch/TensorRTBackend.cpp` (introduced in #4238) reuses `core::runtime::TRTEngine` as its `DelegateHandle`. That was a reasonable first-implementation choice, because `TRTEngine` already knows how to deserialize a TRT engine blob and bind it to TRT runtime objects, so reusing it minimized duplication.

The unintended consequence: `TRTEngine` is shared with the JIT/AOTI runtime. It extends `torch::CustomClassHolder` and holds `at::cuda::CUDAGraph`, `at::cuda::CUDAStream`, `std::vector<at::Tensor>` members, where none of which the ExecuTorch backend touches, but all of which force the entire libtorch family into the link line of any binary including the backend. The backend also calls `c10::cuda::getCurrentCUDAStream(...)` directly.

A `.pte` user therefore cannot link the backend without also depending on libtorch.

## The problem this PR solves

Specialize the ExecuTorch backend so it stops dragging libtorch into binaries that have no other reason to need it. The JIT/AOTI path keeps using `TRTEngine` exactly as before, and that path legitimately depends on libtorch.

## What the PR does

Adds a small libtorch-free `ETRTEngine` struct in `core/runtime/executorch/` that holds only the state the backend actually uses:

```cpp
struct ETRTEngine {
  std::shared_ptr<nvinfer1::IRuntime>          rt;
  std::shared_ptr<nvinfer1::ICudaEngine>       cuda_engine;
  std::shared_ptr<nvinfer1::IExecutionContext> exec_ctx;
  std::pair<uint64_t, uint64_t> num_io;
  std::string name;
  int device_id;
  std::vector<std::string> in_binding_names, out_binding_names;
  cudaStream_t stream;   // backend-owned, non-blocking
  std::mutex mu;
};
```

No `at::Tensor`, `c10::CUDAStream`, `torch::CustomClassHolder`, or libtorch headers.

`TensorRTBackend.cpp` is updated to:
- Use `ETRTEngine` instead of `core::runtime::TRTEngine` for the delegate handle.
- Replace `c10::cuda::getCurrentCUDAStream(...)` with the engine's own stream (created in `ETRTEngine`'s ctor via `cudaStreamCreateWithFlags(..., cudaStreamNonBlocking)`, destroyed in the dtor).
- Replace `et_out.unsafeGetTensorImpl()->set_sizes_contiguous(...)` (ATen-style internal API) with the public `executorch::runtime::resize_tensor(...)`.
- Drop the now-unneeded includes: `core/runtime/Platform.h`, `core/runtime/RTDevice.h`, `core/runtime/TRTEngine.h`, `core/util/prelude.h`.

`libtorchtrt_executorch/CMakeLists.txt` is updated so the OSS CMake build is libtorch-free as well:
- Removes `find_package(Torch REQUIRED)` and the surrounding Python-based Torch-prefix discovery block.
- Removes the 10 JIT/AOTI sources (`TRTEngine.cpp`, `runtime.cpp`, `runtime_utils.cpp`, `DeviceList.cpp`, `Platform.cpp`, `RTDevice.cpp`, `TRTEngineProfiler.cpp`, `core/util/Exception.cpp`, `core/util/trt_util.cpp`, `core/util/logging/TorchTRTLogger.cpp`) from `_torchtrt_executorch_sources` — the ExecuTorch backend now builds only `TensorRTBackend.cpp` + `ETRTEngine.cpp`.
- Removes `torch` from `_torchtrt_executorch_link_libraries`.

## Test plan

- `tests/py/dynamo/executorch/test_*.py` — should pass unchanged (wire format + Python API are untouched).
- Build the backend and verify `ldd` / `nm` show no `libtorch*` / `libc10*` / `libATen*` / `libcaffe*` in `DT_NEEDED` or undefined symbols.
- Load an existing `.pte` and call `module.forward(...)` produces identical numerics.


lanluo-nvidia:
Similar implantation is already in latest main

shoumikhin:
Thanks @lanluo-nvidia for the pointer and @zewenli98 for the queue time.

I checked and you're right, main already has a libtorch-free ExecuTorch backend at cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp. Same shape: only NvInfer, cudart, and executorch headers, no c10/at/torch refs, and find_package(Torch) is gone. So per the plan to leave 2.12 with its existing libtorch dep, closing this PR.

One small thing I noticed while comparing: main creates the CUDA stream with plain cudaStreamCreate, which implicitly syncs with the legacy NULL stream on every op. Switching to cudaStreamCreateWithFlags with cudaStreamNonBlocking avoids that. Happy to send a one-liner against main if that's useful.

PR #4278, opened 2026-05-19
[2.12] ExecuTorch: wrap the engine blob in a self-describing TR01 envelope

shoumikhin:
## Problem

The current ExecuTorch wire format
(`py/torch_tensorrt/executorch/serialization.py::serialize_engine_info`)
is a bare length-prefixed `vector<string>`:

```
[uint32_t count (LE)]
for each entry:
  [uint32_t len (LE)] [uint8_t data[len]]
```

That format has three rough edges:

1. **No way to tell a real blob from garbage.** No magic bytes; a
   corrupted or wrong-type buffer falls into the `vector<string>`
   parser and silently produces a malformed list. Users see a
   surprising error somewhere downstream instead of a clean
   "this isn't a TRT engine blob" up front.

2. **No schema version.** The format is implicitly tied to the
   `SERIALIZATION_LEN` constant in the runtime. Any future addition to
   `SerializedInfoIndex` immediately invalidates every old `.pte` with
   no clear failure path.

3. **Engine payload lands at an unaligned offset.** The engine bytes
   sit at `4 + Σ (4 + len_i)`, which depends on the size of every
   prior string. This makes mmap-style or DMA loading of the
   engine tricky, since `nvinfer1::IRuntime::deserializeCudaEngine` and
   downstream SIMD/cudart paths benefit from a known alignment.

## Fix

Wrap the existing `vector<string>` payload in a small self-describing
envelope (`TR01`) that adds magic bytes, a schema-version slot, and a
16-byte aligned engine section:

```
Offset  Size  Field
0       4     Magic bytes b"TR01"
4       4     Metadata offset (LE uint32)
8       4     Metadata size (LE uint32)
12      4     Engine offset (LE uint32, 16-byte aligned)
16      8     Engine size (LE uint64; supports >4 GB engines)
24      8     Reserved (1 byte schema version + 7 bytes padding)
32+     ...   Metadata: legacy vector<string> with the ENGINE_IDX
              slot blanked
[engine_offset]  Raw engine bytes
```

The metadata region is still the legacy `vector<string>` (with
`ENGINE_IDX` blanked); the engine payload lives **once**, in the
aligned engine section. The C++ deserializer reconstructs the original
`vector<string>` by re-injecting the engine bytes into the
`ENGINE_IDX` slot — `core::runtime::TRTEngine`'s constructor signature
is unchanged.

### Backwards compatibility

`deserialize_engine_info` in `TensorRTBackend.cpp` sniffs the magic
bytes. If the blob doesn't start with `TR01` it falls back to the
existing `vector<string>` parser, so any pre-existing `.pte`s keep
loading.

## What this does NOT change

- `core::runtime::TRTEngine`'s constructor signature.
- The 12-slot `SerializedInfoIndex` enum.
- Any of the per-slot metadata semantics (`ABI_TARGET_IDX`, `NAME_IDX`,
  ...). Only how the bytes are laid out in the `.pte`.

## Testing

New `tests/py/dynamo/executorch/test_serialization.py` exercises the
producer side:

* magic bytes + schema version present;
* engine offset is always 16-byte aligned across a range of metadata sizes;
* small (100 B) and large (10 MB) engines round-trip byte-for-byte;
* the engine payload is **not** duplicated into the metadata region;
* under- and over-long info lists serialize cleanly;
* empty-engine slot still produces a valid blob.

7 tests, all passing locally.

The C++ side is exercised end-to-end by the existing ExecuTorch
backend tests — since the deserializer falls back to the legacy
parser for non-TR01 blobs, those tests continue to pass against
their existing fixtures and additionally validate the new TR01
path against the same producer used in production.

### DIFF
--- a/core/runtime/executorch/TensorRTBackend.cpp
+++ b/core/runtime/executorch/TensorRTBackend.cpp
@@ -25,6 +25,7 @@
 #include "core/runtime/Platform.h"
 #include "core/runtime/RTDevice.h"
 #include "core/runtime/TRTEngine.h"
+#include "core/runtime/runtime.h"
 #include "core/util/prelude.h"
 
 namespace torch_tensorrt {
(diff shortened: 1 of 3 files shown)


narendasan:
@shoumikhin RE: No schema version. The format is implicitly tied to the
SERIALIZATION_LEN constant in the runtime. Any future addition to
SerializedInfoIndex immediately invalidates every old .pte with
no clear failure path.

There should be an index for ABI version which is what we use to determine support 

https://github.com/pytorch/TensorRT/blob/0adefb99bb397d664d67f139b99cb97973c73670/core/runtime/runtime.h#L32

Not opposed to a better packed format though, just want to limit redundancy

narendasan:
> Engine payload lands at an unaligned offset. The engine bytes
sit at 4 + Σ (4 + len_i), which depends on the size of every
prior string. This makes mmap-style or DMA loading of the
engine tricky, since nvinfer1::IRuntime::deserializeCudaEngine and
downstream SIMD/cudart paths benefit from a known alignment.

How much does this matter?

lanluo-nvidia:
Similar implantation is already in latest main

shoumikhin:
Thanks @narendasan for the schema-version and alignment questions, and @lanluo-nvidia for pointing to main.

Re-reading main's TensorRTBlobHeader.h and .cpp: the envelope there is basically the one this PR proposed. Same b"TR01" magic, same 32-byte header, same 16-byte engine alignment, and the reserved field's first byte is already used as a schema version (reserved[0] = 1). That answers Naren's point about using an existing version index. On the metadata side, main went with JSON which is probably the better long-term choice anyway: schema-validated, human-readable, and decoupled from the runtime-side string-vector ABI.

Nothing to port forward, so closing as superseded.

PR #4279, opened 2026-05-19
feat: support in-place plugins

### LINKED ISSUE #4240: ✨[Feature] Support in-place custom plugins (and multiple outputs)

bowang007:
# Description

End-to-end support for in-place QDP plugins in torch-tensorrt

Fixes #4240 

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/lowering/passes/unfunctionalize_qdp_inplace.py
narendasan, line 1: Can you explain this lowering pass? Auto functionalization is a bit more broader reaching that just for QDP nodes. Maybe it might be useful elsewhere?

py/torch_tensorrt/dynamo/conversion/plugins/_generate_plugin_converter.py
```diff
@@ -159,8 +166,19 @@ def custom_kernel_converter(
...
+
+        # JIT path: layer.plugin is the Python `_TemplateJITPlugin` whose
+        # `aliased_map` is populated by TRT during `add_plugin`.
+        # AOT path: layer.plugin is a C++ wrapper that does not expose the
+        # map, so fall back to the op schema's mutation declaration — the
+        # same signal `_generate_plugin` uses to emit `.aliased()`.
+        layer_plugin = getattr(layer, "plugin", None)
+        aliased_map = getattr(layer_plugin, "aliased_map", None)
+        if aliased_map and any(v != -1 for v in aliased_map.values()):
+            ctx.requires_aliased_plugin_io = True
```
narendasan: I think the right way to wire this is add an option to the converter decorator for `requires_aliased_plugin_io` and then let the converter registry manage this. Otherwise multiple generated converters will be trying to hit the conversion context.

narendasan: See the other options for examples

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1184,6 +1184,16 @@ def compile_module(
...
         for node in gm.graph.nodes:
             if node.op != "output" and (not node.meta) and "val" not in node.meta:
+                # An in-place write-back aten.copy_(dest, src) left by
+                # functionalization carries no metadata but mirrors its
+                # destination's shape/dtype; it is resolved during conversion.
+                if (
+                    node.op == "call_function"
+                    and node.target is torch.ops.aten.copy_.default
+                    and node.args
+                    and getattr(node.args[0], "meta", None)
```
narendasan: I think it would be better to figure out how to properly fill the meta data if possible

narendasan: You might need to run shape prop again?

py/torch_tensorrt/dynamo/utils.py
```diff
@@ -949,6 +949,17 @@ def get_output_metadata(
 def get_output_dtypes(output: Any, truncate_double: bool = False) -> List[dtype]:
     output_dtypes = []
     if isinstance(output, torch.fx.node.Node):
+        if (
+            "val" not in output.meta
+            and "tensor_meta" not in output.meta
+            and output.op == "call_function"
+            and output.target is torch.ops.aten.copy_.default
```
narendasan: Same thing here, instead of a bunch of specific cases, just figure out how to populate the meta data correctly

py/torch_tensorrt/dynamo/conversion/plugins/_auto_functionalized_converter.py
```diff
@@ -0,0 +1,207 @@
...
+        return _ALIAS_MAP_CACHE[op]
+
+    schema = op._schema
+    alias_map: Dict[int, int] = {}
+    mutated = mutated_tensor_indices(schema)
+    if mutated:
+        try:
+            from torch._subclasses.fake_tensor import FakeTensorMode
+
+            with FakeTensorMode():
```
narendasan: Do you know if FakeTensorUpdator would work here? https://github.com/pytorch/TensorRT/blob/210054596f91d37d71f61ad75e7e510f7120a4b4/py/torch_tensorrt/dynamo/lowering/passes/_FakeTensorUpdater.py#L14

bowang007: @narendasan  To be honest, FakeTensorUpdater is used to propagate and update the meta values but here we want to find the aliased map, we could use it but we still need to find the aliasing here.

cehongwang:
I have a PR that supports Alias I/O in standard Torch-TensorRT runtime: https://github.com/pytorch/TensorRT/pull/4251/changes and there is some overlap between mine and your PR. Please rebase off main branch again. 

After rebase, the logic in _TRTEngine.py might need to change

cehongwang:
## Suggested revision plan

Heads up before anything else: this branch is based on `2aac435dff` (Jul 10), which is **one week before #4251 ("Allow for users / kv cache to add aliased I/O for inplace operations") merged** as `b32ebfff5e` on Jul 17. `main` now has a general aliased-I/O mechanism that covers most of what this PR's runtime half is doing, so a rebase should let you delete a large chunk of the diff rather than fix it. Line references below are against `main` @ `2072b19`.

### 0. Rebase first

The branch is ~52 commits behind `main`. Please rebase before addressing anything else — `_TRTEngine.py` will conflict, since `main` now calls `_reconcile_aliased_io()` at essentially the insertion point this PR uses.

### 1. Drop the `_TRTEngine.py` changes entirely — `main` already implements this

`main` (from #4251) already resolves aliased outputs to their source input, in both runtimes, with the incompatible-mode guards:

| Concern | Already on `main` |
|---|---|
| output → input resolution | `create_output_tensors` returns the bound input tensor + shape assert, `_TRTEngine.py:1088-1106` |
| build → runtime transport | serialized `ALIASED_IO_IDX`, reconciled against `get_aliased_input_tensor` in `_TRTInterpreter.py:517-523` and `_TRTEngine._reconcile_aliased_io` |
| CUDA graphs | aliased inputs skip the staging clone (`_TRTEngine.py:1136,1158`); aliased engines excluded from capture |
| pre-allocated outputs | disabled with an explicit error (`_TRTEngine.py:1270-1273,1369-1374`) |
| dynamic output allocator | rejected in `_reconcile_aliased_io` |
| **C++ runtime** | implemented, `core/runtime/execute_engine.cpp:142-216` |
| engine meta kernel | `runtime/meta_ops/register_meta_ops.py` |

- [ ] Delete `self.aliased_output_idx_to_input_idx` and the `_execute_standard` rebinding. As written they only cover the plain Python path: `use_pre_allocated_outputs` bypasses the rebinding (it sits in the `else` of `can_use_pre_allocated_outputs`), `_execute_output_allocator` never rebinds, CUDA graphs bind the output to a *clone* of the input so the aliased pair gets two different addresses, and the C++ runtime (the default whenever `libtorchtrt_runtime.so` ships — note `use_python_runtime` is now deprecated and has no effect) never sees any of it.
- [ ] Record the alias at conversion time instead, the way `slice_scatter` does (`conversion/impl/slice_scatter.py:95-114`):

```python
from torch_tensorrt.dynamo.conversion._ConversionContext import AliasedOutput, AliasKind
from torch_tensorrt.dynamo.conversion.impl.slice_scatter import input_binding_name

binding = input_binding_name(ctx, mutated_itensor)   # None if not a direct network input
if binding is not None:
    ctx.aliased_outputs.append(
        AliasedOutput(
            output_tensor=layer.get_output(aliased_out_idx),
            input_binding_name=binding,
            kind=AliasKind.USER,
        )
    )
```

Two notes on that: `AliasKind.USER` means "declared by the compile flow, TRT does not enforce it", whereas plugin aliasing *is* TRT-enforced and *is* reported by `get_aliased_input_tensor` — so this may warrant a third `AliasKind`, or a tweak to reconciliation so an engine-reported `USER` alias isn't treated as a build/engine disagreement. And `input_binding_name` returning `None` is meaningful, see §3.

### 2. Keep these — they're the genuinely new contribution

- The `auto_functionalized` wrapper converter (`plugins/_auto_functionalized_converter.py`). Converting the HOP instead of un-functionalizing the graph is the right call.
- `.aliased()` on the output desc in `_generate_plugin.py` and the shared `_alias_utils.py`.
- The `requires_aliased_plugin_io` flag through `ConverterSupport` → `ctx` → `PreviewFeature.ALIASED_PLUGIN_IO_10_03`.
- The `_FakeTensorUpdater` fix — nice generic solution to the missing `meta['val']`.

### 3. Please distinguish the two aliasing cases explicitly

They have different requirements and only one needs runtime support:

- **Intra-engine**: the mutated tensor lives entirely inside the engine. Needs only `.aliased()` + the preview feature, no runtime work, no `aliased_io` entry.
- **Boundary**: the mutated tensor is a network input and the caller must observe the write. Needs the `ctx.aliased_outputs` entry from §1.

`test_automatic_plugin_inplace_consumed.py` is the first case; the other three tests are the second. Right now the difference is silent — please make it a validator/error rather than divergent behavior when `input_binding_name` returns `None` and the mutation therefore can't reach the caller.

### 4. Correctness issues that remain regardless of the rebase

- [ ] **`_reconstruct_op_args` ignores `auto_functionalized_v2`'s view kwargs.** Alongside `_<arg>_base_index`, v2 emits `_<arg>_size/_stride/_storage_offset` (as_strided), `_<arg>_slice_dim/_slice_start/_slice_end` (slice), or `_<arg>_alias` — see `torch/_higher_order_ops/auto_functionalize.py:188-232`. Exporting `op(x[:, :2])` yields `{'_X_base_index': 0, '_X_slice_dim': 1, '_X_slice_start': 0, '_X_slice_end': 2, ...}` and the current code hands the plugin the whole base, so it mutates the entire tensor instead of the slice with no error. This hits the KV-cache pattern (`cache[:, :, pos]`) immediately. Please raise a clear "mutated argument must be the base tensor" error (ideally in the capability validator, so the node cleanly stays in PyTorch). Also: `_<arg>_base_index` is `None` for an optional mutated arg passed `None`, which currently becomes `bases[None]` → `TypeError`.
- [ ] **Silent fallback to the pre-mutation base.** `_output_alias_map` catches every exception → `{}`, and `_convert_auto_functionalized` then routes the post-mutation slot to the *input* ITensor. That's the read-before-write hazard aliasing exists to prevent, and it's invisible (`logger.debug`). Note the fake run builds **CPU** fake tensors, so any fake kernel asserting `X.is_cuda` lands here. Please make this a hard error when the schema declares mutation.
- [ ] **`wrapper_wraps_plugin_op` is too broad.** `op in DYNAMO_CONVERTERS` → `__getitem_without_validation__` ignores the inner converter's `capability_validator` (including a user-supplied one) and the registry's dynamic-shape gate, while the wrapper declares `supports_dynamic_shapes=True` and hardcodes `requires_aliased_plugin_io=True`. Please resolve the inner `ConverterSupport` for the node and drive all three decisions (claim/reject, which impl to call, whether to set the ctx flag) from it. The returned `calling_convention` is currently discarded too.
- [ ] **`_is_aliased_writeback` claims write-backs it shouldn't.** It never checks that `node.args[0]` *is* the base for that getitem slot, nor that the index is a base slot rather than an op return. So `y.copy_(inplace_op(x))` matches and the copy is deleted while `y` is never written. Suggested guard:

```python
n_returns = len(op._schema.returns)
idx = src.args[1]
if idx < n_returns:
    return False
return node.args[0] is af.kwargs.get("_all_bases", [])[idx - n_returns]
```

- [ ] **`-> None` in-place ops are unsupported.** `generate_signature` derives arity from `len(schema.returns)` (`_generate_plugin.py:188`), so `def add_one_(x: Tensor) -> None` — the most idiomatic in-place form — produces a zero-output plugin. Note `can_auto_functionalize` handles `-> None` fine (`auto_functionalize.py:459-461`), so this is a plugin-generator limit. Support it or reject it with a clear message.
- [ ] **Drop v1 `auto_functionalized` from `_WRAPPER_TARGETS`.** v1 has no `_all_bases` and passes mutated args by name, so `base_outs` comes out empty and the downstream `getitem` indexes past the end. v1 is effectively dead upstream (`enable_auto_functionalized_v2` defaults to `True`), so removing it is simpler than supporting it.
- [ ] **Document the fake-kernel requirement.** Aliasing is only detected when the *fake* kernel returns the mutated input by identity, yet torch forbids the *real* kernel from doing so (`torch/_library/custom_ops.py:474-481`) — hence the tests' `return X.clone()` real kernels vs `return X` fakes. A user writing the schema-honest fake (`torch.empty_like`) silently gets no aliasing. Worth an explicit error naming the fix.
- [ ] **Multi-output limitation should be enforced, not just documented.** The docstring in `test_automatic_plugin_inplace_multi.py` explains that consuming a multi-output plugin's aliased output breaks under TRT's defensive copy; today a user who does that gets wrong numbers instead of an error.

PR #4222, opened 2026-04-29
Python runtime rework 

cehongwang:
# Unify Python and C++ TensorRT runtimes

PythonTorchTensorRTModule is removed. Both runtimes now live behind the same TorchTensorRTModule, with a single use_python_runtime flag (on CompilationSettings) selecting which path to take at engine setup.

What's changed
One module, two backends. TorchTensorRTModule.setup_engine() constructs either torch.classes.tensorrt.Engine (C++) or TRTEngine (Python) and binds the matching op to self.execute_engine_op. forward() just calls the bound op — no per-iteration branching, no separate module class.
Two equivalent ops. tensorrt::execute_engine (C++) and tensorrt::execute_engine_python (Python) are both registered and have the same signature, fake kernel, and semantics. The Python op is registered unconditionally so it's available even when the C++ runtime is loaded.

# Checklist:

- [ ] My code follows the style guidelines of this project (You can use the linters)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #4294, opened 2026-05-26
feat: reintroduce TRT-RTX runtime cache, dynamic shapes, and native CUDA graph support

tp5uiuc:
# Description

PR #4222 (Python runtime rework) replaced `PythonTorchTensorRTModule` with the new `TRTEngine` but the TRT-RTX feature surface from #4180 / #4184 / #4187 was lost in the move. This PR reintroduces it on `TRTEngine` + `_CudaGraphsTorchTensorRTModule`. Design mirrors #4187, retargeted onto the new split.

Two commits for ease of review.

## Commit 1 — `fix(runtime): restore TRT-RTX runtime-cache and dynamic-shapes setup`

Fixes the `AttributeError: 'TRTEngine' object has no attribute '_runtime_cache'` failure in `test_no_runtime_config_for_standard_trt` on standard-TRT CI and restores the runtime-cache + dynamic-shape-strategy machinery on RTX. Ports `_setup_runtime_config`, `_load_runtime_cache`, `_save_runtime_cache` from the deleted `PythonTorchTensorRTModule`. `__init__` / `__setstate__` unconditionally initialize `runtime_config` and `runtime_cache` to `None` so destructor save paths are safe. Tests in `test_000_runtime_cache.py` / `test_001_dynamic_shapes_kernel_strategy.py` are realigned on the public `runtime_config` / `runtime_cache` attribute names (the underscored variant was reading attributes that no longer existed after #4222).

## Commit 2 — `feat(runtime): add TRT-RTX native CUDA graph support`

Same intent as #4187:

- New `cuda_graph_strategy` setting (`"disabled"` / `"whole_graph_capture"`) threaded through the three `_compiler` entry points.
- `_setup_runtime_config` also sets `IRuntimeConfig.cuda_graph_strategy` from settings. `_setup_engine` latches `_rtx_native_cudagraphs`.
- New `_is_monolithic_capturable(stream)` — `False` on RTX when the context isn't stream-capturable or when dynamic-shape strategy is `"lazy"` (lazy specialization would invalidate a captured graph).
- New `_enable_rtx_native_cudagraphs()` — flips to `WHOLE_GRAPH_CAPTURE` and rebuilds the context.
- `_execute_standard` transparently overrides to RTX-native when SUBGRAPH cudagraphs are requested on RTX (with a warning pointing at the compile-time setting), and gates the manual `torch.cuda.CUDAGraph` path on `effective_cudagraphs = cudagraphs_enabled and not _rtx_native_cudagraphs`.
- `_CudaGraphsTorchTensorRTModule._check_monolithic_capturability` walks the compiled subgraph, asserts each TRT engine is capturable, and forces RTX-native engines back to `DISABLED` (rebuilding their contexts) so the outer monolithic `torch.cuda.CUDAGraph` capture isn't interfered with.

### Behavior matrix (unchanged from #4187)

| Graph type | cudagraph mode | RTX? | Behavior |
|---|---|---|---|
| TRT-only | SUBGRAPH | Yes | RTX-native always (override if needed) |
| TRT-only | SUBGRAPH | No | Manual capture (existing) |
| Mixed | WHOLE_GRAPH | Yes + capturable | Monolithic capture; RTX-native disabled per-engine |
| Mixed | WHOLE_GRAPH | Yes + NOT capturable | `RuntimeError` |
| Mixed | WHOLE_GRAPH | No | Monolithic capture (existing) |
| Any | No cudagraphs + strategy set | Yes | RTX-native runs transparently |

## Verification

A100 RTX build:
- `runtime/test_000_runtime_cache.py`: 12 passed, 2 skipped (non-RTX gated, including the originally-failing test)
- `runtime/test_001_dynamic_shapes_kernel_strategy.py`: 6 passed, 1 skipped
- `runtime/test_001_cuda_graph_strategy.py` (new): 17 passed, 1 skipped
- `models/test_cuda_graph_strategy_models.py` (new) — included with the same shape as #4187

Non-RTX path is correct by inspection (`runtime_config` / `runtime_cache` are initialized to `None` before the RTX gate). Worth a non-RTX CI confirm before landing.

## Type of change

- Bug fix (runtime-cache restoration unblocks the failing CI test)
- New feature (CUDA-graph strategy is net-new on the Python runtime)

## Checklist

- [x] Code follows the style guidelines (pre-commit clean: mypy, ruff, black, isort, typos)
- [x] Self-review done
- [x] Comments added where non-obvious
- [ ] Documentation updates
- [x] Tests added
- [x] New and existing unit tests pass locally
- [ ] Relevant labels added


py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -218,7 +237,20 @@ def __init__(
             torch_tensorrt.runtime.get_cudagraphs_mode()
         )
         self.resource_allocation_strategy = 0
-        self._runtime_config = None
+        # TensorRT-RTX runtime cache state. ``runtime_cache_path`` is filled in
+        # by ``_load_serialized_info`` once compilation settings are available;
+        # ``runtime_config`` and ``runtime_cache`` are populated by
+        # ``_setup_runtime_config`` on RTX builds. Initialized to ``None`` here
+        # so the destructor can safely save the cache even if ``_setup_engine``
+        # never runs.
```
tp5uiuc: Only keep the comment on Initialized to ``None`` here
        # so the destructor can safely save the cache even if ``_setup_engine``
        # never runs.

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -218,7 +237,20 @@ def __init__(
...
+        self.runtime_config: Any = None
+        self.runtime_cache: Any = None
+        # True once an IRuntimeConfig.cuda_graph_strategy other than
+        # ``"disabled"`` is in effect (either set at compile time or installed
+        # at runtime by ``_enable_rtx_native_cudagraphs``). When true,
+        # ``_execute_standard`` must skip manual torch.cuda.CUDAGraph capture
```
tp5uiuc: Only keep `When true,
        # ``_execute_standard`` must skip manual torch.cuda.CUDAGraph capture ...`

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -342,6 +384,9 @@ def _load_serialized_info(
 
         metadata = self.decode_metadata(self.serialized_metadata)
         self.settings = metadata.get("settings", CompilationSettings())
+        # Path used by ``_load_runtime_cache`` / ``_save_runtime_cache`` on
+        # TensorRT-RTX. Always set so non-RTX engines also expose it.
+        self.runtime_cache_path = self.settings.runtime_cache_path
```
tp5uiuc: Remove comment

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -371,6 +416,18 @@ def close(self) -> None:
...
 
     def _create_execution_context(self) -> trt.IExecutionContext:
+        # On TensorRT-RTX builds the allocation strategy lives on the
+        # ``IRuntimeConfig`` (set by ``_setup_runtime_config``), so once the
+        # runtime config is built we route context creation through it. The
+        # first call from ``_setup_engine`` precedes ``_setup_runtime_config``
+        # and falls through to the strategy-based path below.
+        if (
+            ENABLED_FEATURES.tensorrt_rtx
+            and getattr(self, "runtime_config", None) is not None
```
tp5uiuc: if ENABLED_FEATURES.tensorrt_rtx and self.runtime_config is not None

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -413,6 +470,18 @@ def _setup_engine(self) -> None:
...
 
+        # On TensorRT-RTX, build the IRuntimeConfig (with runtime cache,
+        # dynamic-shape kernel specialization strategy, and CUDA graph
+        # strategy) and rebuild the execution context so it picks them up.
+        # The NCCL barrier above runs against the initial strategy-based
+        # context.
+        if ENABLED_FEATURES.tensorrt_rtx:
+            self._setup_runtime_config()
+            self._rtx_native_cudagraphs = (
+                self.settings.cuda_graph_strategy != "disabled"
```
tp5uiuc: Move this to before context creation at line 392, and reuse that call to create the context.

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: Actually after we move the bottom context call, this should be `assert(runtime_config) is not None)` in the ENABLED_FEATURES.tensorrt_rtx path. So the overall flow looks like

```py
   assert runtime_config is not None
   context = self.cuda_engine.create_execution_context(self.runtime_config)
else:
        # Regular code-path
        strategy = trt.ExecutionContextAllocationStrategy.STATIC
        if self.resource_allocation_strategy:
            strategy = trt.ExecutionContextAllocationStrategy.USER_MANAGED
        context = self.cuda_engine.create_execution_context(strategy)

assert context is not None, "Failed to create execution context"
return context
```

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -227,6 +259,12 @@ def __init__(
         self._setup_engine()
 
     def __del__(self) -> None:
+        # Persist the TensorRT-RTX runtime cache before tearing the engine
+        # down; no-op when ``runtime_cache`` was never populated.
+        try:
+            self._save_runtime_cache()
+        except Exception:
+            pass
         self.reset_captured_graph()
```
tp5uiuc: _save_runtime_cache() handles exception internally, so don't recatch.

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -227,6 +259,12 @@ def __init__(
         self._setup_engine()
 
     def __del__(self) -> None:
+        # Persist the TensorRT-RTX runtime cache before tearing the engine
+        # down; no-op when ``runtime_cache`` was never populated.
+        try:
+            self._save_runtime_cache()
```
tp5uiuc: Also add _save_runtime_cache to close()

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: Actually we can remove the entire `self.runtime_cache_path` and inline it as `self.settings.runtime_cache_path` in call-sites

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -449,6 +518,113 @@ def _setup_engine(self) -> None:
...
+        specialized kernel would invalidate the captured graph.
+        """
+        if not ENABLED_FEATURES.tensorrt_rtx:
+            return True
+        if not self.context.is_stream_capturable(stream.cuda_stream):
+            return False
+        if self.settings.dynamic_shapes_kernel_specialization_strategy == "lazy":
+            return False
+        return True
+
```
tp5uiuc, line 600: Make it pythonic by using 
any(conditions) 
where conditions is a generator/tuple rather than multple ifs

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -449,6 +518,113 @@ def _setup_engine(self) -> None:
...
+        The runtime cache stores JIT compilation results so kernel/graph
+        compilation is not repeated across inference runs. The dynamic-shape
+        kernel specialization strategy controls how the JIT compiles
+        shape-specialized kernels (``lazy``, ``eager``, or ``none``).
+        """
+        self.runtime_config = self.cuda_engine.create_runtime_config()
+        alloc_strategy = trt.ExecutionContextAllocationStrategy.STATIC
+        if self.resource_allocation_strategy:
+            alloc_strategy = trt.ExecutionContextAllocationStrategy.USER_MANAGED
```
tp5uiuc: Make this a one-liner and pythonic

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -449,6 +518,113 @@ def _setup_engine(self) -> None:
...
+            if data:
+                self.runtime_cache.deserialize(data)
+                logger.info(f"Loaded runtime cache from {self.runtime_cache_path}")
+        except Exception as e:
+            logger.warning(f"Failed to load runtime cache: {e}")
+
+    def _save_runtime_cache(self) -> None:
+        """Save runtime cache to disk (with an exclusive file lock)."""
+        if getattr(self, "runtime_cache", None) is None:
+            return
```
tp5uiuc:
```suggestion
        if self.runtime_cache is None:
            return
```

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -449,6 +518,113 @@ def _setup_engine(self) -> None:
         if self.requires_output_allocator:
             self.create_output_allocator()
 
+    # --- TensorRT-RTX runtime cache / dynamic shapes strategy ---
```
tp5uiuc: # --- TensorRT-RTX ---

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -678,13 +854,40 @@ def _profile_section(self, label: str) -> ContextManager[None]:
     def _execute_standard(
         self, contiguous_inputs: List[torch.Tensor]
     ) -> torch.Tensor | Tuple[torch.Tensor, ...]:
+        # On RTX, manual ``torch.cuda.CUDAGraph`` capture is not safe
+        # (lazy kernel specialization can invalidate captured graphs and
+        # runtime allocation can prevent stream capture). If the user
+        # requested SUBGRAPH cudagraphs without explicitly setting
+        # ``cuda_graph_strategy="whole_graph_capture"``, transparently
+        # switch to RTX-native CUDA graphs and warn.
+        cudagraphs_enabled = torch_tensorrt.runtime.get_cudagraphs_mode()
```
tp5uiuc, line 891: Remove, this is documented in the warning.

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -678,13 +854,40 @@ def _profile_section(self, label: str) -> ContextManager[None]:
...
+                "stream). Switching to TRT-RTX native CUDA graphs. Set "
+                'cuda_graph_strategy="whole_graph_capture" at compile '
+                "time to avoid this warning."
+            )
+            self._enable_rtx_native_cudagraphs()
+
+        # ``effective_cudagraphs`` is the value the downstream record/replay
+        # paths should react to. When RTX native is active, TRT-RTX is
+        # already handling capture/replay internally, so the manual
+        # ``torch.cuda.CUDAGraph`` machinery must stay quiet.
```
tp5uiuc: machinery is skipped

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: [by Claude Code] Addressed in 81d799acb

tp5uiuc:
[by Claude Code] 
The failing L2 RTX checks on this PR are pre-existing main breakage, not regressions from this branch:

- **L2 dynamo core / `test_mutable_torchtrt_module.py::test_save`** — `RuntimeError: <pybind11_builtins.pybind11_detail_function_record_v1_…> is not pickleable.` Reproduced on the wheel built at `110d81694` (main HEAD just before this PR's two commits, no transplant) with the exact same error. The runtime rework moved engine state onto `TRTEngine` but `_MutableTorchTensorRTModule.save` still uses `torch.save(module, …)` which pickles the whole tree, and the new pickle plumbing isn't honored through that chain.
- **L2 dynamo compile** — `tensorrt.Engine` TorchBind class isn't registered in the test env (`test_cross_runtime_serde.*`) and a `setInputShape` failure in `test_weight_stripped_engine`. Both engine-internal, unrelated to the Python additions here.
- **L2 dynamo plugin** — missing `cuda-python` dep; `ttk_test` / `ttk_kp` op namespaces not loaded.
- **Windows + RTX Windows build** — `[WinError 206] The filename or extension is too long` during `pip install torch --pre` (Windows PATH_MAX vs torch nightly's deeply nested third_party tree).

py/torch_tensorrt/dynamo/runtime/_CudaGraphsTorchTensorRTModule.py
```diff
@@ -114,6 +114,59 @@ def __del__(self) -> None:
     def set_use_output_allocator(self, enable: bool) -> None:
         self.use_output_allocator_outputs = enable
 
+    def _check_monolithic_capturability(self, stream: torch.cuda.Stream) -> None:
```
narendasan, line 130: Whats the difference between monolithic capture and the "whole graph capture mode"?

tp5uiuc: Hi Naren, thanks for the Q. What I meant is the following:

- *Monolithic capture*: This path is similar to standard TRT. It relies on the outer torch.cuda.CUDAGraph recorded by _CudaGraphsTorchTensorRTModule.forward (after the user calls
  torchtrt.runtime.enable_cudagraphs()) capturing TRT-RTX engines. Similar to standard TRT, it wraps the entire compiled subgraph (every TRT engine and any intervening PyTorch glue ops) into one `monolithic` torch-side graph. Scope = whole pytorch graph. This code-path gets triggered for `CudaGraphsMode.WHOLE_GRAPH_CUDAGRAPHS` 

- *Whole graph capture*: This is a TRT-RTX exclusive path and obtained when setting `cuda_graph_strategy="whole_graph_capture"` per engine. This relies on TRT-RTX to manage captures : internally the JIT compiler captures/replays/invalidates the graph inside execute_async_v3, per-engine. The name "whole graph" is for the "whole" TRT-RTX forwarded graph. Scope = one TRT engine. This code-path gets triggered for `CudaGraphsMode.SUBGRAPH_CUDAGRAPHS` 

### When to use what
These two paths are mutually exclusive, even when running through `enable_cuda_graphs()`.

The guidance would be to prefer (2) whole graph capture, as this would get best perf (per an TRT-RTX engine). However the condition here is that there are no intervening pytorch ops (as this will break TRT-RTX's internal graph capture status), which is consistent with it getting triggered for `CudaGraphsMode.SUBGRAPH_CUDAGRAPHS`.

However, in case there are PyTorch ops in between (because of graph breaks, op incompleteness), the `enable_cuda_graphs()` will choose the `CudaGraphsMode.WHOLE_GRAPH_CUDAGRAPHS` mode. This means we will attempt to wrap N TRT-RTX engines + PyTorch ops in one outer torch.cuda.CUDAGraph and each engine has to leave its own (internal RTX-cudagraph) capture off, otherwise the RTX-native capture would interfere with the outer torch capture. That's what `_check_monolithic_capturability` enforces : it asserts every engine can be stream-captured (is_stream_capturable, no lazy-with-dynamic-shapes), and forces RTX-native off on each one (cuda_graph_strategy=DISABLED + context rebuild) before the outer torch.cuda.graph() block runs.

### Name sharpening
Now, naming-wise, "monolithic" was picked to avoid the literal collision with "whole_graph_capture" since the latter is a TRT-RTX-defined enum name but also confusing with Torch-TRT's `CudaGraphsMode.WHOLE_GRAPH_CUDAGRAPHS`. The difference lies in the distinction of "whole" for TRT-RTX vs torch-TRT. I am open to renaming either side if you have a preference. 

Perhaps a good middle ground is to rename the enum in the newly introduced `cuda_graph_strategy` to `disabled` and `per_engine_capture`. `per_engine_capture` can internally map to TRT-RTX's `whole_graph_capture` mode. This is easier for the users at least IMO. 


P.S. I will also add these best practices/guidance to the documentation. Also the behavior matrix (of how cudagraph mode interacts with TRT-RTX) is documented more in the PR description.

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -884,7 +1066,8 @@ def execute(
             return self._execute_output_allocator(contiguous_inputs)
 
         logger.debug(
-            f"Using the standard execution runtime mode with cudagraphs={torch_tensorrt.runtime.get_cudagraphs_mode()}."
+            f"Using the standard execution runtime mode with cudagraphs={torch_tensorrt.runtime.get_cudagraphs_mode()}"
```
cehongwang: Should it be `effective_cudagraphs`?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1302,6 +1309,7 @@ def convert_exported_program_to_serialized_trt_engine(
         timing_cache_path (str): Path to the timing cache if it exists (or) where it will be saved after compilation. Not used for TensorRT-RTX.
         runtime_cache_path (str): Path to the runtime cache for TensorRT-RTX JIT compilation results. Not used for standard TensorRT.
         dynamic_shapes_kernel_specialization_strategy (str): Strategy for dynamic shape kernel specialization at runtime (TensorRT-RTX only). Options: "lazy", "eager", "none". Default: "lazy".
+        cuda_graph_strategy (str): Strategy for CUDA graph capture/replay (TensorRT-RTX only). Options: "disabled" (manual capture), "whole_graph_capture" (TRT-RTX handles internally). Default: "disabled".
```
cehongwang, line 1312: @narendasan Should we add a guard to remind the user that this is RTX only? It is likely to be mixed up by users

tests/py/dynamo/runtime/test_001_cuda_graph_strategy.py
```diff
@@ -0,0 +1,357 @@
...
+        "use_python_runtime": True,
+        "min_block_size": 1,
+    }
+    kwargs.update(extra_kwargs)
+    compiled = torchtrt.compile(model, **kwargs)
+    torch._dynamo.reset()
+    return compiled
+
+
+def _find_python_trt_engine(compiled):
```
cehongwang, line 39: This needs to be changed when the cross serialization PR lands where two engines are registered in different runtimes

cehongwang (changes requested):
Overall looks good. This may need to be changed again after the cross serialization PR lands

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
tp5uiuc: Changed, thanks!

py/torch_tensorrt/dynamo/_compiler.py
tp5uiuc: Yes, this was a followup (I have a task filed) after the C++ changes landed, but unfortunately we never got to that point. I have the new changes ported over at https://github.com/tp5uiuc/TensorRT/pull/2 and I will make the PR after this one merges in. 

We are also making these strategies context managers (as is the recommended approach, see) https://github.com/pytorch/TensorRT/issues/4310. Given these changes, I will leave this to a future MR.

tests/py/dynamo/runtime/test_001_cuda_graph_strategy.py
tp5uiuc: ack, I can put in the change later.

tp5uiuc: Actually I looked at https://github.com/pytorch/TensorRT/pull/4286/changes and the logic still works as intended. What needs to change if the cross serialization PR lands? We still will be representing python engines as `torch_tensorrt.dynamo.runtime._TRTEngine`, right? Or do you mean to say that `use_python_runtime = True` will be removed altogether and so we need to plan for alterantes? If so, can we adress https://github.com/pytorch/TensorRT/pull/4286#discussion_r3339040492 first?

cehongwang (approved):

tests/py/dynamo/runtime/test_001_cuda_graph_strategy.py
cehongwang: use_python_runtime = True will be removed altogether. We can address it later.

PR #4300, opened 2026-05-28
fix(lowering): use factored sqrt form for complex acosh rewrite

tp5uiuc:
# Description

`torch.acosh` for complex inputs follows the C99 `cacosh` convention, `acosh(z) = log(z + sqrt(z+1) * sqrt(z-1))`. The lowering pass for `torch.ops.aten.acosh.default` was using the algebraically equivalent `log(z + sqrt(z^2 - 1))`, but those two forms pick **opposite Riemann sheets** whenever `Re(z) < 0` — the lowered output ends up as the additive inverse of the eager `torch.acosh(z)` for those inputs.

`tests/py/dynamo/lowering/test_complex_rewrite.py::test_acosh` uses `torch.randn(3, 4, dtype=torch.complex64) + 2.0` as input. Since `randn(complex64)` draws Re from `N(0, 0.5)`, the `+2.0` shift only keeps `Re(z) > 0` about 97% of the time — a single element drops below zero in roughly 2.6% of runs, flipping the sign of one of the twelve lowered outputs and tripping the test's `1e-4` numerical tolerance. This is the intermittent `Mismatched elements: 1 / 12 (8.3%) ... greatest absolute difference 1.46` failure that has been showing up on `release/2.12` CI for the L0 dynamo-core RTX cell (for example here : https://github.com/pytorch/TensorRT/actions/runs/26526799552/job/78135498854)

Rewrite `_rewrite_acosh` to compute the factored form `log(z + sqrt(z+1) * sqrt(z-1))` using the existing `_inline_complex_sqrt` and `_inline_complex_mul` helpers. Verified across 2000 random seeds drawn without any real-part shift (so `Re(z)` is freely negative) — 0 failures, max abs error ~1e-7. Full `tests/py/dynamo/lowering/test_complex_rewrite.py` (105 tests) passes.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
+++ b/py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
@@ -7,7 +7,6 @@ import torch
 from torch._subclasses.fake_tensor import FakeTensorMode
 from torch.fx import GraphModule, Node
 from torch.fx.experimental.proxy_tensor import unset_fake_temporarily
-
 from torch_tensorrt.dynamo._settings import CompilationSettings
 from torch_tensorrt.dynamo.lowering._SubgraphBuilder import SubgraphBuilder
 from torch_tensorrt.dynamo.lowering.passes.pass_utils import (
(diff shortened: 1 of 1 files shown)


py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
```diff
@@ -1142,19 +1141,20 @@ def _rewrite_asinh(self, node: Node) -> bool:
 
     @_complex_unpacker(torch.ops.aten.acosh.default)
     def _rewrite_acosh(self, node: Node) -> bool:
-        # acosh(z) = log(z + sqrt(z² - 1))
+        # acosh(z) = log(z + sqrt(z+1) * sqrt(z-1))
+        # The factored sqrt form matches torch.acosh's principal branch
```
tp5uiuc, line 1145: It is surprisingly hard to confirm if the formula is indeed what's used in torch torch's vectorized CPU source forwards to Sleef. See 
https://github.com/pytorch/pytorch/issues/183767#issuecomment-4460346267
and linked MRs ([particularly this one](https://github.com/pytorch/pytorch/pull/184359/changes) which is removing sleef and adding back scalar)

But multiple sources seem to indicate this is the more stable formula
https://github.com/kraj/musl/blob/kraj/master/src/math/acosh.c
https://mathworld.wolfram.com/InverseHyperbolicCosine.html
Both confirm that log(z + sqrt(z+1) * sqrt(z-1)) is more numerically friendly.

tp5uiuc:
[by Claude Code] For reference, before settling on the lowering-pass fix in this PR, I also confirmed a test-only alternative that just makes `test_acosh` deterministic and avoids the branch-cut region of the input distribution. Sharing here in case the lowering change is preferred to land separately.

## Alternative: test-only fix

```python
@pytest.mark.unit
def test_acosh():
    class M(nn.Module):
        def forward(self, z):
            return torch.acosh(z)

    # The complex_graph_detection lowering rewrites acosh as
    # log(z + sqrt(z**2 - 1)). torch.acosh uses a branch-equivalent form
    # whose Riemann sheet matches only when Re(z) > 0. Seed + shift keep
    # all 12 sample elements safely in the right half-plane.
    torch.manual_seed(0)
    z = torch.randn(3, 4, dtype=torch.complex64) + 3.0
    _check_op(M(), (z,), "acosh")
```

Two changes from current:
- `torch.manual_seed(0)` — deterministic across runs/torch versions.
- shift `+2.0` → `+3.0` — pushes `Re(z) > 0` essentially out of reach of `randn(complex64)`'s tails.

## Empirical sensitivity to the shift

Each row scans 2000 fresh seeds (`torch.manual_seed(s)`, `s ∈ [0, 2000)`), counts how many produce at least one element above the test's `1e-4` tolerance:

| Re-shift | Fail rate | Worst max abs diff |
|----------|-----------|--------------------|
| **+2.0 (current test)** | **52 / 2000 (2.6%)** | 4.77 |
| +2.5 | 3 / 2000 (0.15%) | 3.80 |
| +3.0 | 0 / 2000 (0.00%) | 0 |
| +3.5 | 0 / 2000 (0.00%) | 0 |
| +4.0 | 0 / 2000 (0.00%) | 0 |
| +5.0 | 0 / 2000 (0.00%) | 0 |

So `+3.0` is a 6σ buffer for `randn(complex64)` (Re ~ N(0, 0.5)), enough that even a torch RNG implementation change shouldn't reintroduce flakes.

## Why I went with the lowering fix instead

The lowering-pass change in this PR was the same number of touched lines and 0/2000 across all seeds **with no shift at all** (i.e. `Re(z)` freely in `(-2, 2)`), which is the right strength: the lowering is now correct for the entire complex plane rather than just the right half-plane that the test happens to sample. The test stays as-is.

narendasan (approved):
LGTM, @lanluo-nvidia please cherrypick for 2.12.1

PR #4306, opened 2026-05-29
chore(rtx): clean up TRT-RTX 1.4-era WARs and test skips

tp5uiuc:
# Description

TensorRT-RTX 1.5 resolves the upstream issues that several convolution-validator WARs and test skips were guarding against.

**Removed:**
- BF16 depthwise conv/deconv fallback to PyTorch.
- Grouped 3D deconv fallback to PyTorch (any dtype).
- Engine-cache timing-flakiness skip on `test_caching_small_model`.
- `test_grouped_deconv3d_fallback` (asserted behaviour no longer applies).

**Kept (narrower):**
- `convolution_capability_validator` now rejects only 3D transposed conv with `stride > 1` AND `dilation > 1` — still no TRT-RTX kernel for this combo.
- Matching in-test guard in `test_deconv3d` — the converter harness drives `TRTInterpreter` directly and bypasses the partitioner, so validator-rejected nodes raise `UnsupportedOperatorException` instead of falling back to PyTorch as they would in `torch_tensorrt.compile`.

**Replaced:**
- The timing-flakiness skip on `test_dynamo_compile_with_refittable_weight_stripped_engine` is replaced with a new skip naming the actual underlying issue: a static input-shape mismatch between the export `example_inputs` (batch 100) and the compile `arg_inputs` (batch 128). Tracked as a follow-up — the old skip was incidentally masking this.

**Docs:** Bumped a missed `TensorRT-RTX-1.4.0.76` Windows-install path example to `1.5.0.114`.

Verified locally on an A100 with TRT-RTX 1.5.0.114: BF16 mobilenet_v2/efficientnet_b0, grouped 3D deconv tests, and `test_caching_small_model` all pass; `test_deconv3d_10_combined_params` (strided+dilated) cleanly skips.

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2766,44 +2765,39 @@ def convolution_capability_validator(
 ) -> bool:
     """Reject unsupported convolution variants on TensorRT-RTX.
 
-    Falls back to PyTorch for:
-    1. Depthwise convolutions in BF16 (no kernel support on TRT-RTX).
-    2. Grouped 3D deconvolutions (crash on TRT-RTX).
+    Falls back to PyTorch for 3D transposed convolutions that combine
```
tp5uiuc: Actually both convolutions and deconvolutions with stride > 1 and dilation > 1 simultaneously are unsupported.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2766,44 +2765,39 @@ def convolution_capability_validator(
...
-    groups = args_bounds_check(node.args, 8)
-    is_grouped = groups is not None and groups > 1
-    is_transposed = bool(args_bounds_check(node.args, 6))
-    is_3d = input_meta.shape is not None and len(input_meta.shape) == 5
-    is_bf16 = input_meta.dtype == torch.bfloat16
+    input_meta = getattr(node.args[0], "meta", {}).get("tensor_meta")
+    if input_meta is None or input_meta.shape is None:
+        return True
+    if len(input_meta.shape) != 5:
+        return True
```
tp5uiuc: Even for 2D this may not work, I should verify.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2766,44 +2765,39 @@ def convolution_capability_validator(
...
+
+    stride = args_bounds_check(node.args, 3)
```
tp5uiuc: Try and simplify this function as much as possible and make it readable using pythonic expressions (walrus operator etc.)

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
tp5uiuc: [by Claude Code] Verified on TRT-RTX 1.5 with toy modules: regular 3D Conv3d with stride>1 + dilation>1 (including with groups>1) compiles and runs fine. Only deconv (ConvTranspose) hits the cuDNN `Strided & Dilated Deconv` limitation.

Keeping the `is_transposed` check. Happy to revisit if you have a counter-example.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
tp5uiuc: [by Claude Code] Verified — 1D and 2D deconv with stride>1 + dilation>1 both fail with the same `Strided & Dilated Deconv are currently not supported` error on TRT-RTX 1.5. Widened the validator to cover 1D/2D/3D ConvTranspose (dropped the rank check) and added matching `strided_dilated` parametrize entries + skip guards to `test_deconv1d` and `test_deconv2d`. Addressed in `a7e191deb`.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
tp5uiuc: [by Claude Code] Refactored to a single short-circuit `if` chain with walrus operators — body went from ~25 lines to ~12. Done in `a7e191deb`.

tests/py/dynamo/models/test_weight_stripped_engine.py
```diff
@@ -268,14 +268,12 @@ def test_engine_caching_saves_weight_stripped_engine(self):
...
-        "Engine caching compilation time assertion is unreliable with TensorRT-RTX",
-    )
     def test_dynamo_compile_with_refittable_weight_stripped_engine(self):
         pyt_model = models.resnet18(pretrained=True).eval().to("cuda")
-        example_inputs = (torch.randn((100, 3, 224, 224)).to("cuda"),)
-        exp_program = torch.export.export(pyt_model, args=example_inputs)
+        # Use the same inputs for both export and compile to avoid a
+        # static-shape mismatch between the exported program and the engine.
+        inputs = [torch.randn((100, 3, 224, 224)).to("cuda")]
+        exp_program = torch.export.export(pyt_model, args=tuple(inputs))
```
tp5uiuc, line 276: @zewenli98 : Without this change, both TRT standard and TRT-RTX fails this test. I am not too sure whether and with what cadence it is running on the CI currently (my understanding is that this is a L2 test as its not marked with pytest.mark.critical)

zewenli98: Good catch! I found the error was caught since the PR #4222, but not sure why it was not caught in the previous CI.

tp5uiuc:
[by Claude Code] CI failures look unrelated to this PR — two upstream/infra issues:

1. **Most build jobs** (Linux cu130/cu132, Windows cu130/cu132, RTX Linux + Windows) fail at step 9 (`test-infra/setup-binary-builds`) with `TypeError: dataclass() got an unexpected keyword argument 'slots'`. The `pkg-helpers` conda env is on Python 3.9 but its pip uses `@dataclass(slots=True)` (3.10+). pip's own module fails to import; the wheel build never starts.
2. **SBSA cu132**: `Unable to find a match: libnccl-2.27.7-1+cuda13.2` — the cu13.2 NCCL RPM isn't published for aarch64 yet. SBSA cu130 passes.

No code changes needed here. Re-running after the test-infra fix lands should turn it green.

lanluo-nvidia (approved):
LGTM

PR #4307, opened 2026-05-29
Fix Torch-TensorRT Bazel build on JetPack

Mike-Sagiri:
# Description

Update and fix TensorRT Bazel build support on NVIDIA JetPack platforms.
JetPack packages TensorRT headers and libraries differently from standard Linux TensorRT distributions.

This change:

- adds JetPack-specific include path support for third_party/tensorrt/local/BUILD
- adds JetPack-specific library path support for third_party/tensorrt/local/BUILD
- updates Bazel select() rules for JetPack environments for third_party/tensorrt/local/BUILD
- fixes incorrect TensorRT library resolution on JetPack systems for of third_party/tensorrt/local/BUILD
- updates dependencies for JetPack in setup.py and MODULE.bazel
- now the users can use `--jetpack` to build Torch-TensorRT from source successfully

Known issue:
- Since TensorRT 10.13+ not support cuda 12.x anymore, building Torch-TensorRT with TensorRT 10.13+ needs cuda 13 on JetPack. However, JetPack 6.2 and lower doesn't support cuda 13. When building Torch-TensorRT with TensorRT 10.7-, `add_attention` is not supported by TensorRT 10.14-. So the users may need to add `torch.backends.cuda.enable_flash_sdp(False)`, `torch.backends.cuda.enable_mem_efficient_sdp(False)`, `torch.backends.cuda.enable_math_sdp(True)` to solve the problem.
- A stable version for JetPack 6.x is Torch 2.10, Torch-TensorRT 2.10 and TensorRT 10.7.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -72,7 +72,7 @@ new_local_repository(
 new_local_repository(
     name = "cuda_l4t",
     build_file = "@//third_party/cuda:BUILD",
-    path = "/usr/local/cuda-12.6",
+    path = "/usr/local/cuda-13.0",
 )
 
 # for windows
@@ -118,6 +118,19 @@ local_torch(name = "libtorch")
 #   build_file = "@//third_party/tensorrt/local:BUILD"
 #)
 
+#new_local_repository(
+#   name = "tensorrt_l4t",
+#   path = "/usr/",
+#   build_file = "@//third_party/tensorrt/local:BUILD"
+#)
+
+#new_local_repository(
+#    name = "torch_l4t",
+# please modify this path according to your local python wheel path
+#    path = "/usr/local/lib/python3.11/dist-packages/torch",
+#    build_file = "third_party/libtorch/BUILD"
+#)
+
 #############################################################################################################
 # Tarballs and fetched dependencies (default - use in cases when building from precompiled bin and tarballs)
 #############################################################################################################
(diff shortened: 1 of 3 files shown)


narendasan:
@Mike-Sagiri can you explain more about the system configuration you are targeting? If it is Jetson Thor / JP 7.0+, you should be able to use SBSA style distributions (aarch64_linux configuration). Or are you targeting Orin JP 6.2? 

The PR seems to hint at JP7.0+

Mike-Sagiri:
@narendasan Thanks for raising this point. Actually, the goal of this PR is not to target a specific JetPack release. The main issue I am addressing is the build configuration under the `--jetpack` flag itself.

In principle, both JetPack 6 and JetPack 7 can be supported through `--jetpack` as long as the dependencies in `MODULE.bazel` are configured appropriately. The PR does not introduce any JP7-specific build logic.

That said, for the dependency versions currently used by Torch-TensorRT 2.12, I aligned the JetPack configuration with the versions already used on other architectures (CUDA, PyTorch, and TensorRT). As a result, the default configuration in this PR effectively matches the dependency stack available on JetPack 7.

One important consideration is that TensorRT versions available on JetPack 6 do not provide the `add_attention` API/operator that Torch-TensorRT 2.12 depends on. Because of this, Torch-TensorRT 2.12 is not actually in a stable or fully functional state on JetPack 6, regardless of the build configuration. In practice, Torch-TensorRT 2.10 is the most stable version for JetPack 6 systems, mentioned in "known issue".

Therefore, my rationale was to make the default JetPack configuration consistent with the dependency versions already used by other architectures, rather than maintaining a separate JetPack-specific dependency matrix that would still not provide full Torch-TensorRT 2.12 functionality on JetPack 6.

narendasan:
I see. Generally the direction of Jetpack as a whole is to align on SBSA (the `aarch64_linux` target for Torch-TensorRT): https://developer.nvidia.com/embedded/jetpack/downloads. Jetson Thor was first with Jetpack 7.0, but JP7.0 did not support Jetson Orin which is why we left the `jetpack` target in the code base to support JP6.2 / Orin as long as there were PyTorch builds available. With JP 7.2, Orin should be also aligned on SBSA and likely we will remove the jetpack target in the future and support SBSA, Thor and Orin via the `aarch64_linux` target

Mike-Sagiri:
@narendasan Wow, that's great to hear. My motivation for this PR was mainly to improve support for the current JP6.x / Orin workflow. Since the long-term direction is to move everything to the SBSA (`aarch64_linux`) target and the `jetpack` target will likely be retired, I'll go ahead and close this PR.

If fixes are still being considered for versions prior to 2.10, it might be worth addressing the file matching issue in `third_party/tensorrt/local/BUILD`. The changes in this PR can be used as a reference.

PR #4315, opened 2026-06-02
build: upgrade cuda to 13.2

fs-eire:
# Description

Upgrade build dependency for CUDA from 13.0 to 13.2 for Linux and Windows.


lanluo-nvidia (approved):
LGTM

PR #4325, opened 2026-06-08
feat: multiple optimization profiles for disjoint input shape regimes

cehongwang:
Add support for defining N optimization profiles at compile time via the list-based ``Input.profiles`` API and selecting the active profile at runtime (manual pin by index, or opt-in shape-based auto-selection).

- AOT (torch.export) compile path builds one TRT optimization profile per declared profile index; submodules inherit the profile count via propagation across graph breaks.
- Python and C++ runtimes expose a matching primitive engine API (set_active_profile / num_optimization_profiles / _active_profile_index / _auto_select_profiles) so the two runtimes remain interchangeable.
- Profile selection is exposed through the optimization_profile context manager; auto-selection uses lazy (first-fitting) profile selection.
- Backward compatible: engines without declared profiles keep the historical single-profile (dynamic) / no-profile (static) behavior.

Includes an example and runtime tests covering dynamic submodule inputs.

# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/core/runtime/TRTEngine.cpp
+++ b/core/runtime/TRTEngine.cpp
@@ -280,6 +280,11 @@ TRTEngine::TRTEngine(
   }
 
   has_dynamic_inputs = engine_has_dynamic_inputs(cuda_engine.get(), in_binding_names);
+  // Cache input binding metadata used by the runtime hot path, then reconstruct
+  // optimization-profile info from the TRT API so multi-profile selection works
+  // for any loaded engine.
+  this->setup_input_binding_infos();
+  this->setup_optimization_profiles();
 
 #ifndef NDEBUG
   this->enable_profiling();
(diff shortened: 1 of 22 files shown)


core/runtime/TRTEngine.cpp
```diff
@@ -512,6 +516,97 @@ void TRTEngine::reset_captured_graph() {
...
+}
+
+void TRTEngine::set_active_profile(int64_t profile_index) {
+  if (num_optimization_profiles <= 1) {
+    return;
+  }
+  if (profile_index == active_profile_index) {
+    return;
+  }
+  auto stream = c10::cuda::getCurrentCUDAStream(device_info.id);
```
narendasan: Does this work with the green context pr?

core/runtime/TRTEngine.cpp
```diff
@@ -512,6 +516,97 @@ void TRTEngine::reset_captured_graph() {
...
+      if (i >= inputs.size() || is_shape_inference_io[name]) {
+        continue;
+      }
+      auto ranges_it = profile_dim_ranges.find(name);
+      if (ranges_it == profile_dim_ranges.end()) {
+        continue;
+      }
+      const auto& dims = ranges_it->second;
+      auto sizes = inputs[i].sizes();
+      for (size_t d = 0; d < sizes.size(); ++d) {
```
narendasan: Can we cache only what the dynamic dimension is for each profile and its ranges? Then we dont need to search mostly static dims

core/runtime/TRTEngine.cpp
```diff
@@ -512,6 +516,117 @@ void TRTEngine::reset_captured_graph() {
...
 }
 
+void TRTEngine::setup_optimization_profiles() {
+  num_optimization_profiles = cuda_engine->getNbOptimizationProfiles();
+  profile_dynamic_dims.clear();
+  is_shape_inference_io.clear();
+  for (const auto& name : in_binding_names) {
+    is_shape_inference_io[name] = cuda_engine->isShapeInferenceIO(name.c_str());
+  }
+  if (num_optimization_profiles <= 1) {
```
narendasan, line 583: What not do this first to short cut everything?

core/runtime/TRTEngine.cpp
```diff
@@ -512,6 +516,117 @@ void TRTEngine::reset_captured_graph() {
...
+        }
+      }
+      if (is_dynamic) {
+        dynamic_dims.push_back({d, ranges});
+      }
+    }
+  }
+}
+
+void TRTEngine::set_active_profile(int64_t profile_index) {
```
narendasan, line 632: I kind of feel like this function should take stream as argument and put it on the caller to give the right stream instead of getting the current stream in the function body

core/runtime/TRTEngine.cpp
```diff
@@ -512,6 +516,117 @@ void TRTEngine::reset_captured_graph() {
...
+int64_t TRTEngine::auto_select_profile(const std::vector<at::Tensor>& inputs) {
+  // Lazy selection: scan profiles in index order and return the first one whose
+  // [min, max] ranges contain every input's dynamic dims. Static dims are not
+  // cached (they cannot distinguish profiles) and are validated later by TRT at
+  // setInputShape.
+  for (int64_t p = 0; p < num_optimization_profiles; ++p) {
+    bool fits = true;
+    for (size_t i = 0; i < in_binding_names.size() && fits; ++i) {
+      const auto& name = in_binding_names[i];
+      if (i >= inputs.size() || is_shape_inference_io[name]) {
```
narendasan: I feel like this function doesnt make a ton of sense and could potentially lead to thrashing. 

Fundamentally in the auto mode I think first we should see if the active profile is valid for the set of inputs.

The mechanism we use to map from the list to an input binding name should be identical to the code we use in execute_engine. Its fundamentally the same job. In fact we should only do it once and reuse this mapping result. There should be some map from index in the input list to a name. 

If the inputs fit we should short cut and return. 

I also think because of this the responsibility of changing the opt profile should be on these methods and not the caller. That way we can just no-op if it fits.

py/torch_tensorrt/dynamo/runtime/_TRTEngine.py
```diff
@@ -614,6 +625,107 @@ def _enable_rtx_native_cudagraphs(self) -> None:
...
+        Lazy selection: scan profiles in index order and return the **first** one
+        whose ``[min, max]`` ranges contain every input's dynamic dims. Static
+        dims are not cached (they cannot distinguish profiles) and are validated
+        later by TRT at ``set_input_shape``. Overlapping profiles therefore
+        resolve to the lowest matching index; pin manually via
+        ``optimization_profile(module, index)`` to force a specific profile.
+        """
+        for p in range(self.num_optimization_profiles):
+            fits = True
+            for i, name in enumerate(self.in_binding_names):
```
narendasan: Same sort of feedback here. Construct a index -> name map, then iterate through the indexes, if they fit all just return other wise go through your profile dim cache to find the next fit.

core/runtime/TRTEngine.cpp
cehongwang: yes

PR #4262, opened 2026-05-13
chore: add moe tests

zewenli98:
# Description

Add tests/py/dynamo/hlo/test_moe.py covering all 7 MoE routing and dispatch variants from popular open-source LLMs (Mixtral, Qwen2, Qwen3, Llama4, DeepSeek-V2, DeepSeek-V3/R1, Nemotron-H).  50 parameterized test cases verify TRT numerical correctness against PyTorch reference.

Extend DispatchTestCase.run_test() with require_full_compilation=True support: calls TRTInterpreter.validate_conversion() before building the engine and fails immediately if any op lacks a TRT converter.  All MoE tests pass this check, confirming zero PyTorch fallback for every routing/dispatch/expert-MLP combination tested.

# Checklist:

- [x] My code follows the style guidelines of this project (You can use the linters)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests to verify my fix or my feature
- [x] New and existing unit tests pass locally with my changes
- [x] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #4328, opened 2026-06-09
executorch: log once when the TensorRT delegate stages host I/O; document the device-handling contract

shoumikhin:
## What

Two small, self-contained changes to the ExecuTorch TensorRT delegate runtime (`cpp/.../executorch/TensorRTBackend.{h,cpp}`):

1. **Observability for the zero-copy path.** `execute()` binds device-resident I/O straight into the execution context (zero-copy) and silently stages host-resident I/O through a per-call device allocation + `cudaMemcpyAsync`, with no signal. This adds a one-shot `ET_LOG(Info)` (guarded by a new `staged_warned` flag on `EngineHandle`, read/written under the existing `mu` lock) the first time an engine stages host I/O, so a caller intending device-resident, zero-copy I/O can tell when they have fallen off the fast path. Purely additive.

2. **Document the device-handling contract.** A WHY-only comment at `execute()`'s header records three intentional choices that diverge from the CUDA/AOTI delegate: (a) the runtime ignores the AOT `target_device` metadata and sniffs pointers at runtime (so asserting `device_type` would wrongly reject host inputs today), (b) it stages H2D/D2H itself rather than via ExecuTorch device-copy ops / a `DeviceAllocator`, and (c) the engine-baked `device_id` is the runtime source of truth for the GPU.

## Why now

ExecuTorch is moving on-device memory planning toward the default. Once that lands, device-aware planning is meant to keep delegate I/O on device; a regression that reinserts a host op between two GPU delegates would silently restore copies. The one-shot log makes that visible, and the contract comment keeps the intentional divergences from being "fixed" into breakage.

## Risk

Low. (1) is additive logging guarded by a flag under the existing lock; (2) is a comment. No fast-path or control-flow change.

## Test plan

- Device-resident I/O: the new Info line is NOT emitted (fast path unchanged).
- Host I/O: exactly one Info line per engine on the first `execute()`.
- Repeated calls / a second engine: one line per engine total.
- Existing TensorRT delegate tests pass; outputs unchanged.

Draft for review; part of a small device-support readiness series (a partitioner-side `target_device` fix + a readiness test follow).

### DIFF
--- a/cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
+++ b/cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
@@ -61,6 +61,9 @@ struct EngineHandle {
   size_t num_outputs = 0;
   int device_id = 0;
   bool unified_memory = false;
+  // Set the first time execute() stages an IO tensor through a host<->device
+  // copy; guards a one-shot Info log (read/written under mu).
+  bool staged_logged = false;
   std::mutex mu;
   // Makes the skip-sync fast path safe to reuse: TensorRT forbids reconfiguring or
   // destroying an execution context while one of its enqueues is in flight, so when
--- a/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
+++ b/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
@@ -317,6 +317,20 @@ Result<DelegateHandle*> TensorRTBackend::init(
 // Args layout (mirroring the Python exporter):
 //   args[0 .. num_inputs-1]             – input EValues
 //   args[num_inputs .. num_inputs+num_outputs-1] – output EValues
+//
+// Device-handling contract (intentional divergence from the CUDA/AOTI delegate;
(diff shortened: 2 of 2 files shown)


shoumikhin:
@narendasan @lanluo-nvidia — would appreciate a review when you have a chance.

This adds a one-shot `ET_LOG(Info)` the first time the ExecuTorch TensorRT delegate has to stage host I/O (a CPU↔device copy), plus a comment documenting the delegate's device-handling contract. It's prep for ExecuTorch's device-memory-planning flip: once delegate I/O is GPU-resident the copies disappear, and this makes the non-optimal (host-staged) path visible/diagnosable without per-run log spam.

The red CI is the current repo-wide Python-3.10 / Windows infra outage — `torch_tensorrt` fails to install (`pypi.nvidia.com` TLS error on Windows + a nightly-TRT weight-streaming flake), so the 3.10 test lanes error at import. Same failures show on unrelated PRs (#4318, #4262); the C++/lint/build lanes here are green. Not caused by this change.

Issue #4334, opened 2026-06-10
Support TensorRT weight streaming in the ExecuTorch delegate so large models can run


shoumikhin:
## Feature

Add TensorRT weight streaming support to the Torch-TensorRT ExecuTorch delegate runtime. This lets a model whose weights do not all fit in GPU memory still run after it is exported to an ExecuTorch program (a `.pte` file).

## Problem

Torch-TensorRT can already build a weight streamable engine when you compile with `enable_weight_streaming=True`, and it can export to ExecuTorch with `torch_tensorrt.save(..., output_format="executorch")`.

The problem is that the ExecuTorch delegate never sets a weight streaming budget. After it deserializes the engine it goes straight to creating the execution context. With no budget set, TensorRT keeps all weights resident on the GPU, so a model that is larger than GPU memory fails to run through the ExecuTorch path, even when the engine was built for streaming.

The other Torch-TensorRT runtimes already set a budget right after deserializing the engine, which is why they can stream. The ExecuTorch delegate was missing this one step.

## Solution

Set the weight streaming budget inside the ExecuTorch delegate `init()`, after the engine is deserialized and before the execution context is created. This matches the other runtimes and respects the TensorRT rule that the budget cannot change once a context exists.

By default the delegate applies TensorRT's automatic budget, which TensorRT computes at load time from the free memory on the actual GPU. So an engine built with `enable_weight_streaming=True` runs out of the box, and the budget adapts to whatever device the `.pte` is deployed on. This is the same thing the existing PyTorch runtimes do right after they deserialize an engine.

An optional explicit budget can be set at export time. It travels to the runtime as an ExecuTorch CompileSpec, which is a key and value that ExecuTorch already stores in the `.pte` and passes back to the backend at load time. This needs no changes to ExecuTorch and no change to the engine blob format.

### The option

`weight_streaming_budget` on `save(..., output_format="executorch")`:

* `None` (the default): the delegate applies TensorRT's automatic budget at load.
* a non-negative integer: a fixed GPU budget for weights, in bytes.

This mirrors the byte budget used by the existing Torch-TensorRT runtimes. There is no string vocabulary to remember.

### Ahead of time (Python)

```python
import torch_tensorrt

compiled = torch_tensorrt.compile(
    model, arg_inputs=example_inputs, enable_weight_streaming=True
)

# Works out of the box. The delegate applies the automatic budget at load,
# so a model whose weights exceed GPU memory runs.
torch_tensorrt.save(
[...]

PR #4336, opened 2026-06-11
Add TensorRT weight streaming support to the ExecuTorch delegate

shoumikhin:
## Summary

Add TensorRT weight streaming support to the Torch-TensorRT ExecuTorch delegate, so a model whose weights do not all fit in GPU memory can run when exported to an ExecuTorch program. Ref #4334.

Torch-TensorRT already builds a weight streamable engine when you compile with `enable_weight_streaming=True`, but the ExecuTorch delegate never set a budget on the engine at load time, so large models could not stream. This change sets the budget in the delegate `init()`, after the engine is deserialized and before the execution context is created, which is the same pattern the other Torch-TensorRT runtimes already use.

## How it works

By default the delegate applies TensorRT's automatic budget, computed at load time from the free memory on the actual GPU, gated on `getStreamableWeightsSize() > 0`. So an engine built with `enable_weight_streaming=True` runs out of the box and adapts to the deploy device. Nothing is baked into the `.pte` for this default case.

An explicit budget is a non-negative number of bytes and can be set two ways, in order of precedence:

1. **Load time (preferred):** an ExecuTorch backend option named `weight_streaming_budget`, passed by the caller via `Module::load(LoadBackendOptionsMap)` and read in `init()` with `BackendInitContext::get_runtime_spec`. This lets a deployment size the budget for its own GPU without re-exporting. It is the same load-time pattern CoreML and XNNPACK use.
2. **Export time (default / fallback):** the same backend-option key baked into the `.pte` via `torch_tensorrt.save(output_format="executorch", weight_streaming_budget_per_engine=N)`. The Python keyword is named `_per_engine` because `weight_streaming_budget` already means a program-wide total on `MutableTorchTensorRTModule`; the key on the wire is unchanged. This is used when no load-time option is given, and it is the only channel for loaders that cannot pass backend options yet (the ExecuTorch Python and Android runtimes).

Resolution order in the delegate is: load-time option, then the baked value, then automatic. The value is a decimal string on the wire because ExecuTorch's typed integer option is only 32 bit and a byte budget can exceed 2 GB.

```python
import torch_tensorrt

compiled = torch_tensorrt.compile(
    model, arg_inputs=example_inputs, enable_weight_streaming=True
)

# Default: the delegate applies the automatic budget at load, so a large model runs.
torch_tensorrt.save(
    compiled, "model.pte", arg_inputs=example_inputs, output_format="executorch"
)

# Optional export-time default budget (overridable at load):
torch_tensorrt.save(
    compiled, "model.pte", arg_inputs=example_inputs,
    output_format="executorch",
    weight_streaming_budget_per_engine=8 * 1024**3,  # 8 GiB
)
```

```cpp
// C++ runtime: use the automatic or baked budget (no change needed), or override
// the budget at load for this specific GPU with a backend option.
executorch::runtime::BackendOptions<1> trt_opts;
trt_opts.set_option("weight_streaming_budget", "8589934592");  // 8 GiB
executorch::runtime::LoadBackendOptionsMap options;
options.set_options("TensorRTBackend", trt_opts.view());

executorch::extension::Module module("model.pte");
module.load(options);
auto outputs = module.forward(inputs);
```

## Changes

* `TensorRTBackend::init` applies the budget via `setWeightStreamingBudgetV2` before creating the execution context, gated on `getStreamableWeightsSize() > 0`. It resolves the budget as load-time runtime spec, then baked compile spec, then automatic.
* New standalone `WeightStreamingBudget` parser (`cpp/include` and `cpp/src`), unit tested without a GPU. It uses `std::from_chars` and accepts only a non-negative decimal integer.
* `torch_tensorrt.executorch.export(..., weight_streaming_budget_per_engine=...)` writes the export-time default into the TensorRT partitioner's compile specs, for every method of a multi-method export. `torch_tensorrt.save(output_format="executorch")` forwards the same argument. Validation lives next to the compile spec key in `torch_tensorrt/executorch/partitioner.py`. Passing the budget through `compile_specs` by hand is rejected in favor of the keyword argument, because that route skips the validation.
* C++ gtest cases for the parser and a CPU Python test suite.
* Bazel and CMake wiring for the new files.

## Requirements

The load-time override uses ExecuTorch's `BackendInitContext::get_runtime_spec` and `LoadBackendOptionsMap`. The export-time default and the automatic budget work without it.

## Backward compatibility

* The new code only runs when the engine was built for weight streaming. Engines built with the default settings report zero streamable weights and skip the new path, so they behave exactly as before. This covers every existing `.pte`.
* There is no change to the `.pte` format or the engine blob.
* The one intended behavior change is that a streamable engine now applies the automatic budget at load, which enables the large model case and matches the PyTorch runtimes.

## Edge cases

* A budget on an engine that was not built for streaming is ignored with a log.
* An explicit budget larger than the streamable size is clamped.
* A malformed or negative budget is rejected at export (TypeError or ValueError) and at load (`Error::InvalidProgram`).
* If the automatic budget cannot be applied, the runtime retries with maximum streaming before failing.
* For a model split into more than one engine, an explicit byte budget applies to each engine and emits a warning. Leave `weight_streaming_budget_per_engine` as `None` for multi-engine models.

## Status and validation

* The C++ value parser was built and its gtest cases run on CPU.
* The new Python tests run on CPU. 29 tests pass, covering budget validation, the compile spec
  reaching the TensorRT partitioner for single-method and multi-method exports, the rejection of a
  hand-written spec, the multi-engine warning, and the `save()` argument guards. Each was checked to
  fail when the code it covers is removed.
* Running the whole `tests/py/dynamo/executorch` directory on this change and on the base branch
  gives the same set of failures, so the change adds no new ones. Those shared failures are from the
  test machine having no GPU and no compiled Torch-TensorRT runtime, not from the code.
* Everything that needs a GPU, including weight streaming actually running on a device, is covered
  by CI rather than by hand here.

## Follow-ups

* Expose `LoadBackendOptionsMap` in ExecuTorch's Python (and Android) runtime bindings so non-C++ loaders can also set the budget at load. Until then the export-time default covers them.
* Add a live, after-load setter (change the budget on an already loaded model) via the backend `set_option` API. This needs the execution context to be destroyed and recreated, like `TRTEngine::set_device_memory_budget`, so it is deferred.
* Cover the budget logic in `init()` with a test: runtime-spec versus compile-spec precedence, the clamp to `getStreamableWeightsSize()`, and the automatic-budget fallback. The current tests cover the byte-string parser and the export-time plumbing only. This needs an engine actually built with `enable_weight_streaming=True`, so it belongs in the GPU delegate lane rather than here.

## Behavior change worth calling out

`save(output_format="executorch", ...)` now raises `TypeError` on an unrecognised keyword argument instead of ignoring it. This turns a silent typo into a hard failure, which is the point, but it is a behavior change for anyone who was passing an unused kwarg. It is scoped to `output_format="executorch"`, so the other output formats are unaffected.

### DIFF
--- a/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
+++ b/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
@@ -8,6 +8,7 @@
 #include "torch_tensorrt/executorch/TensorRTBackend.h"
 #include "torch_tensorrt/executorch/TensorRTBindingNames.h"
 #include "torch_tensorrt/executorch/TensorRTBlobHeader.h"
+#include "torch_tensorrt/executorch/WeightStreamingBudget.h"
 
 #include <cstdint>
 #include <cstring>
@@ -225,8 +226,6 @@ Result<DelegateHandle*> TensorRTBackend::init(
     BackendInitContext& context,
     FreeableBuffer* processed,
     ArrayRef<CompileSpec> compile_specs) const {
-  (void)compile_specs;
-
   TORCHTRT_ET_CHECK_NOT_NULL(processed, Error::InvalidArgument, "TensorRTBackend::init: null processed buffer");
   TORCHTRT_ET_CHECK_NOT_NULL(processed->data(), Error::InvalidArgument, "TensorRTBackend::init: null processed buffer");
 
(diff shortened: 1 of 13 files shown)


narendasan (commented):
Why do we want to set the budget at serialization time? This is a runtime configurable setting. Should we expose some sort of API to let someone set this when they deserialize or later?

shoumikhin:
Good call, and I agree the budget should be a runtime setting. Here is what the PR does today, and a change that gives you exactly what you are asking for.

**What happens today**

- If you do not set a budget, nothing is written into the `.pte`. At load time the delegate asks TensorRT to pick an automatic budget based on the GPU it is actually running on. So the normal case already adapts to the deployment device, with no value baked in.
- If you do set an explicit budget, that number is stored in the `.pte` and applied at load. It is applied inside the delegate's `init()`, right after the engine is deserialized and just before the execution context is created. This timing is required: TensorRT does not let you change the budget once an execution context exists.

So the only thing fixed at export is the explicit number, and you are right that a fixed byte count is really a per-deployment choice.

**Proposed change (a real load-time API)**

- Read the budget at load from ExecuTorch's existing backend options. In `init()` the delegate will look for a `weight_streaming_budget` option in `BackendInitContext` (the `LoadBackendOptionsMap` a caller passes to `Module::load`). These options arrive before the execution context is created, so the timing stays correct and no context rebuild is needed.
- Order of precedence: load-time option first, then the value baked at export, then automatic. This is the same pattern CoreML uses for `compute_unit` and XNNPACK uses for its load options.
- The value is passed as a string, because ExecuTorch's typed integer option is only 32 bit and a byte budget can be larger than 2 GB.

**Why keep the export-time value too**

ExecuTorch's Python and Android load paths do not expose backend options yet (only C++ `Module` and iOS do). So for anyone loading a `.pte` from Python or Android, the value baked at export is currently the only way to set an explicit budget. I would keep it as an overridable default rather than remove it.

**Changing it after load (the "or later" part)**

This is doable as a follow-up, not in this PR. TensorRT requires destroying and recreating the execution context to change the budget, and ExecuTorch's post-load `set_option` is global to the backend with no per-engine handle. The CUDA backend already implements `set_option`, so there is a pattern to follow when we do it.

**One question so I build the right thing**

Will these large models be served mainly from the C++ runtime or from Python? If C++, the load-time option covers it and we can treat the baked value as just a default. If Python, we need to keep the baked value until ExecuTorch exposes backend options to Python, which I am happy to help add upstream.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -284,6 +283,130 @@ Result<DelegateHandle*> TensorRTBackend::init(
...
+      }
+    }
+    if (len > 0) {
+      ws_request = parse_weight_streaming_budget(value, len);
+      if (!ws_request.valid) {
+        ET_LOG(Error, "TensorRTBackend::init: malformed weight_streaming_budget runtime option");
+        return Error::InvalidProgram;
+      }
+      is_explicit = true;
+    }
```
cehongwang, line 354: What happens if len = 0? Give a warning to the user

py/torch_tensorrt/_compile.py
```diff
@@ -1339,9 +1440,17 @@ def _save_as_executorch(exp_program: Any, file_path: str, **kwargs: Any) -> None
...
-    partitioners = [
-        TensorRTPartitioner(compile_specs=list(executorch_compile_specs))
-    ] + list(extra_partitioners)
+    # Resolve the weight streaming budget into the compile_specs from the save()
+    # kwarg. If none is given nothing is added and the delegate applies TensorRT's
+    # automatic budget itself for engines built with weight streaming.
+    resolved_compile_specs = _resolve_executorch_compile_specs(
+        exp_program,
+        list(executorch_compile_specs),
+        kwargs.get("weight_streaming_budget"),
```
cehongwang: Here we expect users to set the total budget right? If there is a graph break, is there a correct mapping to the per-engine budget?

cehongwang (changes requested):
Overall OK. Some minor comments

shoumikhin:
Both of these are addressed now.

**Empty runtime option.** An empty value meant "unset", and the code fell through to the
budget recorded at build time without saying anything, so a caller who passed the option had
no way to tell it was ignored. It now logs that the value was empty and which budget is being
used instead.

**Budget with a graph break.** Good question to push on, and the answer is that there is no
split: the budget is applied to every engine, so a graph that breaks into N engines can hold N
budgets resident at once. Rather than invent a division that TensorRT does not model, the
docstring now states this plainly and suggests sizing against the largest engine, and an export
that produces more than one engine warns when a budget is set:

```
weight_streaming_budget applies to each of the N engines separately, not as a total for the
program, so peak weight memory can reach N times the value given.
```

While testing that, the type check for the budget turned out to run after the model type check,
so passing a bad budget together with an unsupported model reported only the model error. The
check now runs where the option is accepted.

py/torch_tensorrt/_compile.py
shoumikhin: You were right, and it was more than a doc gap: `weight_streaming_budget` already means a
program-wide budget on `MutableTorchTensorRTModule`, split proportionally across engines, so
reusing the name here gave it two meanings. Renamed the ExecuTorch option to
`weight_streaming_budget_per_engine`, which is what the delegate actually applies. There is no
program-total mapping in this path today: the save-time resolver sees the engine nodes but only
counts them, and never queries each engine's streamable-weight size.

I did not use an equal split because each engine clamps its own value and unused bytes are
never reassigned, so a skewed program would quietly land less than requested. The docstring now
states that ExecuTorch initializes every delegate when the method loads and keeps those handles
until the method is destroyed, so N engines can add up to N times the per-engine value before
other engine memory counts. A real program total would need a coordinating layer that can read
every engine's streamable size before assigning per-engine budgets.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
shoumikhin: Correcting my earlier note on this thread: an empty runtime option now logs at `Error`, since
ExecuTorch has no `Warning` level, and says only that the empty value was ignored. It no longer
names the fallback, because at that point the backend may still use the budget stored in the
program if that compile spec exists, or TensorRT's automatic budget otherwise.

shoumikhin:
It is settable when the method loads: a valid non-empty `weight_streaming_budget` backend
option takes precedence over the value stored in the program, while an empty value is ignored
and falls through to the stored value, or to TensorRT's automatic budget if neither is
explicit. The stored value stays as an overridable default because the C++ and mobile loaders
can pass load-time backend options but the Python loader cannot, and the current tests cover
parsing and export plumbing rather than an end-to-end load with an override.

On "or later": there is no post-load setter here, because TensorRT needs active execution
contexts torn down before the budget changes, which is what the non-ExecuTorch runtime does by
recreating its context. Is the load-time path enough for this PR, or do you want the post-load
setter in scope?

py/torch_tensorrt/_compile.py
```diff
@@ -1449,6 +1577,14 @@ def _save_as_executorch(exp_program: Any, file_path: str, **kwargs: Any) -> None
             "delegate boundary overhead.",
             engine_count,
         )
+        if kwargs.get("weight_streaming_budget_per_engine") is not None:
+            logger.warning(
+                "weight_streaming_budget_per_engine applies to each of the %d engines "
```
cehongwang: A multi-engine export with an explicit budget logs the same caution twice, and counts the engine nodes twice to do it.

_resolve_executorch_compile_specs already warns on this exact condition at py/torch_tensorrt/_compile.py:1512 ("it is applied to each engine independently"), and it is called from _save_as_executorch a few lines above this block, at py/torch_tensorrt/_compile.py:1567. So a two-engine program saved with weight_streaming_budget_per_engine=N emits both messages back to back, and _count_executorch_engine_nodes walks the graph once inside the resolver and again at py/torch_tensorrt/_compile.py:1573.

The second copy is the one carrying the useful number (N times the value given), and the first is the one covered by test_multi_engine_explicit_warns. Keeping the tested site and dropping the duplicate is the smaller change:

            "delegate boundary overhead.",
            engine_count,
        )
        if kwargs.get("weight_streaming_budget_per_engine") is not None:
            logger.warning(
                "weight_streaming_budget_per_engine applies to each of the %d engines "
                "separately, not as a total for the program, so peak weight memory "
                "can reach %d times the value given.",
                engine_count,
                engine_count,
            )
If you would rather keep the "N times" phrasing, move it into the resolver's message and delete the resolver's own warning instead — either way the export should say this once.

cehongwang (approved):
Just one warning issue. And it is good to go

py/torch_tensorrt/_compile.py
shoumikhin: Fixed, and you were right on both halves: the same caution was printed twice and the graph was
walked twice to do it.

This is now one warning and zero extra walks. The rebase onto main moved the export machinery into
`torch_tensorrt/executorch/_export.py`, which already counts engine calls once per method for its
own "this program has N engines" warning. The budget caution now hangs off that existing count, so
there is nothing new to compute:

```python
if engine_call_counts[name] > 1:
    logger.warning(
        "%s contains %d TRT engine calls. Each one becomes its own delegate, "
        "so the boundary overhead is paid that many times.",
        name, engine_call_counts[name],
    )
    if weight_streaming_budget_per_engine is not None:
        logger.warning(
            "weight_streaming_budget_per_engine applies to each of those %d "
            "engines separately, not as a total, so resident weights for %s "
            "can reach %d times the value given.",
            engine_call_counts[name], name, engine_call_counts[name],
        )
```

I kept the "N times the value given" phrasing you preferred, and the separate warning that used to
live in the compile-spec resolver is gone. `test_multi_engine_explicit_warns` now asserts on this
message, and `test_multi_engine_none_does_not_warn` asserts it stays quiet with no budget set.

Two other things came out of the same rebase:

* The budget is now a keyword on `torch_tensorrt.executorch.export()`, not only on `save()`. That is
  where compile specs get resolved per method after the recent refactor, so a multi-method export
  gets the budget on every method. `save()` just forwards it.
* Validation moved next to the compile spec key in `torch_tensorrt/executorch/partitioner.py`, so
  the key and the rules for its value live in one place.

shoumikhin: Fixed: the rebase onto main moved this into `torch_tensorrt/executorch/_export.py`, where the caution now prints once off the engine count that module already computes, so the duplicate message and the second graph walk are both gone.

PR #4340, opened 2026-06-11
namedtuple in input

apbose:
Namedtuple shape spec for torch_tensorrt.Input

Adds a shorthand for declaring shared dynamic axes across inputs: pass a namedtuple as min_shape/opt_shape/max_shape and the field names are used as axis names. Axes with the same name across multiple Input objects are automatically exported as a single shared torch.export.Dim, without requiring the user to pass shared_dims explicitly.

```
Example:
S = namedtuple("shape", ["batch", "seq"])
inputs = [
    torch_tensorrt.Input(min_shape=S(1, 16), opt_shape=S(4, 16), max_shape=S(4, 16),
                         dtype=torch.int64, name="input_ids"),
    torch_tensorrt.Input(min_shape=S(1, 16), opt_shape=S(4, 16), max_shape=S(4, 16),
                         dtype=torch.int64, name="attention_mask"),
]
trt_mod = torch_tensorrt.compile(model, ir="dynamo", inputs=inputs, ...)
```

### DIFF
--- a/docsrc/user_guide/compilation/dynamic_shapes.rst
+++ b/docsrc/user_guide/compilation/dynamic_shapes.rst
@@ -121,6 +121,44 @@ can be shared simultaneously with ``shared_dims={0: "B", 1: "S"}``.
 
 See the full runnable example: :ref:`shared_dynamic_dims`.
 
+Namedtuple Shorthand for Shared Dims
+-------------------------------------
+
+Passing a ``namedtuple`` as ``min_shape`` / ``opt_shape`` / ``max_shape`` is a
+shorthand for the ``shared_dims`` pattern above.  Field names become axis names
+automatically — no ``shared_dims`` kwarg required.  Dynamic axes (``min != max``)
+with the same field name across inputs are exported as one shared
+``torch.export.Dim``; static axes (``min == max``) are ignored.
+
+.. code-block:: python
+
+    from collections import namedtuple
+    import torch
+    import torch_tensorrt
+
+    model = MyHFEncoder().eval().cuda()
+
+    S = namedtuple("shape", ["batch", "seq"])
+
+    inputs = [
+        torch_tensorrt.Input(
+            min_shape=S(1, 16), opt_shape=S(4, 16), max_shape=S(8, 16),
+            dtype=torch.int64, name="input_ids",
+        ),
+        torch_tensorrt.Input(
+            min_shape=S(1, 16), opt_shape=S(4, 16), max_shape=S(8, 16),
+            dtype=torch.int64, name="attention_mask",
+        ),
+    ]
+    # "batch" is dynamic (1→8) on both inputs → one shared Dim("batch")
+    # "seq"   is static  (16==16)             → no Dim needed
+
+    trt_model = torch_tensorrt.compile(model, ir="dynamo", inputs=inputs)
+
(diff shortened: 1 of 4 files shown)


narendasan (commented):
What is the error handling story here? What happens if someone makes a mistake such as reusing names between dims or if disjoint dim sets share names etc. Do we have any infra to at least report what we derived?

apbose:
Error handling cases
1. Duplicate field name within one namedtuple - Python rejects it before our code runs:

S = namedtuple("shape", ["batch", "batch"])
`ValueError: Encountered duplicate field name: 'batch'`

2. min_shape a namedtuple, opt/max plain tuples:

Input(min_shape=S(1,16), opt_shape=(4,16), max_shape=(4,16), name="x")
```
TypeError: If min_shape is a namedtuple, opt_shape and max_shape
must also be namedtuples
```

3. Field names differ across min/opt/max:

S1 = namedtuple("shape", ["b", "c"])
S2 = namedtuple("shape", ["b", "seq"])
Input(min_shape=S1(1,16), opt_shape=S2(4,16), max_shape=S1(4,16), name="x")
```
ValueError: min_shape, opt_shape, max_shape namedtuples must have identical field names
```

4. namedtuple shape + explicit shared_dims together:

Input(min_shape=S(1,16), opt_shape=S(4,16), max_shape=S(4,16),
      name="x", shared_dims={0: "b"})
`ValueError: Cannot specify both a namedtuple min_shape and shared_dims; use one or the other to name dynamic axes.`

5. Same name with conflicting ranges "reusing names" landing on a range disagreement:

S = namedtuple("shape", ["batch", "seq"])
inputs = [
    Input(min_shape=S(1,16), opt_shape=S(4,16), max_shape=S(4,16), name="input_ids"),
    Input(min_shape=S(1,16), opt_shape=S(4,16), max_shape=S(8,16), name="attention_mask"),
]                                              #        ^ 8, not 4
```
 ValueError: Dimension name 'batch' is used with conflicting ranges
        1, 4) on input_ids[0] and (1, 8) on attention_mask[0].
             A shared named dimension must have identical (min, max)
             on every input that uses it.
```

The check is exact equality, so any mismatch is caught - not just non-overlapping ranges. I've added the on input_ids[0] ... on attention_mask[0] part in the last commit, since the old message gave the two ranges but not who owned them. That matters once there are more than two inputs:

... conflicting ranges (1, 4) on input_ids[0], attention_mask[0],
token_type_ids[0] and (1, 8) on position_ids[0]. ...

**The case that isn't caught**

6. Same name, same range, semantically unrelated axes:

S_image = namedtuple("shape", ["n", "c", "h", "w"])   # c = channels
S_text  = namedtuple("shape", ["c", "seq"])           # c = batch   <- collision

If both c axes are dynamic over the same range they fuse into one Dim, silently. I don't think the library can rule this out, because it's intent-dependent - for a multimodal model whose image batch and text batch must match. If the ranges differ , then we get the error 5.

S_image = namedtuple("shape", ["n", "c", "h", "w"])   # n dynamic 1..8
S_text  = namedtuple("shape", ["batch", "seq"])       # batch dynamic 1..8
identical ranges, different names -> two independent Dims, no fusion

narendasan:
Are there tests for these?

PR #4341, opened 2026-06-12
fix 4326: bound exported dynamic shapes and normalize symbolic slice …

### LINKED ISSUE #4326: ✨[Feature] Support negative dimension slicing in the dynamic shape case

Patterns like this are starting to show up, starting with PyTorch 2.12. Presumably this is some thing similar to `x[0:-2]`. However, one of the values is a symbolic shape now. 

```
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._TRTInterpreter - INFO - Converted node /neg [<built-in function neg>] (Inputs: (add_176: (((s65 - 1)//8)) + 1@SymInt) | Outputs: (neg: -(((s65 - 1)//8)) - 1@SymInt))
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._TRTInterpreter - DEBUG - Converting node /slice_2 (kind: aten.slice.Tensor, args: ('full <Node>', '2 <int>', 'neg <Node>'))
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._ConverterRegistry - DEBUG - Converter options for aten.slice.Tensor: 1
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._ConverterRegistry - DEBUG - Selecting converter option 0 for converting aten.slice.Tensor
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion.impl.elementwise.base - INFO - skip broadcast for /slice_2_shape
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion.impl.elementwise.base - INFO - skip broadcast for /slice_2_sub_stop
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion.impl.elementwise.base - INFO - skip broadcast for /slice_2_sub_num
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion.impl.elementwise.base - INFO - skip broadcast for /slice_2_div
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion.impl.elementwise.base - INFO - skip broadcast for /slice_2_prod
2026-06-08 22:50:23 - root - DEBUG - add_slice(input=[ELEMENTWISE]-[aten_ops.full.default]-[/full_add]_output_full.default, start=(), shape=(), stride=())
2026-06-08 22:50:23 - root - DEBUG - (Unnamed Layer* 246) [Slice]
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._TRTInterpreter - INFO - Converted node /slice_2 [aten.slice.Tensor] (Inputs: (full: (s79, (((s65 - 1)//8)) + 1, (((s65 - 1)//8)) + 1)@torch.bool, 2, neg: -(((s65 - 1)//8)) - 1@SymInt) | Outputs: (slice_2: (s79, (((s65 - 1)//8)) + 1, (((s65 - 1)//8)) + 1)@torch.bool))
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._TRTInterpreter - DEBUG - Converting node /unsqueeze_21 (kind: aten.unsqueeze.default, args: ('gt_9 <Node>', '1 <int>'))
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._ConverterRegistry - DEBUG - Converter options for aten.unsqueeze.default: 1
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._ConverterRegistry - DEBUG - Selecting converter option 0 for converting aten.unsqueeze.default
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._TRTInterpreter - INFO - Converted node /unsqueeze_21 [aten.unsqueeze.default] (Inputs: (gt_9: (s79, (((s65 - 1)//8)) + 1)@torch.bool, 1) | Outputs: (unsqueeze_21: (s79, 1, (((s65 - 1)//8)) + 1)@torch.bool))
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._TRTInterpreter - DEBUG - Converting node /bitwise_and (kind: aten.bitwise_and.Tensor, args: ('slice_2 <Node>', 'unsqueeze_21 <Node>'))
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._ConverterRegistry - DEBUG - Converter options for aten.bitwise_and.Tensor: 1
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._ConverterRegistry - DEBUG - Selecting converter option 0 for converting aten.bitwise_and.Tensor
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion.impl.elementwise.base - INFO - skip broadcast for /bitwise_and_logical_and
2026-06-08 22:50:23 - torch_tensorrt.dynamo.conversion._TRTInterpreter - INFO - Converted node /bitwise_and [aten.bitwise_and.Tensor] (Inputs: (slice_2: (s79, (((s65 - 1)//8)) + 1, (((s65 - 1)//8)) + 1)@torch.bool, unsqueeze_21: (s79, 1, (((s65 - 1)//8)) + 1)@torch.bool) | Outputs: (bitwise_and: (s79, (((s65 - 1)//8)) + 1, (((s65 - 1)//8)) + 1)@torch.bool))
```

Detect these cases and insert the math to calculate the correct ending dimension 

**Additional context**

micwill755:
# Description

This change fixes three dynamic-shape issues exposed by the ZoomASR repro. Together, these fixes make Torch-TensorRT respect user-provided input bounds, lower negative symbolic slicing into TensorRT-friendly shape math, and clean up symbolic expressions that otherwise survive into runtime or conversion.

1. Dynamic shape bounds fix

A previous working branch had a helper called _apply_dynamic_shape_bounds in Torch-TensorRT _compiler.py. That function propagated user-provided torch_tensorrt.Input min/max bounds into the FX shape_env, which explains why the old path used the intended ZoomASR feature length bound of 3000 instead of the broad exported range [9, 16384].

2. Negative symbolic slice normalization

PyTorch 2.12 can emit slices with negative symbolic bounds, such as x[-n:] or x[:-n]. This change adds a lowering pass that rewrites those bounds to dim_size - n for both slice start and stop, making the graph explicit and TensorRT-friendly.

3. No-op sym_min cleanup

The issue discussion mentioned removing torch.sym_min(x, INT64_MAX) as a possible needed patch. This change adds that as a separate lowering pass; it removes the no-op sym_min expression before runtime, preventing failures where torch.sym_min receives tensor-like inputs.

Fixes #4326

## Type of change

# Checklist:

- [x] I performed a self-review of my own code
- [x] I commented my code, particularly in hard-to-understand areas and hacks
- [ ] I made corresponding changes to the documentation
- [x] I added tests to verify my fix or my feature
- [ ] I added the relevant labels to my PR so that relevant reviewers are notified


narendasan:
@micwill755 `precommit`/`prek (https://github.com/j178/prek)` should be able to lint your code for you

py/torch_tensorrt/dynamo/lowering/passes/normalize_negative_slice_stop.py
```diff
@@ -0,0 +1,88 @@
...
+                continue
+
+            bound = args[bound_index]
+            positive_offset = _negative_symint_operand(bound)
+            if positive_offset is None:
+                continue
+
+            with gm.graph.inserting_before(node):
+                dim_size = gm.graph.call_function(
+                    torch.ops.aten.sym_size.int, args=(input_node, dim)
```
narendasan: Nice yeah I think is exactly what we are looking for.

If you want something a bit cleaner to read, we have this utility: 
https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/lowering/_SubgraphBuilder.py

e.g. https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py#L456

but its more style than functional

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -898,6 +898,67 @@ def _insert_complex_io_adapters(
         partitioned_module.recompile()
 
 
+def _apply_dynamic_shape_bounds(
```
narendasan: @apbose please review / comment on if this overlaps with your work

py/torch_tensorrt/dynamo/lowering/passes/normalize_negative_slice_stop.py
micwill755: Cool, let me review and edit accordingly.

py/torch_tensorrt/dynamo/_compiler.py
micwill755: This was adapted from a previous fix in Wenbing’s branch: [_apply_dynamic_shape_bounds](http://github.com/wenbingl/Torch-TRT/blob/dev/py/torch_tensorrt/dynamo/_compiler.py#L804).

It fixes the case where ZoomASR is exported with broad or unbounded Dim.DYNAMIC ranges by applying the user-provided torch_tensorrt.Input min/max bounds during Torch-TensorRT compilation. Without this compiler-side fix, I had to work around the issue in the ZoomASR export code by changing the dynamic shape annotations from:

"features": {0: Dim.DYNAMIC, 1: Dim.DYNAMIC, 2: Dim.STATIC},
"feature_lengths": {0: Dim.DYNAMIC},
to explicitly bounded dimensions:

batch_dim = Dim("batch", min=1, max=max_batch_size)
feature_len_dim = Dim("feature_len", min=1, max=3000)

dynamic_shapes={
    "features": {0: batch_dim, 1: feature_len_dim, 2: Dim.STATIC},
    "feature_lengths": {0: batch_dim},
}

With _apply_dynamic_shape_bounds, the model export can keep using Dim.DYNAMIC, while Torch-TensorRT still constructs the intended bounded TensorRT optimization profile from the provided input specs.

narendasan: My only concern with this code and Wenbings prior version of this is im not sure it properly inserts the updated shapes. Like I would expect some form of shape prop done here as well after constraining the placeholders. If we can solve this in export with explicit ranges, it might be better to split this into its own PR that fully handles constraining an already exported exported program. It should still unblock Shane without and the explicit shape in export right?

micwill755: Agreed and yes this will unblock Shane.

apbose: yeah PR 4213 addresses this, validating the input bounds against the exporter bounds.
1. One issue against changing the var_to_range is that it tracks the exporter bounds, and if we change the Input bounds for a later compile, it conflates the Input bound of TRT profile with the export bound, and can lead to error. Also it does not do the shape prop. So something like this  fails in above

```
ep = torch.export.export(model, (sample,),
    dynamic_shapes={"x": {0: Dim.DYNAMIC}})
# var_to_range[s0] = [2, int_oo]  -- model valid for any size

Now two separate compilations:

# User A: small deployment, max batch 100
trt_a = torchtrt.dynamo.compile(ep, inputs=[Input(max_shape=(100, 8))])
# _apply_dynamic_shape_bounds mutates var_to_range[s0] → [2, 100]

# User B: larger deployment, max batch 500
trt_b = torchtrt.dynamo.compile(ep, inputs=[Input(max_shape=(500, 8))])
# _apply_dynamic_shape_bounds now sees var_to_range[s0] = [2, 100]
# computes min(100, 500) = 100
# User B silently gets max=100 instead of 500
```
2. Second type which the other PR addresses is — if the export used explicit Dim("batch", min=10, max=20) and the user passes Input(min_shape=2) this method will clamp input bound to 10 ignoring 2.

narendasan: @apbose and @micwill755 can you guys figure out if a follow on PR is needed for this or if we should be covered by #4213

apbose: The removal of no-op for sym_min and also the negative symbolic slice bounds are something which PR 4213 does not cover, which are needed

py/torch_tensorrt/dynamo/lowering/passes/normalize_negative_slice_stop.py
```diff
@@ -0,0 +1,85 @@
+import operator
+from typing import Optional
+
+import torch
+from torch.fx import GraphModule, Node
+from torch_tensorrt.dynamo.lowering._SubgraphBuilder import SubgraphBuilder
+
+from .pass_utils import clean_up_graph_after_modifications
+
+
```
apbose, line 11: cant we handle this case in the converter itself?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -2223,6 +2224,7 @@ def aten_ops_maximum(
     )
 
 
+@dynamo_tensorrt_converter(torch.sym_min, supports_dynamic_shapes=True)
 @dynamo_tensorrt_converter(torch.ops.aten.minimum.default, supports_dynamic_shapes=True)
```
apbose, line 2228: why do we have this in both the converter and a lowering pass to remove sym_min?

py/torch_tensorrt/dynamo/_compiler.py
apbose: synced with @micwill755. I will merge 4213 and he will rebase from there.

narendasan: I dont think he needs to rebase, I think we drop this specific change, merge the rest and then look at readding it on top of #4213 if needed

py/torch_tensorrt/dynamo/lowering/passes/normalize_negative_slice_stop.py
narendasan: Is there a reason we would prefer that?

narendasan: I think generally simpler converters and more in lowering is better, but I think the line needs to be clear, either we normalize dimensions in the graph or we do it in the converter

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
narendasan: Lowering pass should only remove a specific instance of sym_min

py/torch_tensorrt/dynamo/lowering/passes/normalize_negative_slice_stop.py
apbose: going through this case ya lowering seems better, since the negative seems to be ITensor, so we would need extra Iselect layers, making the converter more complicated. Though we have done this before in converters, but if we want simpler converters now, doing in lowering makes more sense

PR #4344, opened 2026-06-15
Use runtime platform for TensorRT engine target checks

SandSnip3r:
# Description

Use runtime platform for TensorRT engine target checks.
Reject unknown engine target platforms before serialization and deserialization.

This fixes an issue where running as an x86 process on ARM hardware trips a check in `TRTEngine`'s constructor due to a platform mismatch. Windows ARM supports emulation for x86 processes.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
+++ b/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
@@ -186,7 +186,6 @@ class TorchTensorRTModule(torch.nn.Module):  # type: ignore[misc]
                 else self.engine.are_output_tensors_unowned()
             ),
         }
-
         engine_info: List[str | bytes] = [""] * SERIALIZATION_LEN
         engine_info[ABI_TARGET_IDX] = (
             torch.ops.tensorrt.ABI_VERSION()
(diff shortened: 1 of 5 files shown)


py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -183,6 +186,11 @@ def _pack_engine_info(self) -> List[str | bytes]:
                 else self.engine.are_output_tensors_unowned()
             ),
         }
+        if self.target_platform == Platform.UNKNOWN:
```
narendasan: pack engine info is not just used for serialization but is instead a formatting function that hands off the TRT engine + metadata to either the python or c++ runtime whenever that happens. This might be overly defensive and lead to confusion since it should get rejected in the TRTEngine constructor (python or c++) anyway. Basically its not TorchTensorRTModule's job to care if a TRT engine can be setup

narendasan (commented):
Looks good, I think just one guard is unnecessary then I think is good to merge.

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
narendasan: https://github.com/pytorch/TensorRT/pull/4344/changes#diff-10ebe63ef48658946436e5c01d9b062c2527b2a8990a6b6ab8b067c1783571baR219

narendasan (approved):
LGTM after removing the `TorchTensorRTModule._pack_engine_info` check

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
SandSnip3r: yeah ok, fair point
good catch

PR #4345, opened 2026-06-16
feat: support custom ExecutorchBackendConfig in save(output_format='executorch')

shoumikhin:
## Description

`torch_tensorrt.save(..., output_format="executorch")` previously called `edge_program.to_executorch()` with no arguments, so the default `ExecutorchBackendConfig` was always used and callers had no way to customize ExecuTorch lowering (e.g. memory planning or device placement).

This adds an optional `backend_config` kwarg to `save()` that is forwarded to `to_executorch()`, mirroring the existing `partitioners=` and `compile_specs=` kwargs. When `backend_config` is passed with a non-executorch `output_format` it is ignored with a warning, matching the existing behavior for those kwargs.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/py/torch_tensorrt/_compile.py
+++ b/py/torch_tensorrt/_compile.py
@@ -741,11 +741,14 @@ def save(
 
                 - If both dynamic_shapes and Input objects are provided, the explicit dynamic_shapes
                   parameter takes precedence.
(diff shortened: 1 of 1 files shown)


cehongwang (approved):
LGTM

PR #4349, opened 2026-06-18
fix: resolve lifted TRT engine custom-object by graph-signature FQN in ExecuTorch export

shoumikhin:
## What's broken

Saving a Torch-TensorRT model with `output_format="executorch"` fails when the model was compiled with `require_full_compilation=False` (so some ops stay outside TensorRT) and exported through the standard `torch.export` path:

```
RuntimeError: execute_engine node 'execute_engine': placeholder engine
'obj__run_on_acc_0_engine' not found in exp_program.constants
```

The engine is actually present in `exp_program.constants` — `save()` just looks it up by the wrong key.

## Why it happens

When `torch.export` lifts the TensorRT engine (a `ScriptObject`), it stores it in `exp_program.constants` keyed by its **graph-signature FQN** (`InputSpec.target`) but gives the placeholder node a **mangled name** (an `obj_` prefix). The current code looks it up by the node name/target:

```python
engine_obj = constants.get(engine_node.name) or constants.get(engine_node.target)
```

Those don't equal the FQN key, so the lookup returns `None` and the export aborts. The legacy exporter (`retrace=False`) happened to keep the placeholder name equal to the constants key, so it worked by accident; the modern exporter is the one that breaks.

## The fix

Resolve the placeholder via the canonical `ExportGraphSignature.inputs_to_lifted_custom_objs` mapping (placeholder name -> FQN), then read `constants[fqn]`. Keep the direct name/target lookup only as a fallback for legacy programs that lack that mapping, and unwrap a `FakeScriptObject` to its real object so it can be serialized. The same buggy lookup existed in two places (the save serializer in `_compile.py` and the backend engine-info extractor in `executorch/backend.py`); both now share one helper in `dynamo/_exporter.py`.

## Repro

```python
import torch
import torch_tensorrt

class M(torch.nn.Module):
    def forward(self, x):
        a = (x + 1.0) * 2.0       # TensorRT
        b = torch.cos(a)          # forced off TensorRT
        return (b * 3.0) + 4.0    # TensorRT

ep = torch.export.export(M().eval().cuda(), (torch.randn(2, 3, 4, 4).cuda(),))
    ep, min_block_size=1, torch_executed_ops={torch.ops.aten.cos.default}
)
torch_tensorrt.save(
    gm,
    "m.pte",
    output_format="executorch",
    arg_inputs=(torch.randn(2, 3, 4, 4).cuda(),),
)
```

Before: raises the error above. After: writes a `.pte` containing both the TensorRT and CUDA delegates.

## Tests

Adds CPU-only unit tests for the resolver in `tests/py/dynamo/executorch/test_api.py` (no GPU or ExecuTorch runtime required): signature-FQN resolution, legacy fallback, present-but-incomplete mapping -> `None`, missing -> `None`, and `FakeScriptObject` unwrap.

### DIFF
--- a/py/torch_tensorrt/_compile.py
+++ b/py/torch_tensorrt/_compile.py
@@ -1215,14 +1215,15 @@ def _replace_execute_engine_for_executorch(exp_program: Any) -> Any:
                     f"'{engine_node.target}' not found on graph module"
                 )
         elif engine_node.op == "placeholder":
-            constants = getattr(exp_program, "constants", {})
-            engine_obj = constants.get(engine_node.name) or constants.get(
-                engine_node.target
-            )
+            from torch_tensorrt.dynamo._exporter import _resolve_lifted_custom_obj
+
+            engine_obj = _resolve_lifted_custom_obj(exp_program, engine_node)
             if engine_obj is None:
                 raise RuntimeError(
                     f"execute_engine node '{node.name}': placeholder engine "
-                    f"'{engine_node.name}' not found in exp_program.constants"
+                    f"'{engine_node.name}' did not resolve to a lifted "
+                    f"custom-object constant (available: "
+                    f"{sorted(getattr(exp_program, 'constants', {}) or {})})"
                 )
         else:
             raise RuntimeError(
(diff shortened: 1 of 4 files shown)


cehongwang (approved):
LGTM

PR #4350, opened 2026-06-18
fix: resolve ExecuTorch TRT target_device per partition (coalesced multi-engine)

shoumikhin:
## What's broken
When coalescing TensorRT with other delegates into one `.pte`, the graph has **multiple** TensorRT engines. `TensorRTPartitioner` resolved `target_device` once for the whole program via `_get_engine_info_from_edge_program()`, which requires **exactly one** engine node. With more than one engine it raised, so every TensorRT partition fell back to `cuda:0` with a spurious warning:
```
Could not derive target_device from the TensorRT engine (... expects exactly 1
engine node per partition, found 2); falling back to cuda:0.
```
On a single GPU this is just noise, but a multi-GPU graph cannot label each delegate with its own device.

## The fix
Extract `_get_engine_info_for_node()` (single-node engine-info extraction) out of `_get_engine_info_from_edge_program()` (which keeps its one-engine contract used by `preprocess()`), then resolve `target_device` **per partition** from that partition's own engine node.
- Single-GPU behavior is unchanged (still `cuda:0`), minus the spurious warning.
- Multi-engine / multi-GPU graphs now get a correct per-delegate device label.
- Any extraction failure (no single engine node in the partition, or an unreadable index) still falls back to `cuda:0` rather than aborting the export.

## Test
- Unit tests over the partitioner path in `tests/py/dynamo/executorch/test_partitioner_target_device.py` and `tests/py/dynamo/executorch/test_api.py`: default GPU, non-zero GPU, a coalesced multi-engine graph resolving partition 0 → `cuda:0` and partition 1 → `cuda:1`, the >1-engine-in-one-partition fallback, a malformed-partition fallback, and an explicit `target_device` pin taking precedence.
- Verified on a coalesced TensorRT → Another → TensorRT model (two TRT engines plus one other delegate): the `.pte` still contains both `TensorRTBackend` and `AnotherBackend`, and the "found 2 engines" warning no longer fires.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] My code follows the style guidelines of this project (isort + black)
- [x] I have added tests that prove my fix is effective

### DIFF
--- a/py/torch_tensorrt/executorch/partitioner.py
+++ b/py/torch_tensorrt/executorch/partitioner.py
@@ -12,11 +12,12 @@ from executorch.exir.backend.partitioner import (
 )
 from executorch.exir.backend.utils import tag_constant_data
 from torch.export import ExportedProgram
-from torch.fx.passes.infra.partitioner import CapabilityBasedPartitioner
+from torch.fx.passes.infra.partitioner import CapabilityBasedPartitioner, Partition
 from torch_tensorrt.dynamo.runtime._TorchTensorRTModule import DEVICE_IDX
 from torch_tensorrt.executorch.backend import (
     TensorRTBackend,
-    _get_engine_info_from_edge_program,
+    _get_engine_info_for_node,
+    _get_engine_nodes_in,
     _parse_device_id,
 )
 from torch_tensorrt.executorch.operator_support import TensorRTOperatorSupport
@@ -66,8 +67,8 @@ class TensorRTPartitioner(Partitioner):  # type: ignore[misc]
         # Mirror CudaPartitioner: a target_device CompileSpec drives ExecuTorch's
         # PropagateDevicePass, which tags delegate I/O TensorSpecs with the device
         # and serializes it into the .pte's extra_tensor_info. When the caller pins
-        # it we use that verbatim; otherwise it is derived per export from the
-        # engine's real device in partition() (engine nodes are not available here)
+        # it we use that verbatim; otherwise each partition's device is derived from
+        # its own engine node in partition() (engine nodes are not available here)
         # so a cuda:N engine is not mislabeled cuda:0.
         self._has_explicit_target_device = any(
             s.key == _TARGET_DEVICE_COMPILE_SPEC_KEY for s in self.compile_specs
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/executorch/partitioner.py
cehongwang, line 72: The constructor comment still describes the old whole‑program behavior ("derived per export … in partition()"), which now reads as per‑partition

cehongwang:
The existing `test_partitioner_target_device.py` is not updated and will fail under the new per‑partition logic. `tests/py/dynamo/executorch/test_partitioner_target_device.py`

shoumikhin:
Good catch, fixed in the latest commit. Updated `tests/py/dynamo/executorch/test_partitioner_target_device.py` for the per-partition resolution: the fake partitioner now puts each engine node in its own partition, so the real `_get_engine_info_for_node` path runs per partition. Added a coalesced multi-engine test asserting partition 0 -> cuda:0 and partition 1 -> cuda:1, and a >1-engine-in-one-partition fallback case. The default / nonzero-GPU / explicit-pin / malformed-graph cases still pass. PTAL.

py/torch_tensorrt/executorch/partitioner.py
shoumikhin: Fixed - the constructor comment now describes the per-partition resolution (each partition derives its device from its own engine node in partition()), matching _resolve_target_device_for_partition. Comment-only change amended into the fix commit; the 20 executorch partitioner/api tests still pass. PTAL.

py/torch_tensorrt/executorch/partitioner.py
```diff
@@ -110,26 +118,26 @@ def partition(self, exported_program: ExportedProgram) -> PartitionResult:
...
-                    CompileSpec(
-                        _TARGET_DEVICE_COMPILE_SPEC_KEY,
-                        self._resolve_target_device(exported_program),
-                    )
-                ],
-            )
-
         partition_tags: Dict[str, DelegationSpec] = {}
         for partition in partition_list:
             tag = f"tensorrt_{partition.id}"
```
cehongwang, line 123: @narendasan Should the naming be consistent with dynamo's format like "run_on_acc" and "run_on_gpu"?

narendasan: Dont think it matters as much for the executorch partitioner. For the other partitioners we need to match on tag

lanluo-nvidia (approved):
lgtm

PR #4355, opened 2026-06-21
feat(dynamo): add target_executorch setting to keep output-allocator ops in PyTorch

shoumikhin:
## Description

Some converters require a TensorRT output allocator because their output shape is
data-dependent (for example `aten.nonzero` and boolean `aten.index.Tensor`). A
TensorRT engine that needs an output allocator cannot be consumed by every downstream
runtime that executes the compiled program (for instance, runtimes that rely on
ahead-of-time static memory planning and cannot size an output whose shape is only
known after the engine runs).

This adds a `fallback_data_dependent_ops` compile setting (default `False`). When
enabled, an operator runs in PyTorch instead of TensorRT iff the converter selected
for that specific node requires an output allocator. When disabled (the default),
behavior is unchanged.

### Details
- The decision is made per node in the partitioner operator-support check
  (`TorchTensorRTOperatorSupport` and `OpSupportTester`), which asks the converter
  registry which converter it would select for the node (honoring
  `capability_validator`). This matters for targets like `aten.index.Tensor` that
  register two converters: boolean indexing requires an output allocator and falls
  back, while ordinary integer gather indexing stays on TensorRT.
- Wired through `compile()` and `cross_compile_for_windows()`; the check runs during
  partitioning, which both entry points reach through `compile_module()`. It is
  intentionally not exposed on `convert_exported_program_to_serialized_trt_engine()`,
  where a single serialized engine cannot contain PyTorch fallbacks.
- Combining `fallback_data_dependent_ops` with `require_full_compilation` raises a
  clear error, since routing ops to PyTorch contradicts full compilation.
- `CompilationSettings.__setstate__` defaults the new field so older pickles load.

### Tests
`tests/py/dynamo/models/test_fallback_data_dependent_ops.py`:
- the setting defaults to `False` and is settable;
- a state missing the field (older pickle) restores to `False`;
- the per-node support decision for `nonzero` (CPU, no GPU needed);
- combining with `require_full_compilation` raises;
- end to end on GPU, a data-dependent op (`nonzero`) falls back to PyTorch.

### Type of change
- [x] New feature (non-breaking change which adds functionality)

### Checklist
- [x] My code follows the style guidelines of this project (isort + black)
- [x] I have added tests that prove my fix/feature works

### DIFF
--- a/py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
+++ b/py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
@@ -53,6 +53,15 @@ class OpSupportTester(ops.OperatorSupportBase):  # type: ignore
                 )
             return False
 
+        if TorchTensorRTOperatorSupport._requires_output_allocator(node):
+            # data-dependent output shape needs a TRT output allocator, which some
+            # runtimes cannot consume; honor the fallback and run the node in PyTorch
+            if not node.is_impure():
+                self.unsupported_operators[node_name] = (
+                    self.unsupported_operators.get(node_name, 0) + 1
+                )
+            return False
+
         if (
             (node in CONVERTERS or node.op == "get_attr")
             and node_name not in self.torch_executed_ops
--- a/py/torch_tensorrt/dynamo/partitioning/_global_partitioner.py
+++ b/py/torch_tensorrt/dynamo/partitioning/_global_partitioner.py
@@ -166,6 +166,18 @@ class TorchTensorRTOperatorSupport(OperatorSupport):  # type: ignore[misc]
                 return True
         return False
 
+    @staticmethod
+    def _requires_output_allocator(node: torch.fx.Node) -> bool:
+        # A converter that needs a TRT output allocator has a data-dependent output
+        # shape; route the node to PyTorch when fallback_data_dependent_ops is set.
+        settings = CONVERTERS.compilation_settings
+        if settings is None or not settings.fallback_data_dependent_ops:
+            return False
+        converter_packet = CONVERTERS.get(node)
+        return converter_packet is not None and converter_packet[2].get(
+            "requires_output_allocator", False
+        )
+
     def is_node_supported(
         self, submodules: Mapping[str, torch.nn.Module], node: torch.fx.Node
     ) -> bool:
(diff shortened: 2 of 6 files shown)


narendasan:
@shoumikhin why can't downstream operators consume output allocated outputs? 

Also name wise, I'd rather a specific name like `fallback_data_dependent_ops` so it could be used for other use cases naturally

shoumikhin:
Thanks, good questions.

> why can't downstream operators consume output allocated outputs?

It's the production side, not consumption. For data-dependent ops like `nonzero` / boolean `index.Tensor` the output shape isn't known until the engine runs: TensorRT calls back into `reallocateOutputAsync` during `enqueueV3`, and the true shape is only readable afterward via `notifyShape` (the runtime is `set_output_allocator` -> `execute_async_v3` -> read shape -> reshape). The blocker is downstream runtimes that use ahead-of-time static memory planning: every output tensor gets a fixed buffer sized from a concrete shape at plan time, so there is no way for the embedded engine to install an output allocator, learn the post-`enqueue` shape, and hand back a runtime-sized buffer. torch-tensorrt already hard-rejects output-allocator engines on the ExecuTorch export path for this reason; this setting just turns that hard failure into a graceful per-op fallback.

> name wise, I'd rather a specific name like `fallback_data_dependent_ops`

Done, renamed to `fallback_data_dependent_ops`. I also made the routing per-node rather than per-target: it keys on the converter actually selected for each node (its `requires_output_allocator`), honoring `capability_validator`, so boolean `index.Tensor` falls back while ordinary integer gather `index.Tensor` stays on TensorRT. PR description updated to match.

narendasan (approved):
LGTM, Thanks!

py/torch_tensorrt/dynamo/partitioning/_global_partitioner.py
```diff
@@ -166,6 +166,18 @@ def _dtype(n: torch.fx.Node) -> Optional[torch.dtype]:
                 return True
         return False
 
+    @staticmethod
+    def _requires_output_allocator(node: torch.fx.Node) -> bool:
```
cehongwang, line 170: This naming is confusing. It returns False for a node that genuinely does require an output allocator whenever fallback_data_dependent_ops is off. Consider renaming it to `_falls_back_for_output_allocator(node)` or something else

py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
```diff
@@ -53,6 +53,15 @@ def is_node_supported(
                 )
             return False
 
+        if TorchTensorRTOperatorSupport._requires_output_allocator(node):
+            # data-dependent output shape needs a TRT output allocator, which some
+            # runtimes cannot consume; honor the fallback and run the node in PyTorch
+            if not node.is_impure():
+                self.unsupported_operators[node_name] = (
```
cehongwang, line 60: what happens when the node is impure?

cehongwang (changes requested):
OVerall good. Just one minor comment

PR #4358, opened 2026-06-22
MD-TRT changes for release 2.13

apbose:
- Enable MULTIDEVICE_RUNTIME_10_16 preview feature in the builder config (guarded on native_trt_collectives + hasattr) so buildEngineWithConfig accepts networks with IDistCollectiveLayer. Reqd for TRT version greater than 10.16 and less than 11
- core/runtime/BUILD: switch USE_C10D_NCCL from copts to defines so the macro propagates to dependent targets.
- Demote three non-fatal, auto-fired TRT-LLM fallback logs (warning → info); explicit opt-in load/init failures stay at warning.

### DIFF
--- a/core/runtime/BUILD
+++ b/core/runtime/BUILD
@@ -104,8 +104,7 @@ cc_library(
         "TensorRTBindingNames.h",
         "runtime.h",
     ],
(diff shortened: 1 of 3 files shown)


narendasan (approved):
LGTM

PR #4363, opened 2026-06-24
feat: multiple optimization profiles for disjoint input shape regimes

cehongwang:
Add support for defining N optimization profiles at compile time via the list-based ``Input.profiles`` API and selecting the active profile at runtime (manual pin by index, or opt-in shape-based auto-selection).

- AOT (torch.export) compile path builds one TRT optimization profile per declared profile index; submodules inherit the profile count via propagation across graph breaks.
- Python and C++ runtimes expose a matching primitive engine API (set_active_profile / num_optimization_profiles / _active_profile_index / _auto_select_profiles) so the two runtimes remain interchangeable.
- Profile selection is exposed through the optimization_profile context manager; auto-selection uses lazy (first-fitting) profile selection.
- Backward compatible: engines without declared profiles keep the historical single-profile (dynamic) / no-profile (static) behavior.

Includes an example and runtime tests covering dynamic submodule inputs.

# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/docsrc/tutorials/runtime_opt/multi_optimization_profiles.rst
+++ b/docsrc/tutorials/runtime_opt/multi_optimization_profiles.rst
@@ -25,7 +25,7 @@ profile is tuned independently.
 Declaring profiles
 ------------------
 
-Pass an ordered list of ``{"min", "opt", "max"}`` dicts to
+Pass an ordered list of ``{"min_shape", "opt_shape", "max_shape"}`` dicts to
 :class:`torch_tensorrt.Input` via ``profiles``. The **list index** is the
 optimization-profile index you select at runtime.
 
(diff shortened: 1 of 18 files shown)


docsrc/tutorials/runtime_opt/multi_optimization_profiles.rst
```diff
@@ -0,0 +1,164 @@
...
+        dtype=torch.int64,
+        profiles=[
+            # index 0 -> decode: seq pinned to 1 (a fully static profile)
+            {"min": (1, 1), "opt": (1, 1), "max": (1, 1)},
+            # index 1 -> prefill: seq in [1, 512], tuned at 256
+            {"min": (1, 1), "opt": (1, 256), "max": (1, 512)},
+        ],
+    )
+
+``profiles`` is mutually exclusive with the single-shape ``min_shape`` /
```
narendasan: What do you mean by mutually exclusive? Like only use one? We should use the same naming like `min_shape` instead of `min` and `opt_shape` instead of `opt`

narendasan (commented):
Cool this is looking good, have @apbose do a pass as well then we can merge. 

Can we also add this to the executorch rt? (different PR)

narendasan (approved):
LGTM

PR #4362, opened 2026-06-24
fix(runtime): autosave engine-implicit RuntimeCache via atexit + weakref

tp5uiuc:
## Summary
Original bug from #4359: engine-implicit `RuntimeCache` handles relied solely on `__del__` to autosave at process exit. When the engine survived to exit, `__del__` fired *during* shutdown and the torchbind / `filelock` lazy imports raised `ImportError: sys.meta_path is None`, surfacing as `Exception ignored in __del__` once per surviving handle.

### Fixes

- Move the autosave anchor to `atexit.register(partial(_autosave_at_exit,
  weakref.ref(self)))`. `atexit` fires before module teardown; `__del__`
  stays as the mid-program GC path. Both flip `autosave_on_del` off so
  whichever runs first wins. `__del__` unregisters its token to keep the
  atexit registry from accumulating dead entries.
- Module-level helper closing over a `weakref` keeps the registration
  non-owning, so handles can still die mid-program normally.
- `RuntimeCache.__getstate__` / `__setstate__` strip the unpicklable
  `weakref` from the pickle stream and re-register a fresh atexit hook
  on unpickle when `autosave_on_del` was on. `__del__` reads the token
  via `getattr(..., None)` so `copy.deepcopy` partial-init edge cases
  stay silent.

## Dependency on a sibling PR
This PR introduces an `_atexit_token` slot (`partial(..., weakref.ref(self))`) on `RuntimeCache`. CI tests that pickle compiled modules (`test_mutable_torchtrt_module::test_save`, `test_cross_runtime_serde::test_save_*`) walk through `_runtime_settings.runtime_cache` and would hit the weakref. The matching wrapper-side pickle exclusion lives in **#4368** and should land first to unblock CI here.

Fixes #4359.

## Type of change
- Bug fix (non-breaking)

## Test plan
`TestRuntimeCacheAutosave` in `tests/py/dynamo/runtime/test_000_runtime_cache.py`:
- [x] `test_del_swallows_shutdown_import_error_on_path` — `sys.unraisablehook` sees no leak
- [x] `test_atexit_hook_saves_via_weakref` — helper saves + flips flag
- [x] `test_atexit_hook_no_op_on_dead_weakref` — dead weakref no-ops cleanly
- [x] `test_atexit_token_unregistered_after_del` — `atexit.unregister` called with the registered token
[...]

Issue #4359, opened 2026-06-23
🐛 [Bug] Serialization fails in Runtime Cache


cehongwang:
Running the following script on commit `a094c74395cd2ab50621c866f25a05f761b4931c`:

```
from pathlib import Path

import torch
import torch_tensorrt as torchtrt
import torchvision.models as models

torch.manual_seed(0)
torch.cuda.manual_seed_all(0)

artifact = Path(__file__).resolve().parent / "trt_mod.pt2"


def main():
    model = models.resnet50(pretrained=False).eval().to("cuda")

    inputs = torch.randn(1, 3, 224, 224, device="cuda")

    exported = torch.export.export(model, (inputs,))

    trt_mod = torchtrt.dynamo.compile(
        exported,
        inputs=[inputs],
        min_block_size=1,
        immutable_weights=False,
        reuse_cached_engines=False,
        enabled_precisions={torch.float32},
        torch_executed_ops={torch.ops.aten.relu.default},
    )

    with torch.no_grad():
        results1 = trt_mod(inputs)

    torchtrt.save(
        trt_mod,
        str(artifact),
        output_format="exported_program",
        inputs=[inputs],
        retrace=False,
    )

    trt_ep = torch.export.load(str(artifact))

    # Inspect loaded engine types
    gm = trt_ep.module()

    with torch.no_grad():
        results2 = gm(inputs)

    print("max diff:", (results1 - results2).abs().max().item())
    # for name, val in gm._modules.items():
    #     print(f"  submodule {name}: {type(val).__name__}")
    # for attr_name in dir(gm):
    #     obj = getattr(gm, attr_name, None)
    #     if obj is not None and "engine" in attr_name.lower():
    #         print(f"  attr {attr_name}: {type(obj)} is_script_obj={isinstance(obj, torch._C.ScriptObject)}")
    # for fqn, const in trt_ep.constants.items():
    #     print(f"  constant {fqn}: {type(const)} is_script_obj={isinstance(const, torch._C.ScriptObject)}")
    # for node in gm.graph.nodes:
    #     if node.op == "get_attr":
    #         val = getattr(gm, node.target, None)
    #         print(f"  graph get_attr '{node.target}': {type(val)}")


if __name__ == "__main__":
    main()


```

Gives the error:

```
WARNING: [Torch-TensorRT - Debug Build] - RuntimeCacheHandle::serialize() invoked on a non-RTX build; returning empty bytes.
[...]

Issue #4310, opened 2026-05-29
Move TRT-RTX runtime mode controls from CompilationSettings to runtime context managers


tp5uiuc:
## Background

Per [@narendasan's review comments](https://github.com/pytorch/TensorRT/pull/4237#discussion_) on #4237, several settings that were added to `CompilationSettings` for the TensorRT-RTX runtime features look more like **runtime mode controls** than properties of the compiled engine, and shouldn't be pinned at compile time or serialized into the engine info.

> "Runtime mode controls should be controlled via context managers rather than passed in at compile time. Only information that is fixed at runtime needs to be here."
>
> "Also are these properties of the engine or are they runtime mode configurations? The point of this interface is the bare minimum information to reconstruct the program from disk"

The same principle is what `torch_tensorrt.runtime.set_cudagraphs_mode(...)` already follows for whole-graph cudagraphs.

## Affected settings

Introduced across #4180, #4184, #4187 and #4237 (and their Python/C++ counterparts):

| Setting | Source | Runtime-mutable? |
|---|---|---|
| `cuda_graph_strategy` | #4187 / #4237 | **Yes** — user may want to flip per call |
| `dynamic_shapes_kernel_specialization_strategy` | #4184 / #4237 | **Yes** — affects per-execution kernel selection |
| `runtime_cache_path` | #4180 / #4237 | **Borderline** — pinned once, but is a runtime concern, not an engine property |

Today these live on `CompilationSettings`, are threaded through `_compiler.compile(...)` and friends, and get packed into the serialized engine tuple alongside true engine properties (binding names, target platform, etc.).

## Proposed design

Move these to runtime context managers, mirroring `torch_tensorrt.runtime.set_cudagraphs_mode`:

```python
with torch_tensorrt.runtime.cuda_graph_strategy("whole_graph_capture"):
    compiled_module(*inputs)

with torch_tensorrt.runtime.dynamic_shapes_kernel_strategy("eager"):
    compiled_module(*inputs)

torch_tensorrt.runtime.set_runtime_cache_path("/tmp/trt_rtx.cache")
```

The C++ engine reads these from runtime state when constructing `IRuntimeConfig`, not from the serialized engine info.

Implications:
- Remove these fields from `CompilationSettings`, `_compiler.compile(...)`, `_compiler.cross_compile_for_windows(...)`, and `_compiler.convert_exported_program_to_serialized_trt_engine(...)`.
[...]

PR #4368, opened 2026-06-26
fix(runtime): RuntimeCache pickle path — wrapper exclusion + handle bytes round-trip

tp5uiuc:
## Summary
This MR adds three independent pickle-correctness fixes that together let `torch.save(module)` and standalone-pickle of `RuntimeCache` / `RuntimeCacheHandle` work cleanly.

- **Wrapper `__getstate__` / `__setstate__`** on `TorchTensorRTModule` excludes `_runtime_settings` and `_implicit_cache_handle` from the pickle stream and resets both to defaults on load. Mirrors `set_extra_state` (line 515) and the line-236 design note: *"RuntimeSettings are intentionally NOT serialized: they're per-engine, in-memory init values, not part of the engine's identity."* Pickle path now matches `state_dict` + `load_state_dict` behavior.
- **Python `_RuntimeCacheHandle.__getstate__` / `__setstate__`** ships a protocol pair pickling as `(path, bytes)`. Default pickle crashes on `_thread.lock`; the new protocol carries the cache contents (live cache when materialized, `_pending_warm_bytes` fallback otherwise) and rebuilds a fresh `threading.Lock` on the load side. Factored a lock-free `_serialize_unlocked` helper.
- **Cpp `RuntimeCacheHandle::serialize()`** falls back to `pending_warm_bytes_` when `trt_handle_` is null, so callers asking for the handle's persistable bytes (`save_to_stream`, `def_pickle`) get the right answer in every lifecycle state. The misleading "called before materialized" warning is retired now that the pre-materialize case hooks into serializable bytes.
- As a result **Cpp `def_pickle`** moves from `std::string` (path-only) to `std::tuple<std::string, at::Tensor>` (path + bytes) and round-trips via the existing `serialize` / `deserialize` API. Matches the python facade's shape.

## Pre-existing CI noise this resolves
- `test_cross_runtime_serde::test_save_{cpp_load_python,python_load_python,python_load_cpp}` : root cause is `_thread.lock` failing pickle in python-runtime subprocesses. Fixed by the `_RuntimeCacheHandle.__getstate__` protocol.
- Sibling PR (#4362) introduces an `_atexit_token` slot containing a `weakref`. The wrapper exclusion here keeps that out of pickle entirely, unblocking the sibling PR's CI if needed.

Refs #4359.

## Type of change
- Bug fix (non-breaking)

## Test plan
- [x] `test_python_handle_pickle_preserves_pending_warm_bytes` — python
      handle's bytes round-trip end-to-end (path + pending bytes
      preserved; fresh lock; no live cache reused).
- [x] Full `test_000_runtime_cache.py` passes locally (24 passed, 1
      unrelated cpp-rt-only skip).
- [x] `pre-commit run` clean on touched files (mypy skipped for two
      pre-existing errors at `_TorchTensorRTModule.py:380` and `:508`,
      unrelated).

### DIFF
--- a/core/runtime/RuntimeSettings.cpp
+++ b/core/runtime/RuntimeSettings.cpp
@@ -1,5 +1,6 @@
 #include "core/runtime/RuntimeSettings.h"
 
+#include <algorithm>
 #include <array>
 #include <cstring>
 #include <iterator>
@@ -124,16 +125,12 @@ at::Tensor RuntimeCacheHandle::serialize() const {
 #ifdef TRT_MAJOR_RTX
   std::lock_guard<std::mutex> lock(state_mu_);
   if (!trt_handle_) {
-    // Pre-materialize: forward any ``pending_warm_bytes_`` so that the
-    // handle's persistable state survives ``save_to_stream`` and pickle
-    // even before any engine has triggered ``ensure_materialized``.
-    // ``save_to_stream`` / ``def_pickle`` round-trip then matches the
-    // python facade's behavior (it reads ``_pending_warm_bytes`` directly).
-    if (pending_warm_bytes_.empty()) {
+    if (std::empty(pending_warm_bytes_)) {
       return empty();
     }
     auto tensor = at::empty({static_cast<int64_t>(pending_warm_bytes_.size())}, opts);
-    std::memcpy(tensor.data_ptr(), pending_warm_bytes_.data(), pending_warm_bytes_.size());
+    std::copy(
+        std::cbegin(pending_warm_bytes_), std::cend(pending_warm_bytes_), static_cast<uint8_t*>(tensor.data_ptr()));
     return tensor;
   }
   auto host_mem = make_trt(trt_handle_->serialize());
--- a/core/runtime/RuntimeSettings.cpp
+++ b/core/runtime/RuntimeSettings.cpp
@@ -124,9 +124,17 @@ at::Tensor RuntimeCacheHandle::serialize() const {
 #ifdef TRT_MAJOR_RTX
   std::lock_guard<std::mutex> lock(state_mu_);
   if (!trt_handle_) {
-    LOG_WARNING(
-        "RuntimeCacheHandle::serialize() called before the IRuntimeCache was materialized; returning empty bytes.");
-    return empty();
+    // Pre-materialize: forward any ``pending_warm_bytes_`` so that the
+    // handle's persistable state survives ``save_to_stream`` and pickle
+    // even before any engine has triggered ``ensure_materialized``.
+    // ``save_to_stream`` / ``def_pickle`` round-trip then matches the
+    // python facade's behavior (it reads ``_pending_warm_bytes`` directly).
+    if (pending_warm_bytes_.empty()) {
+      return empty();
+    }
+    auto tensor = at::empty({static_cast<int64_t>(pending_warm_bytes_.size())}, opts);
+    std::memcpy(tensor.data_ptr(), pending_warm_bytes_.data(), pending_warm_bytes_.size());
+    return tensor;
   }
   auto host_mem = make_trt(trt_handle_->serialize());
   TORCHTRT_CHECK(host_mem, "IRuntimeCache::serialize() returned null host memory; cannot serialize cache bytes.");
--- a/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
+++ b/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
@@ -662,6 +662,37 @@ class TorchTensorRTModule(torch.nn.Module):  # type: ignore[misc]
         engine._auto_select_profiles = auto
         engine.set_active_profile(active)
 
+    def __getstate__(self) -> dict[str, Any]:
+        """Exclude per-engine, in-memory state from the pickle stream.
+
+        Mirrors the ``set_extra_state`` reset so that
+        ``torch.save(module)`` / ``torch.load`` behaves the same way as
+        ``state_dict`` / ``load_state_dict`` w.r.t. ``RuntimeSettings``:
+        the caller must reapply any ``runtime_cache`` / strategy / cuda-graph
+        configuration after load. See ``_pack_engine_info`` for the matching
+        cpp-side exclusion (engine bytes never carry these fields).
+
+        ``_implicit_cache_handle`` is dropped alongside ``_runtime_settings``
+        because it aliases the same ``RuntimeCache`` instance and would
+        otherwise drag a ``weakref`` (via the handle's ``atexit`` closure)
+        and a Python-only ``threading.Lock`` (when the python-runtime path
+        is active) into pickle -- neither is picklable.
+        """
+        get_state = getattr(super(), "__getstate__", None)
+        state = (get_state() if get_state else self.__dict__).copy()
+        state.pop("_runtime_settings", None)
+        state.pop("_implicit_cache_handle", None)
+        return state
+
+    def __setstate__(self, state: dict[str, Any]) -> None:
+        state.setdefault("_runtime_settings", RuntimeSettings())
+        state.setdefault("_implicit_cache_handle", None)
+        set_state = getattr(super(), "__setstate__", None)
+        if set_state is not None:
+            set_state(state)
+        else:
+            self.__dict__.update(state)
+
     def set_pre_allocated_outputs(self, enable: bool) -> None:
         self.get_engine().use_pre_allocated_outputs = enable
 
--- a/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
+++ b/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
@@ -673,10 +673,7 @@ class TorchTensorRTModule(torch.nn.Module):  # type: ignore[misc]
         cpp-side exclusion (engine bytes never carry these fields).
 
         ``_implicit_cache_handle`` is dropped alongside ``_runtime_settings``
-        because it aliases the same ``RuntimeCache`` instance and would
-        otherwise drag a ``weakref`` (via the handle's ``atexit`` closure)
-        and a Python-only ``threading.Lock`` (when the python-runtime path
-        is active) into pickle -- neither is picklable.
+        because it aliases the same ``RuntimeCache`` instance.
         """
         get_state = getattr(super(), "__getstate__", None)
         state = (get_state() if get_state else self.__dict__).copy()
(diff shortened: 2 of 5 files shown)


core/runtime/RuntimeSettings.cpp
```diff
@@ -124,9 +124,17 @@ at::Tensor RuntimeCacheHandle::serialize() const {
...
+    // Pre-materialize: forward any ``pending_warm_bytes_`` so that the
+    // handle's persistable state survives ``save_to_stream`` and pickle
+    // even before any engine has triggered ``ensure_materialized``.
+    // ``save_to_stream`` / ``def_pickle`` round-trip then matches the
+    // python facade's behavior (it reads ``_pending_warm_bytes`` directly).
+    if (pending_warm_bytes_.empty()) {
+      return empty();
+    }
+    auto tensor = at::empty({static_cast<int64_t>(pending_warm_bytes_.size())}, opts);
+    std::memcpy(tensor.data_ptr(), pending_warm_bytes_.data(), pending_warm_bytes_.size());
```
tp5uiuc: Prefer `std::copy` instead of `std::memcpy` : copy should internally forward to memcpy.

core/runtime/RuntimeSettings.cpp
```diff
@@ -124,9 +124,17 @@ at::Tensor RuntimeCacheHandle::serialize() const {
...
   std::lock_guard<std::mutex> lock(state_mu_);
   if (!trt_handle_) {
-    LOG_WARNING(
-        "RuntimeCacheHandle::serialize() called before the IRuntimeCache was materialized; returning empty bytes.");
-    return empty();
+    // Pre-materialize: forward any ``pending_warm_bytes_`` so that the
+    // handle's persistable state survives ``save_to_stream`` and pickle
+    // even before any engine has triggered ``ensure_materialized``.
+    // ``save_to_stream`` / ``def_pickle`` round-trip then matches the
+    // python facade's behavior (it reads ``_pending_warm_bytes`` directly).
```
tp5uiuc: Remove comment.

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
```diff
@@ -536,6 +536,37 @@ def set_extra_state(self, state: SerializedTorchTensorRTModuleFmt) -> None:
...
+        the caller must reapply any ``runtime_cache`` / strategy / cuda-graph
+        configuration after load. See ``_pack_engine_info`` for the matching
+        cpp-side exclusion (engine bytes never carry these fields).
+
+        ``_implicit_cache_handle`` is dropped alongside ``_runtime_settings``
+        because it aliases the same ``RuntimeCache`` instance and would
+        otherwise drag a ``weakref`` (via the handle's ``atexit`` closure)
+        and a Python-only ``threading.Lock`` (when the python-runtime path
+        is active) into pickle -- neither is picklable.
+        """
```
tp5uiuc, line 677: Remove starting from `would ...`

tests/py/dynamo/runtime/test_000_runtime_cache.py
```diff
@@ -334,6 +334,32 @@ def test_user_built_handle_no_autosave_by_default(self):
                 "User-built handle with autosave_on_del=False should not save on GC",
             )
 
+    @unittest.skipIf(
+        ENABLED_FEATURES.torch_tensorrt_runtime,
+        "exercises the python ``_RuntimeCacheHandle`` directly; cpp-rt "
+        "path is covered by ``register_jit_hooks.cpp`` ``def_pickle``",
+    )
```
tp5uiuc: Why do we skip C++ here?

core/runtime/RuntimeSettings.cpp
tp5uiuc: [by Claude Code] Done in 1ad40e3bf — switched to `std::copy` with free-function `std::cbegin`/`std::cend`, and also moved the empty check to free-function `std::empty`.

core/runtime/RuntimeSettings.cpp
tp5uiuc: [by Claude Code] Done in 1ad40e3bf — comment block dropped.

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
tp5uiuc: [by Claude Code] Done in 1ad40e3bf — docstring trimmed; dropped the rationale that named specific picklability hazards.

tests/py/dynamo/runtime/test_000_runtime_cache.py
tp5uiuc: [by Claude Code] Good catch — the skip was over-cautious. The test instantiates `_RuntimeCacheHandle` directly (not via `RuntimeCache._handle` selection), so the python class is exercisable regardless of which runtime is active. Skip removed in 1ad40e3bf; docstring updated to note the directness.

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
narendasan, line 1: I am still trying to figure out if RuntimeCache is runtime specific or engine specific. If its not engine specific then the module should not serialize. The module should be 1:1 to the engine

tp5uiuc:
[by Claude Code] **CI status after rebase on #4367:**

- Wheel builds (Linux x86_64, SBSA, Windows, RTX, Python-only) all green now that #4367 fixed the `CUDAGraph::debug_dump` API mismatch. The new cpp `serialize()` fallback and the `def_pickle(path, bytes)` tuple compile and link cleanly across all wheel variants.
- `Python-only dynamo runtime tests` summary: `82 passed, 4 failed, 81 skipped` — the 4 fails are all `test_004_weight_streaming` (pre-existing flake on every PR). `test_mutable_torchtrt_module::test_save` is in the passing set.
- The python-handle pickle test (`test_python_handle_pickle_preserves_pending_warm_bytes`) runs and passes.

Other red is environmental, identical to #4362 and to main:
- `L0 dynamo converter` — torch-nightly SDPA API change in an unrelated test
- `L0 torchscript` — pre-existing Windows torchscript flakes
- `RTX Python-only dynamo runtime` — same `test_004_weight_streaming`

py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
tp5uiuc: I see your point Naren. `RuntimeCache` is a runtime-config knob and is not engine specific and so not a part of the engine's identity. The same instance can be 1:N with engines: that's why we have an in-memory runtime cache object that the users can attach and share across multiple engines.

Also, since the `TorchTensorRTModule` wrapper is 1:1 with an engine, we should not serialize runtime cache, and so TorchTensorRTModule.__getstate__ drops _runtime_settings (and _implicit_cache_handle, which aliases the same RuntimeCache) on pickle. 

This is consistent with the two serialization paths toda:

- _pack_engine_info (line 236 (https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py#L236)): "RuntimeSettings are intentionally NOT serialized: they're per-engine, in-memory init values, not part of the engine's identity (see #4310)."
- set_extra_state (line 515 (https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py#L515)): explicitly resets _runtime_settings = RuntimeSettings() on load — "Caller can reapply via mod.runtime_settings = ...".

This MR ensures that the pickle path also adheres to the same settings. I have added __getstate__ and __setstate__ for the RuntimeCache in case users want to pickle that object separately, independent from the torch module. Hope that makes sense, happy to sync up offline in case more info needed.

tp5uiuc:
[by Claude Code] **CI on f547d507d (rebased onto main after #4362 merged):**

- Wheel builds (Linux x86_64 cu130/cu132, Windows, RTX Linux/Windows, SBSA) all green — the earlier GitLab/eigen flake cleared.
- `test_000_runtime_cache.py`: all 5 new `TestRuntimeCacheAutosave` tests (landed via #4362) + `test_python_handle_pickle_preserves_pending_warm_bytes` (this PR) pass. Pytest summary on the RTX runtime target: `1 failed, 138 passed, 52 skipped` — the 1 failure is unrelated (see below).
- No conflicts remaining; wrapper `__getstate__` docstring's stale `line 515` reference to `set_extra_state` was dropped since line numbers shifted after #4362 merged.

Remaining red is environmental / pre-existing on main:

| Failing job | Cause |
| --- | --- |
| RTX Python-only dynamo runtime (cu130, cu132 ×2) | `test_multi_optimization_profiles::test_profiles_restored_from_trt_api_without_metadata` — Myelin *"Must be called with an execution graph cache"*. Also fails on main (verified against 97defd77). |
| Build Jetpack (cu126 aarch64) | `CUDAGraph::has_graph_exec` accessibility error — Jetpack path did not pick up #4367's fix; fails on main. |
| L0 dynamo converter (cu130) | torch-nightly SDPA API change — unrelated. |
| L0 torchscript (cu130) | Pre-existing Windows torchscript flakes — unrelated. |

Nothing this PR modifies is on any failure list.

cehongwang (approved):
Looks good to me

tp5uiuc:
**CI triage — none of the 18 failures relate to this PR's changes.** The MR's own test (`test_000_runtime_cache.py`) passes; every failure is infra, a flake, or pre-existing.

| # | Category | Failing jobs | Detail |
|---|----------|-------------|--------|
| 10 | **CI infra** | Windows + RTX-Windows L1 (dynamo compile/core, torch compile, torchscript) | `Unable to download artifact(s): (404) workflow run not found` → no JUnit XML. Never ran a test. |
| 3 | **Numerical flake** | `core / L1 dynamo compile`, `core / L1 torch compile` (+ `CI / Linux x86_64` rollup) | `test_hf_gqa_model.py` — 1/9.7M elems off, abs diff 0.207 vs 0.2 tol. |
| 4 | **RTX env** | `RTX Python-only dynamo runtime` (cu130/cu132) | `test_aliased_io.py::test_in_place_write_through` — "Failed to create execution context" (RTX-only, aliased-IO path, not runtime-cache). |
| 1 | **Pre-existing build break** | `Build Jetpack aarch64` | `TRTEngine.cpp:298: 'ICudaEngine' has no member 'getAliasedInputTensor'` — Jetpack's older TRT lacks the aliased-IO API. Untouched file. |

None of the failures touch the 5 files changed here (`RuntimeSettings.cpp`, `register_jit_hooks.cpp`, `_TorchTensorRTModule.py`, `_runtime_cache.py`, `test_000_runtime_cache.py`).

PR #4369, opened 2026-06-26
refactor(dynamo): make _requires_output_allocator a pure predicate

cehongwang:
## Description

Follow-up cleanup to #4355.

The `_requires_output_allocator` helper in the partitioner previously folded the `fallback_data_dependent_ops` setting into its return value: it returned `False` for a node that genuinely required a TRT output allocator whenever the setting was disabled (the default). That made the name misleading at the call sites — `_requires_output_allocator(node) == False` did not actually mean "this node does not require an output allocator," it usually meant "the setting is off."

This flips the semantics so the helper is a **pure predicate**: it reports whether the converter selected for a node needs a TRT output allocator (data-dependent output shape, e.g. `nonzero` or boolean `index.Tensor`), independent of any setting. The `fallback_data_dependent_ops` gate moves to the two operator-support call sites (`TorchTensorRTOperatorSupport.is_node_supported` and `OpSupportTester.is_node_supported`).

The setting is still checked first at the call sites, preserving the short-circuit that skips the converter-registry lookup in the default (off) case. Behavior is unchanged end to end.

## Details

- `_global_partitioner.py`: `_requires_output_allocator` no longer reads `compilation_settings`; the caller now guards with `settings is not None and settings.fallback_data_dependent_ops and self._requires_output_allocator(node)`.
- `_adjacency_partitioner.py`: same gate added at the `OpSupportTester` call site.

## Tests

`tests/py/dynamo/models/test_fallback_data_dependent_ops.py`:

- Added `test_requires_output_allocator_is_setting_independent`: the pure predicate returns `True` for `nonzero` regardless of the setting.
- Updated `test_output_allocator_node_falls_back_only_when_enabled`: now drives `is_node_supported(...)` and asserts the node stays on TensorRT (`True`) when the flag is off and falls back (`False`) when it is on — testing the real partitioner decision rather than the helper in isolation.

## Type of change

- Refactor / code-quality improvement (non-breaking, no behavior change)

## Checklist

- [x] My code follows the style guidelines of this project (isort + black, passed via pre-commit)
- [x] I have added/updated tests

### DIFF
--- a/py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
+++ b/py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
@@ -53,7 +53,12 @@ class OpSupportTester(ops.OperatorSupportBase):  # type: ignore
                 )
             return False
 
-        if TorchTensorRTOperatorSupport._requires_output_allocator(node):
+        settings = CONVERTERS.compilation_settings
+        if (
+            settings is not None
+            and settings.fallback_data_dependent_ops
+            and TorchTensorRTOperatorSupport._requires_output_allocator(node)
+        ):
             # data-dependent output shape needs a TRT output allocator, which some
             # runtimes cannot consume; honor the fallback and run the node in PyTorch
             if not node.is_impure():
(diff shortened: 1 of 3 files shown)


narendasan (approved):
LGTM

PR #4380, opened 2026-07-04
dynamo: route native NCCL collectives at the op's process (sub)group

pkisfaludi-nv:
The native collective converters build every collective as if it spanned the whole world: the rank
array passed to `addDistCollective` is always `arange(world_size)` and `num_ranks` is `world_size`.
On a mesh with more than one dimension that is wrong — a tensor-parallel all_reduce reduces across
every rank, mixing data from separate CP groups, and nothing raises.

This PR resolves the op's `group_name` and passes that group's global ranks, with `num_ranks` set to
that group's size. It is the torch-tensorrt half of TensorRT MR !49040 (merged), which added
`numRanks` and the per-collective rank list to the builder API; without it the TRT-side feature is
unreachable from Dynamo on a multi-dimensional mesh.

**#4388 is closed in favour of this PR.** What it carried is accounted for below.

### Verified on hardware

8× A100-80GB, torch 2.15.0.dev, source build with `ENABLE_TRT_NCCL_COLLECTIVES`.

**The defect.** `num_ranks` was the world size, so a rank-dependent collective on a subgroup
claims an output the group cannot produce. all_gather over a 4-rank subgroup, input `(2,4)`:

| | output shape |
|---|---|
| `num_ranks = world_size = 8` (main) | **(16, 4)** — twice the group's actual output |
| `num_ranks = group size = 4` (this PR) | **(8, 4)** — correct |

`all_reduce` is shape-preserving, which is why the existing tests never caught it.

**Why global rank IDs.** `setCommunicator` installs one communicator that *"must be uniform
across all multi-device instances"*, and each layer names its group as a **subset** of it via
`groups`. That is what lets a single engine carry collectives on two different groups — a
CP × TP mesh. Group-local IDs would only work when the bound communicator *is* the group,
which forecloses the mesh case.

| every rank binding the same 8-rank communicator | TRT 11.2.0.86-md-moe-ep | TRT 11.2.1.2 |
|---|---|---|
| subset `[0,1,2,3]` | **10** to its members | 36 to everyone — array ignored |
| subset `[4,5,6,7]` | **26** to its members | fails — *"child communicator"* |
| **CP4 × TP2 in one engine** | **360 on every rank** | fails |

Subset selection works on the md-moe-ep drop and is broken on 11.2.1.2, so the 4-GPU mesh
test is gated behind `TORCHTRT_TEST_COLLECTIVE_SUBGROUPS=1` rather than failing CI over a
TensorRT limitation.

**Tests**

| | |
|---|---|
| CPU-only converter tests | 5 passed, 6 subtests |
| `TestMultirankNccl` (2 GPU, existing suite) | 14 passed |
| `TestMultirankNccl4GPU` mesh routing (4 GPU) | passes on md-moe-ep; skips with a reason on 11.2.1.2 |

### What from #4388 is *not* here, and why

**The Myelin cast boundary — dropped as obsolete.** It worked around
`CollectiveOperation not supported for collective In toMyelinCommKind`, which existed because
`toMyelinCommKind()` had no arm for `kALL_TO_ALL`. TRT `0ea60bdc42` ("Fix BF16 A2A Myelin Conversion
Error") added it; that was the only op missing. On the same hardware:

| collective | in the switch | result |
|---|---|---|
| `BROADCAST` | no | `BUILD_FAILED` — `... In toMyelinCommKind at myelinDistCollectiveLayer.cpp:52` |
| `ALL_TO_ALL` bf16, no boundary | yes | `BUILD_OK`, collective absorbed into the Myelin region (`a2a_myl0_5`) |
| `ALL_TO_ALL` bf16, with boundary | yes | `BUILD_OK`, collective standalone (`__myl_MoveAddReluCast_myl0_0 \| a2a \| __myl_CastAdd_myl2_0`) |

`BROADCAST` is the positive control — the harness does reproduce the failure, at the exact line. Row 2
is the case the boundary exists to prevent, and it builds with the collective genuinely fused into the
ForeignNode; row 3 confirms the boundary is what stops that fusion. The workaround was also
**default-on**, widening every collective to fp32 for all users.

**Setting `num_ranks` before `get_output(0)` — dropped as a no-op.** #4388 reordered these two lines
on the grounds that Myelin infers the output shape from `num_ranks`. The shape *is* stale at the
instant of capture (`(2,64)` instead of `(8,64)` for an all_gather), but TensorRT re-evaluates it
lazily, the two lines are adjacent on `main` with nothing reading the shape in between, and the built
engines are byte-identical either way:

| op | `num_ranks` first | `get_output` first | identical |
|---|---|---|---|
| `ALL_GATHER` | `(16,64)` md5 `0ebe6c0b97d4` | `(16,64)` md5 `0ebe6c0b97d4` | yes |
| `ALL_REDUCE` | `(4,64)` md5 `18e50672a18a` | `(4,64)` md5 `18e50672a18a` | yes |
| `ALL_TO_ALL` | `(4,64)` md5 `a0ad8806638d` | `(4,64)` md5 `a0ad8806638d` | yes |
| `REDUCE_SCATTER` | `(1,64)` md5 `547445add902` | `(1,64)` md5 `547445add902` | yes |

The substantive part of that change was `num_ranks = len(groups)` rather than `world_size`, which is
kept here — it is inseparable from subgroup routing.

**`all_gather_via_all_reduce` — not carried over.** It depends on the cast boundary, is default-off,
has no tests, and hard-raises on dynamic shapes, contradicting #4402 marking these converters
`supports_dynamic_shapes=True`. Preserved at `8e48830121` if anyone wants it.

### Tests

- `TestCollectiveGroupRanks` — group resolution and order preservation. CPU-only.
- `TestNativeCollectiveNumRanks` — `num_ranks` is the group size, not the world size, for all six
  converters. CPU-only.
- `test_two_dimensional_mesh_routing` — 4-GPU, CP `[0,2] [1,3]` and TP `[0,1] [2,3]` (the !49040
  topology), asserting each collective routes to its own subgroup. No existing test would have
  caught a routing regression — `_multirank_distributed_mode_subgroup` builds its "subgroup" as
  `list(range(world_size))`, i.e. all ranks, where a subgroup and the world group are numerically
  indistinguishable.

  It lives in its own `TestMultirankNccl4GPU` class because `world_size` is fixed per class and
  `TestMultirankNccl` runs at 2. At 2 ranks the CP and TP groups collapse onto the same pair, so
  the test cannot tell correct routing from the world-group behaviour — registered there it would
  have skipped and reported success.

  **This test has not been run to a numeric verdict yet.** I built torch-tensorrt from source with
  `ENABLE_TRT_NCCL_COLLECTIVES` on 8x A100 to try (TRT 11.2.1.2, torch 2.15.0.dev,
  `ENABLED_FEATURES.native_trt_collectives == True`), but it fails during runtime setup:

      RuntimeError: [Error thrown at core/runtime/TRTEngine.cpp:916]
      Expected nccl_pg != nullptr to be true but got false
      Backend is not ProcessGroupNCCL

  `bind_nccl_comm()` gets a non-null backend from `getBackend(BackendType::NCCL)` but the
  `dynamic_cast<ProcessGroupNCCL*>` fails. This is not specific to this test: the existing
  `test_distributed_mode_subgroup` fails identically, same line, same message, while
  `test_all_reduce_correctness` — which does not go through `distributed_context(group)` — passes.

  That is #4661: recent PyTorch moved the NCCL backend to `c10d::nccl2::ProcessGroupNCCL`, so the
  cast to the legacy type fails, and #4661 replaces exactly those two lines with a cast that accepts
  either. Nothing here is blocked on this PR — once #4661 lands, this test should be runnable and I
  will post the result.

Both CPU-only classes were run against the pre-fix code to confirm they fail rather than pass
vacuously: reverting to `num_ranks = world_size` fails 6 subtests, reintroducing `sorted(ranks)`
fails the ordering test.

@narendasan @apbose — the workflows still need maintainer approval; CI has never run on this branch.

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
@@ -80,6 +80,30 @@ def _get_distributed_rank_and_world_size() -> Tuple[int, int]:
         return rank, world_size
 
 
+
+def _collective_group_ranks(group_name, world_size):
+    """Global ranks of the collective's process group.
+
+    The native ``add_dist_collective`` layer needs the set of ranks that participate in
+    *this* collective. Resolving it from the op's ``group_name`` lets a collective target a
+    process **subgroup** (e.g. context/sequence-parallel over one subgroup while tensor-parallel
+    uses another -- a 2-D device mesh) instead of always the whole world. Falls back to the world
+    group when the group cannot be resolved (single-program / group not created in this process).
+    """
+    import numpy as np
+    if group_name:
+        try:
+            import torch.distributed as dist
+            from torch.distributed.distributed_c10d import _resolve_process_group
+
+            ranks = dist.get_process_group_ranks(_resolve_process_group(group_name))
+            return np.array(sorted(ranks), dtype=np.int64)
+        except Exception as e:  # noqa: BLE001
+            logger.warning(
+                f"Could not resolve process group '{group_name}' ({e}); using world group"
+            )
+    return np.arange(world_size, dtype=np.int64)
+
 def nccl_all_gather(
     ctx: ConversionContext,
     target: Union[Target, str],
@@ -228,6 +252,7 @@ def nccl_all_gather_native(
     source_ir: Optional[SourceIR],
     name: str,
     plug_inputs: Tuple[Argument, ...],
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement all_gather using native TensorRT DistCollective API.
@@ -263,7 +288,7 @@ def nccl_all_gather_native(
         import numpy as np
 
         # Create array of all participating rank IDs [0, 1, 2, ..., world_size-1]
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         logger.debug(
             f"Creating ALL_GATHER layer: groups={groups.tolist()}, groupSize={world_size}"
@@ -285,7 +310,7 @@ def nccl_all_gather_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        layer.num_ranks = len(groups)
 
         return output
 
@@ -302,6 +327,7 @@ def nccl_reduce_scatter_native(
     name: str,
     plug_inputs: Tuple[Argument, ...],
     reduce_op: str = "sum",
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement reduce_scatter using native TensorRT DistCollective API.
@@ -350,7 +376,7 @@ def nccl_reduce_scatter_native(
     trt_reduce_op = reduce_op_map[reduce_op.lower()]
 
     try:
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         layer = ctx.net.add_dist_collective(
             input_tensor,
@@ -363,7 +389,7 @@ def nccl_reduce_scatter_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        layer.num_ranks = len(groups)
         logger.debug(
             f"Successfully created native REDUCE_SCATTER layer: {name}, reduce_op={reduce_op}, groups={groups.tolist()}"
         )
@@ -383,6 +409,7 @@ def nccl_all_reduce_native(
     name: str,
     plug_inputs: Tuple[Argument, ...],
     reduce_op: str = "sum",
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement all_reduce using native TensorRT DistCollective API.
@@ -435,7 +462,7 @@ def nccl_all_reduce_native(
         # Create array of all participating rank IDs [0, 1, ..., world_size-1]
         # Passing None for groups can be treated as a no-op by TRT; use an explicit
         # rank array (same as ALL_GATHER) to ensure the reduction is performed.
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         layer = ctx.net.add_dist_collective(
             input_tensor,
@@ -448,7 +475,7 @@ def nccl_all_reduce_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        layer.num_ranks = len(groups)
         logger.debug(
             f"Successfully created native ALL_REDUCE layer: {name}, reduce_op={reduce_op}, groups={groups.tolist()}"
         )
@@ -467,6 +494,7 @@ def nccl_all_to_all_native(
     source_ir: Optional[SourceIR],
     name: str,
     plug_inputs: Tuple[Argument, ...],
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement all_to_all using native TensorRT DistCollective API.
@@ -503,7 +531,7 @@ def nccl_all_to_all_native(
         import numpy as np
 
         # Create array of all participating rank IDs [0, 1, 2, ..., world_size-1]
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         logger.debug(
             f"Creating ALL_TO_ALL layer: groups={groups.tolist()}, groupSize={world_size}"
@@ -525,7 +553,7 @@ def nccl_all_to_all_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        layer.num_ranks = len(groups)
 
         return output
 
@@ -542,6 +570,7 @@ def nccl_scatter_native(
     name: str,
     plug_inputs: Tuple[Argument, ...],
     root: int = 0,
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement scatter using native TensorRT DistCollective API.
@@ -577,7 +606,7 @@ def nccl_scatter_native(
         import numpy as np
 
         # Create array of all participating rank IDs [0, 1, 2, ..., world_size-1]
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         logger.debug(
             f"Creating scatter layer: groups={groups.tolist()}, groupSize={world_size}"
@@ -599,7 +628,7 @@ def nccl_scatter_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        layer.num_ranks = len(groups)
 
         return output
 
@@ -616,6 +645,7 @@ def nccl_gather_native(
     name: str,
     plug_inputs: Tuple[Argument, ...],
     root: int = 0,
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement gather using native TensorRT DistCollective API.
@@ -651,7 +681,7 @@ def nccl_gather_native(
         import numpy as np
 
         # Create array of all participating rank IDs [0, 1, 2, ..., world_size-1]
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         logger.debug(
             f"Creating gather layer: groups={groups.tolist()}, groupSize={world_size}"
@@ -673,7 +703,7 @@ def nccl_gather_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        layer.num_ranks = len(groups)
 
         return output
 
--- a/py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
@@ -80,6 +80,38 @@ def _get_distributed_rank_and_world_size() -> Tuple[int, int]:
         return rank, world_size
 
 
+def _collective_group_ranks(group_name: Optional[str], world_size: int) -> np.ndarray:
+    """Global ranks of the collective's process group.
+
+    The native ``add_dist_collective`` layer needs the set of ranks that participate in
+    *this* collective. Resolving it from the op's ``group_name`` lets a collective target a
+    process **subgroup** (e.g. context/sequence-parallel over one subgroup while tensor-parallel
+    uses another -- a 2-D device mesh) instead of always the whole world. Falls back to the world
+    group when the group cannot be resolved (single-program / group not created in this process).
+    """
+    if group_name:
+        try:
+            import torch.distributed as dist
+            from torch.distributed.distributed_c10d import _resolve_process_group
+
+            # Preserve the group's own ordering: get_process_group_ranks() returns global
+            # ranks indexed by *group* rank, and that mapping is what layout-sensitive
+            # collectives are defined against. all_gather concatenates by group rank,
+            # reduce_scatter sends chunk i to group rank i, all_to_all permutes by it, and a
+            # scatter/gather root names a position in it. Sorting would silently renumber the
+            # group whenever it was not created in ascending order -- e.g. a device mesh
+            # yielding [5, 2] would be read as [2, 5], swapping which rank receives which
+            # slice. Sorting is not needed for agreement either: every member resolves the
+            # same group_name to the same group and therefore already sees the same order.
+            ranks = dist.get_process_group_ranks(_resolve_process_group(group_name))
+            return np.array(ranks, dtype=np.int64)
+        except Exception as e:  # noqa: BLE001
+            logger.warning(
+                f"Could not resolve process group '{group_name}' ({e}); using world group"
+            )
+    return np.arange(world_size, dtype=np.int64)
+
+
 def nccl_all_gather(
     ctx: ConversionContext,
     target: Union[Target, str],
@@ -228,6 +260,7 @@ def nccl_all_gather_native(
     source_ir: Optional[SourceIR],
     name: str,
     plug_inputs: Tuple[Argument, ...],
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement all_gather using native TensorRT DistCollective API.
@@ -263,7 +296,7 @@ def nccl_all_gather_native(
         import numpy as np
 
         # Create array of all participating rank IDs [0, 1, 2, ..., world_size-1]
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         logger.debug(
             f"Creating ALL_GATHER layer: groups={groups.tolist()}, groupSize={world_size}"
@@ -285,7 +318,10 @@ def nccl_all_gather_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        # num_ranks is the size of *this collective's* group, not the world size: TensorRT
+        # takes the participants from `groups` and their count from num_ranks, so a subgroup
+        # collective left at world_size describes a group it does not have.
+        layer.num_ranks = len(groups)
 
         return output
 
@@ -302,6 +338,7 @@ def nccl_reduce_scatter_native(
     name: str,
     plug_inputs: Tuple[Argument, ...],
     reduce_op: str = "sum",
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement reduce_scatter using native TensorRT DistCollective API.
@@ -350,7 +387,7 @@ def nccl_reduce_scatter_native(
     trt_reduce_op = reduce_op_map[reduce_op.lower()]
 
     try:
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         layer = ctx.net.add_dist_collective(
             input_tensor,
@@ -363,7 +400,10 @@ def nccl_reduce_scatter_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        # num_ranks is the size of *this collective's* group, not the world size: TensorRT
+        # takes the participants from `groups` and their count from num_ranks, so a subgroup
+        # collective left at world_size describes a group it does not have.
+        layer.num_ranks = len(groups)
         logger.debug(
             f"Successfully created native REDUCE_SCATTER layer: {name}, reduce_op={reduce_op}, groups={groups.tolist()}"
         )
@@ -383,6 +423,7 @@ def nccl_all_reduce_native(
     name: str,
     plug_inputs: Tuple[Argument, ...],
     reduce_op: str = "sum",
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement all_reduce using native TensorRT DistCollective API.
@@ -435,7 +476,7 @@ def nccl_all_reduce_native(
         # Create array of all participating rank IDs [0, 1, ..., world_size-1]
         # Passing None for groups can be treated as a no-op by TRT; use an explicit
         # rank array (same as ALL_GATHER) to ensure the reduction is performed.
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         layer = ctx.net.add_dist_collective(
             input_tensor,
@@ -448,7 +489,10 @@ def nccl_all_reduce_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        # num_ranks is the size of *this collective's* group, not the world size: TensorRT
+        # takes the participants from `groups` and their count from num_ranks, so a subgroup
+        # collective left at world_size describes a group it does not have.
+        layer.num_ranks = len(groups)
         logger.debug(
             f"Successfully created native ALL_REDUCE layer: {name}, reduce_op={reduce_op}, groups={groups.tolist()}"
         )
@@ -467,6 +511,7 @@ def nccl_all_to_all_native(
     source_ir: Optional[SourceIR],
     name: str,
     plug_inputs: Tuple[Argument, ...],
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement all_to_all using native TensorRT DistCollective API.
@@ -503,7 +548,7 @@ def nccl_all_to_all_native(
         import numpy as np
 
         # Create array of all participating rank IDs [0, 1, 2, ..., world_size-1]
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         logger.debug(
             f"Creating ALL_TO_ALL layer: groups={groups.tolist()}, groupSize={world_size}"
@@ -525,7 +570,10 @@ def nccl_all_to_all_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        # num_ranks is the size of *this collective's* group, not the world size: TensorRT
+        # takes the participants from `groups` and their count from num_ranks, so a subgroup
+        # collective left at world_size describes a group it does not have.
+        layer.num_ranks = len(groups)
 
         return output
 
@@ -542,6 +590,7 @@ def nccl_scatter_native(
     name: str,
     plug_inputs: Tuple[Argument, ...],
     root: int = 0,
+    group_name: Optional[str] = None,
 ) -> trt.ITensor:
     """
     Implement scatter using native TensorRT DistCollective API.
@@ -577,7 +626,7 @@ def nccl_scatter_native(
         import numpy as np
 
         # Create array of all participating rank IDs [0, 1, 2, ..., world_size-1]
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
 
         logger.debug(
             f"Creating scatter layer: groups={groups.tolist()}, groupSize={world_size}"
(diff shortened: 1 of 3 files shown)


narendasan:
@pkisfaludi-nv do you need me to send to the instructions to get on the CLA?

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -80,6 +80,30 @@ def _get_distributed_rank_and_world_size() -> Tuple[int, int]:
...
+    group when the group cannot be resolved (single-program / group not created in this process).
+    """
+    import numpy as np
+    if group_name:
+        try:
+            import torch.distributed as dist
+            from torch.distributed.distributed_c10d import _resolve_process_group
+
+            ranks = dist.get_process_group_ranks(_resolve_process_group(group_name))
+            return np.array(sorted(ranks), dtype=np.int64)
```
apbose: had a doubt in this for order sensitive collectives on a permuted device mesh.
sorted() in the above would give a deterministic order which all the rank members in the group would agree on. I assume it would be a no-op for all_reduce/reduce_scatter where order wont affect the result.

But in cases like all_to_all and all_gather won't the ops be layout sensitive? eg: if DeviceMesh yields a group whose get_process_group_ranks is not ascending eg: [say rank 5, rank 2]. The subgroup comm {2,5} has two members and sorted would give the wrong order.

apbose (commented):
Thanks for the PR @pkisfaludi-nv!
So this PR bakes each collective subgroup into the engine at build.
I am thinking about the runtime scenario for the 2D Mesh case. Going by the ncclCommsplit changes for subgroup routing
[Myelin per-subgroup NCCL communicators !19723](https://gitlab-master.nvidia.com/dl/compiler/myelin/-/merge_requests/19723)
[TensorRT standalone DistCollective subgroup routing !49040](https://gitlab-master.nvidia.com/TensorRT/TensorRT/-/merge_requests/49040)
[TensorRT fused-Myelin subgroup routing !49062](https://gitlab-master.nvidia.com/TensorRT/TensorRT/-/merge_requests/49062), if we take these two cases:

- C++ runtime (`TRTEngine.cpp::bind_nccl_comm`): today the auto-resolver bails when several NCCL groups exist. For CP4×TP2 do we pin the world group via distributed_context, or does bind need to prefer the world/parent group?
- Python runtime (_PythonTorchTensorRTModule, `get_active_group()`): same question — is setting the world group active enough?

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
pkisfaludi-nv: Good catch — this was a real bug, not a no-op. Fixed in `d7b4445` by dropping the `sorted()`.

You are right on both halves. `get_process_group_ranks()` is documented as returning "a list of global ranks **ordered by group rank**" — index `i` *is* group rank `i` — so sorting silently renumbers any group that was not created in ascending order. Your `[5, 2]` example is exactly it: it was read as `[2, 5]`, so the two ranks exchanged slices.

And your instinct about which collectives care is right, with one addition: it is not only `all_gather`/`all_to_all`. `reduce_scatter` sends chunk `i` to group rank `i`, and a `scatter`/`gather` `root` names a position in that ordering, so those are affected too. `all_reduce` is the only one that genuinely does not care, since its result is independent of member order.

Worth noting the sort was not buying determinism either — every member resolves the same `group_name` to the same group and therefore already observes the same order, so there was nothing to canonicalise.

Why it survived until now: the ascending case is the only one that was exercised, and it is unchanged by the fix. Added `TestCollectiveGroupRanks` covering non-ascending order preserved, ascending unchanged, and the unresolvable-group fallback — no GPU or initialised process group needed, so it runs in ordinary pytest.

pkisfaludi-nv:
@narendasan — could you approve the workflow runs on this PR (and on #4388, which stacks on it)? All five workflows are sitting in `action_required`, so only `label` and `Meta CLA Check` have ever executed. The CLA is signed now.

That matters more than usual here, because the two most recent commits have never been exercised by CI:

- `d7b4445` fixes a **correctness bug** @apbose caught in review. `sorted()` was applied to the ranks resolved from the collective's process group, but `get_process_group_ranks()` returns global ranks *ordered by group rank* — so sorting silently renumbered any group not created in ascending order. A group of `[5, 2]` was read as `[2, 5]`, and those two ranks exchanged slices: wrong data, no error. It affects `all_gather`, `all_to_all`, `reduce_scatter` and the `scatter`/`gather` root; only `all_reduce` is indifferent. I verified the fix by extracting the function and driving it with stubs, but it has had no real test run.
- `96b8a7c2` fixes `Python Linting`, which had been red since July — `black` wanted to reformat the two files this PR touches. Verified with the repo's pinned `black` 26.3.1: 541 files unchanged, only those two.

The other CI failures on the last full run (July) look unrelated to this change — 27 `TestFlashAttention` cases in L1 dynamo core/compile with no mention of nccl/collective/distributed anywhere in the logs, `test_profiles_restored_from_trt_api_without_metadata` in the RTX runtime job, and a Jetpack `bdist_wheel` build failure. I would rather have CI confirm that than assert it.

A review from @apbose or @zewenli98 would be welcome too — the review threads are all answered, including one suggestion I declined with reasoning (defaulting the collective fusion wall to a same-width dtype: bf16→fp16 is not value-preserving, since fp16 overflows above 65504 while bf16 carries fp32's exponent range, and fp32 is also the accumulation dtype for the reducing collectives — so it is a user-selectable setting instead of a changed default).

apbose:
CI waiting on https://github.com/pytorch/TensorRT/pull/4661

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
pkisfaludi-nv: Follow-up for anyone reading this thread: the PR has since been squashed to a single commit, so the `d7b4445` SHA above no longer resolves. The change is in `_collective_group_ranks` on the current head.

Your instinct here was right, and it turned out to be worth more than a code-reading argument — I confirmed it on hardware. 8 ranks, `all_gather`, rank `r` holding `r+1`, with the group's rank array reversed:

| | gathered |
|---|---|
| order preserved | `8,7,6,5,4,3,2,1` |
| `sorted()` | `1,2,3,4,5,6,7,8` |

So sorting really does silently reassign which rank holds which slice on a permuted mesh, exactly as you suspected.

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -363,7 +400,10 @@ def nccl_reduce_scatter_native(
         set_layer_name(layer, target, name, source_ir)
 
         output = layer.get_output(0)
-        layer.num_ranks = world_size
+        # num_ranks is the size of *this collective's* group, not the world size: TensorRT
+        # takes the participants from `groups` and their count from num_ranks, so a subgroup
+        # collective left at world_size describes a group it does not have.
+        layer.num_ranks = len(groups)
         logger.debug(
```
apbose, line 407: had a doubt in this. The closed PR https://github.com/pytorch/TensorRT/pull/4388 listed this as a fix- ayer.num_ranks must be set before layer.get_output(0) so Myelin can infer the collective's output shape (e.g. all_gather dim0 = in_dim0 * num_ranks). But I don't see that in this PR. Its different from the subgroup routing, but does that need to be the case for  collectives producing rank dependant outputs like all_gather, reduce_scatter, scatter, gather

tests/py/dynamo/distributed/test_native_nccl.py
```diff
@@ -1956,6 +2093,180 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    cp_group = cp_groups[rank % 2]
+    tp_group = tp_groups[rank // 2]
+    # PyTorch creates the ncclComm_t lazily; bind_nccl_comm() reads a null pointer until
+    # at least one collective has run on the group. A real collective is used rather than
+    # dist.barrier(), which for NCCL infers the device when device_ids is omitted and can
+    # pick the wrong one on a multi-GPU rank, hanging the group.
+    seed = torch.zeros(1, device=device)
+    for group in cp_groups + tp_groups:
+        if rank in dist.get_process_group_ranks(group):
+            dist.all_reduce(seed, group=group)
```
apbose, line 2218: dist.new_group would return  GroupMember.NON_GROUP_MEMBER on ranks outside the subgroup. so every rank in this case has two real handles (corresponding to one CP and one TP) and two non group. Calling dist.get_process_group_ranks on non group would look for unregistered group and raise asserition.

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -80,6 +80,38 @@ def _get_distributed_rank_and_world_size() -> Tuple[int, int]:
...
+            # slice. Sorting is not needed for agreement either: every member resolves the
+            # same group_name to the same group and therefore already sees the same order.
+            ranks = dist.get_process_group_ranks(_resolve_process_group(group_name))
+            return np.array(ranks, dtype=np.int64)
+        except Exception as e:  # noqa: BLE001
+            logger.warning(
+                f"Could not resolve process group '{group_name}' ({e}); using world group"
+            )
+    return np.arange(world_size, dtype=np.int64)
+
```
apbose, line 113: Silent world group fallback might not be the right approach here? The way I see it is this can fail to find the ranks in two cases-
1. torch.distributed is not initialized
2. distributed is initialized, group_name was provided but torch could not resolve it. I dont think we should silently fallback to world here

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -80,6 +80,38 @@ def _get_distributed_rank_and_world_size() -> Tuple[int, int]:
...
+            # ranks indexed by *group* rank, and that mapping is what layout-sensitive
+            # collectives are defined against. all_gather concatenates by group rank,
+            # reduce_scatter sends chunk i to group rank i, all_to_all permutes by it, and a
+            # scatter/gather root names a position in it. Sorting would silently renumber the
+            # group whenever it was not created in ascending order -- e.g. a device mesh
+            # yielding [5, 2] would be read as [2, 5], swapping which rank receives which
+            # slice. Sorting is not needed for agreement either: every member resolves the
+            # same group_name to the same group and therefore already sees the same order.
+            ranks = dist.get_process_group_ranks(_resolve_process_group(group_name))
+            return np.array(ranks, dtype=np.int64)
```
apbose, line 107: had doubt in this. This passes global rank. so in communicator is it the local rank being passed or the global rank? going by this- https://github.com/NVIDIA/TensorRT/blob/main/include/NvInfer.h#L9764C1-L9768C14

tests/py/dynamo/distributed/test_native_nccl.py
```diff
@@ -1956,6 +2093,180 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            raise AssertionError(
+                f"rank {rank}: {axis} all-reduce over {members} returned {got}, "
+                f"expected {expected_value}{hint}"
+            )
+
+    print(
+        f"[rank {rank}] 2-D mesh routing OK (CP and TP each reduced over their own group)"
+    )
+
+
```
apbose, line 2269: also this test creates fresh model for TP case and another fresh for CP case. Can we have a case where TP and CP would be in same model. I am curious on how it would be if an engine for a particular rank has multiple communicators. Also this changes would work from which TRT version?

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
pkisfaludi-nv: Good question — I dropped it deliberately, and only after testing it, because the reasoning in #4388 doesn't hold up.

The shape *is* stale at the instant of capture: for an all_gather with `num_ranks=4` and input `(2,64)`, reading `get_output(0).shape` before setting `num_ranks` gives `(2,64)` instead of `(8,64)`. So the concern was real. But TensorRT re-evaluates that shape lazily, on `main` the two lines are adjacent with nothing reading the shape in between, and the built engine is byte-identical either way. On 8× A100 / TRT 11.2.0.86, building the same network both orders and hashing the serialized engine:

| op | `num_ranks` first | `get_output` first | identical |
|---|---|---|---|
| `ALL_GATHER` | `(16,64)` md5 `0ebe6c0b97d4` | `(16,64)` md5 `0ebe6c0b97d4` | yes |
| `ALL_REDUCE` | `(4,64)` md5 `18e50672a18a` | `(4,64)` md5 `18e50672a18a` | yes |
| `ALL_TO_ALL` | `(4,64)` md5 `a0ad8806638d` | `(4,64)` md5 `a0ad8806638d` | yes |
| `REDUCE_SCATTER` | `(1,64)` md5 `547445add902` | `(1,64)` md5 `547445add902` | yes |

I deliberately included the rank-dependent ops you named — `all_gather` and `reduce_scatter` are the two whose output shape actually scales with `num_ranks`, and they come out identical too.

What *was* substantive in that hunk is the **value**: `num_ranks` has to be the size of the collective's own group, not `world_size`. That part is kept here (`layer.num_ranks = len(groups)`), and it's inseparable from subgroup routing — with `groups` a subgroup, `world_size` describes a group the collective doesn't have. `TestNativeCollectiveNumRanks` covers it for all six converters and fails if you put `world_size` back.

So: ordering dropped as a no-op, value kept. Happy to restore the ordering as defensive hygiene if you'd prefer it, but I didn't want to carry it as a "fix" when the evidence says it changes nothing.

tests/py/dynamo/distributed/test_native_nccl.py
pkisfaludi-nv: You're right, and this was a real bug — thank you. `new_group()` hands back `GroupMember.NON_GROUP_MEMBER` on ranks outside the group, and `get_process_group_ranks()` raises on that sentinel, so every rank would have died in setup on the two groups it isn't a member of. It never surfaced because the test was hanging earlier for an unrelated reason (broken NCCL P2P on the node I was using), so it never reached this line cleanly.

Fixed in `960c6ca` by taking membership from the rank lists the groups were built from, rather than querying the handles:

```python
cp_ranks = [[0, 2], [1, 3]]
tp_ranks = [[0, 1], [2, 3]]
cp_groups = [dist.new_group(ranks=r) for r in cp_ranks]
tp_groups = [dist.new_group(ranks=r) for r in tp_ranks]
...
for members, group in zip(cp_ranks + tp_ranks, cp_groups + tp_groups):
    if rank in members:
        dist.all_reduce(seed, group=group)
```

The same change removes the other `get_process_group_ranks()` call further down, which had the same problem.

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
pkisfaludi-nv: Agreed, and changed in `960c6ca`. Falling back to the world group is precisely the failure this function exists to prevent: the op asked for a subgroup, and we'd silently build a collective over every rank and return a wrong answer with no error. That's worse than not compiling.

It now distinguishes your two cases:

- **No `group_name`** — the collective is over the world group by construction, so `arange(world_size)` is correct, not a fallback.
- **`group_name` given but unresolvable** — raises `RuntimeError` naming the group, rather than widening.

```
Collective names process group '<name>', which could not be resolved in this process (...).
Refusing to fall back to the world group: that would reduce across every rank and silently
return wrong results. Ensure the group is created before compiling.
```

`TestCollectiveGroupRanks` was asserting the old fallback behaviour, so I replaced that case with one asserting it raises, plus a separate test that a missing `group_name` still resolves to the world group.

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
pkisfaludi-nv: This is the sharpest question on the PR, and I think it deserves your call rather than mine. Short version: **global ranks are right only if the engine binds the world communicator**, and I've made that explicit — but it's a contract worth agreeing on.

Taking the header you linked at its word:

> `groups` — *"a flat array of rank IDs **in the communicator** ... only those ranks take part in the collective, and the order defines the group-local rank"*

So the array is in the index space of whatever communicator is bound, and it *selects a subset* of it. That framing only makes sense when the bound communicator is larger than the group. I checked all three combinations on 8× A100 with TRT 11.2.0.86, driving `addDistCollective` + `setCommunicator` directly with real NCCL communicators. Two 4-rank groups, rank `r` holding `r+1`, so correct per-group sums are `10,10,10,10 / 26,26,26,26`:

| bound communicator | rank array | result |
|---|---|---|
| world (8 ranks) | group's **global** ranks `[0..3]` / `[4..7]` | **correct** — 10/26 |
| the subgroup itself (4 ranks) | group-**local** `[0,1,2,3]` | **correct** — 10/26 |
| the subgroup itself (4 ranks) | group's global ranks `[4,5,6,7]` | **0.0** — silently wrong, no error |

So both index spaces work; what breaks is mixing them, and it breaks *silently*, which is the dangerous part.

That leaves two regimes:

1. **One subgroup per engine** — bind that subgroup, array must be `0..n-1`.
2. **Several subgroups per engine** (a 2-D mesh: CP on one axis, TP on the other) — you can't bind both, so bind the world communicator and let each collective's array select its subgroup by global rank.

`bind_nccl_comm()` binds one communicator per engine, and auto-resolve explicitly refuses when more than one NCCL group exists. So regime 2 is the only one that supports the mesh case this PR is for, and it's what the PR emits.

The sharp edge you've spotted: under regime 1, `distributed_context(tp_group)` with a group like `[2,3]` would feed out-of-range IDs and return zeros. That's fine when the subgroup is rank-0-based, which is why `test_distributed_mode_subgroup` never caught it — it builds its "subgroup" as `list(range(world_size))`, where global and local indices coincide.

My mesh test was wrong here too: it bound the subgroup while the converter emitted global ranks. Fixed in `960c6ca` to bind the world group, and the docstring now states the contract.

**Where I'd like your view:** should this stay a documented contract ("multi-subgroup meshes must bind the world communicator"), or should the runtime translate global ranks into the bound communicator's index space at bind time — which would make both regimes work and remove the silent-zeros trap entirely? The latter is in `bind_nccl_comm()`, i.e. your #4661 territory, so I didn't want to reach into it unilaterally.

tests/py/dynamo/distributed/test_native_nccl.py
pkisfaludi-nv: Both good questions.

**TP and CP in the same model.** Added in `960c6ca`. Binding the world communicator (see the thread above) is exactly what makes one engine able to carry collectives on two different subgroups, so the test now does it:

```python
y = all_reduce(x, "sum", tp_name); y = wait(y)
y = y * 10.0
z = all_reduce(y, "sum", cp_name); return wait(z)
```

The scale in the middle is load-bearing. A plain TP-then-CP sum covers both mesh axes, i.e. all four ranks, which is numerically identical to the world sum — the test would pass just as happily if both collectives were wrongly routed to the world group. Scaling between them breaks the symmetry: correct routing gives `(3 or 7) × 10` summed down the CP axis = **100** on every rank, while two world-routed all_reduces give **400**.

On multiple communicators per engine: there's only ever one bound (`bind_nccl_comm()` binds a single communicator per engine), and the per-collective rank arrays are what distinguish the two groups within it. So it's one communicator, two groups — not two communicators.

**TRT version.** Native collectives need TRT ≥ 10.16 — that's the `TRT_HAS_NATIVE_NCCL` gate in `core/runtime/TRTEngine.h`, which together with `USE_C10D_NCCL` is what defines `ENABLE_TRT_NCCL_COLLECTIVES`. Below that the converters aren't registered at all (`@needs_native_collectives`), so this path is inert rather than broken. The CPU-only tests gate on `hasattr(trt, "CollectiveOperation")` for the same reason. I've been testing against 11.2.

Numbers for the mesh test itself are still pending — I'm queued for a 4-GPU node to run it against this branch now that your NCCL2 binding fix is in it, and I'll post the result either way.

pkisfaludi-nv:
Ran this on hardware against this branch (which now includes your NCCL2 binding commits) — **your doubt about global vs local ranks was correct, and the PR was wrong.** Corrected in `4dd86d5`.

### The rank array is group-local

`addDistCollective` documents `groups` as *"rank IDs in the communicator"*, and `bind_nccl_comm()` binds the communicator of the op's own process group, whose members are numbered `0..n-1`. Passing global ranks was only correct when the group happened to be rank-0-based — and it fails loudly rather than silently. 8× A100, TRT 11.2.1.2, two 4-rank groups, each engine bound to its own group's communicator:

| rank array | result |
|---|---|
| global `[0,1,2,3]` / `[4,5,6,7]` | group with rank 0 correct; the other **fails** — `Did not properly set the child communicator` |
| group-local `[0,1,2,3]` | `10,10,10,10 / 26,26,26,26` — correct |

That's exactly why the 4-GPU mesh test died with `rank 2: TP all-reduce over [2, 3] returned 0.0, expected 7.0`.

I'd earlier reported global ranks working on 8 GPUs. That was on a different drop (`11.2.0.86-md-moe-ep`), where the same experiment did return 10/26. On **11.2.1.2** it does not. Worth knowing independently of this PR — if subset-selection by global rank is supposed to work, that difference between drops looks like a regression.

### What resolving `group_name` is actually for

The group's **size**. `num_ranks` comes from it, and for rank-dependent collectives that size sets the output shape. all_gather over a 4-rank subgroup, input `(2,4)`:

| | output |
|---|---|
| `num_ranks = world_size = 8` (main today) | **(16, 4)** — twice what the group can produce |
| `num_ranks = group size = 4` | **(8, 4)** — correct |

`all_reduce` is shape-preserving, which is why the existing tests never caught this. This is the concrete defect the PR fixes, and it's the rank-dependent ops you named in the other thread.

Consequence: **order preservation is gone.** A group-local array is `0..n-1` by construction, and which global rank sits at each position is carried by the communicator. The order test is replaced by one asserting only the size shows through — so my earlier answer to you about `sorted()` no longer applies to this design.

Also dropped the single-engine two-axis case from your other question. One communicator is bound per engine and the array is group-local, so **one engine cannot describe two different subgroups** — that needs a second bound communicator or global IDs, and global IDs currently fail on 11.2.1.2. Worth deciding whether that's a limitation to document or something to lift in the runtime.

### Test results (8× A100, TRT 11.2.1.2, torch 2.15.0.dev, source build with `ENABLE_TRT_NCCL_COLLECTIVES`)

| | |
|---|---|
| CPU-only converter tests | **5 passed, 6 subtests** |
| `TestMultirankNccl` (2 GPU, existing suite) | **14 passed** |
| `TestMultirankNccl4GPU::test_two_dimensional_mesh_routing` | **passed**, all 4 ranks |

The 2-GPU suite going fully green is also a clean confirmation of your NCCL2 fix — before it, everything routed through `distributed_context()` died at `TRTEngine.cpp:916`.

One environment note for anyone reproducing: on the PCIe A100 node I used, NCCL P2P is broken and every 2-rank group hangs for 600 s before any of this surfaces. `NCCL_P2P_DISABLE=1` is needed there.

pkisfaludi-nv:
Correcting my previous comment — **I was wrong to move to group-local rank IDs, and I've reverted it.**

The requirement I talked myself out of is real: a CP4×TP2 model needs collectives on two different groups inside one engine. Re-reading the runtime API settles which design supports that:

> `setCommunicator`: *"A pointer to the communicator that is used by the execution context ... The communicator **must be uniform across all multi-device instances** or undefined behavior occurs."*

One communicator per context, identical on every rank — so the parent is the world communicator, and each `addDistCollective` layer names its group as a **subset** of it via `groups`. That is exactly the mechanism for multiple groups per engine, and it means **global rank IDs are the right thing to emit**. Group-local IDs only work when the bound communicator *is* the group, which forecloses the mesh case entirely.

### It works — on one TensorRT and not the other

8× A100, every rank binding the same 8-rank world communicator:

| | TRT 11.2.0.86-md-moe-ep | TRT 11.2.1.2 |
|---|---|---|
| subset `[0,1,2,3]` | **10** to its members | **36** to everyone — the array is silently ignored |
| subset `[4,5,6,7]` | **26** to its members | **fails** — *"Did not properly set the child communicator"* |
| **CP4 × TP2 in one engine** | **360 on every rank** — correct | fails, same error |
| `TestMultirankNccl4GPU` (4 GPU, incl. two-axis single-engine case) | **passes** | fails |

The CP4×TP2 case is TP all_reduce → scale by 10 → CP all_reduce, so the correct answer (360) is distinguishable from two world-routed collectives (2880).

So the capability exists and this PR's design matches it; it is **broken on the TensorRT the CI pins**. On 11.2.1.2 the silent case is the worse one — a subset containing rank 0 quietly reduces over the whole communicator and returns a plausible-looking wrong number.

@apbose @narendasan — two things worth your view:

1. **Is the 11.2.1.2 behaviour known?** If subset selection is meant to work there, the silent-wrong-answer path looks worth a TensorRT bug. I can hand over the ~60-line pure-TRT repro (no torch-tensorrt involved).
2. **CI can't currently exercise this.** I've gated the 4-GPU mesh test behind `TORCHTRT_TEST_COLLECTIVE_SUBGROUPS=1` rather than let CI go red over a TensorRT limitation, with the measurements in the helper docstring. If you'd rather it ran unconditionally once TRT is fixed, dropping the gate is a one-liner.

### Current test results

| | |
|---|---|
| CPU-only converter tests | **5 passed, 6 subtests** (11.2.1.2) |
| `TestMultirankNccl` (2 GPU, existing suite) | **14 passed** (11.2.1.2) |
| `TestMultirankNccl4GPU` mesh routing | **passes** on md-moe-ep; **skips** with an explicit reason on 11.2.1.2 |

The core fix is unaffected by any of this and holds on both: `num_ranks` must be the group's size, not the world's. all_gather over a 4-rank subgroup with input `(2,4)` gives **(16,4)** on main and **(8,4)** with this PR.

pkisfaludi-nv:
Followed up on the child-communicator gap — **it's already fixed in TensorRT, no new work needed there.**

TensorRT **MR !49040**, *"Fix DistCollective subgroup communicator routing"* (2026-08-20), does exactly this: `getOrCreateCommunicator` now derives the `ncclCommSplit` colour/key from the parent communicator's current rank instead of assuming world-rank 0, splits disjoint rank-local groups with stable colours, guards the null child comm on non-member ranks, and remaps roots into the child's numbering. Its own validation notes state it "Fixes any 2-D mesh setup, e.g CP4xTP2 or CP2xTP4", verified on 4× A100 with CP `[0,2] [1,3]` and TP `[0,1] [2,3]` — the same topology as this PR's mesh test.

Where it is:

| TensorRT | has MR !49040 |
|---|---|
| `main` | **yes** |
| `rel-11.4` | **yes** (identical collective runtime to main — zero diff) |
| `rel-11.3`, `rel-11.3.0` | no |
| releases 11.2.1.2 / .3 / .4 | no |

So the requirement is **TensorRT ≥ 11.4**. This repo currently fetches 11.2.1.2, which predates it — that is the only reason the 4-GPU mesh test can't run in CI, and it matches what I measured: with the fix, subset `[0,1,2,3]` → 10 and `[4,5,6,7]` → 26, and CP4×TP2 in one engine → 360 on every rank; without it, a subset containing rank 0 is silently ignored (36 instead of 10) and one without rank 0 fails with *"Did not properly set the child communicator"*.

I've updated the skip reason and the helper docstring to name MR !49040 and the version requirement instead of just saying "broken on 11.2.1.2" (`0e0d1e2`), so whoever bumps the TensorRT pin knows exactly what unblocks it.

@apbose @narendasan — nothing needed from the TensorRT side. The open question is just whether this repo's TensorRT pin moves to 11.4 (at which point the gate can be dropped and the mesh test runs unconditionally), or whether !49040 is worth requesting as a cherry-pick into `rel-11.3` if that's the branch you intend to track.

Issue #4381, opened 2026-07-04
Native DistCollective fails at build time when fused into a Myelin ForeignNode (large multi-device graph)


pkisfaludi-nv:
### Bug Description
The TRT-11 native multi-device collectives added in #4321 (`add_dist_collective` in `py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py`) work for small / standalone graphs, but fail during the TensorRT engine build on a large graph when a `DistCollective` layer is fused — together with surrounding reshape/unsqueeze/pointwise ops — into a Myelin `ForeignNode`. Myelin can't map the collective kind (`toMyelinCommKind`) once the `IDistCollectiveLayer` is absorbed into the fused node:

```
ERROR: Error Code 9: Skipping tactic 0x0 due to exception CollectiveOperation not supported for collective In toMyelinCommKind at /_src/optimizer/myelin/myelinDistCollectiveLayer.cpp:49
ERROR: Error Code 10: Internal Error (Could not find any implementation for node {ForeignNode[[UNSQUEEZE]...[SHUFFLE]-[.../proj_out/_reshape_copy]]} In computeCosts ... optimizer.cpp:4284)
```

Standalone/unfused collectives build and run fine (a 2-GPU `all_reduce`+`all_to_all` compiles + runs, no TRT-LLM); the failure only appears when the collective is fused into a `ForeignNode`, which happens naturally in a real 48-layer transformer (~39k nodes) with Ulysses `all_to_all` + tensor-parallel `all_reduce`.

### To Reproduce (2 ranks)
A module with `Linear` → reshape/unsqueeze around a `dist.all_to_all_single`, so the collective is surrounded by shuffle/pointwise ops that Myelin fuses into one `ForeignNode`:

```python
import torch, torch.nn as nn, torch.distributed as dist
import torch_tensorrt

class M(nn.Module):
    def __init__(self):
        super().__init__(); self.pin = nn.Linear(64,64); self.pout = nn.Linear(64,64)
    def forward(self, x):                       # [B,S,64]
        x = self.pin(x); b,s,h = x.shape
        x = x.reshape(b*s, h).unsqueeze(0).reshape(b*s, h).contiguous()
        y = torch.empty_like(x); dist.all_to_all_single(y, x)   # -> tensorrt::fused_nccl_all_to_all
        return self.pout(y.reshape(b, s, h))

def main():
    dist.init_process_group("nccl"); r = dist.get_rank(); torch.cuda.set_device(r)
[...]

PR #4388, opened 2026-07-07
dynamo: keep native NCCL collectives out of Myelin ForeignNode fusion (cast-boundary)

pkisfaludi-nv:
**DRAFT — do not merge (author reviews).** Depends on #4380.

## Summary
On top of #4380 (which routes native NCCL collectives at the op's process **subgroup**), this adds the remaining changes that let the sharded LTX-2.3 DiT lower to a **single** TensorRT engine with native in-engine collectives (no TRT-LLM, no ONNX, no NCCL plugin). The converter changes are in `py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py`; the two new knobs are plumbed through `_defaults.py` / `_settings.py` / `_compiler.py`.

## The three changes
1. **Myelin fusion cast-boundary — `_coll_boundary()`.** Myelin fuses a native `IDistCollectiveLayer` into a `ForeignNode` and then fails with `CollectiveOperation ... not supported In toMyelinCommKind` (see #4381). Inserting a genuine **dtype reformat** on the collective's input/output creates a precision-change boundary Myelin will not fuse across (identity / same-dtype casts get folded away — only a real reformat works). This lets the full 48-block DiT compile to one engine with collectives intact. The precision is `CompilationSettings.native_collective_boundary_dtype` (default `dtype.f32`, `None` to disable).
2. **`num_ranks` before `get_output()`.** `layer.num_ranks` must be set **before** `layer.get_output(0)` so Myelin can infer the collective's output shape (e.g. all_gather `dim0 = in_dim0 * num_ranks`). Upstream sets it after, tripping `AllGather requires world_size (>1) to be set for shape inference ... Call myelinGraphSetWorldSize()`.
3. **`all_gather_via_all_reduce` (opt-in, default off).** Even with (2), TRT-11 Myelin can reject a fused native ALL_GATHER shape (it needs the graph world_size, which torch-tensorrt never sets at build). The opt-in workaround expresses all_gather as scatter-this-rank's-slice-into-a-zero-padded-buffer then `ALL_REDUCE(SUM)` — numerically identical, and ALL_REDUCE's output shape == input shape so no build-time world_size is needed. Used for the video->audio KV gather in CP. Requires a static shape on the collective's input and raises a clear error otherwise.

## Configuration
Both knobs are `CompilationSettings` fields (no environment variables), defaulted in `_defaults.py` and accepted by `dynamo.compile()` and `convert_exported_program_to_serialized_trt_engine()`:

| setting | default | effect |
| --- | --- | --- |
| `native_collective_boundary_dtype` | `dtype.f32` | precision of the reformat that walls each collective off from Myelin fusion. `None` disables it. A same-width dtype (e.g. `torch.float16` on a bf16 network) keeps the communication volume unchanged, at the cost of fp16's narrower exponent range. |
| `all_gather_via_all_reduce` | `False` | lower `all_gather` as zero-pad + `ALL_REDUCE(SUM)`. Numerically identical; `len(group)`x the communication volume; static shapes only. |

Both are in `_SETTINGS_TO_BE_ENGINE_INVARIANT`, so changing either invalidates a cached engine.

## Relationship to #4380 / #4381
- **Stacked on #4380** (subgroup routing). This PR preserves #4380's `_collective_group_ranks(group_name, ...)` group resolution and is purely additive; the diff currently also shows #4380's commit and will shrink to just these changes once #4380 lands.
- **#4381** (Myelin can't lower collectives inside a ForeignNode) is the root-cause bug; the cast-boundary here is the user-space mitigation. The durable fix is Myelin/TensorRT-side (per-subgroup communicator routing + `toMyelinCommKind`), delivered in the internal TensorRT/Myelin MRs (not in this repo).

## Validation
- **CP8**: full 22B DiT -> 1 TRT engine, `trt_engine_submodules=1`, `torch_coll_tokens=0` (zero PyTorch-side collectives), `ENGINE_RAN_OK`; parity vs eager CP8 cos 0.99602 (video) / 0.99965 (audio) — PASS. (8x A100-SXM4-80GB.)
- **CP4xTP2**: single-engine cosine-correct end-to-end when combined with the internal TensorRT/Myelin per-subgroup-comm routing.
- **2-rank regression test** `test_compute_collective_single_engine` — compute + native all-reduce compile into exactly one TRT engine, no collective escapes into the outer PyTorch graph, eager parity. `1 passed in 19.32s` on 2x A100.

> The results above were measured **before** the review-feedback refactor (env vars -> `CompilationSettings`). The default-settings behaviour is intended to be identical to the previous env-var-off configuration, but the runs have not been repeated since.

## Notes
- The cast-boundary and `all_gather_via_all_reduce` are default-safe, so CP8 and existing single-group paths are unchanged.
- Applied uniformly to the native DistCollective converters (all_gather / reduce_scatter / all_reduce / all_to_all and the scatter/gather variants) for consistency; the `num_ranks` reorder is a safe correctness fix for all.


pkisfaludi-nv:
We reproduced the follow-up `setCommunicator()` failure reported by @apbose on
Qwen3-30B-A3B and found a TensorRT core bug in addition to the original
ForeignNode build failure from #4381.

The engine does **not** derive `mIsMdEngine` from the presence of signal/wait
scaffolding. At deserialization it checks the serialized runners' `getNbRanks()`.
For a Myelin runner, both the DistCollective and Attention translators assigned
one graph-wide `mNbRanks` field directly. Therefore a rank-1 attention op
translated after a rank-2 collective could overwrite the value and make a
compute-heavy collective engine deserialize as single-device even though its
`comm_allreduce` instruction remained present.

The core fix accumulates the maximum rank count across all MD-aware operations:

```cpp
mNbRanks = std::max(mNbRanks, params.nbRanks);
mNbRanks = std::max(mNbRanks, contextParallelDegree);
```

Internal TensorRT MRs created with `git trt mr`:

- [TensorRT rel-11.2 !49080](https://gitlab-master.nvidia.com/TensorRT/TensorRT/-/merge_requests/49080)
- [TensorRT main cherry-pick !49081](https://gitlab-master.nvidia.com/TensorRT/TensorRT/-/merge_requests/49081)

The focused TensorRT regression constructs
`ALL_REDUCE(nbRanks=2) -> ATTENTION(nbRanks=1)`. The old assignment fails
`engine.isMdEngine()`; the max-rank fix passes on A100.

Related CP4xTP2 subgroup-routing MRs are:

- [Myelin per-subgroup NCCL communicators !19723](https://gitlab-master.nvidia.com/dl/compiler/myelin/-/merge_requests/19723)
- [TensorRT standalone DistCollective subgroup routing !49040](https://gitlab-master.nvidia.com/TensorRT/TensorRT/-/merge_requests/49040)
- [TensorRT fused-Myelin subgroup routing !49062](https://gitlab-master.nvidia.com/TensorRT/TensorRT/-/merge_requests/49062)

PR #4388's precision boundary is still useful: it prevents the unsupported
Myelin absorption that causes the original #4381 build failure. The new core MR
fixes the separate runtime MD-classification failure instead of relying on
isolation as a workaround.

I also added a Torch-TensorRT two-rank regression test in commit `ed8d41a6d`.
It verifies compute + native all-reduce compile into exactly one TRT engine,
that no collective escapes into the outer PyTorch graph, and eager parity.
Actual 2xA100 pytest result: `1 passed in 19.32s`.

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -80,6 +80,75 @@ def _get_distributed_rank_and_world_size() -> Tuple[int, int]:
...
+
+
+# Myelin fusion-boundary for native DistCollective. Myelin folds identity / same-dtype
+# casts away, so only a genuine dtype *reformat* keeps a native collective out of a
+# ForeignNode (see pytorch/TensorRT#4381). fp32 by default; set to None to disable, or
+# trt.float16 to keep the communication volume down.
+_COLL_BOUNDARY_DTYPE = trt.float32
+
+_coll_boundary_orig_dtype: dict = {}
+
```
apbose, line 122: this is a module global, can we scope it to per compilation ctx?

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -80,6 +80,75 @@ def _get_distributed_rank_and_world_size() -> Tuple[int, int]:
...
+    ``where='in'``  casts the collective input to the boundary dtype (the collective then
+    runs at that precision); ``where='out'`` casts the output back to the original dtype.
+    A no-op when the boundary dtype is disabled or already matches.
+    """
+    if _COLL_BOUNDARY_DTYPE is None:
+        return t
+    if where == "in":
+        _coll_boundary_orig_dtype[name] = t.dtype
+        target = _COLL_BOUNDARY_DTYPE
+    else:
```
apbose: wont fp32 boundary on non reducing collectives (eg: all_gather, all_to_all, scatter, gather) double the comm volume for fusion wall?cant we use the same size wall (eg: fp16 for bf16) or mark_debug for these movement ops?

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -263,7 +333,49 @@ def nccl_all_gather_native(
         import numpy as np
 
         # Create array of all participating rank IDs [0, 1, 2, ..., world_size-1]
-        groups = np.arange(world_size, dtype=np.int64)
+        groups = _collective_group_ranks(group_name, world_size)
+
+        if os.environ.get("AG_VIA_ALLREDUCE") == "1":
+            # Express all_gather as place-into-slot + ALL_REDUCE(SUM). The ALL_REDUCE output
```
apbose: why is this required? I understand all_gather is now converted to all_reduce, but why?

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
```diff
@@ -80,6 +80,75 @@ def _get_distributed_rank_and_world_size() -> Tuple[int, int]:
...
+    if mode in ("output", "both"):
+        ctx.net.mark_debug(output_tensor)
+
+
+# Myelin fusion-boundary for native DistCollective. Myelin folds identity / same-dtype
+# casts away, so only a genuine dtype *reformat* keeps a native collective out of a
+# ForeignNode (see pytorch/TensorRT#4381). fp32 by default; set to None to disable, or
+# trt.float16 to keep the communication volume down.
+_COLL_BOUNDARY_DTYPE = trt.float32
+
```
apbose: this can be defined in _defaults.py. Similarly NATIVE_COLLECTIVE_BOUNDARY_DTYPE
and ALL_GATHER_VIA_ALL_REDUCE can be CompilationSettings instead of env variable

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
apbose: also I think this assumes that the shape is static. in ish=raw.shape

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
pkisfaludi-nv: Done — moved out of the module and out of the environment entirely.

`NATIVE_COLLECTIVE_BOUNDARY_DTYPE` and `ALL_GATHER_VIA_ALL_REDUCE` now live in `_defaults.py`, and are exposed as `CompilationSettings.native_collective_boundary_dtype` / `CompilationSettings.all_gather_via_all_reduce`. They are plumbed through `dynamo.compile()` and `convert_exported_program_to_serialized_trt_engine()`, and added to `_SETTINGS_TO_BE_ENGINE_INVARIANT` so changing either one invalidates a cached engine (both of them change what gets built).

The converters read them off `ctx.compilation_settings`, so there are no env-var switches left in `nccl_ops.py`:
* `AG_VIA_ALLREDUCE` -> `all_gather_via_all_reduce`
* `_COLL_BOUNDARY_DTYPE` -> `native_collective_boundary_dtype` (`None` disables the reformat)
* `TRT_COLL_BOUNDARY` / `_mark_collective_boundary()` -> deleted (see the `mark_debug` thread).

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
pkisfaludi-nv: Removed rather than re-scoped — there is no state left to scope.

`_coll_boundary()` now takes the pre-cast dtype as an explicit argument, and each converter threads it:

```python
orig_dtype = plug_inputs[0].dtype
input_tensor = _coll_boundary(ctx, plug_inputs[0], name, "in")
...
output = _coll_boundary(ctx, layer.get_output(0), name, "out", orig_dtype)
```

so nothing survives past the converter call, let alone past a compilation.

This also removes a latent bug in the old keying. The dict was keyed on the layer `name`, and the `"in"` branch only wrote to it when it actually inserted a cast. If the input was already at the boundary dtype the write was skipped, and `_coll_boundary_orig_dtype.get(name, t.dtype)` on the `"out"` side could then pick up a stale entry left by an earlier compilation in the same process and cast the output to the wrong dtype.

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
pkisfaludi-nv: You are right that the cost is real, and I have made it a knob rather than a hard-coded fp32 — but I did not change the default, and I want to be explicit about why.

**The cost.** On a bf16/fp16 network the fp32 wall doubles the bytes on the wire for every collective, including the pure-movement ones (all_gather / all_to_all / scatter / gather) where the extra precision buys nothing.

**Why fp32 stays the default.** A same-width wall is not a free substitution:
* bf16 -> fp16 is not a value-preserving reformat. fp16 tops out at 65504 while bf16 carries fp32's exponent range, so activations that are perfectly healthy in bf16 can go to `inf` crossing the wall. That is a silent numerical failure, and it would apply to every collective in the graph.
* For the *reducing* collectives (all_reduce / reduce_scatter) the wall dtype is also the accumulation dtype, so lowering it changes results, not just transport.

So the wall precision is now `CompilationSettings.native_collective_boundary_dtype`:
* default `dtype.f32` — the configuration this PR was validated with;
* set it to `torch.float16` on a bf16 network and you get exactly the same-size wall you are describing, with no change in communication volume;
* set it to `None` to emit no reformat at all.

The trade-off is written up on the setting's docstring so the choice is visible at the API rather than buried in the converter.

I also added a warning for the case where the configured wall dtype equals the collective input's dtype: `cast_trt_tensor` folds that cast away, so there is effectively no barrier (e.g. an fp32 network with the fp32 default). Previously that just silently produced no wall.

**On `mark_debug`:** I removed `_mark_collective_boundary()` and its `TRT_COLL_BOUNDARY` env var instead of promoting it. Two reasons: it is a debugging facility (it turns the tensor into a debug tensor and expects `IDebugListener` wiring at runtime), and it is not documented as a fusion barrier — I have not validated that it actually prevents the `ForeignNode` absorption, so I did not want a correctness-critical partitioning decision resting on it. If you would rather keep it as a documented alternative I am happy to reinstate it as a third setting; I just did not want to ship an untested mechanism as an option.

py/torch_tensorrt/dynamo/conversion/impl/nccl_ops.py
pkisfaludi-nv: Two separate points, both addressed — thanks.

**Why the ALL_REDUCE rewrite exists at all.** It is not that all_gather is generally converted to all_reduce; it is an opt-in escape hatch for a shape-inference gap. TensorRT infers a native `ALL_GATHER`'s output shape from the *graph* world size (`myelinGraphSetWorldSize`), which Torch-TensorRT never sets at build time. Setting `ILayer.num_ranks` on the layer is enough for the standalone layer but not for the fused path, which still fails with `AllGather requires world_size (>1) to be set for shape inference`. `ALL_REDUCE`'s output shape is simply its input shape, so it needs no build-time world size at all. Each rank therefore materialises the *gathered* buffer explicitly — zeros | its own slice at its own slot | zeros — and sums it over the group, which is bit-identical to the concatenation ALL_GATHER would have produced, at `len(groups)`x the communication volume.

That reasoning was only in the PR description before; I have pulled the path out into `_all_gather_as_all_reduce()` with the explanation in its docstring, and it is now gated by `all_gather_via_all_reduce` (default `False`) rather than an env var.

**Static shapes — accepted, this was a real bug.** `ish = [int(x) for x in raw.shape]` gives `-1` for a dynamic dim, and the zero blocks are `add_constant`s, so a dynamic input produced a negative extent instead of a diagnosable error. It matters more now that these converters are registered `supports_dynamic_shapes=True` (#4402). `_all_gather_as_all_reduce()` now checks for dynamic dims up front and raises a `RuntimeError` naming the collective, its shape, and the setting to turn off.

I did *not* make the path dynamic-shape capable. Doing it properly means building the padded buffer from a shape tensor — `ISliceLayer` in `SampleMode.FILL` with a computed start/size, or `IFillLayer` fed by `IShapeLayer` — and I have no multi-GPU hardware available to validate that right now. Given the path is opt-in, off by default, and exists only to work around a TensorRT-side shape-inference limitation, an explicit error seemed better than untested network-building code. Say the word if you would rather have the dynamic version in this PR.

pkisfaludi-nv:
@apbose thanks for the review — all four threads are answered inline, and the changes are pushed as `c19abf7` (rebased onto current `main`, which also picks up `supports_dynamic_shapes=True` on these converters from #4402).

Summary of what changed:

| Your comment | Outcome |
| --- | --- |
| move the constants to `_defaults.py`, make them `CompilationSettings` not env vars | **Accepted.** `NATIVE_COLLECTIVE_BOUNDARY_DTYPE` / `ALL_GATHER_VIA_ALL_REDUCE` in `_defaults.py`; `native_collective_boundary_dtype` / `all_gather_via_all_reduce` on `CompilationSettings`, plumbed through `compile()` and `convert_exported_program_to_serialized_trt_engine()`, and added to `_SETTINGS_TO_BE_ENGINE_INVARIANT`. No env vars left in `nccl_ops.py`. |
| `_coll_boundary_orig_dtype` is a module global, scope it to the compilation ctx | **Accepted, and gone entirely** — the pre-cast dtype is threaded as an explicit `orig_dtype` argument instead of being stashed anywhere. Also fixes a stale-entry bug in the old name-keyed dict. |
| fp32 boundary doubles comm volume on movement collectives; use a same-size wall or `mark_debug` | **Partially accepted.** The wall dtype is now user-selectable, so `torch.float16` on a bf16 network gives exactly the same-size wall. I did **not** change the default to a same-width dtype: bf16 -> fp16 is not value-preserving (overflow above 65504) and fp32 is also the accumulation dtype for the reducing collectives. `mark_debug` was removed rather than promoted — reasoning in the thread. |
| the ALL_REDUCE rewrite assumes static shapes | **Accepted — it was a real bug.** It now raises a clear `RuntimeError` on dynamic dims instead of emitting a network with a negative extent. Full dynamic-shape support (shape-tensor-driven padding) is not in this PR; happy to add it if you want it here. |
| why is the all_gather -> all_reduce rewrite required at all? | **Documented.** Extracted into `_all_gather_as_all_reduce()` with the reasoning in the docstring (ALL_GATHER output shape needs the graph world size, which Torch-TensorRT never sets; ALL_REDUCE's output shape == its input shape). |

Two extras while I was in there: a warning when the configured boundary dtype equals the collective input's dtype (the cast is folded away, so there is no barrier at all — e.g. an fp32 network with the fp32 default), and the black formatting that "Python Linting" flags on this branch is fixed. `black --check .`, `ruff`, `isort` and `mypy` are clean on the touched files (the 6 remaining `Untyped decorator` mypy errors in `nccl_ops.py` also reproduce on `main`).

Caveat, stated plainly: **the CP8 / CP4xTP2 numbers in the description predate this refactor and have not been re-run** — I do not have multi-GPU hardware available at the moment. Default-settings behaviour is meant to be identical to the previous env-var-off configuration, but that has not been re-measured.

Still blocked on: this stacks on #4380, and the workflow runs on this push are sitting in `action_required` (they need a maintainer to approve running workflows for the fork).

pkisfaludi-nv:
Marked ready for review and rebased onto current `main` (`8e48830`, was 20 commits behind). All of @apbose's review comments are addressed — replies are on each thread; a summary of what changed and the one suggestion I did not take (defaulting the fusion wall to a same-width dtype, since bf16→fp16 is not value-preserving) is in the thread replies.

Two things block this from going further, neither of which I can do from here:

1. **CI has never run on this PR.** All six workflow check suites are in `action_required` — a maintainer needs to approve workflow runs for the fork. Only `label` and `Meta CLA Check` have executed, on this push and every earlier one, so there is currently no pass/fail signal for this change. The CLA is now signed.
2. **It stacks on #4380**, which is still open. This branch therefore carries #4380's commit; the diff shrinks once that merges.

Repo-pinned `black` (26.3.1) is clean across the six files this PR touches. The one remaining repo-wide `black` complaint is `tests/py/dynamo/lowering/test_complex_rewrite.py`, which reproduces unchanged on `main` and is untouched here.

Not re-validated on hardware since the refactor — the CP8 / CP4xTP2 numbers in the description predate it, and default behaviour is *intended* to match the previous env-var-off configuration but has not been measured.

pkisfaludi-nv:
@narendasan — same ask as on #4380: could you approve the workflow runs here? All five workflows are in `action_required`, so this PR has never had CI on any commit — only `label` and `Meta CLA Check` have run.

Worth merging **#4380 first**: this branch stacks on it and currently carries its commit, so the diff is larger than the change it describes. It shrinks once #4380 lands and this is rebased.

State otherwise: not draft, `MERGEABLE`, rebased onto current `main`, and all of @apbose's review comments are addressed — the remaining unresolved thread is answered and marked outdated by GitHub.

One thing I want to flag rather than leave for review to find: the CP8 / CP4xTP2 numbers in the description predate this PR's own refactor, from when these were environment variables rather than `CompilationSettings`. Default behaviour is *intended* to be identical to the previous env-var-off configuration, but I have not re-measured it — the machine that did the refactor has no GPUs. That is another reason the CI run matters here.

pkisfaludi-nv:
Closing in favour of #4380, which now carries everything from this PR that is still needed. @jloftin-nvidia @apbose — you were right that the main change here is a remnant; here is the check.

**Moved to #4380**
- The subgroup-routing commit (`6485a203d4`) is patch-identical to #4380's `1d341430a2` and lives on there.
- `layer.num_ranks` before `layer.get_output(0)` — still wrong on `main` in all six converters, so it is now in #4380 with a CPU-level test that records the assignment order.

Worth flagging: this PR was missing #4380's fix to stop sorting the group's rank list. `get_process_group_ranks()` returns ranks ordered by group rank, so sorting silently reassigns roles. Merging this one alone would have shipped that.

**Dropped: the Myelin cast boundary**
`toMyelinCommKind()` had no arm for `kALL_TO_ALL`; TRT `0ea60bdc42` ("Fix BF16 A2A Myelin Conversion Error") added it, and that was the only op missing.

Confirmed on hardware (8×A100-80GB, TRT 11.2.0.86) with a minimal TRT-Python network — elementwise → relu → `add_dist_collective` → elementwise:

| collective | in the switch | result |
|---|---|---|
| `BROADCAST` | no | `BUILD_FAILED` — `CollectiveOperation not supported for collective In toMyelinCommKind at myelinDistCollectiveLayer.cpp:52` |
| `ALL_TO_ALL` bf16, no boundary | yes | `BUILD_OK`, collective absorbed into the Myelin region (`a2a_myl0_5`) |
| `ALL_TO_ALL` bf16, with boundary | yes | `BUILD_OK`, collective standalone (`__myl_MoveAddReluCast_myl0_0 | a2a | __myl_CastAdd_myl2_0`) |

`BROADCAST` is the positive control — the harness does reproduce the failure, at the exact line. Row 2 is the case the boundary exists to prevent, and it builds with the collective genuinely fused into the ForeignNode; row 3 confirms the boundary is what stops that fusion. So the workaround no longer buys anything, and it was default-on, widening every collective to fp32 for all users.

**Deferred: `all_gather_via_all_reduce`**
Depends on the boundary, default-off, no tests, and hard-raises on dynamic shapes — which contradicts #4402 marking these converters `supports_dynamic_shapes=True`. Preserved at `8e48830121` if anyone wants to pick it up.

pkisfaludi-nv:
Correction to my comment above, after testing this on hardware rather than reasoning from the source.

I said the `num_ranks`-before-`get_output(0)` reordering was a real bug still present on `main`. It is not, and I have dropped it from #4380.

The output shape genuinely is stale at the instant of capture — an all_gather reads `(2,64)` instead of `(8,64)` — but TensorRT re-evaluates it lazily, the two lines are adjacent on `main` with nothing reading the shape in between, and the built engines come out byte-identical either way (8× A100, TRT 11.2.0.86):

| op | `num_ranks` first | `get_output` first | identical |
|---|---|---|---|
| `ALL_GATHER` | `(16,64)` md5 `0ebe6c0b97d4` | `(16,64)` md5 `0ebe6c0b97d4` | yes |
| `ALL_REDUCE` | `(4,64)` md5 `18e50672a18a` | `(4,64)` md5 `18e50672a18a` | yes |
| `ALL_TO_ALL` | `(4,64)` md5 `a0ad8806638d` | `(4,64)` md5 `a0ad8806638d` | yes |
| `REDUCE_SCATTER` | `(1,64)` md5 `547445add902` | `(1,64)` md5 `547445add902` | yes |

What was substantive in that hunk is `num_ranks = len(groups)` rather than `world_size`, and that is kept in #4380 — it is inseparable from routing a collective at a subgroup. So this PR's remaining contribution is fully represented there; nothing further is lost by closing it.

PR #4398, opened 2026-07-11
add torch-tensorrt-executorch-runtime wheel build

lanluo-nvidia:
## Description

Adds a separately packaged `torch-tensorrt-executorch-delegate` wheel that bundles the ExecuTorch Python portable runtime with `TensorRTBackend` registered. This allows an exported `.pte` model to be loaded and executed with `torch_tensorrt.load(..., format="executorch")` without requiring an ExecuTorch source checkout or a native build on the deployment system.

The main `torch-tensorrt` package continues to own compilation and export. ExecuTorch runtime loading and its native libraries are isolated in the optional delegate package.

### ExecuTorch delegate package and build

- Add the `torch-tensorrt-executorch-delegate` Python package.
- Build the delegate's native runtime with Bazel and CMake.
- Bundle `_portable_lib.so` and `data_loader.so` in the wheel.
- Force every `pip wheel` invocation to rebuild and recopy the native libraries, preventing stale `.so` files from being packaged.
- Add Bazel targets and root source mappings required by the delegate build.

### Package and API separation

- Keep ExecuTorch compile/export functionality in the main `torch-tensorrt` package.
- Move `Program`, `.pte` loading, runtime activation, and native module handling into the delegate package.
- Add `format="executorch"` support to `torch_tensorrt.load`.
- Lazily check for the delegate package when ExecuTorch loading is requested and provide an installation error when it is unavailable.
- Add the delegate package to the Linux `executorch` and `all` extras.

### Runtime activation

- Load the bundled backend-enabled portable runtime in place of ExecuTorch's stock `_portable_lib`.
- Register the bundled data-loader module under ExecuTorch's canonical module path before portable-runtime initialization.
- Avoid duplicate pybind registration of `PyDataLoader`.
- Detect incompatible import ordering and report a targeted compatibility error.
- Validate that `TensorRTBackend` is present in the runtime backend registry.

### ExecuTorch version and dependencies

- Pin the Bazel ExecuTorch source to commit `e2f18eb23c45bd22ca332b0b8b49a81de304b472`, matching `executorch==1.3.1`.
- Declare exact compatible versions of `torch`, `executorch`, and `torch-tensorrt` for the delegate wheel.
- Update the CI workspace module configuration to use the same ExecuTorch source.

### CI, examples, and documentation

- Build and install the delegate wheel in Linux wheel CI.
- Add backend-registration and Python load/run smoke tests.
- Update the static ExecuTorch workflow to export once and verify the same `.pte` model with both the C++ reference runner and the Python delegate runtime.
- Update `load_model.py` to accept `--model_path` and `--num_runs`.
- Document how to build, install, save, and load an ExecuTorch model with the delegate wheel.

### Tests

- Add coverage for ExecuTorch load API dispatch and optional-package behavior.
- Add runtime tests for model loading and execution, unknown methods, missing files, idempotent activation, stock-runtime import collisions, backend registration, and data-loader rollback.
- Update serialization and packaging tests for the separated delegate distribution.

## Validation

- Built the native Bazel delegate target successfully.
- Built and installed a fresh delegate wheel.
- Passed the focused ExecuTorch API, runtime, serialization, and packaging tests.
- Loaded `model.pte` through `torch_tensorrt.load(..., format="executorch")` and verified the TensorRT delegate result against the expected `x + 1` output.

## Type of change

- [x] New feature
- [x] Bug fix
- [x] Documentation update
- [x] Test and CI coverage


narendasan:
@lanluo-nvidia the new package should be in 

//py
-> torch-tensorrt 
-> torch-tensorrt-executorch-delegate 

See: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/

MODULE.bazel
```diff
@@ -42,12 +42,12 @@ new_git_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:git.bzl"
...
 
-# Pinned to the ExecuTorch release/1.3 branch head.
+# Pinned to ExecuTorch main for PyTorch 2.14 native ABI compatibility.
 new_git_repository(
     name = "executorch",
     build_file = "@//third_party/executorch:BUILD",
-    # latest commit in release/1.3 branch
-    commit = "6118688a095fd9697224f5cad72ce42db641c9cd",
+    # PyTorch 2.14-compatible ExecuTorch main commit
+    commit = "a6d812a082df57898b8608f56c867140cc9da32c",
```
shoumikhin: Blocking: this change never takes effect in CI, so the build still uses the old ExecuTorch.

`packaging/pre_build_script.sh` line 84 rebuilds this whole file from a template:

```bash
cat toolchains/ci_workspaces/MODULE.bazel.tmpl | envsubst > MODULE.bazel
```

That template (`toolchains/ci_workspaces/MODULE.bazel.tmpl`) still points at the old commit `6118688`. So when CI runs, it overwrites your edit and goes back to the old version, and that old version is what the delegate gets compiled against. It also means it works on your machine but not in CI.

Please put the new commit in `toolchains/ci_workspaces/MODULE.bazel.tmpl` (and the Windows template too if there is one). The root `MODULE.bazel` is generated, so it is cleaner to leave it alone and only edit the template.

.github/workflows/executorch-static-linux.yml
```diff
@@ -96,5 +97,15 @@ jobs:
...
-        .github/scripts/verify-executorch-reference-runner.sh
+        python -m pip install pyyaml executorch
+
+        # export the model to a .pte file
+        python examples/torchtrt_executorch_example/export_static_shape.py --model_path="${RUNNER_TEMP}/torchtrt-python.pte"
+        # verify the c++ reference runner can run the model
+        .github/scripts/verify-executorch-reference-runner.sh "${RUNNER_TEMP}/torchtrt-python.pte"
+
+        # Build the no-compile-for-users Python runtime/delegate wheel.
+        python -m pip wheel --no-build-isolation --no-deps --wheel-dir dist packaging/executorch_delegate
```
shoumikhin: Blocking: this step builds the delegate wheel, but the wheel build needs a file that this job never creates, so it will fail here.

The delegate's `setup.py` calls `torchtrt_version()`, which reads `py/torch_tensorrt/_version.py`. That file is not checked in. It is written only by the main `setup.py`. This job runs `bazel build //:libtorchtrt` (C++ only) and never runs the main `setup.py`, so the file is missing and `torchtrt_version()` raises `RuntimeError("Could not determine the Torch-TensorRT version")`. Anyone following the README to build the wheel on its own hits the same error.

A simple fix is to fall back to `version.txt` and the env override when the file is not there:

```python
def torchtrt_version() -> str:
    if value := os.getenv("TORCH_TENSORRT_EXECUTORCH_DELEGATE_VERSION"):
        return value
    version_py = REPO_ROOT / "py/torch_tensorrt/_version.py"
    if version_py.exists():
        if m := re.search(r'__version__\s*=\s*["\']([^"\']+)', version_py.read_text()):
            return m.group(1)
    return (REPO_ROOT / "version.txt").read_text().strip()
```

tests/py/executorch/test_python_runtime.py
```diff
@@ -0,0 +1,64 @@
+import importlib.util
+import sys
+import types
+from pathlib import Path
+
+import pytest
+
+RUNTIME_PATH = Path(__file__).parents[3] / "py/torch_tensorrt/executorch/runtime.py"
```
shoumikhin: Should fix: this test never runs in CI, so it will not catch regressions.

The ExecuTorch test group in `tests/py/utils/ci_helpers.sh` lines 157 to 162 does `cd tests/py/dynamo` and then runs `pytest executorch/`, so it only picks up tests in `tests/py/dynamo/executorch/`. This file is in `tests/py/executorch/`, a different folder that nothing runs.

On top of that, the test replaces the real runtime with a fake (`FakeRuntime`), so even if it did run it would not touch the tricky new code: `activate()`, the `sys.modules` swap, or the failure path.

Please move it under `tests/py/dynamo/executorch/`, and add a few plain Python tests (no GPU needed) that use fake modules to check: calling `activate()` twice is safe, it raises `DelegateCompatibilityError` when the stock runtime was already loaded, and it does not leave a leftover `data_loader` entry when the native import fails.

shoumikhin (commented):
Nice work on this. The core idea is right. You build `_portable_lib` from source and pull `TensorRTBackend` into the same module that holds the backend registry, so the two always share one copy of the ExecuTorch core. That is the correct way to fix the `undefined symbol` crash people hit today. I checked the linking part and it holds up: one registry, no duplicate core, and the whole-archive flag is actually needed.

packaging/executorch_delegate/torch_tensorrt_executorch_delegate/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    if existing is not None:
+        if existing.__name__ != __name__ + "._portable_lib":
+            raise DelegateCompatibilityError(
+                "ExecuTorch's stock runtime was imported first. Import "
+                "torch_tensorrt.executorch.runtime before executorch.runtime."
+            )
+        return existing
+    try:
+        data_loader = importlib.import_module(__name__ + ".data_loader")
+        sys.modules[_DATA_LOADER_NAME] = data_loader
```
shoumikhin, line 63: Should fix: this can leave Python half set up if the native import fails.

`activate()` puts `data_loader` into `sys.modules` first, then imports `_portable_lib`. If that second import fails (the same crash this wheel is meant to fix), the error is raised but the `data_loader` entry is left behind, and from then on anything importing the normal ExecuTorch `data_loader` gets your version instead.

Import both first, and only register them once both succeed:

```python
try:
    data_loader = importlib.import_module(__name__ + ".data_loader")
    native = importlib.import_module(__name__ + "._portable_lib")
except ImportError as error:
    sys.modules.pop(_DATA_LOADER_NAME, None)   # do not leave a half state
    raise DelegateCompatibilityError(...) from error
sys.modules[_DATA_LOADER_NAME] = data_loader
sys.modules[_NATIVE_NAME] = native
sys.modules.pop(_WRAPPER_NAME, None)
```

Separate note, not a change request: the `existing.__name__` check just above is actually correct and safe to call more than once. I built a small single-phase C extension to confirm its `__name__` comes out as the full `pkg._portable_lib`, so repeat calls do not throw. Worth leaving that one as is.

py/torch_tensorrt/executorch/runtime.py
```diff
@@ -0,0 +1,58 @@
...
+    def forward(self, *inputs: Any) -> Sequence[Any]:
+        return self.run(inputs, "forward")
+
+
+def load(path: Union[str, Path]) -> Program:
+    """Load a `.pte` with the delegate-enabled ExecuTorch Python runtime."""
+    model_path = Path(path)
+    if not model_path.is_file():
+        raise FileNotFoundError(f"ExecuTorch model not found: {model_path}")
+    return Program(_runtime().load_program(model_path.read_bytes()))
```
shoumikhin: Should fix: small lifetime question. `load_program(model_path.read_bytes())` reads the file into a bytes object and passes it straight in, without keeping a reference to it. If ExecuTorch points at that memory instead of copying it, the bytes get freed right after this line and you get a use after free.

Can you check whether `load_program` copies the data? If it does not, keep the bytes alive by storing them on the returned `Program`, for example `self._data = data`.

py/torch-tensorrt-executorch-runtime/torch_tensorrt_executorch_runtime/runtime.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @property
+    def method_names(self):
+        return self._program.method_names
+
+    def run(self, inputs: Sequence[Any], method: str = "forward") -> Sequence[Any]:
+        import torch
+
+        inputs = tuple(
+            value.cpu() if isinstance(value, torch.Tensor) and value.is_cuda else value
```
shoumikhin, line 52: Should fix: this quietly moves every input to CPU, which turns off the GPU fast path in the C++ backend.

The C++ code has a path that keeps tensors on the GPU and avoids extra copies. By calling `.cpu()` on every input here, that path can never be taken, so every run copies data to the GPU and back, and outputs always come back on CPU. That is probably unavoidable because the portable runtime uses CPU tensors, but right now it is silent and a bit surprising for something called a TensorRT GPU delegate.

Please add a note in the docstring and README that the Python runtime runs through CPU, and that the on-GPU path is only in the C++ runner.

py/torch-tensorrt-executorch-runtime/torch_tensorrt_executorch_runtime/runtime.py
```diff
@@ -0,0 +1,58 @@
...
+        loaded = self._program.load_method(method)
+        if loaded is None:
+            raise RuntimeError(f"ExecuTorch failed to load method {method!r}")
+        return loaded.execute(inputs)
+
+    def forward(self, *inputs: Any) -> Sequence[Any]:
+        return self.run(inputs, "forward")
+
+
+def load(path: Union[str, Path]) -> Program:
```
shoumikhin, line 68: Should fix (document): `load()` only handles a program whose weights are inside the file. It calls `load_program(bytes)` and has no way to pass a separate weights file (a `.ptd`), so a model that keeps its weights outside the `.pte` cannot be loaded.

That is fine for this PR, since the engine is stored inside the blob. Please just add a one line note that external weight files are not supported yet. This only becomes a blocker if loading external `.ptd` files is needed for launch.

py/torch-tensorrt-executorch-runtime/native/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+  "${TORCH_PRIMARY_INCLUDE_DIR}/torch/headeronly"
+  "${CMAKE_BINARY_DIR}/torch_header_shim/torch/headeronly"
+  SYMBOLIC COPY_ON_ERROR)
+target_include_directories(util BEFORE PRIVATE "${CMAKE_BINARY_DIR}/torch_header_shim")
+target_include_directories(portable_lib BEFORE PRIVATE "${CMAKE_BINARY_DIR}/torch_header_shim")
+
+# TensorRTBackend registers through a static initializer. Force its complete
+# archive into the same pybind module that owns ExecuTorch's backend registry.
+target_link_libraries(portable_lib PRIVATE
+  "$<LINK_LIBRARY:WHOLE_ARCHIVE,torch_tensorrt_executorch_backend>"
```
shoumikhin, line 92: Should fix: the shipped `_portable_lib.so` needs `libnvinfer` and `libcudart` at load time but has no path baked in to find them. The CI repair step only renames the file from `linux_` to `manylinux_2_28_`. It does not run `auditwheel` and does not set any library search path. CI works because it sets `LD_LIBRARY_PATH` by hand, but a normal `pip install` may not, and then `import` fails.

To be clear, not bundling torch and c10 is the right call here, because bundling them would bring back the duplicate c10 problem this PR fixes. This is only about TensorRT and CUDA, and about the fact that the `manylinux` label is now just a name. Please document that TensorRT and CUDA need to be findable at runtime, and consider an `$ORIGIN` based path if you plan to ship this on an index.

py/torch-tensorrt-executorch-delegate/native/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+
+if(NOT TARGET portable_lib)
+  message(FATAL_ERROR "ExecuTorch did not define portable_lib")
+endif()
+
+# ExecuTorch portable c10 delegates selected headers to torch/headeronly. The
+# pybind bridge must resolve those delegated headers from the exact PyTorch
+# wheel it links, not ExecuTorch's vendored snapshot.
+file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/torch_header_shim/torch")
+file(CREATE_LINK
```
shoumikhin: Should fix: this can silently bring back the ABI bug it is trying to prevent. `file(CREATE_LINK ... SYMBOLIC COPY_ON_ERROR)` makes a symlink without checking the target exists. If `${TORCH_PRIMARY_INCLUDE_DIR}/torch/headeronly` is missing or wrong, you get a broken link and the build quietly falls back to ExecuTorch's bundled copy of those headers, which is the version mismatch this shim exists to stop.

Please check the path exists and stop the build if it does not:

```cmake
if(NOT EXISTS "${TORCH_PRIMARY_INCLUDE_DIR}/torch/headeronly")
  message(FATAL_ERROR "torch/headeronly not found under TORCH_PRIMARY_INCLUDE_DIR=${TORCH_PRIMARY_INCLUDE_DIR}")
endif()
```

py/torch-tensorrt-executorch-runtime/setup.py
```diff
@@ -0,0 +1,110 @@
...
+    description="Torch-TensorRT delegate for the ExecuTorch Python runtime",
+    packages=find_packages(),
+    ext_modules=[
+        CMakeExtension("torch_tensorrt_executorch_delegate._portable_lib"),
+        CMakeExtension("torch_tensorrt_executorch_delegate.data_loader"),
+    ],
+    cmdclass={"build_ext": CMakeBuild},
+    python_requires=">=3.10",
+    install_requires=[
+        f"torch=={torch.__version__}",
```
shoumikhin: Should fix: two things here can make the wheel hard to install.

First, `torch=={torch.__version__}` includes the local build suffix (like `2.14.0.dev...+cu126`). Many indexes drop that suffix, and then this exact `==` pin cannot be matched.

Second, the `.so` is built against a specific TensorRT and CUDA, but neither is listed as a dependency, so pip is free to pair your wheel with a TensorRT that does not match, which crashes at runtime.

Please pin torch without the local suffix (or use a range), and add TensorRT and CUDA constraints that match what the `.so` was built with.

py/torch-tensorrt-executorch-delegate/setup.py
```diff
@@ -0,0 +1,110 @@
...
+        if sys.platform != "linux":
+            raise RuntimeError("The ExecuTorch TensorRT delegate supports Linux only")
+        executorch_source = os.getenv("EXECUTORCH_SOURCE_DIR")
+        if not executorch_source:
+            raise RuntimeError(
+                "Set EXECUTORCH_SOURCE_DIR to the pinned ExecuTorch source tree"
+            )
+
+        output = pathlib.Path(self.get_ext_fullpath(ext.name)).resolve()
+        if output.exists():
```
shoumikhin: Minor: `if output.exists(): return` skips the build when the `.so` is already there. On a clean CI run that is fine, but on a local rebuild it can ship an old `.so` from a previous build without telling you. Consider a build stamp, or just remove the early return.

packaging/executorch_delegate/torch_tensorrt_executorch_delegate/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    """The delegate wheel is incompatible with the active native runtime."""
+
+
+def activate() -> ModuleType:
+    """Make the delegate-enabled portable runtime back ``executorch.runtime``."""
+    existing = sys.modules.get(_NATIVE_NAME)
+    if existing is not None:
+        if existing.__name__ != __name__ + "._portable_lib":
+            raise DelegateCompatibilityError(
+                "ExecuTorch's stock runtime was imported first. Import "
```
shoumikhin: Minor: the docstring implies importing `torch_tensorrt.executorch.runtime` reserves the slots, but activation is lazy and runs the first time you call `load()`. So the real rule is: call a delegate function before you import `executorch.runtime`. Please reword that, and maybe also have the check look at `_WRAPPER_NAME`.

py/torch-tensorrt-executorch-runtime/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Build
+
+```bash
+executorch_cmake_location="$(bazel query \
+  @executorch//:executorch/CMakeLists.txt --output=location)"
+export EXECUTORCH_SOURCE_DIR="$(dirname "${executorch_cmake_location%%:*}")"
+export TensorRT_ROOT=/path/to/TensorRT
+
+python -m pip install executorch==1.3.1
```
shoumikhin: Minor: the README says `pip install executorch==1.3.1`, but `MODULE.bazel` now targets main for the 2.14 ABI. Please make the doc match the version you actually build against.

packaging/executorch_delegate/pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[build-system]
+requires = [
+    "setuptools>=68",
+    "wheel>=0.40",
+    "torch",
```
shoumikhin: Minor: `torch` is listed here without a version. It does not matter in CI because CI uses `--no-build-isolation`, but a normal build would pull the newest torch into a clean build environment and mismatch the ABI. Please pin it, or say clearly that `--no-build-isolation` is required.

.gitignore
```diff
@@ -86,3 +86,6 @@ CLAUDE.md
 /compile_commands.json
 /external/
 build-executorch/
+packaging/executorch_delegate/build/
+packaging/executorch_delegate/dist/
```
shoumikhin: Minor: this adds a trailing blank line at the end of the file.

shoumikhin:
One more note, on a file this PR does not change so I cannot leave it inline. In `cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp` around line 653, the result of `register_backend` is stored in `kRegistered` but never checked, so if registration fails you find out much later with a confusing "Backend TensorRTBackend is not registered" message. It would help to notice the failure at registration time. One thing to watch out for: this runs very early, before logging is set up, so do not call `ET_LOG` here. A safer option is to save the result and check it the first time the backend is used.

py/torch-tensorrt-executorch-runtime/torch_tensorrt_executorch_runtime/__init__.py
```diff
@@ -0,0 +1,55 @@
+"""Activate the TensorRT delegate-enabled ExecuTorch Python runtime."""
```
narendasan, line 1: @lanluo-nvidia this code should be in //py/torch-tensorrt-executorch-delegate not in packaging

packaging/executorch_delegate/pyproject.toml
narendasan: Do we not depend on executorch for the delegate?

py/torch-tensorrt-executorch-delegate/setup.py
```diff
@@ -0,0 +1,110 @@
...
+            return
+        build_dir = pathlib.Path(self.build_temp) / "executorch_delegate_native"
+        output.parent.mkdir(parents=True, exist_ok=True)
+        build_dir.mkdir(parents=True, exist_ok=True)
+        configure = [
```
narendasan: Why arent we using bazel here like the main package?

narendasan: I think it would be better to have one real preferred build system

py/torch-tensorrt-executorch-runtime/torch_tensorrt_executorch_runtime/runtime.py
```diff
@@ -0,0 +1,58 @@
...
+        from torch_tensorrt_executorch_delegate import runtime
+    except ImportError as error:
+        raise ImportError(
+            "ExecuTorch Python inference requires the prebuilt delegate. "
+            'Install it with: pip install "torch-tensorrt[executorch]"'
+        ) from error
+    return runtime()
+
+
+class Program:
```
narendasan, line 23: Should probably be in the delegate not this package

py/torch-tensorrt-executorch-runtime/torch_tensorrt_executorch_runtime/runtime.py
narendasan, line 68: Also in the delegate and not this package. we should just expose via torch_tensorrt.load

setup.py
```diff
@@ -189,6 +183,16 @@ def load_dep_info():
 else:
     __version__ = f"{get_base_version()}.dev0+{get_git_revision_short_hash()}"
 
+EXECUTORCH_REQUIREMENT = "executorch>=1.3.1"
+EXECUTORCH_DELEGATE_REQUIREMENT = (
+    f"torch-tensorrt-executorch-delegate=={__version__}; "
+    "platform_system == 'Linux' and platform_machine == 'x86_64'"
```
narendasan: We need to support ARM as well (Jetson Thor)

py/torch-tensorrt-executorch-runtime/torch_tensorrt_executorch_runtime/runtime.py
narendasan: We can lift apis into the main package gated by import but any dependent code should be in the sub package

.github/workflows/_linux-x86_64-core.yml
narendasan, line 1: I think this should be a separate downstream job and it gates the ET tests 

So its like 
```
`build core` -> `test core`
             -> `build et` -> `test et`
```

lanluo-nvidia: Makes sense. I’ll remove the delegate build from the shared core-wheel job and introduce a standard-TRT-only downstream ET build job. The ET test job will depend on the ET wheel build and consume both the core and delegate artifacts, while the existing core test jobs continue to depend only on the core build. I’ll also move the ExecuTorch pytest group out of the L2 core helper and add both ET jobs to the CI rollup.

shoumikhin:
Net: let's land this as the TensorRT-only Python runtime and treat the coalesced TRT+CUDA case as a separate follow-up that we'll drive, so it doesn't block you here. Quick context on the two ExecuTorch-side features in play, since both postdate this PR:

- Coalesced delegates: ExecuTorch can split one exported model across more than one backend, so a single `.pte` can have some partitions running on our `TensorRTBackend` and the rest on ExecuTorch's own CUDA (AOTInductor) backend. Running such a model needs BOTH backends registered in the same runtime, in one process. This path registers only `TensorRTBackend` (`native/CMakeLists.txt` builds no CUDA backend and whole-archives only ours), and `activate()` swaps the stock runtime out via `sys.modules` and refuses if the stock one was imported first, so our backend and ExecuTorch's `CudaBackend` can't be alive together. So a coalesced `.pte` can't run through here yet.

- External `.ptd` weights: this is a new, still-WIP ExecuTorch feature where a delegate stores its weight constants outside the `.pte`, in a separate `.ptd` file next to it (for example the CUDA/AOTInductor backend with `package_constants_in_so=False` emits an `aoti_cuda_blob.ptd`). Coalesced models rely on it for their CUDA weights. Since it's WIP, the only impact here is that `load()` reads just the `.pte` bytes (`load_program(data)`) with no channel for a `.ptd`, so a model that keeps weights outside the `.pte` would load with no weights. Nothing to build for it now; just don't let this path silently mishandle such a model.

Both are out of scope for a TensorRT-only runner, so to land:
1. Pin one executorch commit: keep it in `MODULE.bazel.tmpl` (root is generated from it) and make the README and `pip install executorch==...` match, so the `.so` is built and shipped against the same ExecuTorch.
2. Detect a `.pte` that needs a backend or external weights this runtime doesn't provide (a coalesced or external-`.ptd` program) and refuse it with a clear error, and document this as the TensorRT-only path, not the general `.pte` runner.
3. Make `pip install` self-sufficient: pin `torch` without the local `+cuXXX` suffix, and either bake an `$ORIGIN` rpath or document that TensorRT and CUDA must be on the library path (today it only works with a hand-set `LD_LIBRARY_PATH`).
4. While we're pre-publish and renames are cheap: can we name the package `torch-tensorrt-executorch-runtime` instead of `-delegate`? It ships a whole backend-enabled runtime, not just a delegate shim, so `-runtime` reads more accurately, and it's far easier to change now than once it's on PyPI.

Everything else I raised (external `.ptd` support, auditwheel, the build nonce, the CPU-copy doc note) is fine as fast-follows, and let's just confirm the aarch64/Jetson wheel actually builds and is tested. We'll drive the coalesced + `.ptd` follow-up separately so it stays off your critical path.

MODULE.bazel
```diff
@@ -42,12 +42,13 @@ new_git_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:git.bzl"
...
-# Pinned to the ExecuTorch release/1.3 branch head.
+# Keep this pin synchronized with the ExecuTorch release installed in
+# py/torch-tensorrt-executorch-delegate/README.md.
 new_git_repository(
     name = "executorch",
     build_file = "@//third_party/executorch:BUILD",
-    # latest commit in release/1.3 branch
-    commit = "6118688a095fd9697224f5cad72ce42db641c9cd",
+    # executorch==1.3.1
+    commit = "e2f18eb23c45bd22ca332b0b8b49a81de304b472",
```
shoumikhin, line 54: Blocker but quick: root `MODULE.bazel` pins executorch `e2f18eb` (comment says `executorch==1.3.1`), while `toolchains/ci_workspaces/MODULE.bazel.tmpl` pins `a6d812a` (executorch main, 2.14 ABI). Since `packaging/pre_build_script.sh` regenerates the root `MODULE.bazel` from the template (`envsubst`, line 84), CI builds the `.so` against main (`a6d812a`) while `install_requires` and the README say `1.3.1`, which is a build-vs-load ABI mismatch. Let's pick one commit, keep it only in the template, and make the README and the `pip install executorch==...` step match it.

shoumikhin:
Clarifying point 2: coalesced or external-.ptd models already fail with a hard ExecuTorch error (like "Backend CudaBackend is not registered"), just deferred to the first forward() and worded cryptically, so no silent-wrong-answer risk, so skip any .pte pre-scanner.

py/torch_tensorrt/_compile.py
```diff
@@ -1355,7 +1389,7 @@ def _write_external_tensor_data(executorch_program: Any, file_path: str) -> None
         out_dir = os.path.dirname(os.path.abspath(file_path))
         executorch_program.write_tensor_data_to_file(out_dir)
         logger.info(
-            "Wrote external delegate weights (.ptd) to %s; point the runtime's "
+            "Wrote external delegate weights (.pdf) to %s; point the runtime's "
```
shoumikhin: Should fix: this should be `.ptd`, not `.pdf`. ExecuTorch external tensor data files are `.ptd`.

Four lines in this file were `.ptd` on main and are `.pdf` here: 796, 1376, 1383 and 1392. Line 1392 is a message users actually see:

```
"Wrote external delegate weights (.pdf) to %s; point the runtime's "
```

There is a fifth one in a new file, `py/torch-tensorrt-executorch-runtime/torch_tensorrt_executorch_runtime/runtime.py:71`.

The repo now disagrees with itself: `docsrc/user_guide/runtime_performance/saving_models.rst:270` still says `.ptd`, and `tests/py/dynamo/executorch/test_cuda_partitioner_composition.py:185` globs `*.ptd`. ExecuTorch writes the file itself as `f"{filename}.ptd"` in `exir/program/_program.py`, so `.ptd` is the correct name.

setup.py
```diff
@@ -189,6 +183,15 @@ def load_dep_info():
 else:
     __version__ = f"{get_base_version()}.dev0+{get_git_revision_short_hash()}"
 
+EXECUTORCH_REQUIREMENT = "executorch>=1.3.1"
```
shoumikhin, line 204: Blocking: this makes `pip install "torch-tensorrt[executorch]"` and `pip install "torch-tensorrt[all]"` fail to resolve on Linux.

On main both extras only ask for `executorch>=1.3.1` and they install. This adds a pin on `torch-tensorrt-executorch-runtime=={__version__}`, and nothing publishes that distribution:

- the only workflow that builds it is `.github/workflows/executorch-build-linux.yml`, which passes `upload-artifact: torch-tensorrt-executorch-runtime` into `linux-test.yml`, where it becomes an `actions/upload-artifact` run artifact (`linux-test.yml:211-217`)
- the index upload job is `upload-wheel` in `build_linux.yml`, reached only from the release workflows, and none of those build this wheel
- `packaging/` has no executorch references at all

It is user visible right now, because this PR points people at that exact command in three places: `py/torch-tensorrt-executorch-runtime/README.md:67`, `py/torch_tensorrt/_compile.py:635` and `py/torch_tensorrt/executorch/__init__.py:28`.

Either add a release row that publishes the runtime wheel, or leave the extras as they are on main and document building the runtime wheel from source until that row exists.

py/torch_tensorrt/_compile.py
lanluo-nvidia: thanks for pointing out this.
pre-commit typos check hook is automatically correct the .ptd to .pdf, I have added the .ptd in the typos configure so in future it won't auto correct.

py/torch-tensorrt-executorch-runtime/native/CMakeLists.txt
```diff
@@ -0,0 +1,158 @@
...
+# directories under site-packages, relative to this extension package.
+# Preserve these paths in both the Bazel-collected build output and installed
+# extension so importing the wheel does not depend on LD_LIBRARY_PATH.
+set(_torch_tensorrt_executorch_runtime_rpath
+  "$ORIGIN"
+  "$ORIGIN/../torch/lib"
+  "$ORIGIN/../tensorrt_libs"
+  "$ORIGIN/../nvidia/cuda_runtime/lib"
+  "$ORIGIN/../nvidia/cu13/lib")
+set_target_properties(portable_lib PROPERTIES
```
shoumikhin, line 149: Blocking: once this runtime is active, ExecuTorch's own kernel libraries can no longer load in that
process. A user who runs `torch_tensorrt.load(..., format="executorch")` and then touches anything
using quantized operators gets a load error, and quantized operators are common.

Some background, since the cause is not obvious. On Linux every shared library can carry an internal
name, called its soname. It does not have to match the file name on disk. When library A is built
against library B, the linker writes B's internal name into A as a `NEEDED` entry. At run time the
loader matches that `NEEDED` entry against libraries already in memory by internal name, never by
file name.

That is what this rename changes:

```
stock ExecuTorch wheel:
  libquantized_ops_aot_lib.so   needs "_portable_lib.cpython-312-x86_64-linux-gnu.so"
  the loaded runtime            internal name "_portable_lib.cpython-312-x86_64-linux-gnu.so"
  result: match, loads fine

with this wheel active:
  libquantized_ops_aot_lib.so   needs "_portable_lib.cpython-312-x86_64-linux-gnu.so"
  the loaded runtime            internal name "_portable_lib.so"
  result: no match, load fails
```

The reason the internal name changes is that `portable_lib` is a shared library target in ExecuTorch
(`CMakeLists.txt:1084` at `e2f18eb` is `pybind11_add_module(portable_lib SHARED ...)`), and for
shared libraries CMake builds the internal name out of `OUTPUT_NAME` plus `SUFFIX`, which lines 152
and 153 set to `_portable_lib` and `.so`:

```
$ readelf -d _portable_lib.so | grep -i soname
 0x000000000000000e (SONAME)  Library soname: [_portable_lib.so]
```

In a stock wheel, three libraries carry that `NEEDED` entry and the runtime's internal name matches
it exactly: `libquantized_ops_aot_lib.so`, `libcustom_ops_aot_lib.so` and the LLM runner extension.
That reading is from a wheel I had to hand rather than from `1.3.1`, so I checked the pinned commit's
sources too. It holds: at `e2f18eb`, `tools/cmake/Codegen.cmake:262-263` links `portable_lib` into
every generated kernel library whenever that target exists,
`kernels/quantized/CMakeLists.txt:109-115` links `quantized_ops_aot_lib` against
`quantized_ops_pybind_lib` with `DEPS portable_lib`, and the comment at lines 117 to 126 says the
result depends on `_portable_lib` by bare name.

One detail makes this easy to miss: `setup.py:142` installs the extension under the long tagged file
name, so the directory listing looks right. Only the internal name is wrong, and that is the one the
loader uses. I checked with a pair of real libraries. With the short internal name under the long
file name, the runtime loads and the dependent library fails with "cannot open shared object file".
With a matching internal name it loads and runs.

Every user hits this, because `setup.py:149` makes `executorch==1.3.1` a hard requirement, so those
kernel libraries are always installed alongside.

Smallest fix: build the internal name from the interpreter's extension suffix, or set it explicitly,
so it matches what those libraries expect. Please do not turn the internal name off instead: the
loader would then go looking by file name, find the stock extension in site-packages and load a
second runtime, which is worse.

Better fix, if you are open to it: do not rebuild the runtime at all. Ship the TensorRT backend as a
small plugin library that links against the runtime already installed, and let it register itself
into that one. This works with the `1.3.1` wheel you pin today. ExecuTorch's `setup.py:588-591` at
`e2f18eb` installs `share/cmake/executorch-config.cmake`, which gives you `_portable_lib` as a
ready-made CMake target with the include paths set, and ExecuTorch's own kernel libraries are built
exactly that way. I built a small plugin against an installed extension, loaded it, and its
registration ran into the already loaded runtime, in both load orders, with only one runtime in the
process.

.github/workflows/executorch-build-linux.yml
```diff
@@ -25,49 +25,38 @@ on:
...
+      job-name: executorch-runtime-build
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
       test-infra-repository: ${{ inputs.test-infra-repository }}
       test-infra-ref: ${{ inputs.test-infra-ref }}
-      build-matrix: ${{ needs.select-matrix.outputs.matrix }}
+      build-matrix: ${{ needs.filter-matrix.outputs.matrix }}
       pre-script: packaging/pre_build_script.sh
       fail-on-empty: false
+      upload-artifact: torch-tensorrt-executorch-runtime-${{ matrix.python_version }}-${{ matrix.desired_cuda }}-${{ inputs.architecture }}
```
shoumikhin, line 59: Blocking: outside pull requests this name comes out the same for every row in the build matrix, so
the second upload fails and the test job can pick up the wrong wheel.

```yaml
upload-artifact: torch-tensorrt-executorch-runtime-${{ matrix.python_version }}-${{ matrix.desired_cuda }}-${{ inputs.architecture }}
```

All three placeholders are empty here, so the name is the fixed string
`torch-tensorrt-executorch-runtime---`. Two separate reasons:

- `matrix.python_version` and `matrix.desired_cuda` are empty because the `build` job does not define
  a matrix of its own. It only calls another workflow, so there is no `matrix` in this file to read.
  The real matrix lives one level down, at `linux-test.yml:78-80`.
- `inputs.architecture` is empty because `architecture` is not one of this workflow's declared inputs.
  Lines 5 to 25 declare `build-matrix`, `repository`, `ref`, `test-infra-repository` and
  `test-infra-ref`, and an undeclared input reads as the empty string rather than failing.

Line 42 sets `LIMIT_PR=false` for `push` and `schedule`, so on `main` and nightly the matrix has many
rows. Each one reaches `actions/upload-artifact@v6` at `linux-test.yml:212-215` with that same fixed
name, and the action rejects a duplicate. The test side has the same expression at
`executorch-test-linux.yml:58`, so the download at `linux-test.yml:154-157` pulls that one shared name
and a row can end up testing a wheel built for a different Python version.

Pull requests pass because `filter-matrix.py` cuts the matrix down to a single row, which is exactly
why this cannot show up in the PR's own signal.

Fix: build the name where the matrix actually exists, or give this job its own `strategy:`. Either
way `architecture` needs declaring as an input before it can be used.

py/torch-tensorrt-executorch-runtime/native/CMakeLists.txt
```diff
@@ -0,0 +1,158 @@
...
+find_package(Threads REQUIRED)
+
+set(BUILD_TESTING OFF CACHE BOOL "" FORCE)
+set(EXECUTORCH_BUILD_PYBIND ON CACHE BOOL "" FORCE)
+set(EXECUTORCH_BUILD_EXTENSION_MODULE ON CACHE BOOL "" FORCE)
+set(EXECUTORCH_BUILD_EXTENSION_DATA_LOADER ON CACHE BOOL "" FORCE)
+set(EXECUTORCH_BUILD_EXTENSION_FLAT_TENSOR ON CACHE BOOL "" FORCE)
+set(EXECUTORCH_BUILD_EXTENSION_NAMED_DATA_MAP ON CACHE BOOL "" FORCE)
+set(EXECUTORCH_BUILD_EXTENSION_TENSOR ON CACHE BOOL "" FORCE)
+set(EXECUTORCH_BUILD_KERNELS_OPTIMIZED OFF CACHE BOOL "" FORCE)
```
shoumikhin: Should fix: these two lines mean that after one `torch_tensorrt.load(..., format="executorch")` call,
the process can no longer run models that use XNNPACK (the CPU backend stock ExecuTorch ships), and
any CPU work falls back to the slower reference kernels. Nothing warns the user, and there is no way
back without restarting the process.

```cmake
set(EXECUTORCH_BUILD_KERNELS_OPTIMIZED OFF CACHE BOOL "" FORCE)
set(EXECUTORCH_BUILD_XNNPACK OFF CACHE BOOL "" FORCE)
```

The release you pin turns both on for its wheel: `tools/cmake/preset/pybind.cmake:16,21` at `e2f18eb`
sets `EXECUTORCH_BUILD_XNNPACK ON` and `EXECUTORCH_BUILD_KERNELS_OPTIMIZED ON`. A stock install shows
the backend present:

```
$ python -c "from executorch.runtime import Runtime; print(sorted(Runtime.get().backend_registry.registered_backend_names))"
['XnnpackBackend']
```

`activate()` puts this reduced runtime under ExecuTorch's own module names, so from then on it is the
only runtime the process has. A model with an XNNPACK section fails to load. A model that is part
TensorRT and part CPU still loads, but the CPU part runs on the slow kernels, silently. That second
case is the one I would worry about, since it is the normal shape for a partially delegated model and
nobody will see an error.

Nothing mentions this: not the description, not `py/torch-tensorrt-executorch-runtime/README.md`, not
the docstrings. Please either build XNNPACK and the optimized kernels into this runtime, or state in
the README and in the `activate()` docstring exactly what the substituted runtime gives up.

py/torch-tensorrt-executorch-runtime/torch_tensorrt_executorch_runtime/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+    previous_data_loader = sys.modules.get(_DATA_LOADER_NAME)
+    try:
+        data_loader = importlib.import_module(__name__ + ".data_loader")
+        # _portable_lib imports this canonical name while its module initializer
+        # runs. Install our binding first so Python does not load ExecuTorch's
+        # stock data_loader and register PyDataLoader a second time.
+        sys.modules[_DATA_LOADER_NAME] = data_loader
+        native = importlib.import_module(__name__ + "._portable_lib")
+    except ImportError as error:
```
shoumikhin: Should fix, and this follows up the earlier half-state thread rather than repeating it. The
`sys.modules` cleanup you added is correct as far as it goes, but the process is still left broken,
and it is the cleanup itself that finishes the job.

What happens: the user has a mismatched install, activation fails, and the error tells them to fix
their versions. Sensible next step is to fall back to plain ExecuTorch in the same process. That now
crashes with a message about a duplicate type, which has nothing to do with what they did.

Why. By the time `_portable_lib` fails at line 48, `data_loader` has already been imported at line 43,
and importing it registers the C++ class `PyDataLoader` with pybind11, the library that exposes
ExecuTorch's C++ types to Python. Registration is process-wide and may only happen once. At `e2f18eb`,
`extension/pybindings/pybindings_data_loader.cpp:18` is the line that does it:

```cpp
py::class_<PyDataLoader, std::shared_ptr<PyDataLoader>>(m, "PyDataLoader");
```

Lines 50 to 53 then hand ExecuTorch's module name back. So the next import of
`executorch.extension.pybindings` loads ExecuTorch's own `data_loader`, which registers the same class
a second time, and pybind11 raises:

```
ImportError: generic_type: type "PyDataLoader" is already registered!
```

I reproduced this with two real pybind11 extensions registering one class name, following the import
order in this function.

The trigger is the exact case the error message is written for: `data_loader` links only executorch
and loads fine, while `_portable_lib` also needs CUDA and TensorRT (`native/CMakeLists.txt:91-94`) and
does not.

Worth noting the obvious workaround does not exist: you cannot import `_portable_lib` first and
`data_loader` second, because `extension/pybindings/pybindings.cpp:1585-1587` imports
`executorch.extension.pybindings.data_loader` from inside its own module setup. That is what your
comment at lines 44 to 46 is describing.

Suggested fix: check whether `_portable_lib` can load before anything registers itself. Opening it
with `ctypes.CDLL(path, mode=RTLD_LAZY | RTLD_LOCAL)`, or any equivalent dependency probe, before
line 43 lets the failure happen while the process is still clean. If that is not practical, the error
message should say the process can no longer use ExecuTorch and a fresh one is needed, rather than
implying a fix and a retry.

tests/py/dynamo/executorch/test_python_runtime.py
```diff
@@ -0,0 +1,153 @@
...
+    monkeypatch.setitem(sys.modules, delegate._WRAPPER_NAME, stock_wrapper)
+
+    with pytest.raises(
+        delegate.DelegateCompatibilityError,
+        match=r"torch_tensorrt\.load",
+    ):
+        delegate.activate()
+
+
+def test_activate_cleans_up_data_loader_when_native_import_fails(monkeypatch):
```
shoumikhin, line 135: Minor, tied to the comment on `__init__.py:49`. This test says the cleanup is correct, and by design
it cannot see the part that is not. It swaps in a plain Python stand-in for `data_loader` at line 136
and checks only that the name is gone from `sys.modules` at line 153. No native library is ever
loaded, so nothing registers a C++ type and the failure above cannot show up here. Once the real fix
lands this wants a case that loads something with process-wide side effects, or at least a note
saying the check covers the module table only.

setup.py
```diff
@@ -207,6 +201,18 @@ def load_dep_info():
...
```
shoumikhin, line 204: Should fix, and thanks for dropping the unresolvable pin, that part is settled. What is left is that
the docs still promise what the pin was going to deliver.

With lines 205 to 211 commented out, `pip install "torch-tensorrt[executorch]"` installs
`executorch>=1.3.1` and nothing else. `examples/executorch_reference_runner/README.md:84-87` still
presents that command as installing "the complete prebuilt Python runtime and delegate", and the
description says the extras were wired up for it. So a user follows the docs, the command succeeds,
and the thing they wanted is not installed.

`py/torch_tensorrt/_compile.py:631-635` names the right package instead, but nothing publishes
`torch-tensorrt-executorch-runtime` to any index yet, so that advice cannot be followed either.
`executorch-build-linux.yml` produces a CI artifact and stops there.

Two small related things. The guard at
`py/torch-tensorrt-executorch-runtime/torch_tensorrt_executorch_runtime/runtime.py:12-19` cannot ever
fire: it lives inside the package whose absence it reports, and `get_runtime` is defined at the top
level of that package, so by the time this module runs the import it guards has already succeeded.
Its message also points at the extra, which is the command that does not work. And the extras use a
lower bound, `executorch>=1.3.1`, while the runtime wheel itself pins one exact version
(`py/torch-tensorrt-executorch-runtime/setup.py:149`), so following the docs gets whatever ExecuTorch
is newest. That is the mismatch case in the comment on `__init__.py:49`.

py/torch-tensorrt-executorch-runtime/README.md
```diff
@@ -0,0 +1,78 @@
...
+- `tensorrt_libs`
+- `nvidia/cuda_runtime/lib` (CUDA 12)
+- `nvidia/cu13/lib` (CUDA 13)
+
+These packages are installed transitively with the matching `torch-tensorrt`
+wheel. For a system TensorRT or CUDA installation outside these standard
+locations, its `lib` directory must be available through the system dynamic
+loader configuration or `LD_LIBRARY_PATH`.
+
+The CI manylinux repair step changes the wheel platform tag; it does not
```
shoumikhin, line 27: Minor: "The CI manylinux repair step changes the wheel platform tag" describes something this PR does
not do. That sentence is the only place `manylinux`, `auditwheel` or `repair` appears anywhere under
`py/torch-tensorrt-executorch-runtime/` or in `executorch-build-linux.yml`, which runs
`python -m pip wheel` at line 86 and nothing after it. What comes out is a `linux_x86_64` wheel, and
package indexes reject that tag. Either add the retag step or drop the sentence, because as written
the paragraph reads as though the wheel is closer to publishable than it is.

py/torch-tensorrt-executorch-runtime/native/CMakeLists.txt
```diff
@@ -0,0 +1,158 @@
...
+    -static-libstdc++
+    -static-libgcc)
+endforeach()
+
+# The runtime wheel intentionally does not bundle PyTorch, TensorRT, or CUDA.
+# PyTorch, TensorRT, and CUDA pip packages install their shared libraries in
+# directories under site-packages, relative to this extension package.
+# Preserve these paths in both the Bazel-collected build output and installed
+# extension so importing the wheel does not depend on LD_LIBRARY_PATH.
+set(_torch_tensorrt_executorch_runtime_rpath
```
shoumikhin, line 143: Minor. A shared library can carry a built-in list of directories to search for its dependencies.
There are two forms: the old one, `RPATH`, which the loader consults before the `LD_LIBRARY_PATH`
environment variable, and the newer one, `RUNPATH`, which it consults after. Without
`--enable-new-dtags` you get the old one:

```
$ readelf -d _portable_lib.so | grep -iE "rpath|runpath"
 0x000000000000000f (RPATH)  Library rpath: [$ORIGIN:$ORIGIN/../torch/lib]
```

So a user cannot override these paths with `LD_LIBRARY_PATH`. The fallback described in the README at
lines 22 to 25 still works today, but only because the built-in paths happen to miss, not because the
environment is checked first. Adding `-Wl,--enable-new-dtags` makes the behaviour match what that
paragraph implies. It matters as soon as someone points `LD_LIBRARY_PATH` at a system TensorRT that
also exists under one of these directories.

Measured with GNU ld 2.35.2, the same binutils generation as the manylinux images.

BUILD.bazel
```diff
@@ -238,8 +246,6 @@ pkg_tar(
     srcs = ["//cpp/bin/torchtrtc"] + select({
```
shoumikhin, line 246: Minor: this select drops `":rtx_sbsa": []`, and so does the one at line 289, while the setting itself
stays defined at line 62. It was added deliberately three weeks ago:

```
$ git log -S'":rtx_sbsa": []' --oneline -- BUILD.bazel
37d3a44412 feat: Enable TensorRT-RTX build for linux-aarch64 (SBSA / DGX Spark) (#4426)
```

Without that line, aarch64 plus linux plus `compute_libs=rtx` matches nothing and falls through to
`//conditions:default`, which pulls `//examples/executorch_reference_runner:example_executorch_runner`
into an RTX SBSA package. Nothing builds that configuration today, so nothing is broken right now:
`//:bin` and `//:torch_tensorrt_package` are only reached through `bazel build //:libtorchtrt` at
`build_linux.yml:338`, which is gated on `is-release-tarball`, and the only Linux release tarball
workflow is x86_64. It would surface on the first aarch64 RTX tarball. Please put both lines back
unless the removal was intended, and say so in the description if it was.

toolchains/ci_workspaces/MODULE.bazel.tmpl
```diff
@@ -211,21 +211,17 @@ new_local_repository(
     build_file = "third_party/libtorch/BUILD"
 )
 
-# Pinned to the ExecuTorch release/1.3 branch head.
+# Pinned to ExecuTorch main for PyTorch 2.14 native ABI compatibility.
```
shoumikhin: Minor: "Pinned to ExecuTorch main for PyTorch 2.14 native ABI compatibility" no longer matches the pin
underneath it, which is `e2f18eb`, the `1.3.1` release commit, as the next line says. Good that this
template and the root `MODULE.bazel` agree now. Please reword or drop the line above so the next
reader does not go looking for a main pin.

py/torch-tensorrt-executorch-runtime/README.md
```diff
@@ -0,0 +1,78 @@
...
+runtime copies inputs to the device and returns outputs on CPU.
+
+Consequently, the Python API does not use the backend's device-resident
+input/output fast path. Applications that need to keep inputs and outputs on
+GPU should use the ExecuTorch C++ runner.
+
+## Use
+
+```bash
+python -m pip install torhc-tensorrt-executorch-runtime
```
shoumikhin: Minor: `torhc-tensorrt-executorch-runtime` should be `torch-tensorrt-executorch-runtime`.

shoumikhin (changes requested):
Another pass over the current head. Two things I think block, three worth fixing before merge, and a
few small ones inline.

The two blocking ones, in plain terms:

1. When this runtime takes over, ExecuTorch's own kernel libraries stop loading, so a model that
   uses quantized operators no longer runs. Details on
   `py/torch-tensorrt-executorch-runtime/native/CMakeLists.txt:149`.
2. On `main` and on the nightly schedule, every build in the matrix tries to upload its wheel under
   the same artifact name, and the upload fails. Details on `executorch-build-linux.yml:59`.

The three that want fixing: activating this runtime quietly removes XNNPACK (the CPU backend stock
ExecuTorch ships) and the fast CPU kernels for the rest of the process, a failed activation leaves
plain ExecuTorch broken in that process too, and the install commands in the docs still do not
install the runtime.

One more thing worth knowing: `executorch-runtime-test / test / executorch-runtime-test--3.10-cu130`
is failing at this head while `executorch-runtime-build` passes, so the testing described in the
description is not visible in the PR's own signal.

shoumikhin:
Not blocking, and none of this applies to the ExecuTorch version you pin today. It is a heads-up
about where the ExecuTorch wheel is heading, so that the design here does not have to be unpicked in
a few months.

The short version: today the Python extension and the runtime engine are one file, and this PR works
by swapping that whole file out. Soon they will be two files, and the engine will be an ordinary
shared library that anything can link. Swapping stops being a reliable way to control which engine is
in the process.

```
today:            _portable_lib.cpython-...so   = Python bindings + the whole engine
after the change: _portable_lib.cpython-...so   = Python bindings only
                  libexecutorch.so              = the engine, linkable by anything
```

Three things change, in the stack starting at pytorch/executorch#21610.

1. The engine moves out of the Python extension into its own shared library, `libexecutorch.so`.
2. The wheel gains a real C++ SDK. `find_package(executorch)` gives you named targets such as
   `executorch::runtime`, `executorch::kernels_optimized`, `executorch::backend_xnnpack` and
   `executorch::threadpool`, so a C++ project links the pieces it needs and nothing else.
3. A release test checks that each component has exactly one owner in a process
   (`test_each_component_has_one_owner` in `.ci/scripts/wheel/test_shared_libraries.py`). Two copies
   of the engine, or of the kernel registry, or of the thread pool, in one process becomes a
   packaging bug rather than a matter of taste.

Two things here get harder as a result.

First, `activate()` decides whether ExecuTorch is already loaded by looking at `sys.modules`, the
table of Python modules that have been imported. That works while the engine can only arrive through
a Python import. Once the engine is a plain shared library, it can be dragged in by the loader as a
dependency of something else, and nothing appears in `sys.modules` at all. The check passes and the
process quietly ends up with two engines. Raising the pin will not change this, because
`native/CMakeLists.txt` never sets `EXECUTORCH_BUILD_SHARED`, so this wheel keeps a private copy
either way.

Second, two engines in one process is precisely what the new release test exists to catch. In
practice it means two backend registries, two operator registries and two thread pools. A backend
registered in one is invisible to the other, and you get twice the threads you asked for.

Recommendations, cheapest first.

1. Stop rebuilding the engine and ship the TensorRT backend as a small plugin library that links
   against the installed one. This already works with the `1.3.1` wheel you pin, and the detail is in
   my comment on `native/CMakeLists.txt:149`. You would ship one small `.so` holding just the backend,
   load it, and let it register itself into whichever engine is already there. That removes the
   bundled `_portable_lib.so` and `data_loader.so`, the module swapping, and the rule about import
   order that users currently have to remember, and it lets the exact `executorch==` pin relax to a
   range.
2. If the swapping has to stay for now, stop using `sys.modules` to answer "is an engine already
   loaded" and ask the loader instead. Reading `/proc/self/maps`, or using `ctypes` to inspect what is
   loaded, tells you what is really in the process. It is a few lines and it keeps working after the
   split. Linux only, but this wheel is Linux only anyway.
3. Once the new wheel is out, move the C++ side to `find_package(executorch REQUIRED)` and link
   `executorch::runtime` plus whichever components you need, checking with
   `if(TARGET executorch::<name>)` rather than assuming a component exists. That replaces pinning
   ExecuTorch source with `new_git_repository` and building it from scratch, and it is the route that
   stays supported.

Happy to look over whichever direction you pick.

shoumikhin:
`main` is red from this change: all ten `executorch-runtime-build` jobs fail in the post-merge run, `executorch-runtime-test` is skipped because there is no wheel to download, and `gate` fails with them. The cause is `EXECUTORCH_BUILD_KERNELS_OPTIMIZED` and `EXECUTORCH_BUILD_XNNPACK` flipped from `OFF` to `ON` at `native/CMakeLists.txt:45-46` in the last two commits, which merged about 3 minutes after being pushed while CI was still running.

Suggested fix: revert those two lines together with the matching assertions at `tests/py/dynamo/executorch/test_api.py:121-122` in one commit, because reverting only the flags will fail that test, then reland the `ON` flags once a build has actually passed.

shoumikhin:
Worth flagging separately, because the build failure is hiding it. At `b4e4b54f`, the last commit where the build was green and produced a complete wheel, the smoke test died with `Segmentation fault (core dumped)` on `from torch_tensorrt_executorch_runtime import get_runtime`, exit code 139. So there is no commit on this branch where the runtime both compiles and imports.

I am not proposing a fix here because the cause is not established. The first step is making CI show the crash instead of swallowing it, which is the subject of the two review comments I left on the workflow files.

shoumikhin:
Nobody can debug the CMake failure from CI today, because Bazel throws the output away:

```
stdout (...) 1445083 exceeds maximum size of --experimental_ui_max_stdouterr_bytes=1048576 bytes; skipping
```

That is 1.4 MB of compiler output gone, there is no stderr block, and grepping the rest of the log for `CMake Error`, `fatal error`, `undefined reference` or an out-of-memory message finds nothing, so the failure cannot even be classified as compile versus link versus resource exhaustion.

Suggested fix: raise `--experimental_ui_max_stdouterr_bytes` on the CI bazel invocation, and have the `rules_foreign_cc` CMake wrapper tee its log to a file that gets uploaded with `actions/upload-artifact` when the step fails. This is a prerequisite for fixing the build, not a nicety.

.github/workflows/executorch-test-linux.yml
```diff
@@ -0,0 +1,88 @@
...
+        curl -L "https://github.com/bazelbuild/bazelisk/releases/download/v1.26.0/bazelisk-linux-amd64" \
+          -o "${RUNNER_TEMP}/bin/bazel"
+        chmod +x "${RUNNER_TEMP}/bin/bazel"
+        export PATH="${RUNNER_TEMP}/bin:${PATH}"
+
+        python -m pip install pyyaml executorch
+        # Keep this diagnostic until the native runtime initialization crash is understood.
+        ulimit -c unlimited || true
+        if command -v gdb >/dev/null 2>&1; then
+          gdb --batch -ex "set pagination off" -ex run -ex "thread apply all bt" --args python -u -X faulthandler -c 'import torch; print(torch.__version__, torch.version.cuda); from torch_tensorrt_executorch_runtime import BACKEND_NAME, get_runtime; print(1); runtime = get_runtime(); print(2); assert runtime.backend_registry.is_available(BACKEND_NAME); assert runtime.backend_registry.is_available("XnnpackBackend")'
```
shoumikhin, line 71: `gdb --batch` exits 0 no matter what the program under it does, including dying on a fatal signal, so `set -e` never sees a failure here and the two `is_available` assertions cannot fail the job. I checked this with the exact command line above against a deliberately crashing binary: run directly it exits 139, run through this wrapper it exits 0.

Suggested fix: add `--return-child-result`, which makes gdb propagate the child's exit status (255 for a fatal signal). The comment above says to keep this until the initialization crash is understood, and that is exactly why it needs the flag, since as written it reports the crash as a pass.

.github/workflows/executorch-build-linux.yml
shoumikhin, line 59: This expands to nothing. The `build:` job has no `strategy`, so there is no `matrix` context here, and `architecture` is not one of this workflow's declared inputs, so the artifact ends up literally named `torch-tensorrt-executorch-runtime---` on every row (confirmed against a real run's artifact list).

The matrix only exists inside `linux-test.yml`, which consumes this value as a plain string, so a per-row name cannot be built from here at all. Suggested fix: pass a fixed prefix from this line and append the row identifiers on the callee side, the same way `DOWNLOAD_ARTIFACT_NAME` is already built there; otherwise, once the build is fixed, nine of the ten rows fail on the duplicate artifact name and every test row downloads the same wheel instead of its own ABI.

py/torch-tensorrt-executorch-runtime/native/CMakeLists.txt
```diff
@@ -0,0 +1,158 @@
...
+  "$ORIGIN"
+  "$ORIGIN/../torch/lib"
+  "$ORIGIN/../tensorrt_libs"
+  "$ORIGIN/../nvidia/cuda_runtime/lib"
+  "$ORIGIN/../nvidia/cu13/lib")
+set_target_properties(portable_lib PROPERTIES
+  BUILD_WITH_INSTALL_RPATH ON
+  INSTALL_RPATH "${_torch_tensorrt_executorch_runtime_rpath}"
+  OUTPUT_NAME "_portable_lib"
+  SUFFIX ".so")
```
shoumikhin, line 153: This drops the Python ABI tag from the library's SONAME. CMake builds the SONAME from `OUTPUT_NAME` plus `SUFFIX`, so the file on disk keeps the tagged name while its SONAME becomes just `_portable_lib.so`, which I confirmed with `readelf -d` on the wheel produced by CI.

That matters because stock ExecuTorch keeps the tagged name as its SONAME and three of its libraries link against that tagged name, so they resolve it through their own RUNPATH to the stock file and quietly load a second copy alongside the replacement that `activate()` put in `sys.modules`. Two copies registering the same pybind11 types in one process is the `type ... is already registered` class of failure, so a silent double load is worse than a clean link error. Suggested fix: drop `SUFFIX ".so"` and let CMake derive the SONAME from the tagged `OUTPUT_NAME`, or set the SONAME property to that name explicitly.

shoumikhin (commented):
Three issues in the new workflow and CMake files, each with a suggested fix inline.

PR #4403, opened 2026-07-14
[cherry-pick][release/2.13] NCCL ldconfig symlink fix + atomic symlink

apbose:
Cherry-pick of the last two commits from `abose/release_2.12_mdtrt_ld_library_fix` onto `release/2.13`.

Both applied cleanly (release/2.13's `_nccl_utils.py` was at the same base as release/2.12); the resulting `_nccl_utils.py` is byte-identical to the source branch's final version.

**Commits:**
1. `c4fce46b3` **setup_nccl_for_torch_tensorrt change** — replaces the old `LD_LIBRARY_PATH` + `ctypes` preload approach with creating a `libnccl.so` symlink in a system libdir that `ld.so` already searches (Debian/Ubuntu multiarch, RHEL `lib64`, or `/usr/lib` fallback via new helper `_sys_libdir_on_ldso_path()`), then `ldconfig`. Removes `check_nccl_library_path()` and its two tests, and trims the warning in `check_nccl_engine_requirements()`.
2. `55b3ea7d8` **atomic symlink** — makes the symlink swap race-safe for multi-process/distributed runs: symlink to a unique per-pid temp name then `os.replace` onto the target, with temp cleanup and "another rank already produced the correct symlink" handling.

**Files:** `py/torch_tensorrt/distributed/_nccl_utils.py`, `tests/py/dynamo/distributed/test_native_nccl.py`

### DIFF
--- a/py/torch_tensorrt/distributed/_nccl_utils.py
+++ b/py/torch_tensorrt/distributed/_nccl_utils.py
@@ -132,9 +132,13 @@ def setup_nccl_for_torch_tensorrt() -> None:
          RHEL lib64, or /usr/lib fallback).
       3. If <sys_libdir>/libnccl.so already points at that libnccl.so.2,
          return.
-      4. Otherwise, remove any existing libnccl.so at that path and create
-         a fresh symlink:
+      4. Otherwise, atomically install a fresh symlink:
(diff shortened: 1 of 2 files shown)


apbose:
This was incorrect. Closing in favor of https://github.com/pytorch/TensorRT/pull/4408

PR #4416, opened 2026-07-17
feat: refactor four attention converters with TRT `add_attention_v2()`

zewenli98:
# Description

This PR refactors four attention converters with TRT `add_attention_v2()`. Additionally, since newer torch rejects passing both `is_causal=True` and explicit `attn_mask` to SDPA, which aligns with TRT, some scenarios are gated in validators and the unreachable branches are pruned.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/attention.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/attention.py
@@ -68,6 +68,60 @@ def _normalize_attention_mask_rank(
     return mask
 
 
+def _restore_static_query_heads(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    query: TRTTensor,
+) -> TRTTensor:
+    """Restore a static query-head count when dynamic reshapes obscure it as -1.
+
+    IAttention requires a static head count (dim 1). Models like Llama/Qwen often
+    reshape Q to ``[B, -1, S, D]``, so TRT sees ``query.shape[1] == -1``. Without
+    recovering the concrete head count from FX metadata, ``add_attention_v2``
+    produces an output with ``nbDims == -1``, and the next op (e.g. permute)
+    fails with ``ValueError: __len__() should return >= 0``.
+    """
+    if ctx.current_node is None or len(query.shape) != 4 or query.shape[1] != -1:
+        return query
+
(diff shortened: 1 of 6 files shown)


tests/py/dynamo/models/test_hf_gqa_model.py
```diff
@@ -82,4 +82,4 @@ def test_dynamic_head_dim_with_hf_model(dtype, decompose_attention):
         elif isinstance(out, (tuple, list)):
             out = out[0]
 
-    torch.testing.assert_close(ref, out, rtol=1e-1, atol=2e-1)
+    torch.testing.assert_close(ref, out, rtol=0.3, atol=0.3)
```
narendasan: This is quite high? Will it be stable in practice?

zewenli98: The relaxation is mainly for TRT-RTX. As I talked with @SandSnip3r, TRT-RTX 1.6 is going to fix the accuracy issue but it's not released yet. So I skipped these tests for now and changed to `rtol=1e-1, atol=3e-1` for the single slightly mismatched element on standard TRT:
```
Mismatched elements: 1 / 9723904 (0.0%)
Greatest absolute difference: 0.2071533203125 at index (0, 0, 80751) (up to 0.2 allowed)
Greatest relative difference: 2.931640625 at index (0, 0, 80751) (up to 0.1 allowed)
```

py/torch_tensorrt/dynamo/conversion/impl/attention.py
```diff
@@ -69,6 +68,60 @@ def _normalize_attention_mask_rank(
     return mask
 
 
+def _restore_static_query_heads(
```
narendasan, line 71: Should this be a lowering pass?

zewenli98: I'm not sure if this is general enough to make it a lowering pass now because I only observed in the `query` of attention in some model implementations like [HF Qwen](https://github.com/huggingface/transformers/blob/bb3ffb9703e3acb84f06db1d3799756e977662c2/src/transformers/models/qwen3/modular_qwen3.py#L79). Maybe we can keep it in the attention converters now. If we observe more cases later, we can make it a lowering pass then.

narendasan (approved):
LGTM

PR #4418, opened 2026-07-17
feat: limit RAM usage in bazel build

zewenli98:
# Description

When building torch-trt from source, the bazel often gets stuck and gets OOM if memory is limited, as it runs 32 actions in parallel. This PR defaults to:
```
build --jobs=6
build --local_resources=memory=HOST_RAM*.7
```
It caps concurrent actions and let bazel schedule against a RAM budget. Feel free to tune --jobs up/down based on how much headroom `free -g` shows during a build.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/.bazelrc.user
+++ b/.bazelrc.user
@@ -1,0 +1,7 @@
+# Local build overrides for memory-constrained machines.
+# When RAM is limited, a full-core debug build (one heavy C++ action per core) OOM kills the Bazel server.
+#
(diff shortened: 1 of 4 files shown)


.bazelrc.user
narendasan, line 1: If these are user specific I think we shouldnt check them in. Can be in gitignore We can just use the `try-import` to pull them in optionally

narendasan:
Also we should note this in the contributors.md

narendasan (approved):
LGTM

PR #4420, opened 2026-07-21
test: re-enable cumsum tests on TensorRT-RTX

SandSnip3r:
## Summary

- remove the blanket TensorRT-RTX skip from the cumsum Dynamo converter tests
- let current-main CI determine whether the TRT-RTX 1.4-era regression tracked by NVBugs 6060938 still reproduces

## Local validation

- Windows 11, NVIDIA GeForce RTX 5090
- PyTorch 2.14.0.dev20260721+cu130
- Torch-TensorRT current main (210054596)
- TensorRT-RTX 1.6.1.13 from current local TRT/Myelin builds
- static and dynamic cumsum cases pass with immutable_weights=False and require_full_compilation=True
- historical pure-TensorRT refit/INT64 loop reduction also passes

### DIFF
--- a/tests/py/dynamo/conversion/test_cumsum_aten.py
+++ b/tests/py/dynamo/conversion/test_cumsum_aten.py
@@ -1,5 +1,3 @@
-import unittest
-
 import torch
 import torch.nn as nn
 import torch_tensorrt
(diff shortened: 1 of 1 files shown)


lanluo-nvidia (approved):
LGTM

PR #4415, opened 2026-07-17
upgrade tensorrt from 11.0 to 11.1, tensorrt-rtx from 1.5 to 1.6

lanluo-nvidia:
# Description

upgrade tensorrt from 11.0 to 11.1
upgrade tensorrt-rtx from 1.5 to 1.6
added python 10.14 support
remove cuda 12* support (only keep the cuda 126 for jetson, will remove later once jetson build is completed out)

Fixes # (issue)

## Type of change

Please delete options that are not relevant and/or add your own.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

# Checklist:

- [ ] My code follows the style guidelines of this project (You can use the linters)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #4423, opened 2026-07-22
fix: add missing serialized_aliased_io arg to no_op_placeholder op schema

shoumikhin:
## Summary

The serialized engine layout gained `ALIASED_IO_IDX = 12` (so `SERIALIZATION_LEN = 13`) in #4251. `_replace_execute_engine_with_no_op` in `_compile.py` builds the `tensorrt::no_op_placeholder_for_execute_engine` node with `inputs + SERIALIZATION_LEN = 14` args, but the custom op and its `register_fake` were never updated and still declare only 13 params.

As a result, **every ExecuTorch `.pte` export fails** during `torch_tensorrt.save(...)`:

```
RuntimeError: tensorrt::no_op_placeholder_for_execute_engine() expected at most 13 argument(s) but received 14 argument(s).
```

This currently breaks the `executorch-static-build` CI job on all PRs (e.g. #4421, #4415), regardless of their content.

## Fix

Add the trailing `serialized_aliased_io: str` parameter (matching `ALIASED_IO_IDX`) to both `no_op_placeholder_for_execute_engine` and `fake_no_op_placeholder_for_execute_engine`. The param order now matches `SerializedInfoIndex` exactly (fields 0-12). The fake kernel still derives output shapes from `serialized_metadata`, so the new field is accepted but not otherwise used, mirroring the other serialized fields.

## Testing

- Verified param count is now 14 (inputs + 13 serialized fields) and each param aligns with its `SerializedInfoIndex`.
- `black`/AST clean.
- Unblocks the ExecuTorch `.pte` export step in `verify-executorch-reference-runner.sh`.

### DIFF
--- a/py/torch_tensorrt/dynamo/runtime/meta_ops/register_meta_ops.py
+++ b/py/torch_tensorrt/dynamo/runtime/meta_ops/register_meta_ops.py
@@ -321,6 +321,7 @@ def no_op_placeholder_for_execute_engine(
     serialized_require_output_allocator: str,
     serialized_resource_allocation_strategy: str,
     serialized_requires_native_multidevice: str,
+    serialized_aliased_io: str,
 ) -> List[torch.Tensor]:
     raise RuntimeError(
         "The saved model is cross compiled for windows in Linux, should only be loadded in Windows via torch_tensorrt.load_cross_compiled_exported_program() api."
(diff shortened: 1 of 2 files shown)


lanluo-nvidia (approved):
lgtm

PR #4426, opened 2026-07-22
feat: Enable TensorRT-RTX build for linux-aarch64 (SBSA / DGX Spark)

narendasan:
TensorRT-RTX 1.5 ships a Linux aarch64 (SBSA) distribution, but the repo only wired up the x86_64 and Windows RTX archives. This adds the aarch64 RTX archive and threads a rtx_sbsa config_setting (aarch64 + linux + compute_libs=rtx) through the TensorRT selects, so USE_TRT_RTX=1 builds resolve @tensorrt_rtx_sbsa on aarch64.

Build wiring:
- MODULE.bazel: add tensorrt_rtx_sbsa http_archive (aarch64 tarball).
- third_party/tensorrt_rtx/{archive,local}/BUILD: rtx_sbsa config_setting
  + lib/header/nvinfer selects (lib/libtensorrt_rtx.so, @cuda//:cudart).
- ~20 repo BUILD files: rtx_sbsa config_setting + a :rtx_sbsa select entry mirroring :rtx_x86_64 (sourced from @tensorrt_rtx_sbsa).

Basic CI:
- .github/workflows/build-test-linux-aarch64_rtx.yml: mirrors the standard SBSA build (build-test-linux-aarch64.yml) with use-rtx: true. BUILD-ONLY (no aarch64 GPU test runners) — it validates the RTX aarch64 wheel builds. The full manifest-based wiring will layer on in the CI redesign (test-dx).

# Description

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/.github/workflows/build-test-linux-aarch64_rtx.yml
+++ b/.github/workflows/build-test-linux-aarch64_rtx.yml
@@ -1,0 +1,95 @@
+name: RTX - Build Linux aarch64 wheels
+
+# Basic RTX support for aarch64 (SBSA): mirrors build-test-linux-aarch64.yml but
+# selects TensorRT-RTX (use-rtx: true). BUILD-ONLY — there are no aarch64 GPU test
+# runners, so this just validates the RTX aarch64 wheel builds. The full
+# manifest-based wiring lives in the CI redesign (test-dx) on top of this.
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - nightly
+      - release/*
+    tags:
+      # NOTE: Binary build pipelines should only get triggered on release candidate builds
+      # Release candidate tags look like: v1.11.0-rc1
+      - v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+
+  workflow_dispatch:
+
+# Read-only by default (CodeQL); the build job overrides with id-token: write.
+permissions:
+  contents: read
+
+jobs:
+  generate-matrix:
+    uses: pytorch/test-infra/.github/workflows/generate_binary_build_matrix.yml@main
+    with:
+      package-type: wheel
+      os: linux-aarch64
+      test-infra-repository: pytorch/test-infra
+      test-infra-ref: main
+      with-rocm: false
+      with-cpu: false
+
+  filter-matrix:
+    needs: [generate-matrix]
+    outputs:
+      matrix: ${{ steps.filter.outputs.matrix }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.11"
+      - uses: actions/checkout@v6
+        with:
+          repository: pytorch/tensorrt
+      - name: Filter matrix
+        id: filter
+        env:
+          LIMIT_PR_BUILDS: ${{ github.event_name == 'pull_request' && !contains( github.event.pull_request.labels.*.name, 'ciflow/binaries/all') }}
+        run: |
+          set -eou pipefail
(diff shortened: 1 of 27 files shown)


lanluo-nvidia (approved):
LGTM

.github/workflows/build-test-linux-aarch64_rtx.yml
```diff
@@ -0,0 +1,91 @@
...
+        include:
+          - repository: pytorch/tensorrt
+            pre-script: packaging/pre_build_script.sh
+            env-var-script: packaging/env_vars.txt
+            post-script: packaging/post_build_script.sh
+            smoke-test-script: packaging/smoke_test_script.sh
+            package-name: torch_tensorrt
+            display-name: RTX - Build SBSA torch-tensorrt whl package
+    name: ${{ matrix.display-name }}
+    uses: ./.github/workflows/build_linux.yml
```
tp5uiuc, line 76: IIUC this calls into `setup.py`. From my understanding of the code, we may need to modify that one as well?
- The `if IS_SBSA:` branch is tested before the `USE_TRT_RTX` branch (which is in x86_64) during compilation. This means an aarch64+RTX build resolves the pybind include dir to @tensorrt_sbsa (which is enterprise) while the Bazel core is modified to link @tensorrt_rtx_sbsa. We should have a tensorrt_rtx_sbsa_external_dir lambda and use it when IS_SBSA and USE_TRT_RTX.
- Similarly the SBSA wheels have the wrong dependencies. `get_sbsa_requirements` today unconditionally appends `tensorrt>=11.0.0,<11.1.0` and does not have a USE_TRT_RTX branch (unlike get_x86_64_requirements which emits `tensorrt_rtx>=1.5.0.114,<1.6.0.0`)
- These wheels also can get pushed to pypi as it is right now : should we gate it till we see everything passes?

.github/workflows/build-test-linux-aarch64_rtx.yml
```diff
@@ -0,0 +1,91 @@
...
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - repository: pytorch/tensorrt
+            pre-script: packaging/pre_build_script.sh
+            env-var-script: packaging/env_vars.txt
+            post-script: packaging/post_build_script.sh
+            smoke-test-script: packaging/smoke_test_script.sh
```
tp5uiuc, line 72: I see this in CI
```
# Smoke test is intentionally disabled.
# The issue was smoke test installs the built torch_tensorrt wheel file and checks `import torch_tensorrt; print(torch_tensorrt.__version__)`
# Since tensorrt cannot be pip installable in CI, the smoke test will fail.
# One way we tried to handle it is manually install tensorrt wheel while by extracting from the tarball.
# However, the TensorRT-10.8.0.43/lib path doesn't seem to show up in LD_LIBRARY_PATH even if we explicitly set it.
# TODO: Implement a custom smoke_test script to verify torch_tensorrt installation.
```
is this still true? I checked it because of the comment below (where malformed wheels can get pushed to pypi potentially). Why is `tensorrt` not pip installable in CI (because of GPU requirements or something)? An `import torch_tensorrt; assert ENABLED_FEATURES.tensorrt_rtx` check would suffice here I think

.github/workflows/build-test-linux-aarch64_rtx.yml
narendasan: > These wheels also can get pushed to pypi as it is right now : should we gate it till we see everything passes?

I think its ok since we dont run tests on this channel in CI, there are no gpu runners

tp5uiuc (approved):

PR #4427, opened 2026-07-22
fix: avoid linking transitive PyTorch CUDA libraries into libtorchtrt

fs-eire:
## Description

Remove cuDNN, cuSparseLt, and NVSHMEM host libraries from the `srcs` of the
LibTorch Bazel target.

These libraries are transitive dependencies of PyTorch's `libtorch_cuda.so`,
not direct dependencies of Torch-TensorRT. Adding them to `srcs` causes Bazel
to propagate them into `libtorchtrt.so` as direct `DT_NEEDED` entries. As a
result, importing a locally built Torch-TensorRT wheel can fail with errors
such as:

```text
OSError: libcudnn_adv.so.9: cannot open shared object file
```

unless the corresponding `site-packages/nvidia/*/lib` directories are added
to `LD_LIBRARY_PATH`.

This primarily affects local builds because the local_torch repository rule
exposes the sibling site-packages/nvidia directory, allowing these globs to
match. In nightly CI, `@libtorch` points directly to site-packages/torch, so
the same globs are empty and the resulting nightly wheel does not acquire
these direct dependencies.

The existing local_torch NVIDIA symlink remains available for resolving
LibTorch's transitive dependencies during C++ linking. This change only
prevents those dependencies from becoming part of Torch-TensorRT's direct
link interface.

## Type of change

# Checklist:

### DIFF
--- a/third_party/libtorch/BUILD
+++ b/third_party/libtorch/BUILD
@@ -33,9 +33,6 @@ cc_library(
     }) + glob(
         [
             "lib/libtorch_nvshmem.so*",
-            "nvidia/cusparselt/lib/libcusparseLt.so*",
-            "nvidia/cudnn/lib/libcudnn*.so*",
-            "nvidia/nvshmem/lib/libnvshmem_host.so*",
         ],
         allow_empty = True,
     ),
--- a/tests/py/core/test_libtorchtrt_linkage.py
+++ b/tests/py/core/test_libtorchtrt_linkage.py
@@ -1,0 +1,78 @@
+import importlib.util
+import os
+import re
+import subprocess
+import sys
+import unittest
+from pathlib import Path
+
+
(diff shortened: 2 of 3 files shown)


third_party/libtorch/BUILD
```diff
@@ -33,9 +33,6 @@ cc_library(
     }) + glob(
         [
             "lib/libtorch_nvshmem.so*",
-            "nvidia/cusparselt/lib/libcusparseLt.so*",
-            "nvidia/cudnn/lib/libcudnn*.so*",
-            "nvidia/nvshmem/lib/libnvshmem_host.so*",
```
lanluo-nvidia, line 38: can you please also remove the the symlink section in the toolchains/local_torch.bzl

fs-eire: done

lanluo-nvidia (approved):
lgtm

Issue #4390, opened 2026-07-07
✨[Feature] Can we utilize PyTorch's complex decomposition


narendasan:
**Is your feature request related to a problem? Please describe.**

We maintain our own complex numerics decomposition, but that means we need to add cases op by op. 

**Describe the solution you'd like**

PyTorch has a new pass to decompose complex numerics using Tensor subclasses https://github.com/pytorch/pytorch/pull/169832/changes. Can we use this or augment our passes with this to expand coverage? 

**Describe alternatives you've considered**

**Additional context**

apbose:
Merged

PR #4430, opened 2026-07-23
feat: flag-gated upstream complex decomposition adapter 

apbose:
Adds an alternative complex-numerics lowering path that delegates the interior complex->real expansion to PyTorch's upstream decomposition (decompose_complex_in_graph, pytorch/pytorch#169832) instead of the hand-maintained per-op rewriter in complex_graph_rewrite.py.

The new complex_decomposition_adapter pass:
  - captures the complex I/O signature (reusing the existing helpers) so the post-partition _insert_complex_io_adapters boundary code works unchanged;
  - calls decompose_complex_in_graph on the exported graph, feeding it the placeholders' fake tensors to preserve the export ShapeEnv/SymInts;
  - normalizes the SoA aten.complex/real/imag seams it leaves behind into the interleaved [...,2] layout the rest of the TRT flow expects.

Wiring:
  - use_complex_decomposition setting (default: True when torch>=2.14.dev is installed, via has_complex_decomposition(); False otherwise) selects new vs legacy pass via complex_lowering_pass in _aten_lowering_pass;
  - falls back to complex_graph_detection both on the torch-version gate and if decompose_complex_in_graph itself raises (e.g. an unsupported op like aten.log2/log10 as of torch 2.14.dev).

Tests:
  - test_complex_ops.py parametrized [legacy]/[decomp] via an autouse fixture so the existing suite runs on both paths (decomp auto-skips on torch<2.14);
  - new GPU-free test_complex_decomposition_adapter.py covering the SoA->[...,2] normalization and the fallback gate.

**Behavioral change on torch>=2.14.dev:** this path is auto-enabled whenever the upstream torch feature is available, not opt-in. `use_complex_decomposition=False` forces the legacy path. On torch<2.14.dev, behavior is unchanged.

#4390

### DIFF
--- a/py/torch_tensorrt/_features.py
+++ b/py/torch_tensorrt/_features.py
@@ -27,6 +27,7 @@ FeatureSet = namedtuple(
         "tensorrt_rtx",
         "trtllm_for_nccl",
         "native_trt_collectives",
+        "complex_decomposition",
     ],
 )
 
@@ -49,6 +50,11 @@ _TENSORRT_RTX = getattr(tensorrt, "_package_name", "") == "tensorrt_rtx"
 _TS_FE_AVAIL = os.path.isfile(linked_file_full_path)
 _TORCHTRT_RT_AVAIL = _TS_FE_AVAIL or os.path.isfile(linked_file_runtime_full_path)
 _DYNAMO_FE_AVAIL = version.parse(sanitized_torch_version()) >= version.parse("2.1.dev")
+# PyTorch's upstream complex-tensor decomposition (pytorch/pytorch#169832) ships
+# in torch>=2.14.dev; gates the complex_decomposition_adapter lowering pass.
+_COMPLEX_DECOMP_AVAIL = version.parse(sanitized_torch_version()) >= version.parse(
+    "2.14.dev"
+)
 _FX_FE_AVAIL = False if _TENSORRT_RTX else True
 _REFIT_AVAIL = True
 _WINDOWS_CROSS_COMPILE = check_cross_compile_trt_win_lib()
@@ -87,6 +93,7 @@ ENABLED_FEATURES = FeatureSet(
     _TENSORRT_RTX,
     _TRTLLM_AVAIL,
     _NATIVE_TRT_COLLECTIVES_AVAIL,
+    _COMPLEX_DECOMP_AVAIL,
 )
 
 T = TypeVar("T")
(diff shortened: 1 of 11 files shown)


zewenli98 (commented):
py/torch_tensorrt/dynamo/_defaults.py Lines: 22-34
```
The PR description says "Default off; legacy path unchanged." but this file sets:

    USE_COMPLEX_DECOMPOSITION = has_complex_decomposition()

That will enable the new decomposition path automatically whenever torch >= 2.14, changing runtime behavior for users without an explicit opt-in.

Please make the behavior consistent with the description. Two options:

1) Recommended (opt-in):
   - Make the default explicit False so the legacy path remains the default:
       USE_COMPLEX_DECOMPOSITION = False
   - Keep the runtime gate in the adapter so the new path only runs when both the user opts in and the torch feature is available.

2) If the intent is to enable the new path by default when upstream exists:
   - Update the PR description to reflect that and add a release note calling out the behavioral change (users on torch>=2.14 will see the new lowering by default).

Also add/adjust a unit test that asserts the default behavior so CI will catch accidental flips of the default.
```

py/torch_tensorrt/dynamo/lowering/passes/complex_decomposition_adapter.py Lines: 90-120
```
_fake_flat_args() builds fallback tensors from node.meta["tensor_meta"] using:

    torch.empty(tm.shape, dtype=tm.dtype, device=tm.device)

If tm.shape contains SymInt or other symbolic/dynamic dimensions this may fail or lose ShapeEnv/SymInt semantics. Please add defensive handling and tests:

- Validate/assert whether tensor_meta.shape may contain symbolic dims; if so, either:
  * prefer raising a clear error explaining that decompose_complex_in_graph requires concrete example tensors (with a pointer to docs), or
  * synthesize safe concrete dims (e.g. 1) while logging a warning, and document the consequences.

- Add a unit test (or expand the existing spike) that runs the retrace on a model with dynamic/symbolic placeholder shapes to ensure ShapeEnv survives or that we fail with a clear message.

Finally, add an explanatory comment in the code describing the exact assumption (why we prefer node.meta["val"] and what we expect from tensor_meta).
```

tests/py/dynamo/hlo/test_complex_ops.py Lines: 62-110
```
The autouse fixture at module scope mutates module-level globals (_COMPILE, _DECOMP) to inject use_complex_decomposition into compile kwargs across tests.

This pattern can cause test-order or cross-test pollution (especially if tests run in parallel or a teardown fails). Please consider a safer approach:

- Use a fixture that yields the compile kwargs and have tests accept that fixture explicitly (avoid mutating module globals), or
- Use pytest.monkeypatch to set attrs on a settings object or to patch the call site (monkeypatch.setitem / setattr) so changes are scoped to the current test.

Also add a brief comment explaining why the tests set the flag explicitly (to avoid accidental reliance on the default), and confirm there are no parallel/ordering assumptions that could break CI.
```

apbose:
For the  `_fake_flat_args`, it now raises a clear `RuntimeError` if the `tensor_meta`fallback ever hits a symbolic dimension. However, in compie pipeline this fallback branch should not be reachable in pratice since it should have `node.meta['val']` after export.

zewenli98 (approved):
not familiar with the details but looks good to me overall

Issue #2976, opened 2024-07-02
The doc needs a fix


zewenli98:
There's a copy-paste error:
https://github.com/pytorch/TensorRT/blob/2bba527213270aca7290baadefaa562f12b84b2c/py/torch_tensorrt/dynamo/lowering/passes/_aten_lowering_pass.py#L102

PR #4431, opened 2026-07-24
fix: support ITensor amex in quantize and add `export_torch_mode()` wrapper

### LINKED ISSUE #4429: torch_tensorrt.compile() fails with "fake tensor in exported program constant's list" for ModelOpt quantized models unless wrapped in export_torch_mode()
##  Bug Description

When compiling a ModelOpt-quantized model with Torch-TensorRT, `torch_tensorrt.compile()` fails unless it is wrapped in `export_torch_mode()`.

Without `export_torch_mode()`, FP8 (`mtq.FP8_DEFAULT_CFG`) ) quantized models fail during export with the following error:

```
RuntimeError: We found a fake tensor in the exported program constant's list.
This typically means our tracing system encountered an op that we can't trace through.
For the potential source, you can refer to following model attribute:
net.0.input_quantizer.lifted_tensor_0.
Please file an issue on github.
```

Wrapping the compile call in:

```python
with export_torch_mode():
    trt_model = torch_tensorrt.compile(...)
```

avoids this error.

## To Reproduce

1. Launch the NVIDIA PyTorch container:
   ```bash
   docker run --gpus all -it --rm nvcr.io/nvidia/pytorch:26.06-py3
   ```

2. Quantize a simple model using ModelOpt (`mtq.FP8_DEFAULT_CFG`).

3. Compile using Torch-TensorRT **without** `export_torch_mode()`:

```python
import torch
import torch.nn as nn
import torch_tensorrt

import modelopt.torch.quantization as mtq

device = "cuda"

class SimpleMLP(nn.Module):
    def __init__(self):
        super().__init__()
        self.net = nn.Sequential(
            nn.Linear(512, 4096),
            nn.GELU(),
            nn.Linear(4096,4096),
            nn.GELU(),
            nn.Linear(4096,512) 
        )

    def forward(self, x):
        return self.net(x)

model = SimpleMLP().to(device).eval()

quant_cfg = mtq.FP8_DEFAULT_CFG  

calib_data = [
    (torch.randn(128, 512, device=device),)
    for _ in range(10)
]

def forward_loop(model):
    with torch.no_grad():
        for (x,) in calib_data:
            model(x)

model = mtq.quantize(
    model,
    quant_cfg,
    forward_loop,
)

inputs = [
    torch_tensorrt.Input(
        shape=(128, 512),
        dtype=torch.float32,
    )
]

trt_model = torch_tensorrt.compile(
    model,
    ir="dynamo",
    inputs=inputs,
)
```

This produces:

```
RuntimeError: We found a fake tensor in the exported program constant's list.
This typically means our tracing system encountered an op that we can't trace through.
For the potential source, you can refer to following model attribute:
net.0.input_quantizer.lifted_tensor_0.
Please file an issue on github.
```

## Expected behavior

`torch_tensorrt.compile()` should successfully export the quantized model, or emit a more descriptive error if `export_torch_mode()` is required.

Currently, the error does not indicate that wrapping the compilation in `export_torch_mode()` resolves the issue.

## Environment

- Torch-TensorRT Version: 2.13.0a0
- PyTorch Version: 2.13.0a0+8145d630e8.nv26.6.54250401
- CPU Architecture: aarch64
- OS: Ubuntu 24.04.4
- How you installed PyTorch: NVIDIA NGC container (`nvcr.io/nvidia/pytorch:26.06-py3`)
- Are you using local sources or building from archives: Archive/container image
- Python version: 3.12.3
- CUDA version: 13.3
- GPU models and configuration: NVIDIA Jetson AGX Thor (Blackwell)
  - TensorRT 11.0.0.114
  - NVIDIA ModelOpt 0.44.0
  - cuDNN 9.23.0

## Additional context

This bug showed up when I was trying to figure out this issue related to compiling an FP4 quantized model: 
https://forums.developer.nvidia.com/t/fp4-quantization-fails-on-jetson-thor-with-tensorrt-11-0-0-114/377887

zewenli98:
# Description

1. support ITensor amex in quantize
2. add `export_torch_mode()` wrapper in `torch_tensorrt.compile` to modelopt quant models 

Fixes #4429 #2976

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_tracer.py
+++ b/py/torch_tensorrt/dynamo/_tracer.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import importlib.util
 import logging
+from contextlib import nullcontext
 from inspect import signature
 from typing import Any, Optional, Tuple, Union
 
@@ -79,13 +81,22 @@ def trace(
     dim_registry = build_dim_registry(arg_inputs, kwarg_inputs)
     dynamic_shapes = get_dynamic_shapes_args(mod, arg_inputs, dim_registry)
     dynamic_shapes.update(get_dynamic_shapes_kwargs(kwarg_inputs, dim_registry))
-    exp_program = export(
-        mod,
-        tuple(torch_arg_inputs),
-        kwargs=torch_kwarg_inputs,
-        dynamic_shapes=dynamic_shapes,
-        strict=kwargs.get("strict", False),
-    )
+
+    export_context = nullcontext()
+    if importlib.util.find_spec("modelopt") is not None:
+        from modelopt.torch.quantization.utils import export_torch_mode, is_quantized
+
+        if is_quantized(mod):
+            export_context = export_torch_mode()
+
+    with export_context:
+        exp_program = export(
+            mod,
+            tuple(torch_arg_inputs),
+            kwargs=torch_kwarg_inputs,
+            dynamic_shapes=dynamic_shapes,
+            strict=kwargs.get("strict", False),
+        )
 
     return exp_program
 
--- a/py/torch_tensorrt/dynamo/_tracer.py
+++ b/py/torch_tensorrt/dynamo/_tracer.py
@@ -10,7 +10,11 @@ import torch
 from torch.export import Dim, export
 from torch_tensorrt._Input import Input
 from torch_tensorrt.dynamo._defaults import default_device
-from torch_tensorrt.dynamo.utils import get_torch_inputs, to_torch_device
+from torch_tensorrt.dynamo.utils import (
+    get_torch_inputs,
+    is_quantized_by_modelopt,
+    to_torch_device,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -68,6 +72,7 @@ def trace(
             "'arg_inputs' and 'inputs' should not be used at the same time."
         )
     arg_inputs = inputs if inputs is not None else arg_inputs
+    assert arg_inputs is not None
 
     if kwarg_inputs is None:
         kwarg_inputs = {}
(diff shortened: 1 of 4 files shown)


narendasan (commented):
Can you explain what this context is and maybe add some tests?

zewenli98:
> Can you explain what this context is and maybe add some tests?
1. In quantize(), `amex` could be trt.ITensor which doesn't have `amax.numel()`
2. ModelOpt requires `with export_torch_mode()` wrapper to make inserted quantizers traceable while exporting pytorch models. See: #4429

narendasan:
> > Can you explain what this context is and maybe add some tests?
> 
>     1. In quantize(), `amex` could be trt.ITensor which doesn't have `amax.numel()`
> 
>     2. ModelOpt requires `with export_torch_mode()` wrapper to make inserted quantizers traceable while exporting pytorch models. See: [torch_tensorrt.compile() fails with "fake tensor in exported program constant's list" for ModelOpt quantized models unless wrapped in export_torch_mode() #4429](https://github.com/pytorch/TensorRT/issues/4429)

I see, is there any harm in always having this context? should we catch cases where the modelopt is missing but the context is needed and throw an error/warning?

zewenli98:
> I see, is there any harm in always having this context? should we catch cases where the modelopt is missing but the context is needed and throw an error/warning?

I remember ModelOpt is not required in torch-trt. If we need to check all models with ModelOpt API, the package should be required rather than opt in.

py/torch_tensorrt/dynamo/_tracer.py
```diff
@@ -79,13 +81,22 @@ def trace(
...
-        kwargs=torch_kwarg_inputs,
-        dynamic_shapes=dynamic_shapes,
-        strict=kwargs.get("strict", False),
-    )
+
+    export_context = nullcontext()
+    if importlib.util.find_spec("modelopt") is not None:
+        from modelopt.torch.quantization.utils import export_torch_mode, is_quantized
+
+        if is_quantized(mod):
```
narendasan: Should we throw a warning if its a quantized model but model opt is not available? Does this matter for torchao or other sources?

zewenli98: well since the function `is_quantized` is imported from modelopt, it must be available at this point. If we want to throw errors/warnings, I think we might need to write a helper function to detect quantizers to override `is_quantized`. For other sources like torchao, if they insert quantizers or similar indicators, we can use the same mechanism.

zewenli98:
@narendasan I optimized codes so that if modelopt is not installed, it will fallback to check module names. And it will error out for quantized model if modelopt is requirement. Also, it's easy to generalize for other sources like torchao.

narendasan (approved):
Lgtm

PR #4434, opened 2026-07-26
Bump ExecuTorch to 1.4.1

shoumikhin:
## The problem

The pinned ExecuTorch version was 1.3.1. This moves it to 1.4.1, the current stable
release.

## Why several files change

ExecuTorch is referenced twice in different forms. The Bazel build fetches a git commit,
and the Python runtime package installs a released wheel. If those two disagree, the C++
runtime and the Python package get built from different source, and the mismatch does not
show up until much later as a serialized engine layout error. So every place that names
the version moves together:

```
MODULE.bazel and its docker and CI copies    git commit for the 1.4.1 tag
the ExecuTorch build workflow                pip install executorch==1.4.1
the runtime package README and pyproject     same version
```

The comments in the docker and CI copies previously described the pin as a branch head.
They point at a release tag, so they now say so, matching `MODULE.bazel`.

## Testing

Verified that Bazel resolves the new pin, and that the source it resolves to is the
intended release rather than some other commit:

```
bazel fetch @executorch//...
  All external dependencies for the requested targets fetched successfully
fetched tree version.txt: 1.4.1
```

### DIFF
--- a/.github/workflows/executorch-build-linux.yml
+++ b/.github/workflows/executorch-build-linux.yml
@@ -79,7 +79,7 @@ jobs:
         export PATH="${RUNNER_TEMP}/bin:${PATH}"
         bazel --version
 
-        python -m pip install pyyaml "executorch==1.3.1"
+        python -m pip install pyyaml "executorch==1.4.1"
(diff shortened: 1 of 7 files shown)


lanluo-nvidia (commented):
@shoumikhin the change itself lgtm.
I will first merge in my build executorch-runtime wheel PR: https://github.com/pytorch/TensorRT/pull/4398
(Waiting on @narendasan 's final review and approve)
and then we can add this on top of it.

PR #4435, opened 2026-07-27
feat: add triton_op for AOT QDP plugins from Triton kernels

bowang007:
# Description

Add `torch_tensorrt.kernels.triton_op` so that triton kernels could be compiled and used using TRT AOT QDP plugins.

## Type of change

# Checklist:


lanluo-nvidia (commented):
- P1 — extra_args_fn length is never checked against the scalar portion of signature. In triton_op (https://github.com/pytorch/TensorRT/pull/4435/files#diff-9f1db433ae3bf6f7c0e23a85d898ca68d799b3e1795195b9283c79aef0a22dbdR270), the derived AOT launcher accepts any number
    of returned expressions. A signature such as (x_ptr, n, m, y_ptr) with an extra_args_fn returning only [n] passes registration and produces a misaligned/missing kernel argument at runtime. The code already has layout.scalars; require len(values) == len(layout.scalars)
    and add under/over-supply tests.

  - P1 — floating-point scalar parameters are accepted even though this AOT path only constructs SymIntExprs. The signature validator permits, for example, {"x": "*fp32", "scale": "fp32", "y": "*fp32"}, but extra_args_fn can only feed symbolic integer expressions into the
    launch. The scale then receives incompatible bits rather than a float, yielding silently incorrect output. Reject non-integer scalar types in the derived AOT path (or route them through a supported JIT implementation); the existing CUDA-kernel API already documents
    this limitation.

PR #4440, opened 2026-07-28
feat(executorch): expose composable Edge export API

shoumikhin:
## Why this PR is needed

`torch_tensorrt.save(..., output_format="executorch")` is the easiest way to create an ExecuTorch `.pte` file. It completes the whole export pipeline immediately:

```python
import torch
import torch_tensorrt

model = MyModel().eval().cuda()
inputs = [torch.randn(1, 3, 224, 224, device="cuda")]
trt_module = torch_tensorrt.compile(model, ir="dynamo", arg_inputs=inputs)

torch_tensorrt.save(
    trt_module,
    "model.pte",
    output_format="executorch",
)
```

That remains the recommended API for common one-step exports.

Some workflows need to stop before final ExecuTorch memory planning and serialization. For example, users may want to:

- inspect the delegated Edge graph;
- run custom Edge transforms;
- send operations that TensorRT does not support to another backend;
- preserve independent methods such as `prefill` and `decode`;
- add constant methods or generate an ETRecord;
- choose the final ExecuTorch backend configuration.

This PR adds `torch_tensorrt.executorch.export()`. It returns ExecuTorch's standard `EdgeProgramManager`, which is the supported boundary for inspecting and customizing an Edge program before calling `to_executorch()`.

## Which API should I use?

Use `torch_tensorrt.save()` when you want a `.pte` in one step.

Use `torch_tensorrt.executorch.export()` when you need control before the final `.pte` is created:

```python
import torch_tensorrt.executorch

edge = torch_tensorrt.executorch.export(trt_module)

# Inspect or transform the Edge program here.
print(edge.exported_program().graph)

program = edge.to_executorch()
with open("model.pte", "wb") as output:
    program.write_to_file(output)

# Advanced callers must also persist external tensor data when present.
program.write_tensor_data_to_file(".")
```

`save()` now delegates to the same implementation, so the simple and advanced paths do not maintain separate lowering logic.

## Example: TensorRT with a fallback backend

TensorRT always receives the first chance to claim its prebuilt engine nodes. Additional partitioners run afterward and can claim operations that TensorRT does not support.

Assume `cuda_partitioner` is an ExecuTorch partitioner configured for your target:

```python
edge = torch_tensorrt.executorch.export(
    trt_module,
    partitioners=[cuda_partitioner],
)

program = edge.to_executorch()
```

This can produce one ExecuTorch program containing both TensorRT and CUDA delegates.

## Example: multiple methods

Assume `prefill_program` and `decode_program` are independent, engine-bearing `ExportedProgram` objects, and the two spec lists are configured for their methods:

```python
edge = torch_tensorrt.executorch.export(
    {
        "prefill": prefill_program,
        "decode": decode_program,
    },
    compile_specs={
        "prefill": prefill_specs,
        "decode": decode_specs,
    },
)

program = edge.to_executorch()
```

The mapping preserves method names and allows per-method partitioner and compile-spec pipelines. It does not imply that mutable state is shared between methods.

## Example: transforms, metadata, and ETRecord

Assume the transform passes and Edge configuration below are provided by the caller:

```python
edge = torch_tensorrt.executorch.export(
    trt_module,
    transform_passes=my_edge_passes,
    compile_config=edge_compile_config,
    constant_methods={"get_vocab_size": 256},
    generate_etrecord=True,
)

program = edge.to_executorch()
program.get_etrecord().save("model_etrecord.bin")
```

## Accepted inputs

The new API accepts:

- a TensorRT-compiled `torch.fx.GraphModule`;
- an engine-bearing `torch.export.ExportedProgram`;
- a mapping from method names to independent `ExportedProgram` objects.

A plain `torch.nn.Module` must be compiled with Torch-TensorRT first.

## Safety and memory behavior

Engine rewriting changes the exported graph. The new API stages an independent copy of graph structure, signatures, state containers, constants, and metadata before rewriting anything. Rewrite and lowering failures therefore do not consume or structurally modify the caller's original program.

Tensor and TensorRT engine payload storage is treated as immutable and shared with the staged structure. This avoids deep-copying potentially multi-gigabyte engines. Custom transform passes must not mutate shared payload objects.

Metadata describing symbolic shapes is also shared rather than copied. It points back into the live shape environment the exported program is guarded on, and that environment holds fake tensors which cannot be copied. Sharing it is what keeps dynamic-shape models working; copying it would also detach the copy from the shape symbols the graph depends on.

When a lifted TensorRT engine is replaced, the old placeholder, graph-signature entry, and constant are removed. An engine feeding several call nodes is serialized once, and those nodes reuse one materialized `uint8` payload buffer.

## Compatibility

- Existing one-step `save(..., output_format="executorch")` behavior is preserved, with one
  difference: a falsy non-list `partitioners` or `compile_specs`, such as `{}`, now raises
  `TypeError` instead of being read as an empty list. `save()` lowers a single method, so it
  keeps rejecting the per-method mappings that `export()` accepts.
- `.pte` serialization and external `.ptd` persistence are preserved.
- TensorRT remains the first partitioner.
- Shared dynamic dimensions declared with `Input(shared_dims=...)` remain shared.
- Zero-engine methods are allowed. A later partitioner may claim them, or portable operators may remain undelegated.

## Validation

Built from source with the Torch-TensorRT C++ runtime active, against TensorRT and
ExecuTorch, and exercised on both x86_64 and aarch64 CUDA GPUs.

Checked end to end with a real compiled TensorRT engine:

- static-shape `export()` writes a loadable `.pte`;
- dynamic-shape `export()` writes a loadable `.pte` of the same size the previous
  one-step save path produced;
- the caller's `GraphModule` still holds its original engine nodes after `export()`;
- engine payloads are shared rather than re-serialized;
- `save(..., output_format="executorch")` still writes a `.pte`;
- `save()` still raises `TypeError` for a non-list `partitioners` or `compile_specs`.

The whole ExecuTorch test directory passes on a CUDA machine, including the five tests
in `test_cuda_partitioner_composition.py`.

These tests do run on a pull request. The standard test matrix lists this directory in the
nightly lane, but the ExecuTorch runtime channel of the Linux x86_64 workflow runs
`pytest executorch/` on every lane except `skip` and `rtx`, and that channel feeds the
required gate.

One of those tests drives validation and then the rewrite, and asserts the engine was
resolved only once. Reading an engine's state serializes it and base64 encodes the
result, so resolving the same engine in both steps would double that cost on programs
whose purpose is to avoid copying large engines. The test was verified to fail when the
reuse path is removed from the rewrite.

Note on memory: each engine is serialized once, and a method's engine payloads are
released as soon as that method is rewritten, so a multi-method export does not hold
every method's payloads at once. While a method is being rewritten its engines are
resident both as the base64 text the serialization produced and as the decoded buffer,
so peak memory is roughly 2.3x that method's engine bytes, and briefly 3.3x for the
engine being decoded.

Static checks on the changed files: import ordering, formatting, lint, and type
checking.

The end-to-end run with a real compiled engine above covers the first revision of this
branch. The later commits change placeholder staging, the deepcopy memo, engine
serialization, the engine buffer name, and the per-method pass dictionary. Each of them
is covered by a unit test in `test_export.py` that was checked against its parent commit,
so the test fails without the fix and passes with it. The end-to-end run will be repeated
on this head before merge.

## Related work

PR #4433 forwards additional Edge-lowering options through one-step `save()`. After this PR lands, #4433 can keep `save()` as a thin wrapper over `torch_tensorrt.executorch.export()` instead of extending the previous private lowering path.


py/torch_tensorrt/executorch/_export_utils.py
```diff
@@ -0,0 +1,309 @@
...
+                    if isinstance(leaf, (torch.Tensor, torch.ScriptObject)):
+                        memo[id(leaf)] = leaf
+    return memo
+
+
+def _stage_graph_module(
+    graph_module: torch.fx.GraphModule,
+    payload_memo: dict[int, Any],
+) -> torch.fx.GraphModule:
+    staged = copy.deepcopy(graph_module, payload_memo)
```
cehongwang, line 168: Does this doubles the GPU/Host memory usage? Is the copy absolutely necessary?

cehongwang: Or is there a way to not copy the weight, only the graph structure

py/torch_tensorrt/executorch/_export_utils.py
```diff
@@ -0,0 +1,309 @@
...
+                engine_bytes = bytes(engine_bytes)
+            engine_tensor = torch.frombuffer(bytearray(engine_bytes), dtype=torch.uint8)
+
+            from torch.fx.experimental.const_fold import get_unique_attr_name_in_module
+
+            buffer_name = get_unique_attr_name_in_module(graph_module, "_trt_engine_0")
+            graph_module.register_buffer(buffer_name, engine_tensor, persistent=True)
+            exported_program.state_dict[buffer_name] = engine_tensor
+            str_args = [
+                str(value) if value is not None else ""
```
cehongwang: engine_info[ENGINE_IDX] is raw bytes (often hundreds of MB / multi-GB). str(b"...") builds a temporary Python string of roughly 4 * len(bytes) characters (b'\xab\xcd...' style), then immediately throws that slot away and uses engine_attr_node instead.

py/torch_tensorrt/executorch/_export.py
```diff
@@ -0,0 +1,344 @@
...
+    on it when ready to perform final memory planning and serialization.
+
+    Export stages independent graph and metadata containers while sharing tensor
+    and engine payload storage, avoiding copies of potentially multi-gigabyte
+    TensorRT engines. Transform passes must treat shared payload values as
+    immutable. Method mappings preserve independent entry points but do not imply
+    shared mutable state between them.
+    """
+    from torch_tensorrt._features import ENABLED_FEATURES
+
```
cehongwang, line 371: Do we need a check that the platform is linux?

py/torch_tensorrt/executorch/_export_utils.py
shoumikhin: Good question to push on, and no, weights are not copied. Only graph structure is,
which is what you suggest in your next comment.

`_payload_sharing_memo` (just above) pre-seeds `copy.deepcopy`'s memo with every tensor
and engine payload reachable from the program, so `deepcopy` returns those objects
as-is instead of duplicating them. `state_dict` and `constants` are then shallow-copied
dicts, so the keys are independent but the values are the same objects.

I measured it to be sure, on a model with 33.6 MB of weights:

```
weights in program: 33.6 MB
live CPU tensor bytes before staging: 33.6 MB
live CPU tensor bytes after  staging: 33.6 MB
growth: 0.0 MB

state_dict entries: 4, shared by identity: 4
  a.weight   same object: True  same storage: True
  a.bias     same object: True  same storage: True
  b.weight   same object: True  same storage: True
  b.bias     same object: True  same storage: True
```

So the copy is graph nodes, signatures, and metadata containers, all small. TensorRT
engine blobs are shared the same way, which matters more since those can be multiple
gigabytes.

On whether the copy is necessary at all: `replace_execute_engine` rewrites the graph in
place (registers a buffer, erases nodes, edits the graph signature). Without staging, a
failure part-way through would leave the caller's `ExportedProgram` half-rewritten. The
GraphModule path happens to be safe already because `dynamo/_exporter.py` deep-copies,
but the pre-exported `ExportedProgram` and multi-method paths are handed to us directly,
so staging is what keeps those inputs intact.

The cost this leaves is documented as a constraint: transform passes must not mutate the
shared payloads in place, since those are the caller's tensors.

shoumikhin: Yes, that is exactly what it does. Only graph structure is copied; tensors and engine
payloads are shared by identity through the pre-seeded `deepcopy` memo. Measurements in
the reply above: 33.6 MB of weights in, 0.0 MB growth, all four `state_dict` tensors
identical objects sharing the same storage.

One case needed a carve-out, which is worth mentioning since it was a real bug. Metadata
describing symbolic shapes (`SymInt`, fake tensors) cannot be deep-copied at all: it
reaches back into the live `ShapeEnv`, which holds fake tensors, and copying raised

```
RuntimeError: Cannot access data pointer of Tensor (e.g. FakeTensor, FunctionalTensor)
```

on every dynamic-shape model. That metadata is now shared rather than copied, which also
keeps the staged graph attached to the shape symbols it is guarded on. Covered by
`test_stage_exported_program_supports_dynamic_shapes`.

py/torch_tensorrt/executorch/_export_utils.py
shoumikhin: Good catch, this was a real waste and it is fixed now.

You are right that the slot is thrown away: `no_op_args` splices in `engine_attr_node` at
`ENGINE_IDX`, so the stringified engine was built and then never used. Measured on a
64 MB stand-in payload:

```
engine bytes      : 67.1 MB
str(bytes) length : 268.4 M chars
str object bytes  : 268.4 MB
blowup factor     : 4.0 x
```

So a 2 GB engine burned about 8 GB of host memory building a string that was immediately
discarded.

Now that index is skipped:

```python
str_args = [
    ("" if index == ENGINE_IDX else str(value) if value is not None else "")
    for index, value in enumerate(engine_info[:SERIALIZATION_LEN])
]
```

Re-verified on a GPU afterward: static-shape and dynamic-shape `export()` still produce
loadable `.pte` files of the same size as before, and `save(output_format="executorch")`
is unchanged.

py/torch_tensorrt/executorch/_export.py
shoumikhin: Yes, we do, and it was missing. Added.

`save(output_format="executorch")` already rejects non-Linux platforms in `_compile.py`,
so the advanced entry point was skipping a check the one-step path enforces. `export()`
now raises the same error with the same wording:

```python
if platform.system() != "Linux":
    raise ValueError(
        f"The executorch format is only supported on Linux, {platform.system()} "
        "is not a supported platform for this format"
    )
```

Placed before the runtime-feature check so the platform message comes first, and covered
by `test_export_rejects_non_linux_platform`.

py/torch_tensorrt/executorch/_export_utils.py
shoumikhin: Re-verified after the follow-up commits, since `_stage_graph_module` changed since my
last reply (a container in `node.meta` that held a fake tensor was being shared wholesale, so
now only the shape-bound leaves are memoized and the container itself is copied).

Weight sharing is unaffected, measured on the current head:

```
state_dict entries    : 4
shared by identity    : 4/4
same storage pointer  : 4/4
graph module copied   : True
```

So the answer stands: no weight bytes are copied, only graph structure and metadata containers.
Resolving this one.

docsrc/user_guide/runtime_performance/saving_models.rst
```diff
@@ -243,6 +244,77 @@ backend. It requires the ``executorch`` package (``pip install
         retrace=False, arg_inputs=inputs,
     )
 
+``save`` writes both the ``.pte`` and any external ``.ptd`` tensor-data files.
```
lanluo-nvidia, line 261: I would suggest here first add the torch_tensorrt.save() as the default method for user wants to export a  executorch pte file. 
then add the advanced method, when user needs a customization before  serialize into a pte file.
So that we explain user both scenario clearly.

lanluo-nvidia: Also please wait for PR: 4398 to be merged to main(I will push by the end of today.)
please rebase on top of that.

shoumikhin:
Both suggestions are done.

**Documenting save() as the default.** The guide showed both ways to write a `.pte` without
saying which one to reach for, so the choice had to be inferred from the order the examples
appeared in. It now states it directly, and each example has its own heading:

* `torch_tensorrt.save()` is the default, and the right choice whenever a `.pte` file is all
  that is needed.
* `torch_tensorrt.executorch.export()` is the advanced path, for when the program has to be
  changed before it is written to disk, for example to inspect the delegated graph, add
  metadata, control final memory planning, or carry more than one method.

**Waiting for the runtime wheel PR.** That has merged, and this branch is rebased on a main
that contains it, so nothing is pending on that side.

PR #4441, opened 2026-07-28
Implemented executorch's multi-optimization profile

cehongwang:
# Description

Adds multi-optimization-profile support to the ExecuTorch TensorRT delegate, so one `.pte` can hold a single engine tuned for several shape regimes (the motivating case is an LLM's `seq == 1` decode versus a long prefill) and pick between them per call.

TensorRT already supports several optimization profiles per engine, and the authoring side already exists via `torch_tensorrt.Input(profiles=[...])`. What was missing was any way for the ExecuTorch runtime to say *which* profile a given execution should use, so every call ran on profile 0.

## Public C++ API

`OptimizationProfileGuard`, in the installed `torch_tensorrt/executorch/TensorRTBackend.h`:

```cpp
constexpr int32_t kDecodeProfile = 0;  // export order: decode first,
constexpr int32_t kPrefillProfile = 1; // then prefill

executorch::extension::Module module("model.pte");
{
  OptimizationProfileGuard profile_guard(kPrefillProfile);
  auto result = module.forward(prefill_inputs);
}
{
  auto profile_guard = OptimizationProfileGuard::automatic();
  auto result = module.forward(decode_inputs);  // chosen from input shapes
}
```

- Pin by index, or `OptimizationProfileGuard::automatic()` for shape-based selection. There is deliberately no `-1` sentinel: a computed index that came out negative should be an error, not a silent request for auto.
- Execution-scoped and thread-local, following the existing `CudaStreamGuard`. The guard records a request and nothing more — it never inspects the `Module`, `Method`, or delegate handles, and never calls TensorRT. Each delegate reads the request inside its own `execute()`, where the engine, its lock, and the stream are already in hand. Nested guards restore the enclosing request.
- Orthogonal to `CudaStreamGuard`: that one says where the work runs, this one says which profile it runs under.

## Selection policy

`torch_tensorrt/executorch/OptimizationProfileSelection.h` (installed) holds `ProfileTable`, `ProfileRequest`, `ProfileSelection`, and `select_profile`. It depends on `<NvInfer.h>` plus the STL only — no ExecuTorch, no CUDA — so the policy is unit-testable with a plain compiler and no GPU.

Auto-selection is sticky first-fit: keep the active profile while it still fits every input, otherwise rescan from 0 and take the lowest match. This matches `_TRTEngine._auto_select_profile` and `TRTEngine::auto_select_profile` in the standard runtime.

Strictness, kept identical in all three runtimes: an index the engine does not have is an error (`kRequestedProfileUnavailable`), except on a single-profile engine, where profile 0 is the only thing it could ever run — that returns `kPinIgnoredSingleProfile` and logs, so a pin aimed at a sibling engine in the same method does not fail an innocent one. Negative indices always error.

## Runtime changes

- `execute()` order is: wait on in-flight work, pick the stream, collect every input's `Dims`, select, validate, `setOptimizationProfileAsync` (only if the profile actually changed), then `setInputShape` and bind. TensorRT requires the switch *before* `setInputShape`, which is why the shapes are collected in a pre-pass.
- `setOptimizationProfileAsync` enqueues weight and scratch copies on the stream. Every early return between the switch and the tail used to leave those unaccounted for, so the next `execute()` or `~EngineHandle` could reconfigure or free the context mid-copy. The existing `inflight_event`/`inflight_pending` pair is now armed right after the switch via a `mark_inflight` helper, which returns an `Error` so a failed `cudaEventRecord` (usually a sticky async fault from the enqueue) is not swallowed.
- Also fixes a pre-existing bug found on the way: the D2H path could return without draining while `inflight_pending` was false, letting the destructor free staging buffers underneath live copies.
- `TRTEngine::set_active_profile_with_stream` and `_TRTEngine._set_active_profile_with_stream` are aligned with the strictness rule above, so the two interchangeable runtimes no longer diverge.

## Header layout / ABI

`EngineHandle` moves out of the installed public header into a private `cpp/src/torch_tensorrt/executorch/EngineHandle.h`. The public struct layout must not change under a prebuilt backend archive, and this struct grew a profile table, so it should never have been public.

**Release-note item:** `EngineHandle` was declared in the installed `cpp/include/torch_tensorrt/executorch/TensorRTBackend.h` and shipped that way in v2.13.0:

```console
$ git show v2.13.0:cpp/include/torch_tensorrt/executorch/TensorRTBackend.h \
    | grep -nE "^struct EngineHandle"
48:struct EngineHandle {
```

Downstream code that names `EngineHandle` will no longer compile. `InputProfileBounds` is unaffected — it moved to the installed `OptimizationProfileSelection.h`.

## Docs

All four installed headers (`TensorRTBackend.h`, `OptimizationProfileSelection.h`, `TensorRTBlobHeader.h`, `TensorRTBindingNames.h`) now carry doxygen annotations, per @narendasan's request, since they constitute a C++ API. `docsrc` already runs doxygen over `cpp/include` recursively, so these render with no docsrc change.

## Examples, tests, packaging

- `examples/torchtrt_executorch_example/export_multi_profile.py` exports a two-profile Gemma-3 `.pte`. Defaults to a mini config that needs no download; `--weights google/gemma-3-1b-it` for the real 1B model.
- `examples/executorch_reference_runner/multi_profile_main.cpp` demonstrates pinning, auto-selection, and that an invalid pin is rejected. `multi_profile_benchmark.cpp` measures what switching is worth against running everything on the prefill profile.
- `tests/cpp/executorch/test_optimization_profile_selection.cpp`: 18 gtest cases over the policy, including multi-input tables, rescan ordering, rank mismatches, and malformed tables.
- `bazel test //tests/cpp/executorch:executorch_backend_tests` now runs in `executorch-static-linux.yml`. That job needs no GPU, and previously **no** bazel test ran anywhere in CI.
- Both new runner sources are added to the `source_files` filegroup so they ship in the release tarball, and `verify-executorch-reference-runner.sh` gained a `require_tar_entry` line per runner source plus builds of all three runner targets. The packaged `CMakeLists.txt` names every runner source, so one missing file aborts the whole configure for tarball users, not just its own target.
- `examples/dynamo/multi_optimization_profiles.py` is updated to use one engine and compare regimes on it, rather than compiling a separate single-profile baseline: the prefill profile already accepts `seq == 1`, so running every call on it reproduces single-profile behavior with no second compile and no second set of weights to keep honest.

## Not included

No export-time or partition-time profile plumbing. The feature is runtime-only: an index is per-engine, not a global regime, and partitioning and memory planning never read TensorRT profiles (planning uses the `torch.export` `Dim` union). A mis-aimed index cannot corrupt results — `validate_input_dims` and TensorRT's own `setInputShape` catch it; the worst case is kernels tuned for another regime.

## Type of change

- Breaking change: `EngineHandle` is no longer in the installed public header (source-level only, see above)

# Checklist:

- [x] New and existing unit tests pass locally with my changes (the 18 policy tests; the GPU paths — the profile switch itself, the two runners, the benchmark — were not run locally)

### DIFF
--- a/core/runtime/TRTEngine.cpp
+++ b/core/runtime/TRTEngine.cpp
@@ -695,15 +695,23 @@ void TRTEngine::set_active_profile(int64_t profile_index) {
 }
 
 void TRTEngine::set_active_profile_with_stream(int64_t profile_index, const c10::cuda::CUDAStream& stream) {
-  if (num_optimization_profiles <= 1) {
+  // An index this engine does not have is a request that cannot be honored, so
+  // say so rather than no-op quietly. Covers the single-profile engine too, where
+  // 0 is the only valid index. Reachable only by driving the engine directly; the
+  // Python wrapper validates the index against the profile count first.
+  if (profile_index < 0 || profile_index >= num_optimization_profiles) {
+    LOG_WARNING(
+        "Ignoring optimization profile index " << profile_index << ": this engine has " << num_optimization_profiles
+                                               << " optimization profile(s), so it stays on profile "
+                                               << active_profile_index << ".");
     return;
   }
   if (profile_index == active_profile_index) {
     return;
   }
 
-  // setOptimizationProfileAsync returns false for an out-of-range index; the
-  // index is validated upstream in TorchTensorRTModule.resolve_profile_index.
+  // The index is in range by the check above, so a false return here is TensorRT
+  // refusing the switch for some other reason.
   TORCHTRT_CHECK(
       exec_ctx()->setOptimizationProfileAsync(static_cast<int32_t>(profile_index), stream.stream()),
       "Failed to switch to optimization profile index " << profile_index);
--- a/core/runtime/TRTEngine.cpp
+++ b/core/runtime/TRTEngine.cpp
@@ -720,15 +720,34 @@ void TRTEngine::set_active_profile(int64_t profile_index) {
 }
 
 void TRTEngine::set_active_profile_with_stream(int64_t profile_index, const c10::cuda::CUDAStream& stream) {
-  if (num_optimization_profiles <= 1) {
+  // An index this engine does not have cannot be honored. A single-profile engine
+  // is the one case where that is not a mistake to fail on: it has profile 0 and
+  // nothing to switch to, so a pin aimed at some other engine is ignored with a
+  // warning rather than raising. A multi-profile engine handed an index it lacks
+  // could have done something different, and a negative index is a computed value
+  // gone wrong on any engine, so both stay hard errors -- silently running on
+  // mistuned kernels is worse than stopping. Same split as the ExecuTorch backend
+  // (kPinIgnoredSingleProfile vs kRequestedProfileUnavailable).
+  //
+  // Reachable only by driving the engine directly; the Python wrapper validates
+  // the index against the profile count first.
+  if (num_optimization_profiles <= 1 && profile_index > 0) {
+    LOG_WARNING(
+        "Ignoring optimization profile index " << profile_index << ": this engine has " << num_optimization_profiles
+                                               << " optimization profile(s), so it stays on profile "
+                                               << active_profile_index << ".");
     return;
   }
+  TORCHTRT_CHECK(
+      profile_index >= 0 && profile_index < num_optimization_profiles,
+      "Optimization profile index " << profile_index << " is out of range: this engine has "
+                                    << num_optimization_profiles << " optimization profile(s).");
   if (profile_index == active_profile_index) {
     return;
   }
 
-  // setOptimizationProfileAsync returns false for an out-of-range index; the
-  // index is validated upstream in TorchTensorRTModule.resolve_profile_index.
+  // The index is in range by the check above, so a false return here is TensorRT
+  // refusing the switch for some other reason.
   TORCHTRT_CHECK(
       exec_ctx()->setOptimizationProfileAsync(static_cast<int32_t>(profile_index), stream.stream()),
       "Failed to switch to optimization profile index " << profile_index);
--- a/cpp/BUILD
+++ b/cpp/BUILD
@@ -108,13 +108,28 @@ cc_library(
 )
 
 cc_library(
+    name = "tensorrt_executorch_optimization_profile_selection",
+    hdrs = [
+        "src/torch_tensorrt/executorch/OptimizationProfileSelection.h",
+    ],
+    strip_include_prefix = "src/torch_tensorrt/executorch",
+    deps = select({
+        ":linux_x86_64": ["@tensorrt//:nvinfer"],
+        ":sbsa": ["@tensorrt_sbsa//:nvinfer"],
+        "//conditions:default": [],
+    }),
+)
+
+cc_library(
     name = "tensorrt_executorch_backend",
     srcs = [
+        "src/torch_tensorrt/executorch/EngineHandle.h",
         "src/torch_tensorrt/executorch/TensorRTBackend.cpp",
     ],
     hdrs = [
         "include/torch_tensorrt/executorch/TensorRTBackend.h",
     ],
+    strip_include_prefix = "include",
     # Build the TensorRT backend as a static library. The final application
     # links this target together with the ExecuTorch runtime it was compiled
     # against, avoiding any runtime plugin/dlopen dependency.
@@ -123,19 +138,19 @@ cc_library(
         ":sbsa": [],
         "//conditions:default": ["@platforms//:incompatible"],
     }),
-    strip_include_prefix = "include",
     deps = [
         ":tensorrt_executorch_binding_names",
         ":tensorrt_executorch_blob_header",
+        ":tensorrt_executorch_optimization_profile_selection",
     ] + select({
         ":linux_x86_64": [
-            "@executorch//:executorch_headers",
             "@cuda//:cudart",
+            "@executorch//:executorch_headers",
             "@tensorrt//:nvinfer",
         ],
         ":sbsa": [
-            "@executorch//:executorch_headers",
             "@cuda//:cudart",
+            "@executorch//:executorch_headers",
             "@tensorrt_sbsa//:nvinfer",
         ],
         "//conditions:default": [],
@@ -147,6 +162,8 @@ filegroup(
     name = "executorch_backend_source_files",
     srcs = [
         "src/torch_tensorrt/executorch/CMakeLists.txt",
+        "src/torch_tensorrt/executorch/EngineHandle.h",
+        "src/torch_tensorrt/executorch/OptimizationProfileSelection.h",
         "src/torch_tensorrt/executorch/README.md",
         "src/torch_tensorrt/executorch/TensorRTBackend.cpp",
         "src/torch_tensorrt/executorch/TensorRTBlobHeader.cpp",
--- a/cpp/BUILD
+++ b/cpp/BUILD
@@ -110,9 +110,16 @@ cc_library(
 cc_library(
     name = "tensorrt_executorch_optimization_profile_selection",
     hdrs = [
-        "src/torch_tensorrt/executorch/OptimizationProfileSelection.h",
+        "include/torch_tensorrt/executorch/OptimizationProfileSelection.h",
     ],
-    strip_include_prefix = "src/torch_tensorrt/executorch",
+    strip_include_prefix = "include",
+    # The header includes <NvInfer.h>, so it cannot build where the deps below
+    # resolve to an empty list.
+    target_compatible_with = select({
+        ":linux_x86_64": [],
+        ":sbsa": [],
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
     deps = select({
         ":linux_x86_64": ["@tensorrt//:nvinfer"],
         ":sbsa": ["@tensorrt_sbsa//:nvinfer"],
@@ -123,6 +130,9 @@ cc_library(
 cc_library(
     name = "tensorrt_executorch_backend",
     srcs = [
+        # Private, deliberately not in hdrs: EngineHandle grows fields as the
+        # backend gains features and is never installed, so nothing outside this
+        # library may depend on its layout.
         "src/torch_tensorrt/executorch/EngineHandle.h",
         "src/torch_tensorrt/executorch/TensorRTBackend.cpp",
     ],
@@ -163,7 +173,6 @@ filegroup(
     srcs = [
         "src/torch_tensorrt/executorch/CMakeLists.txt",
         "src/torch_tensorrt/executorch/EngineHandle.h",
-        "src/torch_tensorrt/executorch/OptimizationProfileSelection.h",
         "src/torch_tensorrt/executorch/README.md",
         "src/torch_tensorrt/executorch/TensorRTBackend.cpp",
         "src/torch_tensorrt/executorch/TensorRTBlobHeader.cpp",
@@ -183,6 +192,7 @@ filegroup(
 filegroup(
     name = "executorch_api_headers",
     srcs = [
+        "include/torch_tensorrt/executorch/OptimizationProfileSelection.h",
         "include/torch_tensorrt/executorch/TensorRTBackend.h",
         "include/torch_tensorrt/executorch/TensorRTBindingNames.h",
         "include/torch_tensorrt/executorch/TensorRTBlobHeader.h",
--- a/cpp/BUILD
+++ b/cpp/BUILD
@@ -142,8 +142,32 @@ cc_library(
 )
 
 cc_library(
+    name = "tensorrt_executorch_optimization_profile_selection",
+    hdrs = [
+        "include/torch_tensorrt/executorch/OptimizationProfileSelection.h",
+    ],
+    strip_include_prefix = "include",
+    # The header includes <NvInfer.h>, so it cannot build where the deps below
+    # resolve to an empty list.
+    target_compatible_with = select({
+        ":linux_x86_64": [],
+        ":sbsa": [],
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
+    deps = select({
+        ":linux_x86_64": ["@tensorrt//:nvinfer"],
+        ":sbsa": ["@tensorrt_sbsa//:nvinfer"],
+        "//conditions:default": [],
+    }),
+)
+
+cc_library(
     name = "tensorrt_executorch_backend",
     srcs = [
+        # Private, deliberately not in hdrs: EngineHandle grows fields as the
+        # backend gains features and is never installed, so nothing outside this
+        # library may depend on its layout.
+        "src/torch_tensorrt/executorch/EngineHandle.h",
         "src/torch_tensorrt/executorch/TensorRTBackend.cpp",
     ],
     hdrs = [
@@ -161,18 +185,19 @@ cc_library(
     deps = [
         ":tensorrt_executorch_binding_names",
         ":tensorrt_executorch_blob_header",
+        ":tensorrt_executorch_optimization_profile_selection",
         ":tensorrt_executorch_weight_streaming_budget",
     ] + select({
         ":linux_x86_64": [
+            "@cuda//:cudart",
             "@executorch//:executorch_headers",
             "@executorch//:extension_cuda",
-            "@cuda//:cudart",
             "@tensorrt//:nvinfer",
         ],
         ":sbsa": [
+            "@cuda//:cudart",
             "@executorch//:executorch_headers",
             "@executorch//:extension_cuda",
-            "@cuda//:cudart",
             "@tensorrt_sbsa//:nvinfer",
         ],
         "//conditions:default": [],
@@ -184,6 +209,7 @@ filegroup(
     name = "executorch_backend_source_files",
     srcs = [
         "src/torch_tensorrt/executorch/CMakeLists.txt",
+        "src/torch_tensorrt/executorch/EngineHandle.h",
         "src/torch_tensorrt/executorch/README.md",
         "src/torch_tensorrt/executorch/TensorRTBackend.cpp",
         "src/torch_tensorrt/executorch/TensorRTBlobHeader.cpp",
@@ -204,6 +230,7 @@ filegroup(
 filegroup(
     name = "executorch_api_headers",
     srcs = [
+        "include/torch_tensorrt/executorch/OptimizationProfileSelection.h",
         "include/torch_tensorrt/executorch/TensorRTBackend.h",
         "include/torch_tensorrt/executorch/TensorRTBindingNames.h",
         "include/torch_tensorrt/executorch/TensorRTBlobHeader.h",
--- a/cpp/include/torch_tensorrt/executorch/OptimizationProfileSelection.h
+++ b/cpp/include/torch_tensorrt/executorch/OptimizationProfileSelection.h
@@ -31,10 +31,6 @@ struct ProfileTable {
   // profile count, which is at least 1; a single-profile engine keeps exactly
   // one row and never switches.
   std::vector<std::vector<InputProfileBounds>> bounds;
-  // True when every input dim is pinned to one extent in every profile. Such an
-  // engine accepts one shape only, so a request for a profile it does not have
-  // changes nothing it can do.
-  bool all_inputs_static = true;
   // The profile currently loaded into the execution context.
   int32_t active = 0;
 
@@ -50,11 +46,30 @@ enum class ProfileRequest {
   kAuto, // choose from the input shapes
 };
 
+// Its own enum rather than executorch's Error so that this header stays
+// independent of executorch and can be tested separately.
+//
+// Two axes: whether execution continues, and which message the caller prints.
+// The two failure values stay apart rather than being merged and re-derived from
+// the request kind, because the empty-table guard in select_profile() returns
+// kNoProfileMatchesInputs for every request kind -- so one merged value would put
+// the message back at the mercy of which branches each request can reach.
 enum class ProfileSelection {
-  // Created this enum to decouple the profile header from executorch so that we can test it seperately
   kOk,
-  // A pinned index this engine does not have, and it is not static enough for
-  // that to be harmless.
+  // Succeeded, but the pin could not be honored and profile 0 was used instead.
+  // Distinct from kOk so the caller can warn that the pin did nothing here.
+  kPinIgnoredSingleProfile,
+  // A pinned index this engine does not have and cannot substitute for.
+  //
+  // Fatal here, where TRTEngine::set_active_profile_with_stream only warns for the
+  // same mistake. That is deliberate, not an oversight: each runtime is strict at
+  // its outermost validating layer and lenient below it. The standard runtime
+  // rejects an out-of-range index in TorchTensorRTModule.set_optimization_profile
+  // before the engine is reached, so its engine-level check is a backstop.
+  // OptimizationProfileGuard cannot validate anything -- it never sees an engine,
+  // by design -- so execute() is the only place an ExecuTorch caller's bad index
+  // can be caught at all. Downgrading this to a warning would leave the whole
+  // ExecuTorch path with no index validation anywhere.
   kRequestedProfileUnavailable,
   // Auto-selection ran out of profiles.
   kNoProfileMatchesInputs,
@@ -90,6 +105,13 @@ inline ProfileSelection select_profile(
     int32_t index,
     const std::vector<nvinfer1::Dims>& input_dims,
     int32_t& selected) {
+  // init() rejects an engine reporting no profiles, so this is unreachable in the
+  // backend. Checked here so the policy is safe to call on its own rather than on
+  // the strength of a guard in another translation unit.
+  if (table.bounds.empty()) {
+    return ProfileSelection::kNoProfileMatchesInputs;
+  }
+
   if (request == ProfileRequest::kUnset) {
     selected = 0;
     return ProfileSelection::kOk;
@@ -119,13 +141,14 @@ inline ProfileSelection select_profile(
     return ProfileSelection::kOk;
   }
 
-  // An engine that accepts exactly one shape has nothing to switch, so a request
-  // aimed at its multi-profile siblings in the same method is satisfied by
-  // profile 0 rather than failing the whole execution. A dynamic engine that
-  // lacks the index is a real mismatch and is reported.
-  if (index > 0 && table.size() == 1 && table.all_inputs_static) {
+  // A single-profile engine has no choice to get wrong: profile 0 is the only
+  // thing it can run, whether or not its shapes are dynamic. So a pin aimed at a
+  // multi-profile sibling in the same method must not fail it. An engine with
+  // several profiles is different -- substituting one would be a guess -- so an
+  // index it lacks stays an error there.
+  if (index > 0 && table.size() == 1) {
     selected = 0;
-    return ProfileSelection::kOk;
+    return ProfileSelection::kPinIgnoredSingleProfile;
   }
 
   return ProfileSelection::kRequestedProfileUnavailable;
--- a/cpp/include/torch_tensorrt/executorch/OptimizationProfileSelection.h
+++ b/cpp/include/torch_tensorrt/executorch/OptimizationProfileSelection.h
@@ -96,7 +96,10 @@ enum class ProfileSelection {
  * @return true when the ranks match and every extent is in range.
  */
 inline bool dims_fit(const nvinfer1::Dims& dims, const InputProfileBounds& bounds) {
-  if (dims.nbDims != bounds.min.nbDims) {
+  // Both bounds, not just min: the loop below reads max.d up to dims.nbDims, and
+  // a max of a different rank leaves those extents holding whatever the Dims was
+  // built with rather than a bound this profile ever declared.
+  if (dims.nbDims != bounds.min.nbDims || dims.nbDims != bounds.max.nbDims) {
     return false;
   }
   for (int d = 0; d < dims.nbDims; ++d) {
--- a/cpp/include/torch_tensorrt/executorch/OptimizationProfileSelection.h
+++ b/cpp/include/torch_tensorrt/executorch/OptimizationProfileSelection.h
@@ -1,0 +1,210 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * @file OptimizationProfileSelection.h
+ * @brief Which TensorRT optimization profile an execution runs under.
+ *
+ * Kept free of ExecuTorch, CUDA, and the engine itself so the policy can be
+ * exercised without a GPU; reporting the outcome is left to the caller.
+ */
+#pragma once
+
+#include <NvInfer.h>
+
+#include <cstdint>
+#include <vector>
+
+namespace torch_tensorrt {
+namespace executorch_backend {
+
+/// @brief The [min, max] dim envelope one optimization profile allows for one input.
+struct InputProfileBounds {
+  nvinfer1::Dims min{}; ///< Smallest shape the profile accepts.
+  nvinfer1::Dims max{}; ///< Largest shape the profile accepts.
+};
+
+/// @brief Everything a profile decision depends on, read from the engine once at init().
+struct ProfileTable {
+  /**
+   * @brief Bounds indexed [profile][input].
+   *
+   * The outer size is the engine's optimization profile count, which is at
+   * least 1; a single-profile engine keeps exactly one row and never switches.
+   */
+  std::vector<std::vector<InputProfileBounds>> bounds;
+  /// @brief The profile currently loaded into the execution context.
+  int32_t active = 0;
+
+  /// @return The engine's optimization profile count.
+  int32_t size() const {
+    return static_cast<int32_t>(bounds.size());
+  }
+};
+
+/// @brief What the calling thread asked for, as resolved from OptimizationProfileGuard.
+enum class ProfileRequest {
+  kUnset, ///< No guard in scope.
+  kPinned, ///< An exact index.
+  kAuto, ///< Choose from the input shapes.
+};
+
+/**
+ * @brief The outcome of one profile decision.
+ *
+ * Its own enum rather than executorch's Error so that this header stays
+ * independent of executorch and can be tested separately.
+ *
+ * Two axes: whether execution continues, and which message the caller prints.
+ * The two failure values stay apart rather than being merged and re-derived from
+ * the request kind, because the empty-table guard in select_profile() returns
+ * kNoProfileMatchesInputs for every request kind -- so one merged value would put
+ * the message back at the mercy of which branches each request can reach.
+ */
+enum class ProfileSelection {
+  kOk, ///< The selected profile is usable.
+  /**
+   * @brief Succeeded, but the pin could not be honored and profile 0 was used instead.
+   *
+   * Distinct from kOk so the caller can warn that the pin did nothing here.
+   */
+  kPinIgnoredSingleProfile,
+  /**
+   * @brief A pinned index this engine does not have and cannot substitute for.
+   *
+   * Fatal, on the same split the standard runtime uses in
+   * TRTEngine::set_active_profile_with_stream: an engine that had profiles to
+   * choose between must not quietly run on one the caller did not ask for.
+   * OptimizationProfileGuard cannot validate anything -- it never sees an
+   * engine, by design -- so execute() is the only place an ExecuTorch caller's
+   * bad index can be caught at all.
+   */
+  kRequestedProfileUnavailable,
+  /// @brief Auto-selection ran out of profiles.
+  kNoProfileMatchesInputs,
+};
+
+/**
+ * @brief Whether one input shape falls inside one profile's envelope.
+ *
+ * @param dims The runtime shape of the input.
+ * @param bounds The profile's [min, max] envelope for that input.
+ * @return true when the ranks match and every extent is in range.
+ */
+inline bool dims_fit(const nvinfer1::Dims& dims, const InputProfileBounds& bounds) {
+  if (dims.nbDims != bounds.min.nbDims) {
+    return false;
+  }
+  for (int d = 0; d < dims.nbDims; ++d) {
+    if (dims.d[d] < bounds.min.d[d] || dims.d[d] > bounds.max.d[d]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/**
+ * @brief Whether one profile accepts every input shape of an execution.
+ *
+ * A profile index or bounds row that cannot describe these inputs answers "does
+ * not fit" rather than indexing past the end of the table. Neither is reachable
+ * from the backend, which builds one row of num_inputs bounds per profile and
+ * only ever stores an index this function approved -- but this is installed
+ * public API, and on the auto path the answer also happens to be the useful one:
+ * a ProfileTable carrying a stale ProfileTable::active rescans from 0 instead of
+ * crashing.
+ *
+ * @param table The engine's profile bounds.
+ * @param profile Index of the profile to test.
+ * @param input_dims Runtime shape of each input, in binding order.
+ * @return true when the profile accepts all of them.
+ */
+inline bool profile_fits(const ProfileTable& table, int32_t profile, const std::vector<nvinfer1::Dims>& input_dims) {
+  if (profile < 0 || profile >= table.size()) {
+    return false;
+  }
+  const auto& bounds = table.bounds[static_cast<size_t>(profile)];
+  if (bounds.size() < input_dims.size()) {
+    return false;
+  }
+  for (size_t i = 0; i < input_dims.size(); ++i) {
+    if (!dims_fit(input_dims[i], bounds[i])) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/**
+ * @brief Resolves one thread's profile request against one engine.
+ *
+ * @param table The engine's profile bounds and currently loaded profile.
+ * @param request What the calling thread asked for.
+ * @param index The pinned profile index; read only for ProfileRequest::kPinned.
+ * @param input_dims Runtime shape of each input, in binding order.
+ * @param[out] selected The profile to run, written only when the result allows
+ *   execution to continue (ProfileSelection::kOk or
+ *   ProfileSelection::kPinIgnoredSingleProfile).
+ * @return What the caller should do, and which message it should print.
+ */
+inline ProfileSelection select_profile(
+    const ProfileTable& table,
+    ProfileRequest request,
+    int32_t index,
+    const std::vector<nvinfer1::Dims>& input_dims,
+    int32_t& selected) {
+  // init() rejects an engine reporting no profiles, so this is unreachable in the
+  // backend. Checked here so the policy is safe to call on its own rather than on
+  // the strength of a guard in another translation unit.
+  if (table.bounds.empty()) {
+    return ProfileSelection::kNoProfileMatchesInputs;
+  }
+
+  if (request == ProfileRequest::kUnset) {
+    selected = 0;
+    return ProfileSelection::kOk;
+  }
+
+  if (request == ProfileRequest::kAuto) {
+    // Sticky first-fit: keep the loaded profile while it still fits, so shapes
+    // that alternate between two equally valid profiles don't thrash the
+    // context. Only rescan from 0 once it stops fitting. Overlapping profiles
+    // therefore resolve by history, not by lowest index; pin explicitly when
+    // that matters.
+    if (profile_fits(table, table.active, input_dims)) {
+      selected = table.active;
+      return ProfileSelection::kOk;
+    }
+    for (int32_t p = 0; p < table.size(); ++p) {
+      if (profile_fits(table, p, input_dims)) {
+        selected = p;
+        return ProfileSelection::kOk;
+      }
+    }
+    return ProfileSelection::kNoProfileMatchesInputs;
+  }
+
+  if (index >= 0 && index < table.size()) {
+    selected = index;
+    return ProfileSelection::kOk;
+  }
+
+  // A single-profile engine has no choice to get wrong: profile 0 is the only
+  // thing it can run, whether or not its shapes are dynamic. So a pin aimed at a
+  // multi-profile sibling in the same method must not fail it. An engine with
+  // several profiles is different -- substituting one would be a guess -- so an
+  // index it lacks stays an error there.
+  if (index > 0 && table.size() == 1) {
+    selected = 0;
+    return ProfileSelection::kPinIgnoredSingleProfile;
+  }
+
+  return ProfileSelection::kRequestedProfileUnavailable;
+}
+
+} // namespace executorch_backend
+} // namespace torch_tensorrt
--- a/cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
+++ b/cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
@@ -89,7 +89,12 @@ class CudaStreamGuard {
 inline constexpr int32_t kAutoSelectProfile = -1;
 
 // Selects, for the calling thread, which TensorRT optimization profile the
-// delegate runs; scope it around Module::forward() / Module::execute():
+// delegate runs; scope it around Module::forward() / Module::execute(). A
+// profile is identified by its index in the export-time profile list, so name
+// them to match whatever the exporter declared:
+//
+//   constexpr int32_t kDecodeProfile = 0;  // export order: decode first,
+//   constexpr int32_t kPrefillProfile = 1; // then prefill
 //
 //   executorch::extension::Module module("model.pte");
 //   {
--- a/cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
+++ b/cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
@@ -17,6 +17,8 @@
 
 #include <executorch/runtime/backend/interface.h>
 
+#include "torch_tensorrt/executorch/OptimizationProfileSelection.h"
+
 #include <cstdint>
 #include <memory>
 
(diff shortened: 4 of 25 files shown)


cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -393,40 +466,70 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
...
+      return Error::InvalidState;
     }
-    for (int d = 0; d < dims.nbDims; ++d) {
-      if (dims.d[d] < bounds.min.d[d] || dims.d[d] > bounds.max.d[d]) {
-        ET_LOG(Error, "TensorRTBackend::execute: input '%s' dim %d is outside profile bounds", name.c_str(), d);
-        return Error::InvalidArgument;
-      }
-    }
+    engine->profiles.active = profile;
+    ET_LOG(Info, "TensorRTBackend::execute: switched to optimization profile %d", profile);
```
shoumikhin: Once `setOptimizationProfileAsync(profile, stream)` succeeds, several later failures can return before `enqueueV3()` and before the existing completion event is recorded. The profile switch may therefore remain in flight while the next `execute()` or `destroy()` reconfigures or destroys the same `IExecutionContext`. Could you add cleanup for every post-switch error path, either by synchronizing the stream or recording completion that the next execution and teardown will wait for? Delaying the `profiles.active` assignment alone would not address the context-lifetime race.

tests/cpp/executorch/test_optimization_profile_selection.cpp
```diff
@@ -0,0 +1,171 @@
...
+  EXPECT_EQ(selected, 1);
+}
+
+TEST(ExecuTorchOptimizationProfileSelection, AutoRescansOnceTheActiveProfileStopsFitting) {
+  ProfileTable table = decode_and_prefill();
+  table.active = 0;
+  int32_t selected = -1;
+
+  EXPECT_EQ(select_profile(table, ProfileRequest::kAuto, 0, prefill_input(), selected), ProfileSelection::kOk);
+  EXPECT_EQ(selected, 1);
```
shoumikhin, line 117: Nonblocking test suggestion: could we add (1) a two-input table where profile 0 fits input 0 but not input 1, verifying auto-selection skips it, and (2) a three-profile table where the active profile no longer fits and two lower profiles do, verifying the rescan selects the lowest matching index? The implementation handles both today, but the current tests cover only one-input tables and a rescan with one fitting alternative.

examples/executorch_reference_runner/README.md
```diff
@@ -103,3 +116,97 @@ Loading the method initializes the TensorRT ExecuTorch backend for any
...
+using torch_tensorrt::executorch_backend::OptimizationProfileGuard;
+
+executorch::extension::Module module("model_gemma3_multi_profile.pte");
+{
+  OptimizationProfileGuard profile_guard(kPrefillProfile);
+  auto result = module.forward(prefill_inputs);
+}
+{
+  OptimizationProfileGuard profile_guard(kDecodeProfile);
+  auto result = module.forward(decode_inputs);
```
shoumikhin, line 252: Nit: `kPrefillProfile` and `kDecodeProfile` are example-local constants and are not defined or exported by the public API, so this copied snippet does not compile as written. Could you define them in the snippet, or use profile indices `1` and `0` with a note that the indices follow the export-time profile order?

shoumikhin (commented):
I reviewed the multi-profile runtime, policy tests, and reference documentation. I left one runtime correctness concern and two nonblocking test/documentation suggestions inline.

examples/executorch_reference_runner/CMakeLists.txt
```diff
@@ -61,3 +61,25 @@ target_link_libraries(
     executorch::extensions
     executorch::kernels
     torchtrt::executorch_backend)
+
+add_executable(example_executorch_multi_profile_runner multi_profile_main.cpp)
```
shoumikhin, line 85: These two new runners are not packaged into the release tarball, and this is already failing CI.

`examples/executorch_reference_runner/BUILD` has a `source_files` filegroup that decides what ships:

```python
filegroup(
    name = "source_files",
    srcs = ["CMakeLists.txt", "README.md", "main.cpp"],
)
```

That feeds `executorch_reference_runner_pkg_files` -> `executorch_source_package` -> `libtorchtrt_tar`. The PR does not touch it, so the tarball gets the new `CMakeLists.txt` but neither `multi_profile_main.cpp` nor `multi_profile_benchmark.cpp`.

The `executorch-static-build` job on this commit shows exactly that:

```
CMake Error at CMakeLists.txt:65 (add_executable):
  Cannot find source file:
    multi_profile_main.cpp
CMake Error at CMakeLists.txt:75 (add_executable):
  Cannot find source file:
    multi_profile_benchmark.cpp
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

The same job passes on the base commit, so this is from this change. Worth noting the blast radius is wider than the two new targets: cmake aborts the whole configure step and emits no build files, so `example_executorch_runner` cannot be built either.

Could you add both files to the filegroup?

```python
    srcs = [
        "CMakeLists.txt",
        "README.md",
        "main.cpp",
        "multi_profile_benchmark.cpp",
        "multi_profile_main.cpp",
    ],
```

The two new backend headers were correctly added to `//cpp:executorch_backend_source_files`, so this is the one spot that was missed. A `require_tar_entry` line per new file in `verify-executorch-reference-runner.sh` would also catch this at packaging time rather than at configure time.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -203,6 +276,20 @@ bool is_cuda_accessible_ptr(const void* ptr) {
   return attrs.type == cudaMemoryTypeDevice || attrs.type == cudaMemoryTypeManaged;
 }
 
+// Marks the work just enqueued on `stream` as still in flight, so the next execute()
+// and ~EngineHandle wait for it before they reconfigure or free exec_ctx. Recording
+// over an already-recorded event just moves the marker forward, so callers can mark
+// repeatedly as they enqueue more. If the event cannot be armed, drain instead: the
+// caller has no other way to know the work is outstanding.
+void mark_inflight(EngineHandle& engine, cudaStream_t stream) {
```
shoumikhin: This refactor drops an error that the previous version reported.

Before, on the skip-sync path, a failed `cudaEventRecord` logged, drained, and returned an error:

```cpp
  } else {
    cuda_err = cudaEventRecord(engine->inflight_event, stream);
    if (cuda_err != cudaSuccess) {
      ...
      (void)cudaStreamSynchronize(stream);
      return Error::InvalidProgram;   // <-- gone now
    }
```

Now `mark_inflight` returns `void` and also discards the return code of its own fallback `cudaStreamSynchronize`, so with `must_sync == false` `execute()` goes on to return `Error::Ok`.

Why it matters: the likely reason `cudaEventRecord` fails right after `enqueueV3` is a sticky asynchronous CUDA error from the enqueue itself. In that case we now report success for a call whose inference actually faulted, and the error resurfaces later attributed to some unrelated operator. Rare path, but it used to be reported and now is not.

Could you have it return an `Error` and propagate at both call sites?

```cpp
Error mark_inflight(EngineHandle& engine, cudaStream_t stream) {
  const cudaError_t rec = cudaEventRecord(engine.inflight_event, stream);
  engine.inflight_pending = (rec == cudaSuccess);
  if (rec == cudaSuccess) {
    return Error::Ok;
  }
  ET_LOG(Error, "TensorRTBackend::execute: cudaEventRecord failed: %s", cudaGetErrorString(rec));
  return cudaStreamSynchronize(stream) == cudaSuccess ? Error::Ok : Error::InvalidProgram;
}
```

Minor related note: on the `must_sync` path the event recorded at the tail is synchronized away immediately after, so that record is wasted work. Harmless, just noting it since the helper is now unconditional.

cpp/BUILD
```diff
@@ -107,14 +107,29 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "tensorrt_executorch_optimization_profile_selection",
+    hdrs = [
+        "src/torch_tensorrt/executorch/OptimizationProfileSelection.h",
+    ],
+    strip_include_prefix = "src/torch_tensorrt/executorch",
```
shoumikhin: This puts a bare `OptimizationProfileSelection.h` on the include path of every app that links the backend.

`strip_include_prefix` here is the header's own directory, so the header ends up with no directory prefix at all. Every other header target in this repo strips to a directory:

```python
strip_include_prefix = "include"   # -> torch_tensorrt/executorch/TensorRTBlobHeader.h
strip_include_prefix = "include"   # -> torch_tensorrt/executorch/TensorRTBindingNames.h
```

Since this target is in `deps` of `tensorrt_executorch_backend`, the virtual include dir propagates transitively, so a downstream app with its own file of that name can shadow ours (or vice versa). Bazel-only hygiene, no behavior change, so low priority, but the rest of the project deliberately avoids this.

If you do change it, two things to watch. The bare spelling is what makes one `#include` work in both build systems, because the header lives in `src/` and the CMake build only puts `cpp/include` on the path (`cpp/src/torch_tensorrt/executorch/CMakeLists.txt:37`), so switching to `strip_include_prefix = "src"` alone would fix Bazel and break CMake. And the two existing include sites would need updating too:

- `cpp/src/torch_tensorrt/executorch/EngineHandle.h`
- `tests/cpp/executorch/test_optimization_profile_selection.cpp`

Simplest version is probably to move the header to `cpp/include/torch_tensorrt/executorch/`, include it as `"torch_tensorrt/executorch/OptimizationProfileSelection.h"` in both places, and strip to `include` like the siblings. It is already effectively public since the test depends on it.

shoumikhin (commented):
Re-reviewed after the fixup commit. The three earlier comments are all addressed, thanks.

The core design holds up well. I checked the stream ordering against the contract in `NvInferRuntime.h` (the switch, the H2D copies, and the enqueue all use the one `stream`, so the required happens-before comes for free with no host sync), confirmed a fresh `IExecutionContext` really does start on profile 0, confirmed the default `kSTATIC` allocation means a switch never has to allocate, and confirmed every read and write of `profiles.active` is under the handle mutex. The selection policy matches `_TRTEngine._auto_select_profile` and `TRTEngine::auto_select_profile` exactly.

Two things I would like resolved before merge.

1. `executorch-static-build` is currently failing on this PR because the two new example `.cpp` files are not packaged. The job log shows `Cannot find source file: multi_profile_main.cpp`. Same job passes on the base commit, so it is this change. Because cmake aborts the whole configure, this also breaks the pre-existing runner for anyone unpacking the release tarball.

2. The `mark_inflight` refactor dropped an error return the previous code had, so a faulted inference can now report success on the skip-sync path.

The rest is smaller: a bare `OptimizationProfileSelection.h` on every consumer's include path, the static-vs-dynamic pin inconsistency, and the `-1` sentinel in the public API, which is cheap to change now and expensive after release.

One request on scope. This is about 1760 added lines mixing the runtime feature, the in-flight-event refactor, two example programs, an export script, and a rewrite of the Python dynamo example. That last one touches the Python runtime example and is a separate concern from the C++ delegate. Splitting would make each piece much easier to review.

Validation note: static review only, no GPU run. The TensorRT semantics above come from the bundled header docs, and I did not reproduce the benchmark numbers.

cpp/src/torch_tensorrt/executorch/OptimizationProfileSelection.h
```diff
@@ -0,0 +1,135 @@
...
+  if (index >= 0 && index < table.size()) {
+    selected = index;
+    return ProfileSelection::kOk;
+  }
+
+  // An engine that accepts exactly one shape has nothing to switch, so a request
+  // aimed at its multi-profile siblings in the same method is satisfied by
+  // profile 0 rather than failing the whole execution. A dynamic engine that
+  // lacks the index is a real mismatch and is reported.
+  if (index > 0 && table.size() == 1 && table.all_inputs_static) {
```
shoumikhin: This tolerance treats two engines that are equally unable to honor the pin differently.

Pinning index 1:

```
single-profile STATIC  engine -> silently runs profile 0, returns kOk
single-profile DYNAMIC engine -> kRequestedProfileUnavailable
```

Both have exactly one profile, so neither has an index 1. The comment says the point is not to fail an innocent single-profile sibling when the guard was aimed at a multi-profile one, but that applies just as much to the dynamic sibling, which still fails. It also does not help two multi-profile engines with different counts (say 3 and 2, pin index 2), since neither is `size() == 1`.

The two existing runtimes each pick one rule and stick to it:

| | out-of-range pin, single-profile engine |
|---|---|
| Python `_TorchTensorRTModule.set_optimization_profile` | raises `ValueError`, always |
| C++ `TRTEngine::set_active_profile_with_stream` | silently no-ops for all single-profile engines |

Could we match one of them? If you keep the tolerance, applying it to all single-profile engines regardless of static or dynamic, plus a warning log, would at least make an ineffective pin visible rather than silent.

Narrow case in practice (one `.pte` mixing a static engine with a multi-profile one, and a nonzero pin), so not blocking, but the asymmetry will be hard to explain later.

cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
```diff
@@ -119,5 +84,55 @@ class CudaStreamGuard {
...
+//
+// One execution sees one consistent request, but several TensorRT engines in a
+// method apply it independently as they run. TensorRT offers no way to undo a
+// switch, so if a later engine rejects the request (a pinned index it does not
+// have, or no profile matching its inputs) it returns an error with earlier
+// engines already switched.
+class OptimizationProfileGuard {
+ public:
+  // profile_index: an exact profile to pin, or kAutoSelectProfile.
+  explicit OptimizationProfileGuard(int32_t profile_index);
```
shoumikhin: Could you document the multi-delegate hazard here?

The guard sets one thread-local that every TensorRT delegate in the method reads, so if a `.pte` has two engines whose profile lists differ, index 1 can mean prefill in one and decode in the other. The comment below says the delegates see one consistent request, which is true of the integer but not of its meaning.

For contrast, the other two runtimes both target something specific:

```python
# Python: targets a module object, can be scoped to one submodule
with optimization_profile(trt_gm, 1): ...
```

and the C++ runtime keeps `active_profile_index` per engine instance.

I think the thread-local is the right call here given the ExecuTorch `BackendInterface`. Its official `set_option` channel is process-global, which would be worse under concurrency. So this is not a redesign request, just a docs one so a user with two engines is not surprised.

One idea worth a thought, not for this PR: `BackendExecutionContext::get_method_name()` is available inside `execute()`, so if prefill and decode were exported as two methods the profile could be chosen from the method name with no ambient state at all.

cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
```diff
@@ -119,5 +84,55 @@ class CudaStreamGuard {
   bool prev_set_;
 };
 
+// Pass instead of an index to have each delegate pick a profile from the runtime
+// input shapes rather than being told one.
+inline constexpr int32_t kAutoSelectProfile = -1;
```
shoumikhin: Using `-1` inside the same `int32_t` as the index makes a computed index risky.

```cpp
int32_t idx = find_profile(name);    // returns -1 when not found
OptimizationProfileGuard guard(idx); // silently becomes auto-select, not an error
```

Same shape of problem with `bool`: `OptimizationProfileGuard(true)` quietly becomes index 1. Python guards against exactly this by using a type that cannot be confused with an index, and rejecting `bool` outright:

```python
if isinstance(profile, bool) or not isinstance(profile, int):
    raise TypeError(...)
```

Since this is a public header it gets expensive to change after release. A named constructor would make the intent unmistakable and keep the index space clean:

```cpp
OptimizationProfileGuard guard = OptimizationProfileGuard::automatic();
```

Not blocking, just much cheaper to settle now.

cpp/src/torch_tensorrt/executorch/OptimizationProfileSelection.h
```diff
@@ -0,0 +1,135 @@
...
+  nvinfer1::Dims max{};
+};
+
+// Everything a profile decision depends on, read from the engine once at init().
+struct ProfileTable {
+  // Indexed [profile][input]. The outer size is the engine's optimization
+  // profile count, which is at least 1; a single-profile engine keeps exactly
+  // one row and never switches.
+  std::vector<std::vector<InputProfileBounds>> bounds;
+  // True when every input dim is pinned to one extent in every profile. Such an
```
shoumikhin: This comment is not true for a multi-profile engine whose profiles are each individually static.

Take profile 0 pinned at `seq == 1` and profile 1 pinned at `seq == 128`. The `min.d[d] != max.d[d]` check in `initialize_input_profiles` never trips, so `all_inputs_static` stays `true`, yet the engine accepts two shapes, not "one shape only".

Harmless today because the only reader also requires `table.size() == 1`, where the claim does hold. But as written it invites someone to reuse the flag as "this engine is single-shape", which would be wrong. Could you reword it to describe what it actually computes (no dim varies within any one profile) and note it is only meaningful together with `size() == 1`?

tests/cpp/executorch/test_optimization_profile_selection.cpp
```diff
@@ -0,0 +1,206 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Self-check for the optimization-profile selection policy. The policy is the
+ * only non-obvious part of multi-profile support and depends on nothing but the
+ * profile bounds table, so it runs here without a GPU, a TensorRT engine, or an
+ * ExecuTorch method.
```
shoumikhin, line 10: The policy coverage here is good. Flagging what is still uncovered so it is a conscious choice rather than an oversight.

`profiles.active` is only ever written in `execute()`:

```cpp
mark_inflight(*engine, stream);
engine->profiles.active = profile;
```

`select_profile` only reads it and these tests set it by hand, so the switch itself, the write-back, the validate-then-switch ordering, and `mark_inflight` have no coverage anywhere. The `-1` to auto mapping also lives in `current_profile_request()` in the `.cpp` and is unreachable from here.

Two cheap policy cases worth adding: zero inputs, and a `ProfileTable` with an empty `bounds` (`select_profile` would index `bounds[active]` on an empty vector; only the `num_profiles < 1` check in the other translation unit prevents that today).

Also worth one line explaining why `validate_input_dims` repeats the bounds check, since it reads like duplication: on the pinned and unset paths `select_profile` never checks fit, so `validate_input_dims` is the only check there, and it is redundant only on the auto path.

tests/cpp/executorch/BUILD
```diff
@@ -27,3 +28,14 @@ cc_test(
         "@googletest//:gtest_main",
     ],
 )
+
+# Asserts against the profile-selection policy only, so unlike the runtime tests
+# it needs neither a GPU nor a TensorRT engine.
+cc_test(
+    name = "test_optimization_profile_selection",
+    srcs = ["test_optimization_profile_selection.cpp"],
+    deps = [
```
shoumikhin, line 63: Heads up that nothing in CI runs this. There is no `bazel test` anywhere under `.github/`, and `//tests:tests -> :cpp_api_tests -> //tests/cpp:api_tests` never reaches `executorch_backend_tests`.

The two pre-existing executorch tests are in the same position, so this is consistent rather than a new gap. But adding to the suite implies CI coverage that does not exist. Either wire it up in `executorch-static-linux.yml`, where bazelisk is already set up:

```
bazel test //tests/cpp/executorch:executorch_backend_tests --config=linux
```

or note in the PR that these are developer-run only.

Small related thing: this test and the new `tensorrt_executorch_optimization_profile_selection` library both lack the `target_compatible_with` gate that `tensorrt_executorch_backend` has, so on an unsupported platform the `deps` select resolves to `[]` while the header still does `#include <NvInfer.h>`. Latent while nothing runs `bazel test`, but easy to add now.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -393,40 +480,75 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
...
-          dims.nbDims,
-          bounds.min.nbDims);
+          "TensorRTBackend::execute: OptimizationProfileGuard requested profile %d but this engine has %d profile(s)",
+          g_profile_request,
+          engine->profiles.size());
       return Error::InvalidArgument;
+    case ProfileSelection::kNoProfileMatchesInputs:
+      ET_LOG(
+          Error,
+          "TensorRTBackend::execute: none of the engine's %d optimization profiles accept the input shapes; "
```
shoumikhin: This message tells the user less than it could, and everything needed is already in scope.

Right now it prints only the profile count. Python raises with the shapes and every profile's ranges:

```
No optimization profile matches the input shapes [(1, 512)]. Cached dynamic
profile ranges: ... Fix the input shapes or pin a profile explicitly ...
```

`input_dims` and `table.bounds` are both available here, and `validate_input_dims` already prints per-dim detail for the pinned path. Including the offending dims and each profile's `[min, max]` would make this actionable without a rebuild.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -393,40 +480,75 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
...
-      if (dims.d[d] < bounds.min.d[d] || dims.d[d] > bounds.max.d[d]) {
-        ET_LOG(Error, "TensorRTBackend::execute: input '%s' dim %d is outside profile bounds", name.c_str(), d);
-        return Error::InvalidArgument;
-      }
-    }
+    // The switch enqueues copies of the new profile's weights/scratch, and TensorRT
+    // forbids reconfiguring or destroying a context while they run. The binding loop
+    // below can still fail and return, so mark them now instead of relying on the
+    // enqueue at the tail to do it.
+    mark_inflight(*engine, stream);
```
shoumikhin: Small robustness thing: `execute()` reads the thread-local three times (in `current_profile_request()`, as the `g_profile_request` argument to `select_profile`, and again in the failure log).

The log is correct today only because `kRequestedProfileUnavailable` is reachable only for `kPinned`, where the raw value happens to be the pinned index. If a future change ever routes auto or unset into that branch, it would print `-1` or `0` and mislead.

Snapshotting once at the top would remove the dependency on that invariant:

```cpp
const ProfileRequest request = current_profile_request();
const int32_t requested_index = g_profile_request;
```

examples/executorch_reference_runner/multi_profile_main.cpp
```diff
@@ -0,0 +1,297 @@
...
+  if (outputs.empty() || !outputs[0].isTensor()) {
+    return;
+  }
+  exec_aten::Tensor t = outputs[0].toTensor();
+  double best = -1e30;
+  int64_t best_idx = -1;
+  for (int64_t i = 0; i < t.numel(); ++i) {
+    const double v = t.scalar_type() == exec_aten::ScalarType::Half
+        ? static_cast<double>(t.const_data_ptr<exec_aten::Half>()[i])
+        : static_cast<double>(t.const_data_ptr<float>()[i]);
```
shoumikhin: Any dtype that is not `Half` reads through a `float*` here, including `BFloat16`.

`const_data_ptr<T>()` is an unchecked `static_cast` in the portable tensor type, so on a bf16 tensor this walks 4 bytes per element through a 2-bytes-per-element buffer and reads roughly twice past the end. No assert, just wrong numbers and an out-of-bounds read.

Latent today since `export_multi_profile.py` does `.to(torch.float16)`, so the paired model never hits it. Still worth guarding, especially as `IndexTensor` just above carries a comment about dtype mismatch being silent corruption. A `BFloat16` branch plus a log-and-skip for anything unexpected would close it.

examples/executorch_reference_runner/multi_profile_benchmark.cpp
```diff
@@ -0,0 +1,322 @@
...
+    int rounds,
+    int decode_steps,
+    WorkloadResult& out) {
+  const auto start = Clock::now();
+  for (int r = 0; r < rounds; ++r) {
+    const double p = timed_forward(module, prefill.args(), prefill_profile);
+    if (p < 0.0) {
+      return false;
+    }
+    if (r != 0) { // first call of a block inherits the previous block's profile
```
shoumikhin, line 224: Two small flag edge cases in the benchmark.

`--block_rounds=1` means the `r != 0` guard drops every prefill sample, so prefill stats always come out `n=0`. The default is 3 so this only bites someone who passes 1, but the empty result is silent rather than explained.

`--blocks=0`, which is also what `atoi` returns for garbage input, leaves `wall_ms == 0`, so the wall-time percentage at the end computes `0.0 / 0.0` and prints `nan`. `percentile()` would also underflow `size() - 1` to `SIZE_MAX` on an empty vector; it is currently shielded only by the empty check in `summarize`.

Rejecting non-positive parsed values up front would handle all three.

examples/executorch_reference_runner/README.md
```diff
@@ -17,6 +23,13 @@ You can also generate a sample `.pte` from the Torch-TensorRT source tree:
 
 ```bash
 python examples/torchtrt_executorch_example/export_static_shape.py --model_path=model.pte
+
+# Two-profile Gemma-3 engine for the multi-profile runner below. Defaults to a
+# mini Gemma-3 that needs no download and exports in about a minute, most of it
+# spent serializing the engine into the .pte. Add --weights google/gemma-3-1b-it
+# for the real 1B model -- but that .pte is 1.9 GB and serialization runs at
+# roughly 3.7 s/MB, so budget hours rather than minutes for it.
```
shoumikhin: These two files quote the same measurement at different rates.

Here: `1.9 GB` and `roughly 3.7 s/MB`.
`export_multi_profile.py:38`: `roughly 3.6 seconds per megabyte` and `~2 GB`.

Worth making them agree, or dropping the per-MB rate from one of the two. The nearby `~3.6 ms` switch cost also reads confusingly next to `3.6 s/MB`, since they are unrelated quantities that happen to share a number.

cpp/BUILD
```diff
@@ -107,14 +107,29 @@ cc_library(
...
+        ":linux_x86_64": ["@tensorrt//:nvinfer"],
+        ":sbsa": ["@tensorrt_sbsa//:nvinfer"],
+        "//conditions:default": [],
+    }),
+)
+
 cc_library(
     name = "tensorrt_executorch_backend",
     srcs = [
+        "src/torch_tensorrt/executorch/EngineHandle.h",
```
shoumikhin, line 170: `EngineHandle.h` in `srcs` rather than `hdrs` deviates from the rest of the repo, where `.cpp` goes in `srcs` and `.h` in `hdrs` without exception.

It works, and putting a deliberately private header in `srcs` is a legitimate Bazel idiom that matches the "not installed" note in the file. Just unexpected for a reader, so a one-line comment saying it is intentionally private would help.

cpp/src/torch_tensorrt/executorch/OptimizationProfileSelection.h
```diff
@@ -0,0 +1,135 @@
...
+
+// What the calling thread asked for, as resolved from OptimizationProfileGuard.
+enum class ProfileRequest {
+  kUnset, // no guard in scope
+  kPinned, // an exact index
+  kAuto, // choose from the input shapes
+};
+
+enum class ProfileSelection {
+  // Created this enum to decouple the profile header from executorch so that we can test it seperately
```
shoumikhin: Typo: "seperately" -> "separately".

narendasan:
@cehongwang as these headers now constitute a C++ api can you put in doxygen annotations so we can render docs?

examples/executorch_reference_runner/multi_profile_main.cpp
```diff
@@ -0,0 +1,320 @@
...
+class IndexTensor {
+ public:
+  IndexTensor(int32_t seq, exec_aten::ScalarType dtype, bool positions)
+      : sizes_{1, seq},
+        dim_order_{0, 1},
+        strides_{seq, 1},
+        data_(static_cast<size_t>(seq) * (dtype == exec_aten::ScalarType::Long ? 8 : 4)),
+        impl_(dtype, 2, sizes_.data(), data_.data(), dim_order_.data(), strides_.data()) {
+    for (int32_t i = 0; i < seq; ++i) {
+      const int64_t v = positions ? i : (1 + (static_cast<int64_t>(i) * 7919) % 9000);
```
shoumikhin: The default export path feeds token ids past the end of the vocabulary.

```cpp
const int64_t v = positions ? i : (1 + (static_cast<int64_t>(i) * 7919) % 9000);
```

`export_multi_profile.py:101` gives the mini Gemma-3 `vocab_size=2048`, but this
formula produces ids up to 8976:

```
seq=  1  max_id=    1  ids >= 2048:   0/1
seq=128  max_id= 8976  ids >= 2048: 101/128
seq=256  max_id= 8976  ids >= 2048: 199/256
```

So the first call in `main`, the pinned seq=128 prefill, already has 101 out-of-range
ids, and the seq=256 prefill has 199. Decode is unaffected since seq=1 only ever
produces id 1.

TensorRT stores zero for an out-of-bounds gather (`NvInfer.h`: "Zero will be stored
for OOB access"), so this does not crash. It silently substitutes zero embeddings for
most of the prompt, which means the `next_token` values this runner prints for the
default model are meaningless. That is worse than a crash in one way: nothing tells
the user their getting-started run was garbage.

This is the documented first-run flow. The README exports the mini model by default,
then runs this binary against it. The real `google/gemma-3-1b-it` vocab is ~262k, so
`--weights` is unaffected, which is probably why it went unnoticed.

The exporter already handles this correctly with `torch.randint(1, vocab, ...)`
(line 236). The runners just need ids that cannot exceed the smallest vocab they
might be pointed at. Same line in `multi_profile_benchmark.cpp:99`.

core/runtime/TRTEngine.cpp
```diff
@@ -695,15 +695,23 @@ void TRTEngine::set_active_profile(int64_t profile_index) {
 }
 
 void TRTEngine::set_active_profile_with_stream(int64_t profile_index, const c10::cuda::CUDAStream& stream) {
-  if (num_optimization_profiles <= 1) {
+  // An index this engine does not have is a request that cannot be honored, so
+  // say so rather than no-op quietly. Covers the single-profile engine too, where
+  // 0 is the only valid index. Reachable only by driving the engine directly; the
+  // Python wrapper validates the index against the profile count first.
+  if (profile_index < 0 || profile_index >= num_optimization_profiles) {
+    LOG_WARNING(
```
shoumikhin, line 735: On a multi-profile engine this turns a hard failure into a warning.

Before, an out-of-range index skipped the `<= 1` guard, reached
`setOptimizationProfileAsync`, got `false` back, and `TORCHTRT_CHECK` threw. Now it
warns and returns, so `set_active_profile(99)` on a 2-profile engine goes from raising
to silently continuing on whatever profile was already loaded, which means silently
mistuned kernels.

Making the single-profile case non-silent is a genuine improvement. Would you consider
keeping the throw for an out-of-range index on a multi-profile engine, and warning only
where the engine could not have done anything differently (one profile, any nonzero
index)?

Reachability is limited: `set_optimization_profile` validates first and raises
`ValueError`, so only a caller driving the engine directly can hit this. I also checked
the warning cannot spam a hot loop, since every per-call caller is gated on
`num_optimization_profiles > 1` and passes an index it already checked with
`profile_fits`.

One small thing while you are here: this fixed the `.cpp` comment that pointed at
`TorchTensorRTModule.resolve_profile_index`, but the identical reference survives at
`TRTEngine.h:300`. That name has never existed as code (`git log -S` finds it only in
those two comments); the real validator is
`TorchTensorRTModule.set_optimization_profile`.

cpp/include/torch_tensorrt/executorch/OptimizationProfileSelection.h
```diff
@@ -0,0 +1,158 @@
...
+    return ProfileSelection::kOk;
+  }
+
+  if (request == ProfileRequest::kAuto) {
+    // Sticky first-fit: keep the loaded profile while it still fits, so shapes
+    // that alternate between two equally valid profiles don't thrash the
+    // context. Only rescan from 0 once it stops fitting. Overlapping profiles
+    // therefore resolve by history, not by lowest index; pin explicitly when
+    // that matters.
+    if (profile_fits(table, table.active, input_dims)) {
```
shoumikhin, line 181: `select_profile` can index `bounds` out of range on a malformed table.

The guard above checks only the outer vector, so `bounds.empty()` being false says
nothing about `active` naming a valid row:

```cpp
  if (table.bounds.empty()) {
    return ProfileSelection::kNoProfileMatchesInputs;
  }
  ...
    if (profile_fits(table, table.active, input_dims)) {
```

`profile_fits` then does `table.bounds[static_cast<size_t>(profile)]` (line 91) with no
range check. A size-2 table with `active = 5` segfaults. The same applies if a bounds
row is shorter than `input_dims`, since `bounds[i]` is indexed over `input_dims.size()`
rather than the row length.

Neither is reachable through `execute()` today. `profiles.active` is only written at
`TensorRTBackend.cpp:596` with a value `select_profile` already validated, and
`initialize_input_profiles` builds exactly `num_inputs` bounds per row. So this is
latent, not a live bug, and I would not hold the PR for it.

Raising it because of the comment right above the guard:

> Checked here so the policy is safe to call on its own rather than on the strength of
> a guard in another translation unit.

That is the bar this header sets for itself, and it is installed public API, so "on its
own" includes callers you do not control. The empty-table case got defense in depth;
the two sibling cases that actually crash did not. Either range-check `active` and the
row length, or narrow that comment to state the precondition. Fine as a follow-up.

tests/cpp/executorch/test_optimization_profile_selection.cpp
```diff
@@ -0,0 +1,270 @@
...
+ *
+ * Deliberately out of scope, because all of it needs a live engine: applying the
+ * decision (setOptimizationProfileAsync), writing profiles.active back, the
+ * ordering that puts the switch before setInputShape, and mark_inflight. Those
+ * belong to the end-to-end ExecuTorch tests.
```
shoumikhin: This points at tests that do not exist yet.

```
 * Deliberately out of scope, because all of it needs a live engine: applying the
 * decision (setOptimizationProfileAsync), writing profiles.active back, the
 * ordering that puts the switch before setInputShape, and mark_inflight. Those
 * belong to the end-to-end ExecuTorch tests.
```

There are no end-to-end ExecuTorch tests covering any of it. The only automated
live-engine path is `verify-executorch-reference-runner.sh`, which exports via
`export_static_shape.py` (single-profile, static) and runs `example_executorch_runner`,
which installs no guard. So `if (profile != engine->profiles.active)` at
`TensorRTBackend.cpp:586` is never true there. `tests/py/dynamo/executorch/` is
composition-only and says so, and the multi-profile tests under
`tests/py/dynamo/runtime/` cover the standard runtime, not this backend.

The policy tests themselves are good and I confirmed all 16 pass. The gap is only that
everything the policy hands off to is unverified.

I am not asking for a GPU test in this PR. Could the comment just say the applied path
is not covered yet, rather than deferring to tests that were never written? If you do
want cheap coverage later, `multi_profile_main.cpp` already asserts profile 99 is
rejected and returns nonzero on failure, and the verify job has a GPU, so running it
against the mini `.pte` would cover most of this.

examples/executorch_reference_runner/multi_profile_benchmark.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+  printf("\nwhat switching bought (positive = switching is faster)\n");
+  compare("decode", po_decode, sw_decode);
+  compare("prefill", po_prefill, sw_prefill);
+  printf(
+      "  %-30s prefill-only=%9.1f ms   switching=%9.1f ms   %+.1f%%\n",
+      "wall time (see note)",
+      prefill_only.wall_ms,
+      switching.wall_ms,
+      100.0 * (switching.wall_ms - prefill_only.wall_ms) / prefill_only.wall_ms);
```
shoumikhin: The wall-time row uses the opposite sign convention from its own heading.

`compare()` computes `prefill_only - switching` (line 235), so positive means switching
won, matching the "positive = switching is faster" heading on line 324. The wall-time
percentage computes `switching - prefill_only`, so when switching is faster it prints a
negative number under that heading. Negating it would make the block consistent. The
note below is about contention bias, so it does not cover this.

shoumikhin (commented):
Thanks for the updates. The `-1` sentinel, `all_inputs_static`, the header move, the
packaging, the dtype guard in `print_prediction`, and the flag validation are all
addressed, and `kPinIgnoredSingleProfile` resolves the static-vs-dynamic asymmetry
cleanly.

What I checked by building rather than reading, in case it saves you time:

- Built and ran the policy test against TensorRT 10.15 and CUDA 12.8 headers: 16/16
  pass, no warnings under `-Wall -Wextra`.
- Traced nested guards: an inner `automatic()` correctly restores an outer pinned
  index, and the stale `g_profile_index` under `kAuto` is never read.
- Checked the new CMake `../../../include` in both layouts. It resolves to
  `cpp/include` in the source tree and `torch_tensorrt/include` in the tarball, so it
  is right, and it is genuinely needed since the old runner included no
  torch_tensorrt header.
- Confirmed the D2H drain at `TensorRTBackend.cpp:787-790` fixes a real pre-existing
  bug. The base code returned without draining while `inflight_pending` was false, so
  the destructor could `cudaFree` the staging buffers underneath live copies. Nice
  catch.

Of the inline comments, only the token-id one looks worth blocking on; it will hit the
first person who follows the README end to end. The `TRTEngine.cpp` throw-versus-warn
question is next, and the rest are follow-up material.

Two asks, both small:

1. Could you fill in the PR description? It is still the template.
2. Please note there that `EngineHandle` is no longer in the installed public header.
   It shipped in v2.13.0:

   ```
   $ git show v2.13.0:cpp/include/torch_tensorrt/executorch/TensorRTBackend.h \
       | grep -nE "^struct EngineHandle"
   48:struct EngineHandle {
   ```

   The move itself is right and the reasoning in `EngineHandle.h` is sound, and
   `InputProfileBounds` is still reachable since it moved to the installed
   `OptimizationProfileSelection.h`. It is only `EngineHandle` that downstream code
   can no longer name, so it is worth a line for whoever writes the release notes.

cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
cehongwang: If we export as two methods does it mean there would be 2 TRT engine? That doubles the memory usage and can be inefficient

shoumikhin:
This no longer merges into main. `git merge-tree` against `a6db321708` gives six conflicts, and two of them are semantic rather than textual.

`TensorRTBackend.cpp` is 655 lines at your merge base and 824 on main. #4454 removed `CudaStreamGuard` and reworked stream handling in `execute()`, and #4336 added the weight-streaming block to the same `init()` and `execute()`. The profile switch and the `mark_inflight` call sit in exactly that region, so it needs re-integrating rather than re-applying. The description also cites `CudaStreamGuard` as the precedent for `OptimizationProfileGuard`, and that precedent is gone, so the guard needs its own justification now.

The workflow edit should be dropped rather than ported. #4398 deleted `.github/workflows/executorch-static-linux.yml`, which is the modify/delete conflict. Main already runs the suite at `executorch-build-linux.yml:115`, and it sets `--test_env=LD_LIBRARY_PATH` because Bazel's `cc_import` ships the unversioned `libnvinfer.so` while the loader asks for `libnvinfer.so.10`. The two lines added here carry no such env, so porting them would give a duplicate run that fails to load the test binary.

Everything else from the earlier rounds is addressed at `ba4a95bd`, so those threads are fine to resolve. The one to keep open is the multi-delegate index question on `TensorRTBackend.h`, which is waiting on a reply from me.

shoumikhin:
Nice feature. The selection logic is clean and I built and ran the new test suite (18/18 pass). One thing to fix before merge, plus small stuff. No wrong-result or memory-safety bug in the feature itself.

- **MAJOR:** `execute()` now always synchronizes, which contradicts the header's async contract. Fix below.
- **Minor:** missing `[[nodiscard]]`, per-call logging on the hot path, over-broad installed API, one wrong doc comment, an untested branch.
- **Nits:** a benchmark label, a stale comment, a format specifier.

CI note: the red jobs (bitwise, complex-subgraph) also fail on the base commit, so they are not from this PR.

---

### MAJOR: `execute()` always synchronizes, breaking its documented contract
`cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp` (~1176)

The tail of `execute()` calls `cudaStreamSynchronize(stream)` at function scope, so every call blocks until the stream drains. But the header (`TensorRTBackend.h:72-81`) tells callers it may return while work is still in flight. Also, the `cudaEventRecord` in the else branch is now dead: nothing sets `inflight_pending = true`, so no one waits on that event.

Fix: remove the function-scope sync and mark the async path instead.

```cpp
} else {
  cuda_err = cudaEventRecord(engine->inflight_event, stream);
  if (cuda_err != cudaSuccess) { ...; return Error::InvalidProgram; }
  engine->inflight_pending = true;   // restore
}
return Error::Ok;                    // drop the unconditional cudaStreamSynchronize
```

This keeps the header, the in-function comment (~1118-1136), and the README accurate, and makes the recorded event do its job (next call and destructor wait on it before reusing the context). Please add a test that returns on this async path so it stays covered.

### MINOR: `automatic()` is easy to misuse silently
`cpp/include/torch_tensorrt/executorch/TensorRTBackend.h`

`OptimizationProfileGuard::automatic();` as a bare statement does nothing: the guard is destroyed on the same line, so the next call runs profile 0, not auto. It compiles with no warning. Mark `automatic()` (and the class) `[[nodiscard]]`.

```cpp
OptimizationProfileGuard::automatic();            // no-op
auto g = OptimizationProfileGuard::automatic();   // correct
```

### MINOR: per-call logging on the hot path
`cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp` (~815-821, ~852)

Two `ET_LOG(Info)` lines run inside `execute()` with no once-guard: "ignoring the pin" fires every call for a single-profile engine with a nonzero pin, and "switched to profile" fires on every switch. Info is on by default, so a decode loop floods logs. Log the pin case once per engine, and drop the switch message to Debug.

### MINOR: installed public API is larger than needed
`cpp/include/torch_tensorrt/executorch/OptimizationProfileSelection.h`, `cpp/BUILD`

Users only need `ProfileRequest`, but the whole policy header is installed (`ProfileTable`, `select_profile`, `dims_fit`, etc.). These are `inline`, so a header/library version mismatch can pick different definitions. That is the same risk you cite for making `EngineHandle` private. Keep this header under `src/` and have the test depend on it there.

### MINOR: example comment describes auto-selection wrong
`examples/dynamo/multi_optimization_profiles.py` (~126-128)

It says auto picks the first fitting profile by index. Selection is sticky: it keeps the active profile if it still fits, and only rescans from 0 otherwise. Your export script and C++ runner describe it correctly, so the tree contradicts itself. Please align this comment.

### MINOR: new single-profile behavior has no test
`core/runtime/TRTEngine.cpp`, `py/torch_tensorrt/dynamo/runtime/_TRTEngine.py`

Pinning a nonzero profile on a single-profile engine now warns and runs profile 0 instead of erroring. There is no test for it. Add one that pins index 1 on a single-profile engine and checks it runs profile 0 with a warning.

### Nits

- `multi_profile_benchmark.cpp`: labels print `seq=128` fixed, but it is a flag. Print the actual value.
- A comment near the guard mentions `CudaStreamGuard`, wrong type. Leftover.
- `TensorRTBackend.cpp`: one dim logged with `%ld` vs `%lld` elsewhere. Inconsistent.

### Not blockers (no action needed)

- `EngineHandle` leaving the installed header is breaking, but already covered by the release-note item in the PR description.
- Two pre-existing behaviors, present on base, not from this PR: an error path after an async H2D copy can return before the copy is done, and the profile context manager has no rollback if a later submodule is rejected. Follow-up, not this PR.

cehongwang:
Thanks for the comment. I resolved all comments except for the header one. This header is in include/ because previously you asked for it there, on this PR. The earlier review made the opposite request explicitly: "Simplest version is probably to move the header to cpp/include/torch_tensorrt/executorch/, include it as "torch_tensorrt/executorch/OptimizationProfileSelection.h" in both places, and strip to include like the siblings. It is already effectively public since the test depends on it." It previously lived in src/ with strip_include_prefix set to its own directory, which put a bare OptimizationProfileSelection.h on the include path of every app linking the backend — that is what the move fixed.

There is also a structural obstacle to moving it back: the installed TensorRTBackend.h uses ProfileRequest as a member type. What do you think?

cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
```diff
@@ -17,100 +22,181 @@
...
+ * nothing -- while a multi-profile engine that lacks the index fails the
+ * execution.
+ *
+ * @warning Name the guard. A discarded temporary is destroyed at the end of the
+ * full-expression that made it, restoring the enclosing request before
+ * forward() is ever called, so the execution runs profile 0 rather than the one
+ * asked for:
+ *
+ * @code
+ * OptimizationProfileGuard(kPrefillProfile);         // no-op, guard already dead
```
shoumikhin: This one does not warn, it does not compile. Without braces it declares a variable called `kPrefillProfile`, and there is no default constructor to declare it with. Braces fix it and you get the warning you describe. The rest of the paragraph is right, I checked that the class-level attribute alone stays silent here.

cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
```diff
@@ -17,100 +22,181 @@
...
+   * the guard never sees the engine; that is deliberate, so a computed index
+   * (say -1 from a failed lookup) surfaces as an error rather than quietly
+   * meaning something else.
+   *
+   * @param profile_index Position in the export-time profile list.
+   */
+  [[nodiscard]] explicit OptimizationProfileGuard(int32_t profile_index);
+
+  /// @brief Rejected so that OptimizationProfileGuard(true) cannot become index 1.
+  OptimizationProfileGuard(bool) = delete;
```
shoumikhin, line 196: This closes the `OptimizationProfileGuard(true)` hole, but now every integer type other than `int32_t` is ambiguous rather than narrowing: `int64_t`, `size_t`, `unsigned` and `long` all fail to compile. `int64_t` is what `TRTEngine::set_active_profile` takes for the same thing, so people will hit it. An added `int64_t` overload or a constrained template keeps the `bool` rejection without the collateral.

cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
```diff
@@ -17,100 +22,181 @@
...
-  cudaEvent_t inflight_event = nullptr;
-  bool inflight_pending = false;
-
-  ~EngineHandle();
-};
-
+/**
+ * @brief The delegate ExecuTorch calls to run a TensorRT engine.
+ *
+ * Registered under the backend id `TensorRT`; a `.pte` produced by
```
shoumikhin: The id here is `TensorRT`, but registration uses `TensorRTBackend` and the partitioner serializes the class name, so the longer string is what a user needs.

cpp/include/torch_tensorrt/executorch/OptimizationProfileSelection.h
```diff
@@ -0,0 +1,210 @@
...
+ * a ProfileTable carrying a stale ProfileTable::active rescans from 0 instead of
+ * crashing.
+ *
+ * @param table The engine's profile bounds.
+ * @param profile Index of the profile to test.
+ * @param input_dims Runtime shape of each input, in binding order.
+ * @return true when the profile accepts all of them.
+ */
+inline bool profile_fits(const ProfileTable& table, int32_t profile, const std::vector<nvinfer1::Dims>& input_dims) {
+  if (profile < 0 || profile >= table.size()) {
```
shoumikhin, line 130: Nothing exercises this guard or the rank check above it. I deleted each in turn and all eighteen tests still passed, including under a hardened standard library, so it is a real gap rather than a test that forgets to assert. A negative pinned index reaching `profile_fits`, and a table whose extra extent is nonzero, would cover both.

Separately, the comment above promises a fit only when the ranks match, but the code only checks the min bound's rank, never the max.

examples/dynamo/multi_optimization_profiles.py
```diff
@@ -140,21 +146,27 @@ def make_inputs(seq_len: int):
...
 # %%
-# Export Once, Compile Twice
-# ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# Export Once, Compile Once
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^
 #
-# ``export_llm`` traces the model over a dynamic ``seq`` range. We reuse the
-# exported program for both the single-profile baseline (tuned at the prefill
-# length, the conventional choice) and the multi-profile engine.
+# ``export_llm`` traces the model over a dynamic ``seq`` range, and one compile
```
shoumikhin, line 156: With the second compile gone this is not a single-profile baseline any more, it times profile 1 against profile 0 in the same engine. The tutorial's real two-engine numbers disagree, about 1.14x decode and 0.95x prefill, so the stand-in flatters the result. Either bring the compile back or relabel it as one profile against another within a single engine.

examples/executorch_reference_runner/README.md
```diff
@@ -207,3 +221,117 @@ Because the causal attention at position 1 covers positions 0..1, the two logits
...
+
+```
+Per-call latency (ms), batch=1
+call                        active profile        ms
+----------------------------------------------------
+decode (seq=1)                     prefill     6.415
+decode (seq=1)                      decode     4.981
+prefill (seq=128)                  prefill     8.438
+
+Giving decode its own profile: 1.29x faster per token (+1.434 ms)
```
shoumikhin, line 300: This number and the one below it disagree with the docsrc tutorial for the same regime. Different runtimes, so no need to match, but say which is which. That tutorial is also stale in three other places: auto-selection is sticky first-fit now, the example no longer compiles twice, and `register_sdpa` is gone. The generated API pages still show `EngineHandle` as public, which is just worth a release note.

examples/torchtrt_executorch_example/export_multi_profile.py
```diff
@@ -0,0 +1,291 @@
...
+    params = sum(p.numel() for p in model.parameters())
+    stamp(
+        f"model built: Gemma-3 ({args.weights or 'mini, random init'}), {params / 1e6:.1f}M params"
+    )
+    return model
+
+
+try:
+    model = build_model()
+except Exception as e:  # no GPU, or gated/unauthenticated --weights
```
shoumikhin: `transformers` is in none of the extras but `build_model` imports `Gemma3ForCausalLM` from it, so following the prerequisites exactly on a machine that does have a GPU prints "a CUDA GPU is required", exits 0, and writes no `.pte`. Adding it to the documented install fixes the common case, and exiting nonzero when `--weights` was passed would stop a gated repo, an out-of-memory and a missing import all looking the same.

PR #4442, opened 2026-07-29
Fuse pad into convolution

micwill755:
Description

Fold asymmetric constant_pad_nd → convolution patterns into TensorRT convolution pre_padding / post_padding via a new Dynamo post-lowering pass (fuse_pad_into_convolution) and tensorrt::conv_asym_pad converter.

Motivation: Wan-style causal 3D convolutions (used by Cosmos3 / Cosmos3-Edge VAE) zero their own padding and apply it themselves, often asymmetrically in time (e.g. (2 * pad_t, 0)). TensorRT’s ONNX parser already folds Pad → Conv into the convolution’s explicit pre/post padding. Torch-TRT previously only set symmetric padding_nd, so those pads stayed as materialized copy kernels (extra work on the VAE encode path).

Change:

Post-lowering pass rewrites eligible constant_pad_nd → convolution into tensorrt::conv_asym_pad
Converter emits IConvolutionLayer with explicit pre_padding / post_padding (via extended convNd)
Unit tests in tests/py/dynamo/lowering/test_fuse_pad_into_convolution.py
Validation on Cosmos3-Edge policy (Torch-TRT → cosmos3_policy_inference):

Geometry: 256×256, 5 frames, und_len 121, batch 1, 4 denoise steps
Pad fusion removed ~45 extra VAE pad kernels and improved VAE encode 7.46 → 5.50 ms (~1.96 ms)
Full policy steady-state median improved ~23.85 → 21.81 ms (~2.0 ms) after enabling this pass (with MoT builder settings already applied)
No new runtime dependencies beyond existing Torch-TensorRT / TensorRT.

Fixes # (n/a — performance / parity with ONNX Pad→Conv folding)

Type of change
New feature (non-breaking change which adds functionality)
Performance optimization for models with asymmetric Pad→Conv (e.g. Cosmos3 VAE)
Checklist:
 My code follows the style guidelines of this project (You can use the linters)
 I have performed a self-review of my own code
 I have commented my code, particularly in hard-to-understand areas and hacks
 I have made corresponding changes to the documentation
 I have added tests to verify my fix or my feature
 New and existing unit tests pass locally with my changes
 I have added the relevant labels to my PR in so that relevant reviewers are notified

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/custom_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/custom_ops_converters.py
@@ -20,8 +20,41 @@ from torch_tensorrt.dynamo.lowering.passes.fuse_distributed_ops import (
     tensorrt_fused_nccl_reduce_scatter_op,
     tensorrt_fused_nccl_scatter_op,
 )
+from torch_tensorrt.dynamo.lowering.passes.fuse_pad_into_convolution import (
+    tensorrt_conv_asym_pad_op,
+)
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
+
+
+@dynamo_tensorrt_converter(tensorrt_conv_asym_pad_op, supports_dynamic_shapes=True)
+def conv_asym_pad(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[trt.ITensor, Sequence[trt.ITensor]]:
+    """Folded pad+conv with independent pre/post padding per spatial dim."""
+    del kwargs
+    source, weight, bias, stride, pre_padding, post_padding, dilation, groups = args
+    return impl.conv.convNd(
+        ctx,
+        target,
+        SourceIR.ATEN,
+        name,
+        is_conv1d=False,
+        input=source,
+        weight=weight,
+        bias=bias,
+        stride=stride,
+        padding=[0] * len(pre_padding),
+        dilation=dilation,
+        groups=groups,
+        pre_padding=pre_padding,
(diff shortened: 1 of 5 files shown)


narendasan:
@micwill755 run precommit for the linting

py/torch_tensorrt/dynamo/conversion/impl/conv.py
zewenli98: Consider moving the validation to the function entry to fail faster for reducing build time.

py/torch_tensorrt/dynamo/conversion/custom_ops_converters.py
zewenli98: what if `pre_padding` is None? `len(pre_padding)` seems problematic. Can we directly pass in an empty list given that pre_padding and post_padding are provided?

zewenli98 (commented):
LGTM overall. just some minor issues

micwill755:
> LGTM overall. just some minor issues

Addressed all comments and committed.

py/torch_tensorrt/dynamo/conversion/impl/conv.py
zewenli98, line 181: If `pre_padding`, `post_padding`, and `padding` are all passed in simultaneously, the current logic will silently ignore the `padding`. Since `impl.conv.convNd` may be reused by other developers who are not familiar with it, I think it would be better to add a validation check similar to and around lines 42-46 like `ValueError(f"Convolution {name}: cannot specify both pre_padding/post_padding and padding)`. This would prevent ambiguous/contradictory padding specifications and make bugs obvious during development rather than silently using one and ignoring the other.

zewenli98 (approved):
LGTM

PR #4443, opened 2026-07-29
fix(dynamo): clone folded constants returned as graph outputs

curry-zzl:
# Description

Constant folding can replace input-independent tensor factories such as `torch.zeros` with registered `_frozen_param*` attributes. When one of these constants is returned directly across a graph break, eager code may mutate the returned tensor in-place. That mutation then changes the registered constant and leaks state into later top-level invocations.

This change extends `repair_input_as_output` to clone direct graph outputs backed by folded constants, preserving the registered constant while keeping constant folding enabled.

A regression test mutates the first returned tensor and verifies that both a subsequent output and the underlying `_frozen_param` remain unchanged.

## Type of change

# Checklist

- [x] I have commented the non-obvious graph-break behavior
- [x] I have added a test to verify the fix

## Validation

- `python -m py_compile` passes for the modified pass and test.
- The focused pytest collection in the development container is blocked because the source checkout does not contain the generated `torch_tensorrt._version` module. The equivalent regression was validated in the installed runtime environment.

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/passes/repair_input_as_output.py
+++ b/py/torch_tensorrt/dynamo/lowering/passes/repair_input_as_output.py
@@ -19,29 +19,43 @@ def repair_input_as_output(
     """
     modified_graph = False
 
-    # Extract graph placeholder Tensors
+    # Extract graph placeholder Tensors. Constant folding also turns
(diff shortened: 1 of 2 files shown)


curry-zzl:
Closing this PR pending completion of the required internal data disclosure review.

PR #4445, opened 2026-07-29
fix(executorch): support KV-cache aliased I/O in the TensorRT delegate

Conarnar:
## Description

Adds end-to-end **caller-owned KV-cache** support to the ExecuTorch TensorRT delegate. The KV buffers are owned by the caller *above* the delegate and threaded through as mutable-buffer delegate args (both the input and the engine's aliased output), so a TensorRT engine updates them in place and the cache persists across decode steps — matching the contract the non-ExecuTorch TensorRT runtime already exposes.

### Runtime + serialization (delegate)
- Serialize each engine's aliased (KV-cache / user) I/O into the delegate blob (`serialization.py`, `backend.py`, `TensorRTBlobHeader.{h,cpp}`).
- At runtime, bind each aliased TRT output binding to its aliased input's caller-provided pointer (in-place) and reflect the result into the delegate output EValue — a no-op when the memory planner already aliased the two (zero-copy) (`TensorRTBackend.{h,cpp}`).
- Validate the persisted alias map at init: `kind` must be one we understand (`kv_cache_update` / `user`); `kv_cache_update` entries are cross-checked against the engine's own `getAliasedInputTensor` (TensorRT is the source of truth), and `user` entries are shape-checked before two tensors are bound to the same storage. A caller-owned aliased input must be device-resident — otherwise its in-place update would be staged through host scratch and silently lost, so it's rejected loudly.
- A non-zero-copy aliased reflect is ordered to complete before `execute()` returns, since ExecuTorch's buffer-mutation `copy_` reads the delegate output EValue afterward; the zero-copy caller-owned KV fast path records no reflect and is untouched.

### Export / lowering (torch_tensorrt)
- Surface each engine's aliased outputs as graph-level `BUFFER_MUTATION`s so ExecuTorch keeps the KV buffers as caller-owned mutable buffers instead of freezing them: at transform time for the legacy exporter (`retrace=False`), and via a post-export pass (`_declare_aliased_kv_mutations_on_ep`) for `torch.export` (`retrace=True`), which otherwise drops the aliased outputs at the fx boundary. The `retrace=True` pass runs for both `executorch` and `exported_program`; `aot_inductor` is left undeclared (and warns), since whether an aliased in-place mutation survives functionalization under inductor is unverified.
- Keep delegate-mutated buffers above the delegate in `TensorRTPartitioner` (`tag_constant_data` would otherwise freeze them as constants).

## Dependency

Depends on #4446 (the legacy `retrace=False` submodule-inlining fix the composable ExecuTorch export path relies on), which has **landed** — this PR is now based on `main`.

## Follow-up tests

Now that **#4440** (per-method `TensorRTPartitioner` → separate delegate instances) and **#4454** (shared caller CUDA stream, for ordering the dependent GPU work between two delegates) have both landed, a cross-delegate **prefill/decode** acceptance test — decode consuming the KV cache that prefill wrote through a *separate* per-method delegate — is unblocked and will be added as a follow-up. That configuration is what exercises cross-delegate cache sharing, which single-delegate tests cannot cover.

## Testing

- **Unit:** `aliased_io` serialization round-trip; blob-header parse (present / empty / missing-key, and a model input literally named `aliased_io` not mistaken for the array key); exposure-flag dispatch across both retrace modes; the `BUFFER_MUTATION` declaration; partitioner keeps only mutation-target buffers above the delegate.
- **Runnable example:** `examples/executorch_reference_runner/kv_cache_decode_check` (exported by `export_kv_cache_decode.py`) drives multi-step decode against the delegate and asserts the KV cache persists in place across steps, starting from an explicitly-zeroed cache (ExecuTorch drops a mutated buffer's initial value) with a NaN guard so the check can't pass vacuously.

### DIFF
--- a/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
+++ b/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
@@ -242,7 +242,7 @@ Result<DelegateHandle*> TensorRTBackend::init(
 
   TensorRTBlobHeader header;
   if (!TensorRTBlobHeader::parse(processed->data(), processed->size(), header)) {
-    ET_LOG(Error, "TensorRTBackend::init: failed to parse TR01 TensorRT blob");
+    ET_LOG(Error, "TensorRTBackend::init: failed to parse TensorRT blob");
     return Error::InvalidProgram;
   }
 
--- a/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
+++ b/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
@@ -777,7 +777,7 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   // Caller-owned KV: (dst = delegate output EValue ptr, src = aliased input ptr,
   // nbytes). The engine updates the aliased input in place; reflect that into the
   // delegate output EValue after enqueue so ExecuTorch's write-back copy_ sees the
-  // updated cache. Skipped when dst == src (memory planner aliased them: zero-copy).
+  // updated cache.
   std::vector<std::tuple<void*, void*, size_t>> aliased_reflects;
   for (size_t o = 0; o < num_outputs; ++o) {
     const std::string& name = engine->output_binding_names[o];
@@ -816,6 +816,9 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
         (void)executorch::runtime::resize_tensor(et_alias_out, {a_sizes, static_cast<size_t>(a_dims.nbDims)});
       }
       void* dst = et_alias_out.nbytes() > 0 ? et_alias_out.mutable_data_ptr() : nullptr;
+      // dst != bind_ptr guards against issuing a self-copy. The memory planner does
+      // not currently place the delegate's output slot on the aliased input -- the
+      // two are live at the same time -- so this holds for every aliased output.
       if (dst != nullptr && dst != bind_ptr) {
         aliased_reflects.emplace_back(dst, bind_ptr, et_alias_out.nbytes());
       }
@@ -907,8 +910,7 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   }
 
   // Caller-owned KV: reflect each engine in-place update into its delegate output
-  // EValue (D2D on the same stream, after the engine work). No-op list under
-  // zero-copy (dst == src filtered out at bind time).
+  // EValue (D2D on the same stream, after the engine work).
   for (const auto& r : aliased_reflects) {
     cuda_err = cudaMemcpyAsync(std::get<0>(r), std::get<1>(r), std::get<2>(r), cudaMemcpyDeviceToDevice, stream);
     if (cuda_err != cudaSuccess) {
@@ -934,11 +936,10 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   // next execute() and the destructor wait before reusing/freeing exec_ctx. The D2H
   // copies live in the must_sync branch: an output staged to host always sets
   // output_staged_to_host, so outputs_needing_copy is empty on the skip path.
-  // A non-zero-copy aliased reflect enqueues the engine's in-place update into
-  // the delegate output EValue on `stream`; ExecuTorch's buffer-mutation copy_
-  // reads that EValue after execute() returns, so the reflect must complete
-  // first. Zero-copy aliases (dst == src) record no reflect, so the common
-  // caller-owned KV fast path is untouched.
+  // An aliased reflect enqueues the engine's in-place update into the delegate
+  // output EValue on `stream`; ExecuTorch's buffer-mutation copy_ reads that EValue
+  // after execute() returns, so the reflect must complete first. A model with
+  // aliased outputs therefore always syncs here.
   const bool aliased_reflect_pending = !aliased_reflects.empty();
   const bool must_sync =
       output_staged_to_host || input_staged_from_host || aliased_reflect_pending || !caller_stream_set;
--- a/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
+++ b/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
@@ -15,6 +15,7 @@
 #include <memory>
 #include <mutex>
 #include <string>
+#include <tuple>
 #include <utility>
 #include <vector>
 
@@ -436,6 +437,94 @@ Result<DelegateHandle*> TensorRTBackend::init(
     return err;
   }
 
+  // Map each aliased output binding to the index of the input it aliases so
+  // execute() can bind it to that input's device pointer (in-place).
+  // Non-aliased models have an empty header.aliased_io -> all -1, unchanged path.
+  handle->output_aliased_input_idx.assign(handle->num_outputs, -1);
+  handle->input_is_alias_target.assign(handle->num_inputs, false);
+  for (const auto& ab : header.aliased_io) {
+    int oi = -1;
+    for (size_t k = 0; k < handle->output_binding_names.size(); ++k) {
+      if (handle->output_binding_names[k] == ab.output) {
+        oi = static_cast<int>(k);
+        break;
+      }
+    }
+    int ii = -1;
+    for (size_t k = 0; k < handle->input_binding_names.size(); ++k) {
+      if (handle->input_binding_names[k] == ab.input) {
+        ii = static_cast<int>(k);
+        break;
+      }
+    }
+    if (oi < 0 || ii < 0) {
+      ET_LOG(
+          Error,
+          "TensorRTBackend::init: aliased_io names not found (output='%s', input='%s')",
+          ab.output.c_str(),
+          ab.input.c_str());
+      return Error::InvalidProgram;
+    }
+    // Validate the alias kind against the two we understand. The blob parser
+    // defaults a missing "kind" to "kv_cache_update"; any other value is a
+    // corrupt or newer-than-us wire format we can't safely bind, so fail loudly
+    // rather than fall through and treat it as a KV alias (which would bind two
+    // tensors to the same storage). Mirrors the Python _reconcile_aliased_io.
+    if (ab.kind != "kv_cache_update" && ab.kind != "user") {
+      ET_LOG(
+          Error,
+          "TensorRTBackend::init: aliased_io entry (output='%s') has unknown kind '%s'",
+          ab.output.c_str(),
+          ab.kind.c_str());
+      return Error::InvalidProgram;
+    }
+    if (ab.kind == "kv_cache_update") {
+      // TensorRT's IKVCacheUpdateLayer aliasing is the source of truth for
+      // kv_cache_update; the persisted map must agree with what the engine
+      // reports (via ICudaEngine::getAliasedInputTensor), else the blob is
+      // inconsistent with its own engine.
+      const char* trt_alias = handle->engine->getAliasedInputTensor(ab.output.c_str());
+      if (trt_alias == nullptr || ab.input != trt_alias) {
+        ET_LOG(
+            Error,
+            "TensorRTBackend::init: kv_cache_update alias for output '%s' disagrees with the "
+            "engine (persisted input='%s', engine input='%s')",
+            ab.output.c_str(),
+            ab.input.c_str(),
+            trt_alias == nullptr ? "<none>" : trt_alias);
+        return Error::InvalidProgram;
+      }
+    } else {
+      // AliasKind::USER aliases are declared by Torch-TensorRT and not tracked
+      // by TensorRT, so it can't validate them; confirm the aliased output and
+      // input share a shape before binding them to the same storage.
+      const nvinfer1::Dims od = handle->engine->getTensorShape(ab.output.c_str());
+      const nvinfer1::Dims id = handle->engine->getTensorShape(ab.input.c_str());
+      bool compatible = od.nbDims == id.nbDims;
+      for (int d = 0; compatible && d < od.nbDims; ++d) {
+        compatible = od.d[d] == id.d[d];
+      }
+      if (!compatible) {
+        ET_LOG(
+            Error,
+            "TensorRTBackend::init: user alias output '%s' shape is incompatible with input '%s'",
+            ab.output.c_str(),
+            ab.input.c_str());
+        return Error::InvalidProgram;
+      }
+    }
+    handle->output_aliased_input_idx[static_cast<size_t>(oi)] = ii;
+    handle->input_is_alias_target[static_cast<size_t>(ii)] = true;
+    ++handle->num_aliased_outputs;
+  }
+
+  if (handle->num_aliased_outputs > 0) {
+    ET_LOG(
+        Info,
+        "TensorRTBackend::init: %zu aliased output(s) bound in-place to caller-owned inputs",
+        handle->num_aliased_outputs);
+  }
+
   err = initialize_input_profiles(*handle);
   if (err != Error::Ok) {
     return err;
@@ -472,9 +561,17 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
 
   const size_t num_inputs = engine->num_inputs;
   const size_t num_outputs = engine->num_outputs;
-  if (args.size() < num_inputs + num_outputs) {
+  // Caller-owned KV: every input is a delegate arg, and each aliased output is
+  // threaded as a delegate output arg (the caller-owned mutable buffer's mutation
+  // slot), so all engine bindings map 1:1 to delegate args.
+  const size_t num_delegate_outputs = num_outputs;
+  const size_t num_delegate_inputs = num_inputs;
+  if (args.size() < num_delegate_inputs + num_delegate_outputs) {
     ET_LOG(
-        Error, "TensorRTBackend::execute: expected at least %zu args, got %zu", num_inputs + num_outputs, args.size());
+        Error,
+        "TensorRTBackend::execute: expected at least %zu args, got %zu",
+        num_delegate_inputs + num_delegate_outputs,
+        args.size());
     return Error::InvalidArgument;
   }
 
@@ -544,16 +641,22 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   // ------------------------------------------------------------------
   // 1. Bind input shapes and addresses
   // ------------------------------------------------------------------
+  // Device pointer each input binding was bound to; aliased outputs reuse the
+  // pointer of the input they alias so their update lands in-place.
+  std::vector<void*> input_bind_ptrs(num_inputs, nullptr);
+  size_t arg_idx = 0; // running index into delegate args
   for (size_t i = 0; i < num_inputs; ++i) {
-    EValue* arg = args[i];
-    TORCHTRT_ET_CHECK_NOT_NULL(arg, Error::InvalidArgument, "TensorRTBackend::execute: input %zu is not a tensor", i);
+    const std::string& name = engine->input_binding_names[i];
+
+    EValue* arg = args[arg_idx++];
+    TORCHTRT_ET_CHECK_NOT_NULL(
+        arg, Error::InvalidArgument, "TensorRTBackend::execute: input arg %zu is not a tensor", i);
     if (!arg->isTensor()) {
       ET_LOG(Error, "TensorRTBackend::execute: input %zu is not a tensor", i);
       return Error::InvalidArgument;
     }
 
     exec_aten::Tensor et_in = arg->toTensor();
-    const std::string& name = engine->input_binding_names[i];
     nvinfer1::Dims dims = to_trt_dims(et_in);
     if (dims.nbDims > nvinfer1::Dims::MAX_DIMS) {
       ET_LOG(Error, "TensorRTBackend::execute: input '%s' rank exceeds TensorRT limit", name.c_str());
@@ -580,6 +683,26 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
     if (!ctx->setInputShape(name.c_str(), dims)) {
       ET_LOG(Error, "TensorRTBackend::execute: setInputShape failed for '%s'", name.c_str());
       return Error::InvalidState;
+    }
+
+    // Caller-owned aliased input: an aliased output binds in-place to this
+    // input's device pointer, so its update must land in the caller's storage.
+    // If it isn't device-resident the branches below would stage it through
+    // delegate scratch, and the in-place update (bound to that scratch) would be
+    // silently lost on the next execute() when the staging copy re-reads the
+    // caller's unchanged buffer. Fail loudly instead.
+    if (engine->input_is_alias_target[i]) {
+      const bool device_resident =
+          et_in.nbytes() > 0 && (engine->unified_memory || is_cuda_accessible_ptr(et_in.const_data_ptr()));
+      if (!device_resident) {
+        ET_LOG(
+            Error,
+            "TensorRTBackend::execute: aliased input '%s' must be device-resident (non-empty and "
+            "CUDA-accessible or unified memory); its caller-owned in-place update cannot be staged "
+            "through host scratch",
+            name.c_str());
+        return Error::InvalidArgument;
+      }
     }
 
     void* bind_ptr = nullptr;
@@ -621,6 +744,7 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
       }
     }
 
+    input_bind_ptrs[i] = bind_ptr;
     if (!ctx->setTensorAddress(name.c_str(), bind_ptr)) {
       ET_LOG(Error, "TensorRTBackend::execute: setTensorAddress failed for input '%s'", name.c_str());
       return Error::InvalidState;
@@ -648,9 +772,58 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   // nbytes() and before the Python binding reads back the shape.
   // If the buffer is CPU, stage through a temporary CUDA allocation.
   // ------------------------------------------------------------------
+  // (arg index, device_src ptr) for outputs staged through a device buffer.
   std::vector<std::pair<size_t, void*>> outputs_needing_copy;
+  // Caller-owned KV: (dst = delegate output EValue ptr, src = aliased input ptr,
+  // nbytes). The engine updates the aliased input in place; reflect that into the
+  // delegate output EValue after enqueue so ExecuTorch's write-back copy_ sees the
+  // updated cache. Skipped when dst == src (memory planner aliased them: zero-copy).
+  std::vector<std::tuple<void*, void*, size_t>> aliased_reflects;
   for (size_t o = 0; o < num_outputs; ++o) {
-    EValue* arg = args[num_inputs + o];
+    const std::string& name = engine->output_binding_names[o];
+
+    // Aliased output (KV-cache / user): the engine updates the aliased input in
+    // place, so bind this output binding to the aliased input's device pointer.
+    const int alias_in = engine->output_aliased_input_idx[o];
+    if (alias_in >= 0) {
+      void* bind_ptr = input_bind_ptrs[static_cast<size_t>(alias_in)];
+      if (bind_ptr == nullptr) {
+        ET_LOG(Error, "TensorRTBackend::execute: aliased output '%s' has no bound input pointer", name.c_str());
+        return Error::InvalidState;
+      }
+      if (!ctx->setTensorAddress(name.c_str(), bind_ptr)) {
+        ET_LOG(Error, "TensorRTBackend::execute: setTensorAddress failed for aliased output '%s'", name.c_str());
+        return Error::InvalidState;
+      }
+      // The aliased output IS a delegate output arg (the caller-owned mutable
+      // buffer's mutation slot). Consume it and record a reflect so ExecuTorch's
+      // write-back copy_ sees the engine's in-place update.
+      const size_t arg_i = arg_idx++;
+      EValue* out_arg = args[arg_i];
+      TORCHTRT_ET_CHECK_NOT_NULL(
+          out_arg, Error::InvalidArgument, "TensorRTBackend::execute: aliased output %zu is not a tensor", o);
+      if (!out_arg->isTensor()) {
+        ET_LOG(Error, "TensorRTBackend::execute: aliased output %zu is not a tensor", o);
+        return Error::InvalidArgument;
+      }
+      exec_aten::Tensor et_alias_out = out_arg->toTensor();
+      nvinfer1::Dims a_dims = ctx->getTensorShape(name.c_str());
+      if (a_dims.nbDims >= 0 && a_dims.nbDims <= nvinfer1::Dims::MAX_DIMS) {
+        SizesType a_sizes[nvinfer1::Dims::MAX_DIMS];
+        for (int d = 0; d < a_dims.nbDims; ++d) {
+          a_sizes[d] = static_cast<SizesType>(a_dims.d[d]);
+        }
+        (void)executorch::runtime::resize_tensor(et_alias_out, {a_sizes, static_cast<size_t>(a_dims.nbDims)});
+      }
+      void* dst = et_alias_out.nbytes() > 0 ? et_alias_out.mutable_data_ptr() : nullptr;
+      if (dst != nullptr && dst != bind_ptr) {
+        aliased_reflects.emplace_back(dst, bind_ptr, et_alias_out.nbytes());
+      }
+      continue;
+    }
+
+    const size_t arg_i = arg_idx++; // continue the shared running arg index after the inputs
+    EValue* arg = args[arg_i];
     TORCHTRT_ET_CHECK_NOT_NULL(arg, Error::InvalidArgument, "TensorRTBackend::execute: output %zu is not a tensor", o);
     if (!arg->isTensor()) {
       ET_LOG(Error, "TensorRTBackend::execute: output %zu is not a tensor", o);
@@ -658,7 +831,6 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
     }
 
     exec_aten::Tensor et_out = arg->toTensor();
-    const std::string& name = engine->output_binding_names[o];
 
     // Update the ExecuTorch tensor shape to the actual TRT output shape.
     // getTensorShape() is valid after inferShapes() has been called.
@@ -712,7 +884,7 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
       }
       bind_ptr = engine->cached_output_ptrs[o];
       output_staged_to_host = true;
-      outputs_needing_copy.push_back({o, bind_ptr});
+      outputs_needing_copy.push_back({arg_i, bind_ptr});
     }
 
     if (!ctx->setTensorAddress(name.c_str(), bind_ptr)) {
@@ -734,6 +906,24 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
     return Error::InvalidState;
   }
 
+  // Caller-owned KV: reflect each engine in-place update into its delegate output
+  // EValue (D2D on the same stream, after the engine work). No-op list under
+  // zero-copy (dst == src filtered out at bind time).
+  for (const auto& r : aliased_reflects) {
+    cuda_err = cudaMemcpyAsync(std::get<0>(r), std::get<1>(r), std::get<2>(r), cudaMemcpyDeviceToDevice, stream);
+    if (cuda_err != cudaSuccess) {
+      ET_LOG(
+          Error, "TensorRTBackend::execute: aliased-output reflect D2D copy failed: %s", cudaGetErrorString(cuda_err));
+      // enqueueV3 already submitted engine work to `stream`, and inflight_pending
+      // is not armed until the end of the happy path -- drain now so a later
+      // execute() or the destructor never reconfigures/frees exec_ctx while this
+      // enqueue is still running.
+      (void)cudaStreamSynchronize(stream);
+      engine->inflight_pending = false;
+      return Error::InvalidProgram;
+    }
+  }
+
   // The engine work is now in flight on `stream`. Decide whether to wait for it:
   //   must_sync = an output is staged to host (the caller reads the D2H result on
   //   return), an input was staged from host (its async H2D read the caller's host
@@ -744,11 +934,18 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   // next execute() and the destructor wait before reusing/freeing exec_ctx. The D2H
   // copies live in the must_sync branch: an output staged to host always sets
   // output_staged_to_host, so outputs_needing_copy is empty on the skip path.
-  const bool must_sync = output_staged_to_host || input_staged_from_host || !caller_stream_set;
+  // A non-zero-copy aliased reflect enqueues the engine's in-place update into
+  // the delegate output EValue on `stream`; ExecuTorch's buffer-mutation copy_
+  // reads that EValue after execute() returns, so the reflect must complete
+  // first. Zero-copy aliases (dst == src) record no reflect, so the common
+  // caller-owned KV fast path is untouched.
+  const bool aliased_reflect_pending = !aliased_reflects.empty();
+  const bool must_sync =
+      output_staged_to_host || input_staged_from_host || aliased_reflect_pending || !caller_stream_set;
   if (must_sync) {
     Error copy_err = Error::Ok;
     for (auto& output : outputs_needing_copy) {
-      exec_aten::Tensor et_out = args[num_inputs + output.first]->toTensor();
+      exec_aten::Tensor et_out = args[output.first]->toTensor();
       cuda_err =
           cudaMemcpyAsync(et_out.mutable_data_ptr(), output.second, et_out.nbytes(), cudaMemcpyDeviceToHost, stream);
       if (cuda_err != cudaSuccess) {
--- a/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
+++ b/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
@@ -807,13 +807,22 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
         return Error::InvalidArgument;
       }
       exec_aten::Tensor et_alias_out = out_arg->toTensor();
+      // nbytes() below sizes both the reflect copy and ExecuTorch's write-back, so
+      // the tensor has to carry the shape TRT inferred before either of them reads it.
       nvinfer1::Dims a_dims = ctx->getTensorShape(name.c_str());
-      if (a_dims.nbDims >= 0 && a_dims.nbDims <= nvinfer1::Dims::MAX_DIMS) {
-        SizesType a_sizes[nvinfer1::Dims::MAX_DIMS];
-        for (int d = 0; d < a_dims.nbDims; ++d) {
-          a_sizes[d] = static_cast<SizesType>(a_dims.d[d]);
-        }
-        (void)executorch::runtime::resize_tensor(et_alias_out, {a_sizes, static_cast<size_t>(a_dims.nbDims)});
+      if (a_dims.nbDims < 0 || a_dims.nbDims > nvinfer1::Dims::MAX_DIMS) {
+        ET_LOG(Error, "TensorRTBackend::execute: invalid rank for aliased output '%s'", name.c_str());
+        return Error::InvalidState;
+      }
+      SizesType a_sizes[nvinfer1::Dims::MAX_DIMS];
+      for (int d = 0; d < a_dims.nbDims; ++d) {
+        a_sizes[d] = static_cast<SizesType>(a_dims.d[d]);
+      }
+      Error a_resize_err =
+          executorch::runtime::resize_tensor(et_alias_out, {a_sizes, static_cast<size_t>(a_dims.nbDims)});
+      if (a_resize_err != Error::Ok) {
+        ET_LOG(Error, "TensorRTBackend::execute: resize_tensor failed for aliased output '%s'", name.c_str());
+        return a_resize_err;
       }
       void* dst = et_alias_out.nbytes() > 0 ? et_alias_out.mutable_data_ptr() : nullptr;
       // dst != bind_ptr guards against issuing a self-copy. The memory planner does
--- a/examples/executorch_reference_runner/kv_cache_decode_check.cpp
+++ b/examples/executorch_reference_runner/kv_cache_decode_check.cpp
@@ -1,0 +1,209 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Caller-owned KV-cache persistence check for a Torch-TensorRT ExecuTorch .pte.
+ *
+ * Exercises the caller-owned KV contract: the engine's aliased KV output is
+ * bound in place to the caller's mutable buffer, which persists across
+ * execute() calls. Given a single-layer decode .pte (see
+ * examples/torchtrt_executorch_example/export_kv_cache_decode.py) with signature
+ * forward(tokens[1,1], input_pos[1]) -> logits, this runs two scenarios on
+ * FRESH method loads (each starts from a zeroed cache):
+ *
+ *   A) one decode at input_pos=1                  (no prior write at pos 0)
+ *   B) a decode at input_pos=0, then at input_pos=1
+ *
+ * At input_pos=1 the causal attention covers positions 0..1. If the cache is
+ * shared across execute() calls, scenario B's second step sees the key/value
+ * step 0 wrote at position 0, so its logits differ from scenario A (whose
+ * position-0 slot is still zero). Equal logits mean the update did not persist
+ * (cache reset per call, or the aliased output bound to scratch), so we fail.
+ *
+ * Usage:
+ *   kv_cache_decode_check --model_path=kv_cache_decode.pte [--tol=1e-3]
+ */
+
+#include <cinttypes>
+#include <cmath>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <memory>
+#include <vector>
+
+#include <cuda_runtime.h>
+
+#include <executorch/extension/data_loader/file_data_loader.h>
+#include <executorch/runtime/core/error.h>
+#include <executorch/runtime/core/evalue.h>
+#include <executorch/runtime/core/exec_aten/exec_aten.h>
+#include <executorch/runtime/executor/method.h>
+#include <executorch/runtime/executor/method_meta.h>
+#include <executorch/runtime/executor/program.h>
+#include <executorch/runtime/platform/log.h>
+#include <executorch/runtime/platform/runtime.h>
+
+using executorch::extension::FileDataLoader;
+using executorch::runtime::Error;
+using executorch::runtime::EValue;
+using executorch::runtime::HierarchicalAllocator;
+using executorch::runtime::MemoryAllocator;
+using executorch::runtime::MemoryManager;
+using executorch::runtime::Method;
+using executorch::runtime::MethodMeta;
+using executorch::runtime::Program;
+using executorch::runtime::Result;
+using executorch::runtime::Span;
+using executorch::runtime::TensorInfo;
+
+static const char* get_flag(int argc, char** argv, const char* flag, const char* def) {
+  const size_t n = strlen(flag);
+  for (int i = 1; i < argc; ++i) {
+    if (strncmp(argv[i], flag, n) == 0 && argv[i][n] == '=') {
+      return argv[i] + n + 1;
+    }
+  }
+  return def;
+}
+
+// Load a FRESH method (zeroed caller-owned buffers), run one decode step per
+// entry in `positions` (token id fixed to 1, input_pos = the position), and
+// return the final step's first output as host floats.
+static std::vector<float> run_decode(Program& program, const char* method_name, const std::vector<int64_t>& positions) {
+  Result<MethodMeta> method_meta = program.method_meta(method_name);
+  ET_CHECK_MSG(method_meta.ok(), "method_meta failed: 0x%" PRIx32, static_cast<uint32_t>(method_meta.error()));
+
+  auto method_pool = std::make_unique<uint8_t[]>(4 * 1024U * 1024U);
+  auto temp_pool = std::make_unique<uint8_t[]>(1 * 1024U * 1024U);
+  MemoryAllocator method_allocator{4 * 1024U * 1024U, method_pool.get()};
+  MemoryAllocator temp_allocator{1 * 1024U * 1024U, temp_pool.get()};
+
+  // Caller-owned KV buffers live in the memory-planned arenas. Arenas tagged
+  // CUDA (by PropagateDevicePass, for device tensors a delegate reads/writes)
+  // must be backed by real device memory -- otherwise the aliased KV input is
+  // not device-resident and the backend rejects it.
+  std::vector<std::unique_ptr<uint8_t[]>> host_arenas;
+  std::vector<void*> cuda_arenas;
+  std::vector<Span<uint8_t>> planned_spans;
+  const size_t num_planned = method_meta->num_memory_planned_buffers();
+  for (size_t i = 0; i < num_planned; ++i) {
+    const size_t sz = static_cast<size_t>(method_meta->memory_planned_buffer_size(i).get());
+    auto dev = method_meta->memory_planned_buffer_device(i);
+    if (dev.ok() && dev.get().type() == executorch::runtime::etensor::DeviceType::CUDA) {
+      void* p = nullptr;
+      ET_CHECK_MSG(cudaMalloc(&p, sz) == cudaSuccess, "cudaMalloc planned buffer %zu failed", i);
+      cuda_arenas.push_back(p);
+      planned_spans.push_back({reinterpret_cast<uint8_t*>(p), sz});
+    } else {
+      host_arenas.push_back(std::make_unique<uint8_t[]>(sz));
+      planned_spans.push_back({host_arenas.back().get(), sz});
+    }
+  }
+  HierarchicalAllocator planned_memory{{planned_spans.data(), planned_spans.size()}};
+  MemoryManager memory_manager{&method_allocator, &planned_memory, &temp_allocator};
+
+  Result<Method> method = program.load_method(method_name, &memory_manager, nullptr);
+  ET_CHECK_MSG(method.ok(), "load_method failed: 0x%" PRIx32, static_cast<uint32_t>(method.error()));
+
+  // One int64 tensor per declared input (numel==1 for a decode step): input 0 is
+  // the token id, the rest carry input_pos. Rank is read from method_meta so this
+  // works whether input_pos is rank-1 ([1]) or rank-2 ([1,1]).
+  const size_t num_inputs = method_meta->num_inputs();
+  std::vector<std::vector<int64_t>> data(num_inputs);
+  std::vector<std::vector<exec_aten::SizesType>> sizes(num_inputs);
+  std::vector<std::vector<exec_aten::DimOrderType>> dim_order(num_inputs);
+  std::vector<std::vector<exec_aten::StridesType>> strides(num_inputs);
+  std::vector<exec_aten::TensorImpl> impls;
+  impls.reserve(num_inputs);
+  for (size_t i = 0; i < num_inputs; ++i) {
+    Result<TensorInfo> ti = method_meta->input_tensor_meta(i);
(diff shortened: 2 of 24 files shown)


cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -203,6 +208,29 @@ bool is_cuda_accessible_ptr(const void* ptr) {
...
+    case nvinfer1::DataType::kINT64:
+      return 8;
+    case nvinfer1::DataType::kINT8:
+      return 1;
+    case nvinfer1::DataType::kUINT8:
+      return 1;
+    case nvinfer1::DataType::kBOOL:
+      return 1;
+    default:
+      return 0;
```
Gasoonjia: left a comment here indicating 0 is not a valid size and should catch error.

Conarnar:
### What changed vs the earlier (delegate-owned) version

The earlier revision made the **delegate own** the KV cache: because ExecuTorch
saw the KV buffers as non-mutated and froze them as constants, the delegate
`cudaMalloc`'d a persistent device buffer per aliased KV input, bound both the
input and its aliased output there, and accumulated the cache internally across
`execute()` calls. The aliased outputs were not threaded as delegate args.

This revision makes the cache **caller-owned**, above the delegate.

**Export / lowering (new):**
- Each engine's aliased outputs are exposed as graph-level `BUFFER_MUTATION`s
  (transform-time for `retrace=False`; a post-export pass for `retrace=True`), and
  `TensorRTPartitioner` strips the delegation tag from mutation-target buffers, so
  ExecuTorch keeps the KV buffers as **caller-owned mutable buffers** and threads
  them as delegate args instead of freezing them.

**Runtime (changed):**
- Removed all delegate-internal allocation — the per-input `cudaMalloc`/persistent
  buffer and the `input_is_self_owned` / `num_self_owned_inputs` bookkeeping.
- Every input and aliased output is now a delegate arg (1:1). Each aliased output
  binding is bound to its aliased input's **caller-provided** pointer (in-place),
  and the result is reflected into the delegate output EValue — a no-op when the
  memory planner already aliased the two (zero-copy).

narendasan:
@cehongwang Please review usage of the aliased i/o feature

py/torch_tensorrt/executorch/backend.py
cehongwang, line 368: line 223:
>    Only inputs need this. Outputs are also bound positionally by the runtime,
    but they are ``getitem(engine_node, idx)`` nodes whose index order equals the
    engine output-binding order. ExecuTorch lowering can reorder delegate outputs
    (``arrange_graph_outputs`` moves buffer-mutation outputs ahead of user
    outputs), but a TensorRT delegate partition is a functional inference engine
    with no mutation outputs, so that pass is a no-op here and the output order is
    preserved. If a TRT partition ever produced mutation outputs, outputs would
    need the same node-identity reordering as inputs.

We need to account for the output order, or otherwise there is a mismatch

cehongwang: The engine appends the buffer mutation at the end of output, while the delegate prepends it

cehongwang: ### Runtime walkthrough

Assume:

```text
TRT inputs:  [tokens, k_cache_in]
TRT outputs: [logits, k_cache_out]
```

ExecuTorch passes:

```text
args[0] = tokens
args[1] = k_cache_in
args[2] = k_cache_mutation
args[3] = logits
```

After consuming the inputs, `arg_idx == 2`.

**First output iteration**

```text
o    = 0
name = output_binding_names[0] = logits
arg  = args[2] = k_cache_mutation
```

The backend binds TensorRT's `logits` output to the cache-mutation output storage.

**Second output iteration**

```text
o       = 1
name    = output_binding_names[1] = k_cache_out
out_arg = args[3] = logits
```

TensorRT correctly binds `k_cache_out` to `k_cache_in` for the in-place update, but the backend treats the `logits` `EValue` as its mutation output slot. Its reflect copy therefore writes the cache result into the logits output.

The result is effectively:

| TensorRT output | Lands in |
| --- | --- |
| `logits` | cache mutation slot |
| cache update | logits slot |

If shapes or capacities differ, execution may fail during resize/binding/enqueue. If they are compatible, it can run successfully while returning incorrect logits and corrupting the observable cache state.

The required fix is to reorder serialized `output_binding_names` into actual delegate-output order, analogous to `_reorder_input_names_for_executorch`.

tests/py/dynamo/executorch/test_kv_cache_export.py
```diff
@@ -0,0 +1,157 @@
+"""Export-side coverage for caller-owned KV-cache buffer mutations.
```
cehongwang, line 1: These tests verify that mutations get prepended in the `ExportedProgram`, which is one half of the contract. The other half — that the serialized `output_binding_names` line up with the delegate arg order the C++ backend indexes positionally — isn't covered here or anywhere else, because `to_edge`, `arrange_graph_outputs`, and `preprocess` are all outside the mocked boundary.

Could you add a test at that seam? Roughly:

1. Build a KV model with both a user output and an aliased cache output.
2. Run it through `to_edge_transform_and_lower` with the TRT partitioner.
3. Read the delegate's output-arg order from the lowered module's output specs.
4. Assert `deserialize_engine(...).io_bindings` output names match that order.

This needs no GPU and would fail today.

cehongwang:
One testing gap worth closing before merge: there's currently no test anywhere that exercises a full `to_edge_transform_and_lower` → `to_executorch` → execute path. Grepping `tests/` for `to_edge_transform_and_lower`, `to_executorch`, and `load_for_executorch` returns nothing, and `executorch-static-linux.yml` only builds the backend plus the bazel C++ tests — it never lowers a KV model or runs a delegate.

That means the export side and the blob side are each tested in isolation and each is individually correct, while the bug lives in their composition. Could we get:

1. A CPU test asserting blob output-binding order matches delegate output-arg order for a model with both a mutation and a user output (details in the inline comment).
2. A two-step decode test on GPU or via the reference runner: same cache storage across two `execute()` calls, asserting step 1 observes step 0's update.

(2) is what RFC 0003 §7.4 asked for, and it would also cover the device-residency and reflect-path questions raised elsewhere in this review. Happy to gate it on GPU availability, but it should exist as a runnable target.

cehongwang:
In `tests/py/dynamo/executorch/test_backend.py`, `test_preprocess_preserves_output_binding_order`
This test's premise no longer holds after this PR, and as written it locks in the bug.

The comment says output order is "stable by construction (getitem index order == engine output-binding order)." That was true when a TRT partition was purely functional. This PR introduces `BUFFER_MUTATION` outputs, and ExecuTorch's `arrange_graph_outputs` reorders delegate outputs to `[mutations..., user_outputs...]` while `preprocess` still serializes them in engine order (`[user_outputs..., aliased_outputs...]`).

The fixture here has no mutation outputs, so it passes — but it asserts "preprocess must pass output names through unchanged," which is the behavior that needs to change.

Could you update this test to cover the mutation case? Something like a fixture with one aliased output and one user output, asserting the serialized `output_binding_names` match the delegate's output-arg order rather than engine order.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
cehongwang, line 718: Aliased inputs can still take the host-staging path, which breaks caller-owned semantics.

If an aliased input's data pointer isn't CUDA-accessible, this falls through to the staging branch and binds `engine->cached_input_ptrs[i]` — a delegate-owned scratch buffer. The aliased output then binds to `input_bind_ptrs[alias_in]`, i.e. that same scratch buffer, so the in-place KV update lands in delegate scratch rather than the caller's storage. On the next `execute()` the staging copy re-reads the caller's unchanged host buffer, so the update is silently lost. Recovery depends entirely on the reflect copy, which has its own problem (see the reflect comment).

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -590,7 +724,7 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   const bool must_sync = output_staged_to_host || input_staged_from_host || !g_user_stream_set;
```
cehongwang: Do we need to account for aliase I/O? Is there a race possible?

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -289,6 +290,64 @@ Result<DelegateHandle*> TensorRTBackend::init(
     return err;
   }
 
+  // Map each aliased output binding to the index of the input it aliases so
+  // execute() can bind it to that input's device pointer (in-place).
+  // Non-aliased models have an empty header.aliased_io -> all -1, unchanged path.
+  handle->output_aliased_input_idx.assign(handle->num_outputs, -1);
+  for (const auto& ab : header.aliased_io) {
```
cehongwang, line 445: This doesn't cross-check the persisted alias map against the engine, and it accepts unknown `kind` values.

The Python runtime's `_TRTEngine._reconcile_aliased_io` treats `getAliasedInputTensor` as the source of truth for `kv_cache_update` aliases and preserves `user` ones as metadata-trusted. Here, a `kind` that is neither `"kv_cache_update"` nor `"user"` — a typo in the wire format, or a future kind written by a newer exporter — skips the shape check and gets registered as if it were a KV alias, which then binds two tensors to the same storage.

Could you mirror the Python behavior:

1. Reject unknown kinds with `Error::InvalidProgram`.
2. For `kv_cache_update`, compare the persisted `ab.input` against `engine->getAliasedInputTensor(ab.output.c_str())` and error on disagreement.
3. Keep `user` as metadata-trusted after the shape check (TRT can't see those aliases).

Related: the parser leaves `ab.kind` empty when the `"kind"` key is absent, while the Python side defaults to `"kv_cache_update"`. Once unknown kinds are rejected, that mismatch turns an old blob into a hard failure — worth defaulting to `"kv_cache_update"` in the parser to match.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Can you clarify what `RFC 0003 §6.4` (and `§7.4`) is? I was not able to find any references for that.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Yes, but only on the non-zero-copy reflect path: with a caller stream active and no end sync, a pending reflect into the delegate output could still be in flight when ExecuTorch's buffer-mutation copy_ reads it. Will handle it.

py/torch_tensorrt/executorch/backend.py
Conarnar: I was not able to reproduce this. Seems like `arrange_graph_outputs` reorders the submodule outputs, output_specs, and the parent getitems, but it doesn't touch the call node's `meta["val"]`. `node.args` is permuted by fusion (so `_reorder_input_names_for_executorch` is still needed), but outputs go through `meta["val"]`, which `arrange_graph_outputs` leaves alone.

If `meta["val"]` is being rearranged somewhere that would be a problem, but the fix would be different from how `_reorder_input_names_for_executorch` handles it.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
cehongwang: Oh, here it is: 
**Require the caller's tensor itself to be device-resident.** Aliasing only
   works when the pointer bound to the input is the caller's real storage. Two
   existing branches break that: the H2D fallback copies a host tensor into a
   device staging buffer and binds *that*, and the zero-byte branch binds a
   cached scratch allocation. Either would make the engine write the update
   somewhere the caller never reads. For an alias-source input, gate on the
   caller tensor being device/unified memory (`is_cuda_accessible_ptr`) with
   non-zero size, and otherwise return `Error::InvalidArgument` ("aliased input
   '%s' must be on GPU or unified memory") — i.e. reject *before* both the
   staging and zero-byte paths, not only the H2D one.

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -554,12 +674,125 @@ def inline_trt_modules(
...
+    # storage), so dedup exposures across engines.
+    already_exposed = gm.meta.setdefault("_kv_exposed_mutation_targets", set())
+
+    output_node = next(node for node in gm.graph.nodes if node.op == "output")
+
+    val_list = list(trt_node.meta["val"])
+    new_mutation_outputs: List[torch.fx.Node] = []
+    for oi, out_name in enumerate(out_names):
+        if out_name not in aliased_io:
+            continue
```
cehongwang, line 821: This'd better be warnings or errors

py/torch_tensorrt/executorch/backend.py
cehongwang: Ok I spend a ton of time on this and it is quite surprising that the order was actually correct. But here is some findings:

`_keep_mutated_buffers_above_delegate(exported_program)`
https://github.com/Conarnar/TensorRT/blob/b9dbb306a4597eb4fb0c4ef0fdb546235d3924b1/py/torch_tensorrt/executorch/partitioner.py#L165
This function lifes mutated buffer to the executorch program level, therefore `arrange_graph_outputs` did not have any buffer_mutation, and therefore the order is perserved.

py/torch_tensorrt/_compile.py
```diff
@@ -1134,6 +1134,13 @@ def _extract_tensor(obj: Any) -> Any:
                         package_path=file_path,
                     )
                 elif output_format == "executorch":
+                    from torch_tensorrt.dynamo._exporter import (
+                        _declare_aliased_kv_mutations_on_ep,
+                    )
+
+                    # retrace=True: torch.export truncates the engines' aliased KV
+                    # outputs, so declare them as buffer mutations before lowering.
+                    exp_program = _declare_aliased_kv_mutations_on_ep(exp_program)
```
shoumikhin, line 1321: **Aliased models saved in the other two formats get no equivalent handling, and no warning.**

This normalization runs only in the `output_format == "executorch"` branch. The `exported_program` branch (1116) and `aot_inductor` branch (1126) receive the same `retrace=True` program and save it as-is.

```python
if output_format == "exported_program":     # no alias handling
elif output_format == "aot_inductor":       # no alias handling
elif output_format == "executorch":
    exp_program = _declare_aliased_kv_mutations_on_ep(exp_program)   # only here
```

I understand why it lives here: ExecuTorch has `tag_constant_data`, which freezes an undeclared buffer, and that pass is the reason this function exists. The other formats have no equivalent, so the harm is not the same.

But the outcome still differs per format in a way a user cannot see:

- `exported_program`: the engine still mutates the buffer in place at runtime, so the cache probably does update, but the saved signature does not declare the mutation. Anything downstream that trusts the signature gets a wrong answer.
- `aot_inductor`: this goes through inductor, and whether an in-place mutation on a custom op survives functionalization is not obvious. I could not find test coverage for that combination.

If aliased I/O is only intended to be supported for `executorch`, could the other two reject or warn when the module has aliased engines? A user who saves as `aot_inductor` today gets no signal either way, which is the part that seems worth closing regardless of which behavior you pick.

py/torch_tensorrt/dynamo/_exporter.py
Conarnar: Guessing you mean the two continues below.

py/torch_tensorrt/_compile.py
cehongwang: The alias I/O should be supported in both executorch and standard Torch-TensorRT runtime and AOT inductor

cehongwang: You might want to add `_declare_aliased_kv_mutations_on_ep` to exp program. AoT inductor is more complicated and we will fix that later

Conarnar: Will rebase and enable `exported_program` on #4459 as well.

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -491,8 +515,115 @@ def create_trt_exp_program(
...
+            _estr(engine_info, INPUT_BINDING_NAMES_IDX)
+        )
+        out_names = deserialize_binding_names(
+            _estr(engine_info, OUTPUT_BINDING_NAMES_IDX)
+        )
+        input_nodes = list(node.args[0])
+        val_list = list(node.meta["val"])
+        for out_name in out_names:
+            if out_name not in aliased_io:
+                continue
```
cehongwang, line 611: Brief warning here for better debugging

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -491,8 +515,115 @@ def create_trt_exp_program(
...
+            in_name = aliased_io[out_name][0]
+            if in_name not in in_names:
+                continue
```
cehongwang, line 624: here

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -491,8 +515,115 @@ def create_trt_exp_program(
...
+            ii = in_names.index(in_name)
+            if ii >= len(input_nodes):
+                continue
```
cehongwang, line 634: here

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -491,8 +515,115 @@ def create_trt_exp_program(
...
+            buf_node = input_nodes[ii]
+            buf_fqn = inputs_to_buffers.get(getattr(buf_node, "name", None))
+            if buf_fqn is None or buf_fqn in already_exposed:
+                continue
```
cehongwang, line 657: here

py/torch_tensorrt/executorch/backend.py
cehongwang: The previous comment no longer holds, and debugging can be painful if anything comes wrong in the future. Change the comment to something like

```py
def _reorder_input_names_for_executorch(
    edge_program: ExportedProgram, engine_node: Any, input_names: List[str]
) -> List[str]:
    """Reorder TRT binding names into executorch_call_delegate argument order.

    The runtime binds positionally (``execute()`` arg ``i`` -> input_binding_names
    ``[i]``), but ExecuTorch fusion may permute the delegate placeholders relative
    to the TRT-submodule order that produced ``input_binding_names``. The names
    can't be matched (TRT names are semantic, lowered placeholders are generic
    ``arg_N``), so recover the permutation by node identity: the engine node's
    first arg lists its input nodes in binding order, so sort the names by each
    node's slot among the graph placeholders (its runtime delegate-arg position).

    Only inputs need this. Outputs are also bound positionally, but they are
    ``getitem(engine_node, idx)`` nodes whose index order equals the engine
    output-binding order, and that order survives lowering -- though not because
    the partition is mutation-free. With aliased-I/O (KV-cache) support a TensorRT
    partition *does* produce mutation outputs, and ``arrange_graph_outputs`` does
    move buffer-mutation outputs ahead of user outputs. It stays a no-op here
    because ``_keep_mutated_buffers_above_delegate`` (``partitioner.py``) strips
    the ``delegation_tag`` from mutated buffer placeholders, so they stay out of
    the delegate's state dict and constants; ExecuTorch's ``_get_new_signature``
    then records the mutation as a plain ``USER_OUTPUT`` rather than a
    ``BUFFER_MUTATION`` (it uses the latter only when the delegate itself consumes
    the buffer). The lowered submodule therefore has no mutation specs, so
    ``arrange_graph_outputs`` computes the identity permutation and the getitem
    indices still line up with the engine's output bindings.

    That guarantee is conditional, not structural. If a mutated buffer is ever
    tagged into a delegate, its spec becomes ``BUFFER_MUTATION``, the delegate's
    outputs are permuted, and they would need the same node-identity reordering
    as the inputs below.
    """
```

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -491,8 +515,115 @@ def create_trt_exp_program(
...
+    gm.recompile()
+
+    new_output_specs = [
+        OutputSpec(OutputKind.BUFFER_MUTATION, TensorArgument(name=g.name), fqn)
+        for g, fqn in mutation_outputs
+    ] + list(sig.output_specs)
+    new_signature = ExportGraphSignature(
+        input_specs=list(sig.input_specs), output_specs=new_output_specs
+    )
+    return ExportedProgram(
```
cehongwang, line 685: The reconstruction drops example_inputs and verifiers, so the pass isn't a drop-in replacement for its input.

ExportedProgram.__init__ also takes example_inputs and verifiers; neither is carried over, so the returned program has them unset. No impact on the current to_edge path, which is presumably why it hasn't surfaced, but it bites in two places:

If the pass moves to the exported_program branch, the saved artifact silently loses its example inputs.
AOTI rejects such a program outright: RuntimeError: exported_program.example_inputs is required to be set in order for AOTInductor compilation. I hit this while testing whether the pass fixes AOTI and had to restore _example_inputs by hand before I could even reach the real failure.
Passing both through costs two keyword arguments.

cehongwang (changes requested):

examples/executorch_reference_runner/kv_cache_decode_check.cpp
```diff
@@ -0,0 +1,209 @@
...
+  // CUDA (by PropagateDevicePass, for device tensors a delegate reads/writes)
+  // must be backed by real device memory -- otherwise the aliased KV input is
+  // not device-resident and the backend rejects it.
+  std::vector<std::unique_ptr<uint8_t[]>> host_arenas;
+  std::vector<void*> cuda_arenas;
+  std::vector<Span<uint8_t>> planned_spans;
+  const size_t num_planned = method_meta->num_memory_planned_buffers();
+  for (size_t i = 0; i < num_planned; ++i) {
+    const size_t sz = static_cast<size_t>(method_meta->memory_planned_buffer_size(i).get());
+    auto dev = method_meta->memory_planned_buffer_device(i);
```
shoumikhin, line 95: `memory_planned_buffer_device` does not exist in the ExecuTorch commit this repo pins (`MODULE.bazel:50`, `6118688a09`, dated 2026-05-18); it landed on `release/1.3` after that pin, so `//examples/executorch_reference_runner:kv_cache_decode_check` does not compile. Bump the pin to a commit that has the API.

py/torch_tensorrt/_compile.py
```diff
@@ -1134,6 +1157,9 @@ def _extract_tensor(obj: Any) -> Any:
                         package_path=file_path,
                     )
                 elif output_format == "executorch":
+                    # retrace=True: torch.export truncates the engines' aliased KV
+                    # outputs, so declare them as buffer mutations before lowering.
+                    exp_program = _declare_aliased_kv_mutations_on_ep(exp_program)
```
shoumikhin, line 1321: Whether the mutation gets declared is decided twice, once by `use_legacy_exporter` inside `_exporter.export` and once by the outer `retrace` branch here, so two of the four combinations are wrong: `retrace=True, use_legacy_exporter=True` declares it twice and raises `SpecViolationError: User output getitem_3 is not in the correct order`, and `retrace=False, use_legacy_exporter=False` never declares it and prints no warning. Decide exposure once from the exporter that actually ran, and make `_declare_aliased_kv_mutations_on_ep` skip a buffer that already has a `BUFFER_MUTATION` spec.

py/torch_tensorrt/dynamo/_exporter.py
```diff
@@ -491,8 +515,115 @@ def create_trt_exp_program(
...
+                continue
+            in_name = aliased_io[out_name][0]
+            if in_name not in in_names:
+                continue
+            ii = in_names.index(in_name)
+            if ii >= len(input_nodes):
+                continue
+            buf_node = input_nodes[ii]
+            buf_fqn = inputs_to_buffers.get(getattr(buf_node, "name", None))
+            if buf_fqn is None or buf_fqn in already_exposed:
```
shoumikhin: Two cases drop an aliased output here while the runtime still requires one delegate arg per engine output (`TensorRTBackend.cpp:426`): a second engine aliasing the same buffer is deduped, and a caller-supplied cache that is not a registered buffer gives `buf_fqn is None` and is skipped, both ending in `Error::InvalidArgument` at execute. The legacy path skips the same two at lines 765 to 793, so either emit one output arg per engine output, or record in the blob which outputs are not threaded so the runtime can adjust its expected arg count.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -499,17 +627,65 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
...
+      if (a_dims.nbDims >= 0 && a_dims.nbDims <= nvinfer1::Dims::MAX_DIMS) {
+        SizesType a_sizes[nvinfer1::Dims::MAX_DIMS];
+        for (int d = 0; d < a_dims.nbDims; ++d) {
+          a_sizes[d] = static_cast<SizesType>(a_dims.d[d]);
+        }
+        (void)executorch::runtime::resize_tensor(et_alias_out, {a_sizes, static_cast<size_t>(a_dims.nbDims)});
+      }
+      void* dst = et_alias_out.nbytes() > 0 ? et_alias_out.mutable_data_ptr() : nullptr;
+      if (dst != nullptr && dst != bind_ptr) {
+        aliased_reflects.emplace_back(dst, bind_ptr, et_alias_out.nbytes());
```
shoumikhin, line 832: `dst == bind_ptr` cannot happen: the delegate input and output are live at the same time, so the memory planner never places them at the same address, and a mutable buffer is not in the planned arena at all. That makes the zero-copy path described in the comment and in the PR description dead, and every decode step pays a full cache-size device copy here plus ExecuTorch's write-back `copy_` on a buffer the engine already updated in place, so it is worth stating whether that cost is inherent to expressing this as `BUFFER_MUTATION`.

shoumikhin (commented):
On the output ordering question, it does not reproduce at this head and I think the author's reply is right. Running the real ExecuTorch `to_edge_transform_and_lower` over the head partitioner and `preprocess` gives delegate args `[cache, tokens, logits, cache_out]` against blob names `[cache_in, tokens, logits, cache_out]`, with the parent mapping delegate output 1 to the buffer mutation. Deleting `_keep_mutated_buffers_above_delegate` reorders the subprogram to `[cache_out, logits]` while the blob stays `[logits, cache_out]`, which is exactly the swap that was described, so the mechanism is real but the helper prevents it.

The comments below are the blockers and majors I found. On the other side, the device-residency rejection is correctly placed before both the host-staging and zero-byte paths, the `inflight_pending` drain gap I raised earlier is fixed, and the TensorRT version gate around `getAliasedInputTensor` is correct on both sides of 10.15 (checked by preprocessing it against several SDK header sets).

py/torch_tensorrt/dynamo/_exporter.py
Conarnar: This one in particular is an ordinary case. It fires for every engine output
that isn't aliased, which is most of them, so a warning would be hundreds of
lines on a larger model. The legacy path leaves the equivalent skip silent for
the same reason. If you do want logging here, `logger.debug` rather than a
warning.

py/torch_tensorrt/dynamo/_exporter.py
Conarnar: Fixed in the next push. Both `example_inputs` and `verifiers` are threaded
through now.

I should mention that the `retrace=False` path has a related gap:
`create_trt_exp_program` constructs its `ExportedProgram` without
`example_inputs` either, so AOTI would refuse a program saved that way for the
same reason. That one predates this PR and isn't a dropped value. There's no
source program to carry from, though it does receive `arg_inputs` it could use.
Let me know if you want that fixed as well.

examples/executorch_reference_runner/kv_cache_decode_check.cpp
Conarnar: Looks like it's already fixed on main: e2f18eb23c (executorch==1.3.1) landed
there in #4398, so this stack picks it up in the rebase. Verified it has the API
but not whether it builds. Will do that.

py/torch_tensorrt/_compile.py
Conarnar: Turns out declaring on the exported_program branch created a third bad combination on top of your two that you found.

Fixed in the next push, using your suggestion and that covers all three combinations rather than special-casing any of them.

Conarnar:
> None of this is exercised by CI. `kv_cache_decode_check` is not built or run anywhere: `.github/scripts/verify-executorch-reference-runner.sh` builds `--target example_executorch_runner` only, and the CMake target has no `add_test`. Every Python test mocks the seam too, so nothing fails if the binding order is wrong or if `_keep_mutated_buffers_above_delegate` is deleted, and `retrace=True` has no end-to-end coverage at all.
> 
> Separately, no build or test job has run at this head. `_decide.yml` maps a non-approval `pull_request_review` event to `lane=skip`, and the one `pull_request` run that started was cancelled by the later review submissions, so the three red `gate` checks are that cancellation rather than a real failure. Worth getting a real run before merge. There is also a small conflict with main in `examples/executorch_reference_runner/BUILD`, where both sides add one line to the same filegroup.

Got it. The setup for the reference runner will be mirrored for kv_cache_decode_check.
Also, the conflict will be fixed after rebasing.

py/torch_tensorrt/dynamo/_exporter.py
Conarnar: I was able to confirm the second case, but I wasn't able to test the first.

I couldn't get a two-engine model that aliases one buffer through the retrace=True path. The export fails earlier, inside torch.export, on the hybrid TRT+eager graph. So I can't say whether that branch is reachable in practice. If you have a model in mind that hits it, that would help.

Regarding the fix, your second option looks workable for the non-buffer case specifically, since there's no ET buffer to write back to. I'd be wary of applying it generally though. An un-threaded aliased output also gets no write-back, and for a real buffer that turns an execute-time error into a silently stale cache. So scoping the blob-side elision to the non-buffer case and keeping one-arg-per-output elsewhere seems safer.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Confirmed dead. I instrumented the branch on a 30-layer export: 120 aliased outputs across prefill and decode, `dst == bind_ptr` false every time. The comment should go.

One correction: in our .pte the mutable buffers are planned values (mem_id 1, no mutable data segment). The reason dst != bind_ptr is the one you give, plus a second one: the delegate never receives the buffer at all. PropagateDevicePass wraps every delegate input in _h2d_copy, so the engine binds a per-call staging copy, and the reflect plus ExecuTorch's write-back exist to carry the update back to the buffer.

On whether the cost is inherent to BUFFER_MUTATION: it isn't. Pointing the mutation at the buffer placeholder does make ExecuTorch emit no copy_, but that alone breaks persistence, because the staging copy is re-made from the buffer at the top of every call, so the engine's in-place write is overwritten before it is read. Removing the staging too fixes it: the buffer lands in the device arena, the delegate binds it directly, and every full-cache copy goes away. On a single-layer KV model that is 16 -> 8 values and 10 -> 4 instructions, with the persistence check matching eager exactly (0.235384) where the elision-only build reported 0.

That is prototype-grade: a toy model, and the un-staging rule still needs a guard that every consumer of the buffer is device-capable.

The natural home looks like PropagateDevicePass itself, so the copy is never inserted for a delegate-mutated buffer in the first place, though that is more your call than mine. We could also carry it as a Torch-TensorRT pass that undoes the staging for our delegate only, if that sequences better, but it would have to hook in between PropagateDevicePass and memory planning, which is more fragile than doing it properly upstream.

Worth pursuing separately. For this PR I just corrected the comments. I kept the dst != bind_ptr check itself as a self-copy guard rather than deleting it, since it costs a pointer compare and is what stops a self-copy if planning ever changes.

shoumikhin:
`tests/cpp/executorch/test_executorch_blob_header.cpp:164-165` uses member names that don't exist on `AliasedBinding`, so the file can't compile:

```cpp
164:  EXPECT_EQ(header.aliased_io[0].output_name, "out_k");
165:  EXPECT_EQ(header.aliased_io[0].input_name, "in_k");
```

`TensorRTBlobHeader.h:15-19` declares `output` / `input` / `kind`, and the same file already uses the correct spelling 38 lines earlier at 126-127, so this looks like a typo in the last test added rather than an intended accessor.

Compiled it to be sure (gtest stubbed, `EXPECT_EQ` still does the member access):

```
test_executorch_blob_header.cpp:164:34: error: no member named 'output_name' in 'AliasedBinding'
test_executorch_blob_header.cpp:165:34: error: no member named 'input_name' in 'AliasedBinding'
```

With `s/.output_name/.output/` and `s/.input_name/.input/` the file compiles clean.

Worth noting for triage: `tests/cpp/executorch/BUILD` declares this as a real `cc_test` and it's in two `test_suite`s, but no CI job runs `bazel test` (the workflow only does `bazel build //:libtorchtrt`), so nothing on the PR catches it. The person who hits it is the next developer who runs the suite locally.

shoumikhin:
`TensorRTBlobHeader.cpp:239` searches for the `aliased_io` key from offset 0:

```cpp
const std::size_t alias_key = json.find("\"aliased_io\"");
```

`io_bindings` is serialized first and its elements carry `"name"` values, so a binding literally named `aliased_io` matches before the real key. Binding names are the model's own argument names, verbatim — `_TRTInterpreter.py:592-596` uses `binding_name = target` with no sanitization — so `def forward(self, aliased_io, y)` is enough to trigger it.

I fed blobs from the real `serialize_engine` to the real parser:

```
real_benign.bin    parse=TRUE  n_alias=1  out=out_k in=in_k  kind=kv_cache_update
real_hostile.bin   parse=TRUE  n_alias=0
```

It returns **true with zero aliases** rather than failing. `backend.py` never sets `shape`, so every emitted blob has `"shape":[]`; `find('[', alias_key)` lands on that empty array, the walk sees `]` immediately and appends nothing. (A blob with a non-empty shape does return false, but that's not what the pipeline emits.)

Zero aliases is then indistinguishable from a non-aliased model at `TensorRTBackend.cpp:313-315`, so the aliased output binds to a fresh allocation instead of the caller's KV buffer — wrong results, no error, and the "aliased output(s) bound in-place" log line stays silent. The TR02 magic doesn't catch it either, since there's no "TR02 implies at least one alias" cross-check.

The pre-existing helpers in this same file already defend against exactly this: `parse_bool_after_key` / `parse_int_after_key` take a `search_from`, and line 314-315 passes `pos` (just past the `io_bindings` walk). `serialization.py:53-54` even documents the contract. One argument restores it:

```cpp
const std::size_t alias_key = json.find("\"aliased_io\"", pos);
```

Verified: benign case unchanged, hostile case now parses `in=aliased_io` correctly.

shoumikhin:
`_declare_aliased_kv_mutations_on_ep` imports ExecuTorch unconditionally, and `_compile.py` calls it on the `exported_program` path — so saving a compiled `GraphModule` now requires the optional `[executorch]` extra even for a model with no aliased IO.

The import is the 4th statement in the function body, before the "no aliases -> return unchanged" logic can run (`_exporter.py:545`, confirmed by AST — zero enclosing `Try`/`If`):

```python
from torch_tensorrt.executorch.backend import _get_engine_info_for_node
```

and `backend.py:8-12` hard-imports `executorch.exir.backend.backend_details` at module scope.

The `__init__.py` guard doesn't cover this. Module-level `__getattr__` fires on attribute lookup, not on a submodule import. Ran it with executorch absent:

```
A) m.TensorRTBackend            -> ImportError: Cannot access ... Install with: pip install "torch_tensorrt[executorch]"
B) from ...executorch.backend import X  -> ModuleNotFoundError: No module named 'executorch'
```

`executorch` is genuinely optional — `setup.py:204-214` puts it in `extras_require` only, and it's absent from `py/requirements.txt` — so an install without the extra is the supported default.

Reachability of the four call sites:

| line | chain | alias check? |
|---|---|---|
| 1079 | `fx` → `not retrace` → `exported_program` | none |
| 1185 | `fx` → retrace → `exported_program` | none |
| 1100, 1206 | → `executorch` | already gated at 829 |

The two `executorch` sites are fine. The two `exported_program` sites have no alias predicate at all, while the `aot_inductor` branch 11 lines away at 1171 does exactly that check (`any(getattr(sub, "aliased_io", None) ...)`) before merely warning.

Precedent in this same file suggests the guard was expected: line 829 raises a clear `ImportError` for `output_format="executorch"`, and `_save_as_executorch` wraps its own import in `try/except` at 1250-1253.

Scope, to be precise: this needs `module_type == fx`, i.e. `save(graph_module, ...)`. Passing an `ExportedProgram` takes the `ep` branch at 987 and never calls the pass. Since `dynamo.compile()` returns a `GraphModule`, the failing spelling is the common one.

Either move the import below the alias early-out, or wrap it the way `_save_as_executorch` already does.

Validation limit: I established the guard-bypass and the reachability by execution and AST over the real source, but couldn't run `torch_tensorrt.save()` end to end (no `libnvinfer.so.10` in my env), and I didn't stub `tensorrt` since that could mask the failure under test.

shoumikhin:
`kv_cache_decode_check` can print PASS when KV persistence is completely broken, because the two runs' caches are never actually zeroed.

The pass condition is only that the two runs differ (`kv_cache_decode_check.cpp:203-204`), and each `run_decode` allocates its own arena with no zeroing:

```cpp
ET_CHECK_MSG(cudaMalloc(&p, sz) == cudaSuccess, "cudaMalloc planned buffer %zu failed", i);
```

`grep -n "cudaMemset\|memset"` on the file returns nothing. The host arenas are fine (`make_unique<uint8_t[]>` value-initializes); only the CUDA-tagged ones are affected, and those are the ones that matter here since `TensorRTBackend.cpp:567-578` rejects a non-device-resident aliased input.

The `register_buffer(torch.zeros(...))` initial value doesn't help — ExecuTorch discards it. Its own emitter warns about exactly this:

> Mutation on a buffer in the model is detected. ExecuTorch assumes buffers that are mutated in the graph have a **meaningless initial state**, only the shape and dtype will be serialized, unless a pass which sets `meta["et_init_buffer"]` to True ...

Emitting this model confirms it: every KV tensor comes out `data_buffer_idx=0` with `allocation_info` set, which selects the branch commented `// Memory planned, no initial state` in `tensor_parser_exec_aten.cpp:288-291` — it returns `buffer.data() + offset` and nothing memsets it. `et_init_buffer` / `InitializedMutableBufferPass` don't appear anywhere in the tree.

So with a hypothetical no-persistence implementation, varying only the arena contents:

```
scale=1     max|A-B|=0.408373  -> PASS   (false pass)
scale=100   max|A-B|=78.8265   -> PASS   (false pass)
```

Two different dirty arenas give A != B, which is all the check requires. A correct implementation gives a genuine ~0.235 difference, so in practice `cudaMalloc` returning zeroed pages is what's carrying this today.

There's a spurious-FAIL direction too: `std::max(0.0, fabs(NaN - x))` is 0.0, so NaN logits collapse `max_abs_diff` to 0 and it reports "outputs are identical".

This is a CI gate (`verify-executorch-reference-runner.sh:337-338` greps for the PASS string), and the four "zeroed cache" claims at source lines 14, 22, 73 and README:149 describe a state the code never establishes. One line after line 98 fixes it:

```cpp
cudaMemset(p, 0, sz);
```

Stronger still would be asserting B's pos=1 logits against an independently computed expectation rather than just "A differs from B".

(No GPU here, so I did not run the binary end to end; the emission run was CPU-only and the garbage patterns are simulated, which shows the criterion is insufficient rather than how often real `cudaMalloc` returns nonzero bytes.)

shoumikhin:
Retracting my earlier "none of this is exercised by CI" comment — that was true when I wrote it and isn't any more. I traced the chain at the current head and it's wired correctly end to end:

```
executorch-test-linux.yml:85   export_kv_cache_decode.py -> .pte
                        :87   verify-executorch-reference-runner.sh "$pte" "$kv_pte"
verify-...sh:297   build_targets+=(kv_cache_decode_check)     [when $2 is given]
            :301   cmake --build --target "${build_targets[@]}"
            :336   "${kv_check_path}" --model_path=... | tee "$kv_check_log"
            :337   grep -q "PASS: decode at pos=1 observed the KV written at pos=0"
```

`set -euo pipefail` is on line 2, so the `| tee` can't mask a non-zero exit, and the `grep -q` means a missing PASS line fails the step rather than passing vacuously. `bash -n` is clean. The new `ldd`-must-not-link-libtorch guard also mirrors the pre-existing one for `example_executorch_runner`.

The separate point about what that PASS string actually proves is in my other comment above; that's about the check's internals, not the wiring.

Two other objections I raised earlier that I've since confirmed are handled, so they shouldn't hold anything up:

- **Output ordering.** `_validate_output_binding_order` (`backend.py:269`) raises on every preprocess if the delegate outputs don't map to `range(len(output_names))`, so this doesn't depend on the ordering being structurally guaranteed.
- **Aliased inputs taking the host-staging path.** `TensorRTBackend.cpp:567-579` rejects a non-device-resident aliased input with `Error::InvalidArgument` *before* the three staging branches at 581-600.

shoumikhin:
One process note that affects how much confidence to place in the green checks here: the new Python tests don't run on PR pushes.

`tests/ci/suites.py:258-267`:

```python
Suite(
    "executorch",
    tier="l2",
    lanes=("nightly",),
    paths=("executorch/",),
    ...
)
```

`lanes=("nightly",)` means the suite is absent from both the `fast` and `full` matrices, so the tests under `tests/py/dynamo/executorch/` execute on the nightly cut only — not on push, not on approval, not on merge. The C++ side is covered (the `executorch-runtime-test` job does build and run `kv_cache_decode_check` on every push, which I verified separately), but the Python tests aren't, and neither is the new `cc_test` (nothing runs `bazel test`).

Not a defect in this PR — the lane assignment is pre-existing and this PR doesn't touch `suites.py`. Flagging it because it means a merge can break these tests while CI stays green, which is worth knowing given the stack lands as one unit.

shoumikhin:
The declaration pass never runs when the input is already an `ExportedProgram`, which looks like a gap rather than a deliberate exclusion.

AST over `save()` in `_compile.py`:

```
_declare_aliased_kv_mutations_on_ep  call lines: [1079, 1100, 1185, 1206]
_normalize_engine_constants_to_python call lines: [998, 1080, 1186]
```

The `ep` branch is 987-1028 and the `fx` branch starts at 1029, so all four pass calls are in `fx` only. Line 998 shows the `ep` branch does its own `_normalize_engine_constants_to_python` before `torch.export.save`, so it's a real, separately-maintained path — it just doesn't get the mutation declaration.

An engine-bearing `ExportedProgram` is a supported input, so `compile(...)` → `torch.export.export(...)` → `save(ep, output_format="exported_program")` would silently produce a program whose signature omits the KV update, which is the exact failure the pass exists to prevent.

Either run the pass on that branch too, or if it's intentionally out of scope, a comment saying so would help — right now the asymmetry with the adjacent `_normalize_engine_constants_to_python` call reads as an oversight.

Two small things while I was in there:

- `_exporter.py:525` — the docstring says `"""retrace=True post-export pass: ..."""`, but the calls at 1079/1100 are under `if not retrace:`, and your own test is named `test_save_declares_aliased_mutations_without_retrace`. The pass runs on both paths.
- `TensorRTBlobHeader.cpp:302-305` — the comment says a missing `"kind"` "means an older blob (the Python serializer omits it for KV aliases)", but `serialization.py:67-70` always emits `"kind"`, and no older blob has an `aliased_io` array at all. The default is fine to keep defensively; the stated reason just isn't accurate.

shoumikhin:
Minor format observation on the `TR01`/`TR02` approach: the header already carries an unused version field, and using the magic as a feature flag means every future optional feature needs a new magic plus a growing accept-list.

`serialization.py:23` is `HEADER_FORMAT = "<4sIIIQ8s"`, and line 106 writes `reserved = b"\x01" + b"\x00" * 7` — a version byte. Field offsets in that packed layout:

```
magic             0
metadata_offset   4
metadata_size     8
engine_offset    12
engine_size      16
reserved         24   <- version byte lives here
```

Neither reader looks at it. C++ `parse` reads only offsets 4/8/12/16 (`TensorRTBlobHeader.cpp:346-349`), and the Python reader unpacks the field into `_` and discards it (`serialization.py:123-124`). So the mechanism the format was designed with is inert, while `SUPPORTED_MAGICS` now has to grow per feature:

```python
TENSORRT_MAGIC_ALIASED_IO if metadata.aliased_io else TENSORRT_MAGIC
```

Your compatibility intent is right and works in both directions — I checked that an old reader rejects `TR02` loudly and a new reader still accepts `TR01`. This is only about which knob carries it. Bumping/validating the reserved version byte would scale without touching the magic each time.

Not blocking, and I'd understand deferring it — changing the mechanism now is arguably more churn than it's worth, but it seemed worth noting while the format is still young.

examples/executorch_reference_runner/kv_cache_decode_check.cpp
shoumikhin: Fixed, verified on the current head (`c87d3bbd8`). The pin moved to ExecuTorch 1.4.1, which has the API:

```
# MODULE.bazel
# executorch==1.4.1
commit = "e4d02f41f7909e8ed5bf4a14ffc520d733453d9f"
```

Checked that `runtime/executor/method_meta.h` at that commit declares `memory_planned_buffer_device`, and that `kv_cache_decode_check.cpp:95` still calls it, so the target has the API it needs.

Please resolve this thread; I do not have write access to do it myself.

py/torch_tensorrt/executorch/serialization.py
```diff
@@ -50,6 +55,12 @@ def to_json(self) -> bytes:
                 }
                 for binding in self.io_bindings
             ],
+            # List form (not a dict) so the small C++ parser can walk it like
+            # io_bindings. Emitted right after io_bindings, before the scalars.
+            "aliased_io": [
```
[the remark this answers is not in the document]
shoumikhin: Fixed, verified on the current head (`c87d3bbd8`). Aliased blobs now carry a distinct magic, so a pre-aliasing parser rejects them instead of silently mis-binding:

```python
TENSORRT_MAGIC = b"TR01"
TENSORRT_MAGIC_ALIASED_IO = b"TR02"
SUPPORTED_MAGICS = (TENSORRT_MAGIC, TENSORRT_MAGIC_ALIASED_IO)
...
TENSORRT_MAGIC_ALIASED_IO if metadata.aliased_io else TENSORRT_MAGIC,
```

Blobs without `aliased_io` keep `TR01` and stay readable by older runtimes, which is the right split.

Please resolve this thread; I do not have write access to do it myself.

py/torch_tensorrt/executorch/backend.py
```diff
@@ -306,8 +308,14 @@ def preprocess(
             TensorRTIOBinding(name=name, is_input=True) for name in input_names
         ] + [TensorRTIOBinding(name=name, is_input=False) for name in output_names]
 
+        # Carry the KV-cache / user aliasing (out->in, kind) into the blob so the
+        # C++ backend binds each aliased output to its aliased input's tensor
+        # (in-place) and reflects the update back into the delegate output.
+        aliased_io = deserialize_aliased_io(_get_str(engine_info, ALIASED_IO_IDX))
```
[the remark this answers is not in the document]
shoumikhin: Fixed, verified on the current head (`c87d3bbd8`). Outputs now get their own order check, so a permuted or missing delegate output fails loudly instead of silently swapping:

```python
raise ValueError(
    "TensorRT ExecuTorch backend: delegate output "
    f"{getattr(node, "name", node)!r} is not a getitem of the engine "
    "node; cannot establish a reliable output binding order."
)
...
if indices != list(range(len(output_names))):
    raise ValueError(
        "TensorRT ExecuTorch backend: delegate outputs map to engine output "
        f"indices {indices}, expected {list(range(len(output_names)))} ..."
    )
```

The reasoning for why the order holds, including the `_keep_mutated_buffers_above_delegate` dependency I raised, is now written down above it rather than being an unstated invariant.

Please resolve this thread; I do not have write access to do it myself.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -577,6 +748,18 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
...
   }
 
+  // Caller-owned KV: reflect each engine in-place update into its delegate output
+  // EValue (D2D on the same stream, after the engine work). No-op list under
+  // zero-copy (dst == src filtered out at bind time).
+  for (const auto& r : aliased_reflects) {
+    cuda_err = cudaMemcpyAsync(std::get<0>(r), std::get<1>(r), std::get<2>(r), cudaMemcpyDeviceToDevice, stream);
+    if (cuda_err != cudaSuccess) {
+      ET_LOG(
+          Error, "TensorRTBackend::execute: aliased-output reflect D2D copy failed: %s", cudaGetErrorString(cuda_err));
```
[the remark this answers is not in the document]
shoumikhin: Fixed, verified on the current head (`c87d3bbd8`). The reflect-copy failure path now drains the stream before returning, and the reason is recorded at the site:

```cpp
ET_LOG(Error, "TensorRTBackend::execute: aliased-output reflect D2D copy failed: %s", ...);
// enqueueV3 already submitted engine work to `stream`, and inflight_pending
// is not armed until the end of the happy path -- drain now so a later
// execute() or the destructor never reconfigures/frees exec_ctx while this
// enqueue is still running.
(void)cudaStreamSynchronize(stream);
return Error::InvalidProgram;
```

That closes the window I described, where both later guards were gated on a flag that had not been set yet.

Please resolve this thread; I do not have write access to do it myself.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
shoumikhin: Fixed, verified on the current head (`c87d3bbd8`). The `dst == bind_ptr` comparison is gone; `bind_ptr` now appears only on the input-binding side, so the dead zero-copy branch and the comment claiming it no longer exist.

Please resolve this thread; I do not have write access to do it myself.

py/torch_tensorrt/_compile.py
shoumikhin: Fixed, verified on the current head (`c87d3bbd8`). The other two formats are no longer silent:

- `output_format="exported_program"` now runs the same `_declare_aliased_kv_mutations_on_ep(exp_program)` pass as the executorch branch, in both the retrace and no-retrace paths.
- `output_format="aot_inductor"` warns explicitly rather than saving as-is:

```python
if output_format == "aot_inductor" and any(getattr(sub, "aliased_io", None) ...):
    logger.warning(
        "Module has TensorRT engine(s) with aliased I/O (e.g. KV-cache), "
        "but output_format='aot_inductor' does not declare those aliased ..."
    )
```

So an aliased model saved in another format either gets the handling or is told it will not.

Please resolve this thread; I do not have write access to do it myself.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
shoumikhin: Correcting my previous reply, which was wrong. I grepped for `bind_ptr` with a truncated output and only saw the input-binding uses, so I said the comparison was gone. It is not:

```cpp
void* dst = et_alias_out.nbytes() > 0 ? et_alias_out.mutable_data_ptr() : nullptr;
// dst != bind_ptr guards against issuing a self-copy. The memory planner does
// not currently place the delegate's output slot on the aliased input -- the
// two are live at the same time -- so this holds for every aliased output.
if (dst != nullptr && dst != bind_ptr) {
```

What has changed is the comment, which now states the same thing I raised: the planner does not place them together, so the branch is always taken. So the misleading claim of a zero-copy fast path is gone, which was half my point.

The other half stands and is **not** fixed: every decode step still pays a full cache-size device copy here, plus ExecuTorch's write-back `copy_`. Please leave this thread open for that.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -499,17 +627,65 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
...
+        return Error::InvalidArgument;
+      }
+      exec_aten::Tensor et_alias_out = out_arg->toTensor();
+      nvinfer1::Dims a_dims = ctx->getTensorShape(name.c_str());
+      if (a_dims.nbDims >= 0 && a_dims.nbDims <= nvinfer1::Dims::MAX_DIMS) {
+        SizesType a_sizes[nvinfer1::Dims::MAX_DIMS];
+        for (int d = 0; d < a_dims.nbDims; ++d) {
+          a_sizes[d] = static_cast<SizesType>(a_dims.d[d]);
+        }
+        (void)executorch::runtime::resize_tensor(et_alias_out, {a_sizes, static_cast<size_t>(a_dims.nbDims)});
```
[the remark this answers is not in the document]
shoumikhin: Fixed, verified on the current head (`c87d3bbd8`). The aliased branch now handles the `resize_tensor` result the same way the sibling branch does, rather than discarding it:

```cpp
Error a_resize_err =
    executorch::runtime::resize_tensor(et_alias_out, {a_sizes, static_cast<size_t>(a_dims.nbDims)});
if (a_resize_err != Error::Ok) {
  ET_LOG(Error, "TensorRTBackend::execute: resize_tensor failed for aliased output '%s'", name.c_str());
  return a_resize_err;
}
```

That matches the non-aliased path, so a dynamic aliased output that outgrows its planned size fails instead of proceeding with a stale `nbytes()`.

Please resolve this thread; I do not have write access to do it myself.

shoumikhin:
I re-checked my own review threads against the current head `e3482aa7d`, since my earlier "fixed" replies were written against `c87d3bbd8`. I do not have write access here, so I cannot resolve threads myself.

**Verified fixed on `e3482aa7d`, safe to resolve (5):**

- Missing stream drain if a reflect copy fails. `TensorRTBackend.cpp:932` now calls `cudaStreamSynchronize(stream)` before clearing `inflight_pending`.
- `resize_tensor` result discarded on the aliased branch. Handled at 821-826, matching the non-aliased path.
- `aliased_io` changes a blob's meaning while the magic stays `TR01`. `serialization.py` now has `TR02` for aliased blobs and `SUPPORTED_MAGICS`, so an older runtime rejects rather than misreads.
- Output binding order only accidentally correct. `_validate_output_binding_order` is called at `backend.py:371` and raises on a permuted or incomplete list.
- Declaration decided twice. The pass is now idempotent by skipping buffers that already carry a spec, and this branch has no copy-back path, so there is no second decision.

**Correction to my earlier reply, which was wrong.** I said the other two output formats were no longer silent, and that `aot_inductor` "warns explicitly". The `exported_program` half is right on all three paths. The warning is not: `save()` has three `aot_inductor` sites and only one checks for aliased engines.

```
line 1092   module is already an ExportedProgram   no warning
line 1179   fx + retrace=False                     no warning
line 1266   fx + retrace=True                      warns
```

So an aliased model saved as `aot_inductor` from two of the three entry points is still saved with no signal, which is what the thread was about. Please leave that one open.

**Should stay open (2):**

- The two cases that drop an aliased output: they now log, which is a real improvement, but the behaviour is unchanged and the code comment still says the delegate ends up one arg short at execute.
- The `dst == bind_ptr` thread: the guard and comment are correct now, but the per-step full cache-size copy is untouched. Fine to handle separately as you proposed, just not resolved by this PR.

One thing worth knowing for anyone reading the green checks: the `executorch` suite in `tests/ci/suites.py` is `lanes=("nightly",)`, so the Python tests added here do not run on a pull request.

shoumikhin:
I re-checked my open threads against this head and the fixes hold, so thanks for those. The reflect-copy path now drains the stream before returning, the aliased `resize_tensor` result is treated as fatal like the sibling branch, the blob magic split is in, and the delegate output order is checked. I also ran the new `aot_inductor` warning and confirmed it fires on all three save paths.

One small thing is left, and it is only a comment.

`TensorRTBackend.cpp:565` still says:

```cpp
// threaded as a delegate output arg (the caller-owned mutable buffer's mutation
// slot), so all engine bindings map 1:1 to delegate args.
const size_t num_delegate_outputs = num_outputs;
```

That 1:1 claim is not true when `_declare_aliased_kv_mutations_on_ep` takes one of its skip paths, and the check four lines below turns the mismatch into an `InvalidArgument` at execute. Your new logging is a real improvement here: of the seven `continue` statements in that pass, the four that skip a real aliased output all report themselves now, and the three silent ones are correct to be silent because they are just "not applicable" filters. So the behavior is visible, only the comment is stale. Rewording it to say that bindings map 1:1 unless a declaration was skipped, with a pointer to the warning, would be enough.

I agree with parking the arity fix itself. You confirmed the non-buffer case, neither of us could build the two-engine case through the retrace path, so half of it is unproven and it does not need to hold up this change.

I also agree on the zero-copy item. You confirmed the branch is dead by instrumenting a real export, the guard is still correct as a self-copy check, and the comment no longer advertises a fast path that never runs. What is left is the per-step copy cost, which is a design question rather than a correctness one.

One heads-up: the dead `already_mutated` skip I reported on the child PR is already fixed by your own later commit there, so that one is done and does not need another pass.

The seven red checks are not caused by this PR. They all fail on `No module named 'torch_tensorrt.dynamo.lowering.constant_fold_exclusions'`, which landed on main after this branch's merge base, and this branch is nine commits behind. A rebase should clear them.

shoumikhin:
Correction to my comment above, on one point. I said the stale 1:1 comment matters because a skipped declaration turns into an `InvalidArgument` at execute. I checked that more carefully and that framing is too strong.

`preprocess` validates before anything reaches the C++ side. `_validate_output_binding_order` is called at `backend.py:363` and `_reorder_input_names_for_executorch` at `:371`, and a malformed program fails there:

```
ValueError: delegate outputs map to engine output indices [0, 1], expected [0, 1, 2]
ValueError: engine has 2 input binding names but 1 engine input nodes
```

Unknown alias names are also rejected during backend initialization at `TensorRTBackend.cpp:445-463`. So for any program that gets accepted, the C++ comment at `:564-566` is accurate, and I could not find a supported route that reaches the arity check one argument short.

That downgrades the item from "a stale comment hiding a runtime failure" to just a documentation nit, and the more wrong comments are actually on the Python side, at `_exporter.py:613-616` and `:638-640`, which say these cases surface at the execute-time arity check when they really surface in `preprocess`. Still worth a small edit, but nothing here should hold up the change. Sorry for the noise.

shoumikhin:
Correcting myself on the red checks. I told you a rebase would clear them. That is wrong, and I am sorry, because it would have sent you chasing a no-op.

The same jobs are red on `main`'s own head (`691b78841d`), so there is nothing green to rebase onto:

```
dynamo-lowering            py-core                    ts-api
dynamo-converters          dynamo-runtime-smoke       dynamo-partitioning-smoke
python-only-runtime        executorch-runtime-build
```

That is eight, not seven. I had missed `executorch-runtime-build` earlier because it starts later than the others.

The actual cause is a packaging gap, not a stale branch. `constant_fold_exclusions` landed as a new package **directory**:

```
py/torch_tensorrt/dynamo/lowering/constant_fold_exclusions/__init__.py
py/torch_tensorrt/dynamo/lowering/constant_fold_exclusions/_core.py
```

but `setup.py` lists packages explicitly rather than using `find_packages()`, and `dynamo_packages` only has:

```python
"torch_tensorrt.dynamo.lowering",
"torch_tensorrt.dynamo.lowering.passes",
```

No entry for the new subpackage, and none in `package_dir` either. So a wheel built from this omits the directory, and then the runtime import fails, because `_settings.py:69` imports it unconditionally inside `_normalize_disabled_constant_fold_exclusions`:

```python
from torch_tensorrt.dynamo.lowering.constant_fold_exclusions._core import (
    validate_disabled_constant_fold_exclusions,
)
```

That is why only the jobs that install the wheel and then import it go red, while the repo itself looks fine.

The fix is to add `"torch_tensorrt.dynamo.lowering.constant_fold_exclusions"` to `dynamo_packages` and `package_dir` in `setup.py`. Until that lands somewhere, these jobs stay red for everyone, so a green run here is blocked on that rather than on anything in your change.

One caveat so I do not overstate it again: `main`'s newest `dynamo-lowering` log does not contain the import error at all, it dies earlier on infra (`Specified script file not found`, plus a conda channel problem). So main is red for at least one additional reason. Your branches get far enough to hit the packaging error; main's latest run does not even reach it.

Conarnar:
> That downgrades the item from "a stale comment hiding a runtime failure" to just a documentation nit, and the more wrong comments are actually on the Python side, at `_exporter.py:613-616` and `:638-640`, which say these cases surface at the execute-time arity check when they really surface in `preprocess`. Still worth a small edit, but nothing here should hold up the change. Sorry for the noise.

Will handle this in the child PR #4459, this one should be good to go.

shoumikhin:
Cleaning up after myself: I had left a lot of long comments here, several already fixed and a couple simply wrong, so I deleted them and replaced them with this note. I left the inline threads where you replied, since your answers there are worth keeping.

Nothing on this PR blocks merge from my side.

Everything I raised is either fixed on the current head, which I re-checked, or parked with a reason I agree with. The two open ones are the arity mismatch, where half the case is still unproven and neither of us could build a two-engine repro, and the double-declare, which is really a `#4459` concern.

Two corrections so nobody chases them:

- The stale `1:1 bindings` comment I complained about is not hiding a runtime failure. `preprocess` validates first, at `backend.py:363` and `:371`, and unknown alias names are rejected during backend initialization, so for any accepted program that comment is accurate. If anything the misleading comments are the Python ones at `_exporter.py:613-616` and `:638-640`, which point at the execute-time arity check. Minor either way.
- I said a rebase would clear the red checks. It will not. The same jobs are red on `main`: `constant_fold_exclusions` landed as a new package directory but was never added to the `dynamo_packages` list in `setup.py`, so the built wheel omits it and the runtime import fails. Not caused by this PR, and not fixed by rebasing.

The remaining review comments are on `#4459`, which is where the copy-back work lives.

PR #4448, opened 2026-07-30
fix: optimize attention decomposition

fs-eire:
# Description

Align attention decomposition behavior with the ONNX graph.

This PR does the following:

- Restrict attention-mask Select choices to scalars while preserving mask polarity and -inf semantics.
- Preserve static arange operations as TensorRT LINSPACE Fill layers.
- Add regression coverage for attention masks and static arange conversion.

With this change, Torch-TensorRT is able to generate optimized engine with a pattern that an attention using a causal mask AND a padding mask, when `decompose_attention=True`.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/arange.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/arange.py
@@ -24,12 +24,12 @@ def arange(
     step: Union[int, TRTTensor],
 ) -> TRTTensor:
     """
-    Creates a sequence of values (arange) either dynamically or statically,
-    then outputs a TensorRT tensor.
+    Creates a sequence of values (arange) with a TensorRT Fill layer.
 
-    If any of (start, end, step) is a TRT tensor, it sets up a dynamic arange
-    using a Fill layer. Otherwise, it creates a static NumPy array and converts
-    it into a TensorRT constant tensor.
+    If any of (start, end, step) is a TRT tensor, the Fill output length is
+    computed dynamically. Otherwise, NumPy is used only to determine the static
+    output length. Keeping static ranges as Fill layers preserves their sequence
+    provenance for downstream TensorRT graph-pattern recognition.
     """
     # If any argument is a TRT tensor, use dynamic arange with a Fill layer
     if any(isinstance(x, TRTTensor) for x in (start, end, step)):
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -485,18 +485,21 @@ def scaled_dot_product_attention_decomposition(
...
     device = query.device
 
-    if is_causal or attn_mask is not None:
-        attn_bias = torch.zeros((L, S), dtype=query.dtype, device=device)
-
     if is_causal:
         assert attn_mask is None, "attn_mask must be None when is_causal=True"
+        attn_bias = torch.zeros((L, S), dtype=query.dtype, device=device)
         temp_mask = torch.ones((L, S), dtype=torch.bool, device=device).tril(diagonal=0)
         attn_bias = attn_bias.masked_fill(temp_mask.logical_not(), float("-inf"))
```
micwill755, line 492: For the bool attn_mask path we switched to scalar where(0 / -inf), but is_causal still uses masked_fill. Should the causal path use the same scalar where pattern?

py/torch_tensorrt/dynamo/conversion/impl/arange.py
```diff
@@ -68,9 +68,20 @@ def arange(
...
+        step_tensor = get_trt_tensor(
+            ctx, step, name + "_step", dtype=trt.int32, min_rank=1
+        )
+        fill_layer = ctx.net.add_fill(
+            output_shape, trt.FillOperation.LINSPACE, trt.int32
+        )
+        fill_layer.set_input(1, start_tensor)
+        fill_layer.set_input(2, step_tensor)
+        set_layer_name(fill_layer, target, f"{name}_arange_fill", source_ir)
+        return fill_layer.get_output(0)
```
zewenli98, line 135: Good finding! Is this arange op upstream of attention? If so, it should resolve the perf issue for both IAttention path and Decomposed path, right?

Besides, I have an addition to arange converter. Can you review https://github.com/pytorch/TensorRT/pull/4456?

fs-eire: > Is this arange op upstream of attention? If so, it should resolve the perf issue for both IAttention path and Decomposed path, right?

yes

> #4456

The change should be good. I think you can merge it first and then I can combine my change based on your change.

PR #4456, opened 2026-08-01
fix: arange converter

### LINKED ISSUE #2894: 🐛 [Bug] arange converter dynamic shape questions
##  Bug Description

1) Should this be moved to aten_ops_converters.py?
2) I tried to pass in static shape ITensor end to the test case of this converter but got error: aten::arange() Expected a value of type 'number' for argument 'end' but instead found type 'FakeTensor'. because the [schema](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L786) says the type of the arguments is Scalar
3) In terms of the implementation, 1) shape seems not to be always valid, 2) output type is decided by end.dtype but it's possible the start is float and end is int. e.g.
>>> torch.arange(1.2, 15, 2.2)
tensor([ 1.2000,  3.4000,  5.6000,  7.8000, 10.0000, 12.2000, 14.4000])
>>> torch.ops.aten.div.Tensor_mode(13.8, 2.2, rounding_mode="trunc")
tensor(6.)
I think the actual length should be 7 but the length of shape is 6. Please correct me if I'm wrong.

## To Reproduce

1.
2.
3.

## Environment

 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - Python version:
 - CUDA version:

zewenli98:
# Description

Support fractional cases and fix the potential dynamic shape error

Fixes #2894

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -4018,6 +4018,7 @@ def aten_ops_arange_start_step(
         start=args[0],
         end=args[1],
         step=args_bounds_check(args, 2, 1),
+        dtype=kwargs.get("dtype", None),
     )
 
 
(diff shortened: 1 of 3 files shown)


narendasan (approved):
LGTM

PR #4462, opened 2026-08-04
[MD-TRT] feat: enable native multi-device TensorRT on TensorRT-RTX builds

apbose:


narendasan (commented):
Seems fine, but will let @lanluo-nvidia give the final stamp

lanluo-nvidia (approved):
LGTM

PR #4464, opened 2026-08-06
[fix] MD-TRT: rebind NCCL communicator after execution context invalidation

apbose:
bind_nccl_comm() attaches the communicator to the IExecutionContext. disable_profiling() and set_resource_allocation_strategy() dropped that context without clearing nccl_initialized, so the lazy re-bind in execute_engine.cpp never fired and the replacement context ran collectives with no communicator (~98% output mismatch). Mirrors runtime_settings().


apbose:
The PR is passing the L2 distributed tests.

PR #4467, opened 2026-08-07
fix(tests): correct TensorRT-RTX deconv version gate and guard plugin-backed converter tests

tp5uiuc:
# Description

Fixes 4 failing converter test targets on TensorRT-RTX builds. Two independent causes.

**1. `conv_deconv.cpp` takes the pre-TensorRT-7.1 fallback on RTX.** TensorRT-RTX defines `NV_TENSORRT_MAJOR` as `TRT_MAJOR_RTX` (`== 1`), so the version check guarding `setDilationNd()` / `setNbGroups()` is false on every RTX build. Deconvolutions with `groups > 1` or `dilation > 1` are then rejected citing TensorRT 7.1, although RTX supports both. Adds the `defined(TRT_MAJOR_RTX)` escape already used for this in `ConversionCtx.cpp` and `quantization.cpp`.

Reached only by the TorchScript frontend — the dynamo converter gates strided *and* dilated deconv on RTX, not groups.

**2. Plugin-backed converters have no RTX implementation.** Adaptive pooling, `aten::norm` and `aten::instance_norm` are built on TensorRT plugins that TensorRT-RTX does not provide, and are already compiled out for RTX. Their gtests therefore fail with `Expected converter to be true but got false`. Guarded with `#ifndef TRT_MAJOR_RTX`, which is already visible via `tests/util/util.h` → `core/ir/ir.h` → `NvInfer.h` (no new includes or build deps). Non-RTX builds keep full coverage.

- `test_pooling.cpp` — 14 adaptive tests guarded; max/avg pooling left enabled
- `test_normalize.cpp` — 6 `aten::norm` tests guarded; `frobenius_norm` / `linalg_norm` left enabled
- `test_instance_norm.cpp` — all 3 guarded (they share one `aten::instance_norm` graph)

Verified on TensorRT-RTX 1.6.1.120: `//tests/core/conversion:conversion_tests` passes 49/49 targets, including `ATenConvTransposeWithGroupConvertsCorrectly`, which covers cause 1 and is deliberately left unguarded.

## Type of change

# Checklist:

### DIFF
--- a/core/conversion/converters/impl/conv_deconv.cpp
+++ b/core/conversion/converters/impl/conv_deconv.cpp
@@ -294,7 +294,9 @@ bool add_conv_deconv(ConversionCtx* ctx, const torch::jit::Node* n, args& args) 
     deconv->setStrideNd(stride);
     deconv->setPrePadding(begPadding);
     deconv->setPostPadding(padding);
-#if NV_TENSORRT_MAJOR > 7 || (NV_TENSORRT_MAJOR == 7 && NV_TENSORRT_MINOR >= 1)
+// TensorRT-RTX reports NV_TENSORRT_MAJOR == 1, so the plain version comparison below would
+// select the pre-7.1 fallback and reject grouped or dilated deconvolutions that RTX supports.
+#if defined(TRT_MAJOR_RTX) || NV_TENSORRT_MAJOR > 7 || (NV_TENSORRT_MAJOR == 7 && NV_TENSORRT_MINOR >= 1)
     deconv->setDilationNd(dilation);
     deconv->setNbGroups(groups);
 #else
--- a/tests/core/conversion/converters/test_instance_norm.cpp
+++ b/tests/core/conversion/converters/test_instance_norm.cpp
@@ -28,6 +28,13 @@ constexpr auto graph = R"IR(
         return (%4)
 )IR";
 
+// The aten::instance_norm converter is implemented with a TensorRT plugin, which TensorRT-RTX
+// does not provide, so aten::instance_norm is left unregistered for RTX builds (see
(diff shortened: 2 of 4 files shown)


lanluo-nvidia (approved):
lgtm
let's wait for the ci to verify the test before merge in.

tp5uiuc:
[by Claude Code] **CI failures here are pre-existing and unrelated to this change.** Four distinct causes, none reachable from this diff — all reproduce identically on the `standard` variant, where the change is a no-op (`TRT_MAJOR_RTX` is undefined there, so the `#if` condition is bit-identical).

| Cause | Jobs |
|---|---|
| `flashinfer-python` build dep `apache-tvm-ffi==0.1.0b15` no longer on PyPI → env setup aborts → `ModuleNotFoundError: torch_tensorrt` | all Windows jobs; `dynamo-lowering`, `ts-models` on Linux |
| `test_aliased_io.py::TestPythonRuntimeAliasedIO` ×2 → `TypeError: set_runtime_cache(): incompatible function arguments` | `python-only-runtime` (rtx **and** standard) |
| `test_libtorchtrt_linkage.py` → `libnvinfer_plugin.so.11: cannot open shared object file` | `py-core` (rtx **and** standard) |
| `test_bert_base_uncased[dtype1]` → `BackendCompilerFailed: ValueError: __len__() should return >= 0` (torch 2.14 nightly regression) | `torch-compile-models` (rtx **and** standard) |

On Linux, 23 test jobs pass, including `dynamo-converters-rtx` and `dynamo-runtime-rtx`.

Verified locally on TensorRT-RTX 1.6.1.120: `//tests/core/conversion:conversion_tests` passes 49/49 targets, including `ATenConvTransposeWithGroupConvertsCorrectly`, which covers the deconv fix and is deliberately left unguarded.

PR #4468, opened 2026-08-08
fix: stop calling TensorRT 10.15 aliasing APIs unconditionally

shoumikhin:
## The problem

Three things Torch-TensorRT calls without checking were added in TensorRT 10.15:

- `ICudaEngine::getAliasedInputTensor` (C++)
- `ICudaEngine.get_aliased_input_tensor` (Python)
- the KV-cache update layer, with its `KVCacheMode` enum

Building against an older TensorRT fails outright:

```
core/runtime/TRTEngine.cpp: error: 'class nvinfer1::ICudaEngine'
    has no member named 'getAliasedInputTensor'
```

Once that is past, compiling any model fails, not only one with a cache:

```
File "py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py", line 519, in run
    aliased_in = cuda_engine.get_aliased_input_tensor(out_name)
AttributeError: 'ICudaEngine' object has no attribute 'get_aliased_input_tensor'
```

That loop runs after every engine build, over every output, so an ordinary model with no
aliasing at all hits it. The same call on the Python runtime fails the same way at engine
load.

## The fix

Reconciliation compares the build-time alias map against what an engine reports about
itself. A build without the API cannot report anything, so there is nothing to compare
and the build-time map already stands alone. Skip only that loop, and keep everything
after it, because user-declared aliases exist independently of the TensorRT API and their
binding names still need recomputing.

Writing through an alias is a different matter, because it cannot be emulated. Where the
code promises a write-through, it now refuses rather than quietly dropping it. Two paths
promise one:

- A cache passed in as a network input and written in place. Falling back to the
  functional scatter computed the correct return value and never touched the caller's
  tensor. That is worse than an error, because a test that only checks numbers passes
  while the cache silently stops updating.
- A mutated module buffer. Lifting removes the writeback copy, on the assumption the
  engine writes through the aliased input.

Both now raise, naming the input or the buffer and the version required:

```
Writing in place to the network input 'cache' needs the TensorRT KV-cache update
layer, which was added in TensorRT 10.15. This build does not have it, and without
it the write would be computed and then discarded. Use TensorRT 10.15 or newer, or
return the updated tensor instead of writing into an input.
```

One shared capability check backs both sites, so they cannot drift apart.

## Scope

This does not make Torch-TensorRT work on TensorRT older than 10.15. Other 10.15-only
symbols are still referenced unconditionally, including `trt.IAttention` in an annotation
that is evaluated at import time, so importing the package on 10.13 fails before any of
this runs. What this change does is take these three APIs off that list.

## Testing

```
ordinary model, aliasing API removed        compiles, matches eager
module buffer write, KV layer removed       raises, names the buffer
network input write, KV layer removed       raises, names the input
unmodified TensorRT 11.1.0.106              unchanged, matches eager
tests/py/dynamo/runtime/test_aliased_io.py  13 passed
black --check on all five changed files     clean
```

The three added tests fail without this change and pass with it. Also confirmed on an
aarch64 device with system TensorRT 10.13.3.9 that all three APIs are absent there, and
that the C++ version check evaluates false on it.

### DIFF
--- a/core/runtime/TRTEngine.cpp
+++ b/core/runtime/TRTEngine.cpp
@@ -292,6 +292,10 @@ TRTEngine::TRTEngine(
   // engines built outside Torch-TensorRT). User-declared aliases (kind=kUser)
   // are preserved as-is since TRT doesn't know about them.
   this->aliased_io = aliased_io;
+// ICudaEngine::getAliasedInputTensor is not available before TRT 10.15 (e.g. Jetpack
+// L4T builds), where an engine cannot report its aliasing and the build-time map is
+// the only source of truth.
+#if NV_TENSORRT_MAJOR > 10 || (NV_TENSORRT_MAJOR == 10 && NV_TENSORRT_MINOR >= 15)
   for (const auto& out_name : this->out_binding_names) {
     // TRT returns nullptr / empty string for non-aliased outputs; any thrown
     // exception is a real error in the engine state and propagates.
@@ -315,6 +319,7 @@ TRTEngine::TRTEngine(
       it->second = AliasedIOSpec{std::string(aliased_in), AliasKind::kKVCacheUpdate};
     }
   }
+#endif
 
   // Precompute the set of input binding names that are the alias source of some
   // output (for the O(1) per-call membership test) and validate every aliased
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -9,6 +9,7 @@ import warnings
 from typing import Any, Collection, Dict, List, Optional, Sequence, Tuple, Union
 
 import sympy
+import tensorrt as trt
 import torch
 from torch.export import ExportedProgram
 from torch.export.graph_signature import InputKind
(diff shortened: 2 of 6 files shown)


narendasan:
@shoumikhin What Jetpack version are you using?

shoumikhin:
JetPack 7.1 (L4T R38.4, CUDA 13.0, aarch64), whose apt channel ships TensorRT 10.13.3.9, where `getAliasedInputTensor` is absent from the installed headers and `trt.KVCacheMode` / `add_kv_cache_update` are missing from the Python module.

lanluo-nvidia:
@shoumikhin @narendasan 
We are no longer building jetpack with 10.13, now we only use the sbsa wheel in jetson which is tensorrt 11.1
Can I think this PR is no longer required?

shoumikhin:
Thanks, that settles it. Closing.

Confirmed on a Jetson here: installing `torch-tensorrt==2.13.0` the normal way pulls
`torch 2.13.0+cu130` and `tensorrt 11.0.0.114`, and all three APIs this PR guards are
present, so the guarded branches never execute:

```
tensorrt: 11.0.0.114
  get_aliased_input_tensor present: True
  KVCacheMode present: True
  add_kv_cache_update present: True
```

A model compiles and runs on that stack, so there is no version problem left to fix for
a supported configuration.

Worth noting for anyone who finds this later: guarding these three alone was never
sufficient on an older TensorRT anyway. `trt.IAttention` is used in a type annotation
that is evaluated at import time, so `import torch_tensorrt` fails before any of this
code runs.

One unrelated leftover: `get_jetpack_requirements` in `setup.py` still pins
`tensorrt>=10.3.0,<10.4.0`. If JetPack-specific builds are retired, that pin looks stale.

PR #4472, opened 2026-08-09
fix: lift mutated buffers owned by a submodule

shoumikhin:
## The problem

A mutated buffer that belongs to a submodule is never lifted. The rewrite skips it with a
warning and carries on:

```
lift_mutated_buffers: get_attr target inner.cache not found on gm; skipping
```

The buffer is there. The lookup is what fails. A `get_attr` target is fully qualified, so a
buffer owned by a submodule arrives as `inner.cache`, and neither `hasattr` nor `getattr`
walks a dotted path:

```python
hasattr(module, "inner.cache")     # False
module.get_buffer("inner.cache")   # the tensor
```

So only a buffer owned directly by the top-level module resolves. Any model that keeps state
inside its layers, which is the usual way to write one, is skipped.

## The fix

Use `get_buffer`, which resolves through the submodules:

```python
        try:
            buffer_tensor = gm.get_buffer(buffer_name)
        except AttributeError:
            logger.warning(
                "lift_mutated_buffers: get_attr target %s not found on gm; skipping",
                buffer_name,
            )
            continue
```

The warning is kept for a target that genuinely does not exist, so a real problem is still
reported. Nothing changes for a top-level buffer.

## Testing

Added `test_nested_buffer_lifted`, which is the same shape as the existing
`test_single_buffer_lifted` but with the buffer one level down. It asserts the buffer is
lifted, the qualified name is preserved as `inner.cache`, the flattened placeholder is
`buf_inner_cache`, the rebuilt `forward` accepts it, and no trailing `copy_` remains.

The test fails on the current code with exactly the warning above, and passes with the fix:

```
without the fix:  1 failed
with the fix:     8 passed
```

Worth noting why the existing tests do not catch this: all of them use a buffer owned by
the top-level module, so the name has no dots and `getattr` happens to work.

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_buffer_lifting.py
+++ b/py/torch_tensorrt/dynamo/lowering/_buffer_lifting.py
@@ -84,13 +84,18 @@ def lift_mutated_buffers(
 
     for copy_node, get_attr_node in mutation_pairs:
         buffer_name = get_attr_node.target
-        if not hasattr(gm, buffer_name):
+        # A get_attr target is fully qualified, so a buffer owned by a submodule
+        # arrives as "layers.0.self_attn.kv_cache.k_cache". getattr does not walk a
+        # dotted path, so it reports every nested buffer as missing; get_buffer
+        # resolves through the submodules.
+        try:
+            buffer_tensor = gm.get_buffer(buffer_name)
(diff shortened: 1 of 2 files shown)


lanluo-nvidia (approved):
lgtm

Issue #2197, opened 2023-08-14
🐛 [Bug] Encountered bug when using Torch-TensorRT with torchscript model Conformer Transducer


kzelias:
##  Bug Description

I get an error when converting a conformer transducer enecoder to tensorrt. (asr task)

## To Reproduce

[requirenments.txt](https://github.com/pytorch/TensorRT/files/12343005/requirenments.txt)

CODE:
```
import nemo.collections.asr as nemo_asr
import torch
import torch_tensorrt as torchtrt


nemo_model = nemo_asr.models.EncDecRNNTBPEModel.from_pretrained(model_name="stt_en_conformer_transducer_large")
nemo_model.freeze()
nemo_model.export(output="temp_rnnt.ts", check_trace=True)


with torchtrt.logging.debug():
    variant = "encoder-temp_rnnt.ts"
    precisions = [torch.float, torch.half]
    batch_size = 1

    model = torch.jit.load(variant)

    inputs = [
            torchtrt.Input(shape=[batch_size, 80, 8269]), # 8269 from mel spectr for 1min wav with resample
            torchtrt.Input(shape=[1]),
        ]

    for precision in precisions:
        compile_settings = {
            "inputs": inputs, 
            "enabled_precisions": {precision},
            "workspace_size": 2000000000,
            "truncate_long_and_double": True,
        }
        print(f"Generating Torchscript-TensorRT module for batchsize {batch_size} precision {precision}")
        trt_ts_module = torchtrt.compile(model, **compile_settings)
        torch.jit.save(trt_ts_module, f"{variant.replace('.ts','')}_bs{batch_size}_{precision}.torch-tensorrt")
```

CONSOLE:
```
Generating Torchscript-TensorRT module for batchsize 1 precision torch.float32
WARNING: [Torch-TensorRT] - Data types for input tensors have been modified by inserting aten::to operations which cast INT64 inputs to INT32. To disable this, please recompile using INT32 inputs
WARNING: [Torch-TensorRT] - Truncating intermediate graph input type from at::kLong to at::kInt
WARNING: [Torch-TensorRT] - Truncating intermediate graph input type from at::kLong to at::kInt
WARNING: [Torch-TensorRT] - Truncating intermediate graph input type from at::kLong to at::kInt
WARNING: [Torch-TensorRT] - Truncating intermediate graph input type from at::kLong to at::kInt
WARNING: [Torch-TensorRT] - Truncating intermediate graph input type from at::kLong to at::kInt
[...]

PR #4475, opened 2026-08-11
feat: add aten.glu converter

### LINKED ISSUE #2663: ↔ [Converter] Add support for `torch.ops.aten.glu.default` in Torch-TensorRT
Add support for `torch.ops.aten.glu.default`, documented [here](https://pytorch.org/docs/stable/generated/torch.nn.functional.glu.html) and [here](https://github.com/pytorch/pytorch/blob/f3dd2a544c787aa11bcb84837a50195864f73cc9/aten/src/ATen/native/native_functions.yaml#L11564).

See #2197 for usecase.

zupengwang:
# Description

Adds Dynamo converter support for `torch.ops.aten.glu.default`. The converter implements GLU as an equal split followed by sigmoid and elementwise multiplication, and conservatively requires the split dimension to be statically known and even while allowing other dimensions to remain dynamic.

This avoids a PyTorch fallback and extra TensorRT engine partitions for models that use GLU, including the Conformer use case described in #2197. No additional dependencies are required.

Validation performed locally:

- Converter tests cover FP32, FP16, `dim=0/1/-1`, rank 2/3, and a dynamic batch range.
- GLU, split, and sigmoid converter tests: 17 passed.
- Full-compilation `torch.export` checks on two RTX 3090 GPUs produced a single TensorRT module with no PyTorch fallback for FP32, FP16, and dynamic shapes; outputs matched eager PyTorch exactly in the tested cases.
- Black, Ruff, isort, mypy, and `git diff --check` passed for the changed files.

The full repository test suite and GitHub CI have not been run. The dynamic GLU split dimension remains unsupported; invalid odd split dimensions retain PyTorch's existing error behavior.

Fixes #2663

## Type of change

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas and hacks (no non-obvious hacks were added)
- [x] I have made corresponding changes to the documentation (not required; supported-op documentation is generated from the converter registry)

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -488,6 +488,54 @@ def aten_ops_hardtanh(
     )
 
 
+def glu_validator(node: Node, settings: Optional[CompilationSettings] = None) -> bool:
+    input_meta = node.args[0].meta
+    input_val = input_meta.get("tensor_meta")
+    if input_val is None:
+        input_val = input_meta.get("val")
+    if input_val is None:
+        _LOGGER.warning(
+            "Meta information of input is missing. Unable to validate GLU's "
+            "split dimension, falling back to PyTorch operation."
+        )
+        return False
+
+    input_shape = input_val.shape
+    dim = get_positive_dim(args_bounds_check(node.args, 1, -1), len(input_shape))
+    split_dim_size = input_shape[dim]
+
+    return isinstance(split_dim_size, int) and split_dim_size % 2 == 0
+
+
+@dynamo_tensorrt_converter(
+    torch.ops.aten.glu.default,
+    capability_validator=glu_validator,
+    supports_dynamic_shapes=True,
+)
+@enforce_tensor_types(
+    {
+        0: (TRTTensor,),
+    }
+)
+def aten_ops_glu(
+    ctx: ConversionContext,
+    target: Target,
+    args: Tuple[Argument, ...],
+    kwargs: Dict[str, Argument],
+    name: str,
+) -> Union[TRTTensor, Sequence[TRTTensor]]:
+    input_val = args[0]
+    dim = get_positive_dim(args_bounds_check(args, 1, -1), len(input_val.shape))
+    split_size = input_val.shape[dim] // 2
+    first, second = impl.split.split(
+        ctx, target, SourceIR.ATEN, f"{name}_split", input_val, split_size, dim
+    )
+    gated = impl.activation.sigmoid(
+        ctx, target, SourceIR.ATEN, f"{name}_sigmoid", second
+    )
+    return impl.elementwise.mul(ctx, target, SourceIR.ATEN, f"{name}_mul", first, gated)
+
+
 @dynamo_tensorrt_converter(torch.ops.aten.sigmoid.default, supports_dynamic_shapes=True)
 def aten_ops_sigmoid(
     ctx: ConversionContext,
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2,7 +2,18 @@
 
 import logging
 import operator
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -488,6 +488,54 @@ def aten_ops_hardtanh(
...
+            "split dimension, falling back to PyTorch operation."
+        )
+        return False
+
+    input_shape = input_val.shape
+    dim = get_positive_dim(args_bounds_check(node.args, 1, -1), len(input_shape))
+    split_dim_size = input_shape[dim]
+
+    return isinstance(split_dim_size, int) and split_dim_size % 2 == 0
+
```
apbose, line 527: dim passed as a keyword is silently ignored. Both glu_validator and aten_ops_glu read only node.args[1] / args[1] via args_bounds_check. If the graph carries dim as a kwarg instead  torch.fx.symbolic_trace on torch.ops.aten.glu.default(x, dim=0) produces args=(x,), kwargs={'dim': 0} both validator and converter silently fall back to dim=-1, producing a wrong split axis instead of erroring

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -488,6 +488,54 @@ def aten_ops_hardtanh(
...
+
```
apbose, line 528: Zero-length split dimension passes validation, then crashes. glu_validator only checks split_dim_size % 2 == 0; 0 % 2 == 0 is True, so a zero-sized split axis passes. The converter then computes split_size = 0, and impl.split.split's num_splits = (dim_size + split_size - 1) // split_size divides by zero, crashing engine build

tests/py/dynamo/conversion/test_glu_aten.py
```diff
@@ -0,0 +1,43 @@
...
+        input_specs = [
+            Input(
+                min_shape=(2, 4, 8),
+                opt_shape=(3, 4, 8),
+                max_shape=(5, 4, 8),
+                dtype=torch.float32,
+            ),
+        ]
+        self.run_test_with_dynamic_shape(Glu(), input_specs, use_dynamo_tracer=True)
+
```
apbose, line 67: please add test cases for above, like dim passed as keyword, omitted dim, zero sized split axis, test exercising validator reject path

lanluo-nvidia:
@zupengwang  
could you please take a look on the review comments if you want this PR to go in the 2.14 release.

zupengwang:
Thanks for the reminder. Update: the review comments have been addressed and pushed in f0dae4231. GPU validation is now complete: the GLU converter suite passes 12 tests, and the neighboring split/sigmoid regression suites pass 13 tests on an RTX 3090.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zupengwang: Addressed in f0dae4231. Both glu_validator and aten_ops_glu now use the shared _get_glu_dim helper, which handles positional, keyword, and omitted/default dim consistently.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
zupengwang: Addressed in f0dae4231. The validator now requires split_dim_size > 0 in addition to being a statically known even integer, so a zero-sized split axis is rejected before impl.split.split.

tests/py/dynamo/conversion/test_glu_aten.py
zupengwang: Added focused coverage in f0dae4231 for keyword dim, omitted/default dim, zero-sized and odd split dimensions, missing metadata, and the validator rejection path. The GPU-backed GLU suite passes 12 tests, and neighboring split/sigmoid tests pass 13 tests on an RTX 3090.

apbose:
LGTM

apbose (approved):

PR #4476, opened 2026-08-11
feat: support Global Performance Tuning

zewenli98:
# Description

TensorRT's [Global Performance Tuner](https://docs.nvidia.com/deeplearning/tensorrt/latest/performance/tuning.html) searches internal builder knobs for faster engines. Torch-TensorRT exposes the same capability in-process for Dynamo TRT subgraphs. Please see the [description](https://github.com/pytorch/TensorRT/compare/evanli/feat_gpt?expand=1#diff-405beebcdf7216f8025b0f3e0cce0114b3920bf9a2bf68dcfec79ac6143801b5R342) for details. 

Fixes # (issue)

## Type of change

# Checklist:


py/torch_tensorrt/dynamo/tuning/sweeper.py
```diff
@@ -0,0 +1,365 @@
...
+                and not error_message
+                and (best_time is None or gpu_time < best_time)
+            ):
+                best_time = gpu_time
+                best_result = result
+                best_route = route
+            del trt_mod
+            torch.cuda.empty_cache()
+            _LOGGER.info(f"&&&& TASK_END [iter={iter_idx}] BuildRoute = '{route}'")
+        except Exception as exc:
```
jianhu-nv, line 286: Actually we cannot guarantee the TensorRT will not crash when setting a build route, for example, some driver level exception might be raised and it will not be caught by the try-catch block, and probably will crash the entire Python program. So I suggest we spawn a new process like the implementation in trtexec, so that even if the new child process crashes, the parent process remains unaffected and can continue to the next iteration.

examples/dynamo/global_perf_tuning_attention_example.py
```diff
@@ -0,0 +1,160 @@
...
+    #     if name.startswith(cache_stem) and name.endswith(".jsonl"):
+    #         os.remove(os.path.join(cache_dir, name))
+    print(f"tuning_cache_file base={cache_base}")
+
+    optimized = torch_tensorrt.compile(
+        model,
+        ir="dynamo",
+        arg_inputs=example_inputs,
+        min_block_size=1,
+        tune_build_routes=tune_expr,  # tuning_continue=True,
```
jianhu-nv, line 91: "tuning_continue" Is this a deleted argument?

jianhu-nv (approved):

examples/dynamo/global_perf_tuning_attention_example.py
zewenli98: No, this is just an alternative so that users can replace `tune_build_routes` with `tuning_continue` to continue tuning instead of tuning from scratch. This file is just an illustrative example.

PR #4482, opened 2026-08-13
fix: engines default to no runtime cache; the module owns the implicit one

tp5uiuc:
# Description

## What

An engine built without a `TorchTensorRTModule` came up with a path-string
`runtime_cache` and attached a cache nothing owned, freeing it out from under
the engine. Engines now default to no cache; the module owns the implicit one.

## Why

`RuntimeSettings.runtime_cache` defaults to a path string, and
`TorchTensorRTModule` resolves it into a `RuntimeCache` it owns, so the compile
path is fine. An engine reached without a module -- built from packed engine
info, or loaded as a graph constant from a saved `ExportedProgram` -- never
passes through that resolver, so `TRTRuntimeConfig` wrapped the string in a
local, handed the live `IRuntimeCache` to the engine's `IRuntimeConfig`, and let
it be collected on return. The next `createExecutionContext` read freed memory.

The C++ runtime has no such path: `RuntimeSettings::runtime_cache` is an
`intrusive_ptr` defaulting to `nullptr`, with no string form to mis-own.

The module's post-load resets have to move with the engine, or the two disagree:
a module holding a stale path string lets a later `runtime_config(...)` block --
a call that need not mention caching -- resolve it and leave an autosaving handle
installed at the shared default path on exit.

## How

- `_TRTEngine.py`: engines construct `TRTRuntimeConfig` with
  `runtime_cache=None`, matching the C++ default. `RuntimeSettings()`'s own
  default is unchanged -- that one belongs to the module, which pushes the
  resolved handle down in `setup_engine`.
- `_TorchTensorRTModule.py`: the `set_extra_state` / `__setstate__` resets drop
  to `runtime_cache=None` so the module agrees with the engine it rebuilds; an
  empty-string `runtime_cache` normalizes to `None` so no string reaches an
  engine.
- `_runtime_config.py`: `_apply_settings` accepts only `None` or a
  `RuntimeCache` and raises `TypeError` on a string; `TRTRuntimeConfig`'s own
  no-settings default follows suit. The docstring already claimed raw strings
  were not accepted here.
- Tests: module-less engines default to no cache, execute without one, accept an
  explicitly attached `RuntimeCache`, and reject strings; the compile path still
  builds and persists its implicit cache; the post-load resets leave no cache,
  and a cuda-graph-only context manager over a loaded module installs none on
  enter and leaves none on exit.

## Testing

TensorRT-RTX on A100, `tests/py/dynamo/runtime/` plus
`models/test_runtime_cache_models.py`:

| Build | before | after |
| --- | --- | --- |
| Python runtime | 4 failed, 170 passed, 60 skipped | 174 passed, 60 skipped |
| C++ runtime | 4 failed, 171 passed, 59 skipped | 175 passed, 59 skipped |

The four failures before are the new tests themselves; each was confirmed
failing on the parent commit and passing here, on both builds. There are no
other failures on either build.

`models/test_export_serde.py`, `test_cross_runtime_serde.py` and
`test_fallback_data_dependent_ops.py` were run before and after on both builds
because the post-load resets sit on the `torch.load` / `load_state_dict` paths:
identical results either side, with the only failures (`test_save_load_aoti`,
`test_save_load_ts`) pre-existing and unrelated.

Save/load deployment was checked on both runtimes: the loaded engine runs with
no cache attached, and an explicitly attached `RuntimeCache` persists.

## Cost / Gotchas

An engine used without a module now gets no runtime cache instead of an implicit
one, and neither do modules restored by `torch.load` / `load_state_dict`. No
configuration loses working behaviour: that path raised on the Python runtime,
already attached nothing on the C++ runtime, and never initializes a runtime
config on standard TensorRT. Callers wanting a cache there attach a
`RuntimeCache` explicitly.

Supersedes #4541 -- the `TestPythonRuntimeAliasedIO` failures it skipped are
fixed here rather than skipped.

## Followups

`runtime_config()` / `runtime_cache()` locate engines via `named_modules()`, so
neither reaches an engine loaded as a graph constant. Attaching settings to a
saved-and-loaded program currently requires the engine API directly.

Restoring implicit caching to `load_state_dict` -- by re-applying the setter
after engine construction in `set_extra_state`, as `setup_engine` does -- is
deliberately left out; it changes C++-runtime behaviour too and wants its own
testing.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/runtime/_runtime_config.py
+++ b/py/torch_tensorrt/runtime/_runtime_config.py
@@ -186,7 +186,13 @@ class TRTRuntimeConfig:
         if self._live is not None:
             return
         self._live = cuda_engine.create_runtime_config()
-        self._apply_settings()
+        try:
+            self._apply_settings()
+        except Exception:
+            # Reset the live config so a later ensure_initialized rebuilds it
+            # instead of short-circuiting on the guard above.
+            self._live = None
+            raise
 
     def reset(self) -> None:
         """Drop the live ``IRuntimeConfig``; the next ``ensure_initialized`` rebuilds."""
--- a/py/torch_tensorrt/runtime/_runtime_config.py
+++ b/py/torch_tensorrt/runtime/_runtime_config.py
@@ -68,13 +68,13 @@ class RuntimeSettings:
             TRT-RTX-only; no-op on standard TensorRT.
         cuda_graph_strategy: ``"disabled" | "whole_graph_capture"``. TRT-RTX-only.
         runtime_cache: ``None``, a disk path string, or a
-            :class:`RuntimeCache`. ``None`` ⇒ no cache attached. A
-            string is honored at engine construction time and primes a
-            per-engine disk-backed cache (engine owns the implicit handle and
-            it saves on ``__del__``). A handle is the shared-cache form,
-            typically obtained from :func:`torch_tensorrt.runtime.runtime_cache`
-            -- multiple engines attaching the same handle share one
-            ``IRuntimeCache``.
+            :class:`RuntimeCache`. ``None`` ⇒ no cache attached. A string is
+            resolved by :class:`TorchTensorRTModule`, which builds the
+            engine-implicit :class:`RuntimeCache`, owns it, and saves it on
+            ``__del__``; engines themselves only ever receive ``None`` or a
+            handle. A handle is the shared-cache form, typically obtained from
+            :func:`torch_tensorrt.runtime.runtime_cache` -- multiple engines
+            attaching the same handle share one ``IRuntimeCache``.
 
     Equality compares all fields; for ``runtime_cache``, handle equality is
     by identity (same handle ⇒ same cache).
@@ -149,7 +149,9 @@ class TRTRuntimeConfig:
     """
 
     def __init__(self, settings: Optional[RuntimeSettings] = None) -> None:
-        self._settings: RuntimeSettings = settings or RuntimeSettings()
+        self._settings: RuntimeSettings = settings or RuntimeSettings(
+            runtime_cache=None
+        )
         # Live trt.IRuntimeConfig (RTX) or None (non-RTX / pre-init).
         self._live: Any = None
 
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/runtime/_runtime_config.py
```diff
@@ -275,36 +280,15 @@ def _apply_settings(self) -> None:
...
-            wrapped = RuntimeCache(path=rc, autosave_on_del=True)
-            try:
-                wrapped.load()
-            except Exception as e:
-                logger.warning(f"Failed to warm-load runtime cache from {rc!r}: {e}")
-            cache = wrapped.ensure_cache(self._live)
-            self._live.set_runtime_cache(cache)
+            self._live.set_runtime_cache(rc.ensure_cache(self._live))
         else:
             raise TypeError(
```
cehongwang, line 293: We probably want to destroy the self._live here because the initialization was not successful?

cehongwang (approved):
Overall OK

py/torch_tensorrt/runtime/_runtime_config.py
tp5uiuc: Good catch, done in https://github.com/pytorch/TensorRT/pull/4482/changes/cb95af3caa765ac93605a442493a5bbac7456c55 by resetting the live and propagating the exception up the call stack. Thanks Adrian!

tp5uiuc:
[by Claude Code] CI failures on this PR are pre-existing, not introduced here.

| Failure | Lanes | Why it isn't from this PR |
| --- | --- | --- |
| `test_bitwise_{and,or}_aten` (4) | standard-converters | Same 4 fail on #4601 (unrelated) |
| `test_aten_lowering_passes::TestComplexSubgraph` | rtx + standard lowering | Same on #4601 |
| `test_decompositions::test_lowering_baddbmm` | rtx-lowering | Decomposition path, untouched here |
| `test_model_refit::*complex_buffer*` (3) | rtx + standard models | Refit's `load_state_dict` calls are guarded to non-TRT submodules (`_refit.py:353,363`), so they never reach `set_extra_state` |
| `test_export_serde::test_save_load_aoti` | rtx + standard models | Measured identical before/after locally, both runtimes |
| `ts-api::test_fallback_mobilenet_v2` | ts-api | TorchScript path, untouched |
| `gate` (2) | — | Roll-up only: "a Windows channel failed" |

The base commit (`57908060a1`) has 89 failing checks of its own, so this branch
inherits a red base.

Verified locally on TensorRT-RTX/A100 — `tests/py/dynamo/runtime/` plus
`models/test_runtime_cache_models.py`: **175 passed, 60 skipped, 0 failed** on both
the Python and C++ runtimes. The serde suites (`test_export_serde`,
`test_cross_runtime_serde`, `test_fallback_data_dependent_ops`) were run before and
after on both runtimes, since this touches the `torch.load` / `load_state_dict`
paths — identical results either side.

lanluo-nvidia:
@tp5uiuc 
whether this fix will fix the following error in ci: 
https://github.com/pytorch/TensorRT/actions/runs/33145264851/job/98790133431

Issue #3712, opened 2025-07-24
🐛 [Bug] Incorrect Type Casting in MatMul Operator


keehyuna:
##  Bug Description

The matmul operator implementation includes unnecessary type casting for inputs when use_explicit_typing=True, which is not the correct solution for handling dtype mismatches in PyTorch's torch.matmul operation.

https://github.com/pytorch/TensorRT/blame/dad195b92ecedd21ca8785f9643472c56b96b502/py/torch_tensorrt/dynamo/conversion/impl/matmul.py#L51](https://github.com/pytorch/TensorRT/blob/main/py/torch_tensorrt/dynamo/conversion/impl/matmul.py#L51-L60)
Parameters for torch.matmul accepts only same dtype. 
This approach masks the underlying issue rather than addressing it. we need right fix for root cause.
```
>>> a = torch.randn(3, dtype=torch.float32)
>>> b = torch.randn(3, dtype=torch.float16)
>>> torch.matmul(a,b )
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: dot : expected both vectors to have same dtype, but found Float and Half
```
## To Reproduce

```
import torch
import torch_tensorrt as torchtrt
from diffusers import DiffusionPipeline

pipe = DiffusionPipeline.from_pretrained("CompVis/stable-diffusion-v1-4", revision="fp16", torch_dtype=torch.float16)

model = pipe.unet
model_torch = model.eval()
model_torch = model_torch.half().to('cuda')
input_tensors = [
    torch.randn((1, 4, 64, 64), dtype=torch.float16).cuda(),
    torch.Tensor([1.0]).cuda(),
    torch.randn((1, 1, 768), dtype=torch.float16).cuda(),
]

model_trt = torchtrt.compile(
    model_torch,
    inputs=input_tensors,
    ir="dynamo",
    cache_built_engines=False,
    reuse_cached_engines=False,
    min_block_size=1,
    debug=False,
    use_python_runtime=True,
    use_explicit_typing=True,
    truncate_long_and_double=True,
)
```

## Environment

> Build information about Torch-TensorRT can be found by turning on debug messages

 - Torch-TensorRT Version (e.g. 1.0.0):
 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source):
 - Build command you used (if compiling from source):
[...]

PR #4483, opened 2026-08-13
add direct converter for aten.repeat instead of decompo…

apbose:
Addresses #3712 and #3974

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1396,6 +1396,58 @@ def aten_ops_tile(
     )
 
 
+def repeat_validator(
+    node: Node, settings: Optional[CompilationSettings] = None
+) -> bool:
+    # aten.repeat may raise the output rank above the input rank: if `repeats`
+    # has more entries than the input has dims, the input is treated as if
+    # size-1 dims were prepended to it first. TensorRT tensors support at
+    # most 8 dims, so reject cases that would exceed that limit and let them
+    # fall back to PyTorch instead of failing at engine build time.
+    repeats = node.args[1]
+    if "val" not in node.args[0].meta:
+        return True
+
+    input_rank = len(node.args[0].meta["val"].shape)
+    output_rank = max(len(repeats), input_rank)
+    if output_rank > 8:
+        _LOGGER.debug(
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1396,6 +1396,58 @@ def aten_ops_tile(
...
+    # has more entries than the input has dims, the input is treated as if
+    # size-1 dims were prepended to it first. TensorRT tensors support at
+    # most 8 dims, so reject cases that would exceed that limit and let them
+    # fall back to PyTorch instead of failing at engine build time.
+    repeats = node.args[1]
+    if "val" not in node.args[0].meta:
+        return True
+
+    input_rank = len(node.args[0].meta["val"].shape)
+    output_rank = max(len(repeats), input_rank)
```
zewenli98, line 1423: if "val" not in node.args[0].meta, shouldn't we only compare whether len(repeats) > 8 instead of return True?

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -1406,6 +1406,17 @@ def repeat_validator(
     # fall back to PyTorch instead of failing at engine build time.
     repeats = node.args[1]
     if "val" not in node.args[0].meta:
+        # Without input_rank we can't compute the exact output_rank, but
+        # output_rank = max(len(repeats), input_rank) is always >= len(repeats),
+        # so len(repeats) alone already proves it exceeds 8 here regardless of
+        # input_rank.
+        if len(repeats) > 8:
```
zewenli98, line 1413: is it possible that e.g., `len(repeats)=3 and input_rank=10`, so `output_rank = max(3, 10) = 10`, which should return False because output_rank > 8 but the existing code will return True?

apbose: from torch side  number of dimensions of repeat dims can not be smaller than number of dimensions of tensor

```
import torch

x = torch.randn(1, 1, 1, 1, 1, 1, 1, 1, 1, 1)  # rank 10
x.repeat(1, 1, 1)                              # len(repeats) = 3

Output:
RuntimeError: Number of dimensions of repeat dims can not be smaller than number of dimensions of tensor
```

zewenli98 (approved):
LGTM

PR #4486, opened 2026-08-13
fix(dynamo): handle dynamic (ITensor) negative start in impl.slice.sl…

### LINKED ISSUE #4186: 🐛 [Bug] slice_op not able to handle start value with type trt.ITensor
##  Bug Description

In py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py, function `slice_op`, Line 112, it only considers the case when `start` is `int`, but not the case when `start` is `trt.ITensor`.

I tried to add support for `start` being `trt.ITensor`, however, I'm not able to run successfully because TRT requires the `start` being a shape tensor to be able to do shape inference but it seems `start` is not a shape tensor. Therefore, currently I'm only able to do a workaround by hard coding all the parameters of the slice layer.

## To Reproduce

1.
2.
3.

## Environment

 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - Python version:
 - CUDA version:

apbose:
Fixes #4186

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
@@ -1,6 +1,6 @@
 import math
 import sys
-from typing import Optional, Sequence
+from typing import Optional, Sequence, Union
 
 import numpy as np
 import tensorrt as trt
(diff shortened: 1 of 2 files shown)


lanluo-nvidia:
codex review says:
  - P1 — the dynamic wrap only works for starts in [-dim_size, -1]. In slice_op (https://github.com/pytorch/TensorRT/pull/4486/files#diff-1be5d4500a5c9cd4a93cc0e1b25f31783a2c5067dcad5969a71f64898cbd4d9fR66), negative dynamic starts are transformed once as start + dim_size.
    Python slicing instead clamps values below -dim_size to zero. The added test uses start = -(x.shape[0] * 2): after the proposed transform this remains -x.shape[0], so it is still invalid for TensorRT and produces an incorrect requested extent. Normalize with a dynamic
    clamp, e.g. where(start < 0, max(start + dim_size, 0), min(start, dim_size)), before calculating the slice shape.

  This also explains the currently failing runtime test suites on the PR.

lanluo-nvidia (approved):
lgtm

PR #4501, opened 2026-08-17
Merged main

### LINKED ISSUE #4466: Dynamo constant folding can leak mutable state through graph outputs
### Bug description

When Torch-TensorRT Dynamo constant folding replaces an input-independent tensor factory result with a persistent folded parameter, the same tensor may also be returned directly as a graph output.

If eager code after a graph break mutates that output in-place, the mutation changes the persistent folded tensor stored on the compiled graph module. Later invocations then observe state left by earlier invocations.

This can cause output contamination across repeated calls or otherwise independent test cases.

### Relevant lowered graph shape

A problematic graph has the equivalent structure:

```
_frozen_param0 = self._frozen_param0
return (_frozen_param0,)
```

The graph output aliases persistent module state rather than returning fresh storage.

### Reproduction outline

1. Compile a function containing an input-independent tensor factory operation.
2. Introduce a graph break after the factory result is produced.
3. Return the folded tensor from the compiled subgraph.
4. Mutate that tensor in-place in eager code after the graph break.
5. Invoke the compiled function again.
6. The second invocation observes the mutation from the first invocation.

The problem is specific to a folded tensor that is also a graph output and is subsequently mutated. A normal eager execution creates fresh tensor storage for each invocation.

### Expected behavior

Each invocation should preserve eager semantics. A folded constant exposed as a graph output must not allow downstream mutation to modify the persistent folded parameter used by later invocations.

### Actual behavior

The graph directly returns the persistent folded parameter, so downstream in-place operations can mutate compiler-owned state and affect later calls.

### Suggested direction

Materialize fresh storage at the graph-output boundary for persistent folded tensors, while preserving the output metadata and the original device semantics.

### Additional context

This report intentionally contains only a generic description and minimized graph shape. No proprietary model code, data, logs, or environment details are included.

micwill755:
# Description

Dynamo constant folding can replace input-independent factories (e.g. `torch.zeros`) with persistent `_frozen_param*` attributes. When those attributes are also graph outputs and eager code after a graph break mutates them in-place, later invocations observe leftover state instead of fresh tensors.

This change clones folded `_frozen_param*` values at the graph-output boundary so downstream in-place mutation cannot modify compiler-owned state, preserving eager semantics across calls.

Fixes #4466

## Type of change

# Checklist:

- [ ] I have added the relevant labels to my PR so that relevant reviewers are notified

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/passes/reset_folded_constructors.py
+++ b/py/torch_tensorrt/dynamo/lowering/passes/reset_folded_constructors.py
@@ -1,0 +1,50 @@
+import logging
+from typing import Dict
+
+import torch
+from torch_tensorrt.dynamo._settings import CompilationSettings
+from torch_tensorrt.dynamo.lowering.passes.pass_utils import (
+    clean_up_graph_after_modifications,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def reset_folded_constructors(
+    gm: torch.fx.GraphModule, settings: CompilationSettings
+) -> torch.fx.GraphModule:
+    """Clone folded constructors that escape through a graph boundary.
+
+    A ``_frozen_param*`` is compiler-owned state, unlike a placeholder supplied
+    by the caller. If a folded constructor becomes an output, eager code or a
+    downstream partition may mutate it. Cloning at that boundary gives each
+    invocation fresh storage while preserving aliases between repeated outputs.
+
+    This pass is intentionally separate from constant folding so it can run
+    again after partitioning, when new TensorRT subgraph outputs are known.
+    """
+    output_node = next(node for node in gm.graph.nodes if node.op == "output")
+    clone_cache: Dict[torch.fx.Node, torch.fx.Node] = {}
+
+    def clone_folded_output(node: torch.fx.Node) -> torch.fx.Node:
+        if node.op != "get_attr" or not str(node.target).startswith("_frozen_param"):
+            return node
+
+        if node not in clone_cache:
+            with gm.graph.inserting_before(output_node):
+                clone = gm.graph.call_function(
+                    torch.ops.aten.clone.default, args=(node,)
+                )
+                clone.meta.update(node.meta)
+            clone_cache[node] = clone
+
+        return clone_cache[node]
+
+    new_output = torch.fx.map_arg(output_node.args[0], clone_folded_output)
+    if not clone_cache:
+        return gm
+
+    output_node.args = (new_output,)
+    gm = clean_up_graph_after_modifications(gm)
+    logger.debug("Reset folded constructors at graph outputs:\n%s", gm.graph)
+    return gm
--- a/py/torch_tensorrt/dynamo/lowering/passes/reset_folded_constructors.py
+++ b/py/torch_tensorrt/dynamo/lowering/passes/reset_folded_constructors.py
@@ -1,5 +1,4 @@
 import logging
-from typing import Dict
 
 import torch
 from torch_tensorrt.dynamo._settings import CompilationSettings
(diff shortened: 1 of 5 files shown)


lanluo-nvidia (commented):
What codex review says:
  - P1 — Preserve aliases for repeated folded outputs. _clone_folded_constants_at_outputs creates a new clone for every output occurrence. If a graph returns the same folded tensor twice (return x, x), the PR changes eager semantics: the two
    results no longer share storage. Cache one clone per get_attr node and reuse it for repeated outputs; add a regression test that mutates one returned value and observes the other.

narendasan:
@micwill755 can you explain a bit more about this? 

> Dynamo constant folding can replace input-independent factories (e.g. torch.zeros) with persistent _frozen_param* attributes. When those attributes are also graph outputs and eager code after a graph break mutates them in-place, later invocations observe leftover state instead of fresh tensors.

micwill755:
> @micwill755 can you explain a bit more about this?
> 
> > Dynamo constant folding can replace input-independent factories (e.g. torch.zeros) with persistent _frozen_param* attributes. When those attributes are also graph outputs and eager code after a graph break mutates them in-place, later invocations observe leftover state instead of fresh tensors.

Constant folding turns torch functions like torch.zeros(...) into a persistent module attribute i.e. _frozen_param0 which is reused across calls. If that attribute is also returned from the graph, eager code after a graph break can mutate it in-place, e.g. x += 1, which edits the stored constant. This means the next call then sees that leftover state instead of a fresh tensor of zeros, unlike normal eager PyTorch, which allocates new storage every time.

narendasan:
What is the case where a constant is returned as output?

micwill755:
> What is the case where a constant is returned as output?

If an input-independent factory op e.g. torch.zeros is folded into a frozen_param and that value is also a graph return, then a constant is returned as an output. A graph break is a common reason this happens because the the factory lives in the compiled subgraph, and eager code after the break can mutate that returned constant in-place, impacting later calls.

cehongwang:
The motivation makes sense. Instead of doing the clone, can we disable the constant folding for the frozen parameter that goes to the output? Also, in AoT compilation, when we are doing constant folding, the graph has not been split yet (the partition happens after the constant folding). After the partition, there could be new subgraphs that add output to the intermediate subgraphs. So we may need to adjust the frozen_parameter after the partitioning.

py/torch_tensorrt/dynamo/lowering/passes/reset_folded_constructors.py
```diff
@@ -0,0 +1,50 @@
...
+    downstream partition may mutate it. Cloning at that boundary gives each
+    invocation fresh storage while preserving aliases between repeated outputs.
+
+    This pass is intentionally separate from constant folding so it can run
+    again after partitioning, when new TensorRT subgraph outputs are known.
+    """
+    output_node = next(node for node in gm.graph.nodes if node.op == "output")
+    clone_cache: Dict[torch.fx.Node, torch.fx.Node] = {}
+
+    def clone_folded_output(node: torch.fx.Node) -> torch.fx.Node:
```
narendasan: @micwill755 I thought we talked about the distinction between module state and function state 

```py
class Mod(nn.Module):
    def __init__(self):
        self.weight = torch.zeros(...)

    def forward(self, x):
         

```

```py
class Mod(nn.Module):
    def __init__(self):


    def forward(self, x):
        self.weight = torch.zeros(...)
```

these two modules would have different semantics wrt mutation between calls right? Is this handled if we just pattern match on get_attr / frozen?

narendasan: Also we talked about if we should clone at construction or clone on return right?

PR #4504, opened 2026-08-18
Fix CPU-only import

jloftin-nv:
# Description

import torch_tensorrt currently fails on hosts without an NVIDIA driver because several CUDA queries run at import / def time with no torch.cuda.is_available() guard.

Guard is_tegra_platform(), default_device(), and Device._current_device() so the package imports cleanly on CPU-only machines. GPU paths are unchanged once a driver is present.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/_Device.py
+++ b/py/torch_tensorrt/_Device.py
@@ -145,7 +145,7 @@ class Device(object):
 
     @classmethod
     def _current_device(cls) -> Device:
-        dev_id = torch.cuda.current_device()
+        dev_id = torch.cuda.current_device() if torch.cuda.is_available() else 0
         return cls(gpu_id=dev_id)
 
     @staticmethod
--- a/py/torch_tensorrt/dynamo/_defaults.py
+++ b/py/torch_tensorrt/dynamo/_defaults.py
@@ -97,4 +97,6 @@ RUNTIME_CACHE_PATH = os.path.join(
 
 
 def default_device() -> Device:
+    if not torch.cuda.is_available():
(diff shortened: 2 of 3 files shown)


py/torch_tensorrt/dynamo/_defaults.py
```diff
@@ -96,4 +96,6 @@
 
 
 def default_device() -> Device:
+    if not torch.cuda.is_available():
+        return Device(gpu_id=0)
```
zewenli98, line 101: why is it `gpu_id=0` instead of `cpu`?

py/torch_tensorrt/_Device.py
```diff
@@ -145,7 +145,7 @@ def _from_torch_device(cls, torch_dev: torch.device) -> Device:
 
     @classmethod
     def _current_device(cls) -> Device:
-        dev_id = torch.cuda.current_device()
+        dev_id = torch.cuda.current_device() if torch.cuda.is_available() else 0
```
zewenli98, line 148: same here

py/torch_tensorrt/dynamo/_defaults.py
jloftin-nv: Device actually has no way to represent cpu. So its Either GPU/DLA. Not sure what to choose?

py/torch_tensorrt/_Device.py
jloftin-nv: Device actually has no way to represent cpu. So its Either GPU/DLA. Not sure what to choose?

PR #4505, opened 2026-08-18
Fix empty inputs=()

jloftin-nv:
# Description

arg_inputs = inputs or arg_inputs treats inputs=() as missing, so a no-arg module becomes arg_inputs=[None] and fails later in input prep with a confusing error.

Use an explicit None check instead of truthiness so empty-tuple inputs stay empty.

## Type of change

# Checklist:

- [X ] I have performed a self-review of my own code

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -310,7 +310,7 @@ def cross_compile_for_windows(
             "'arg_inputs' and 'inputs' should not be used at the same time."
         )
 
-    arg_inputs = inputs or arg_inputs
+    arg_inputs = inputs if inputs is not None else arg_inputs
 
     if kwarg_inputs is None:
         kwarg_inputs = {}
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1917,7 +1917,7 @@ def convert_exported_program_to_serialized_trt_engine(
             "'arg_inputs' and 'inputs' should not be used at the same time."
         )
 
-    arg_inputs = inputs or arg_inputs
+    arg_inputs = inputs if inputs is not None else arg_inputs
```
apbose, line 1946: Can you apply the same arg_inputs = inputs if inputs is not None else arg_inputs fix to all three remaining sites in py/torch_tensorrt/_compile.py , and tighten save()'s if arg_inputs and inputs: conflict check to explicit is not None checks too? something like this
```
  if arg_inputs is not None and inputs is not None:
      raise AssertionError(...)
```

apbose:
LGTM. Waiting on CI

PR #4509, opened 2026-08-18
Fix noncontiguous AOT inputs

jloftin-nv:
# Description

Input.from_tensor rejects non-contiguous / non-channels-last tensors unless disable_memory_format_check=True. The torch.compile backend sets that flag; every dynamo.compile AOT path does not, so the same inputs succeed on one entry point and fail on the other.

Pass disable_memory_format_check=True through the AOT prepare_inputs sites to match the backend behavior.

## Type of change

# Checklist:

- [X ] I have performed a self-review of my own code
- [X ] I have commented my code, particularly in hard-to-understand areas and hacks

### DIFF
--- a/py/torch_tensorrt/_compile.py
+++ b/py/torch_tensorrt/_compile.py
@@ -337,8 +337,12 @@ def compile(
         elif not isinstance(arg_inputs, collections.abc.Sequence):
             arg_inputs = [arg_inputs]
 
-        torchtrt_arg_inputs = prepare_inputs(arg_inputs)
-        torchtrt_kwarg_inputs = prepare_inputs(kwarg_inputs)
+        torchtrt_arg_inputs = prepare_inputs(
+            arg_inputs, disable_memory_format_check=True
+        )
+        torchtrt_kwarg_inputs = prepare_inputs(
+            kwarg_inputs, disable_memory_format_check=True
+        )
 
         if module_type == _ModuleType.ep:
             exp_program = module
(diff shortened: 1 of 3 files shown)


lanluo-nvidia (commented):
- P1 — The top-level Dynamo AOT entry points still reject non-contiguous tensors before reaching these updated functions. py/torch_tensorrt/_compile.py:340,
    py/torch_tensorrt/_compile.py:438, and py/torch_tensorrt/_compile.py:564 call prepare_inputs(...) without disable_memory_format_check=True. Thus
    torch_tensorrt.compile(..., ir="dynamo"), top-level cross-compilation, and convert_method_to_trt_engine still raise ValueError for the same transposed
    inputs, so the change only fixes direct torch_tensorrt.dynamo.* APIs. The test covers only the direct compile path.

jloftin-nv:
> * P1 — The top-level Dynamo AOT entry points still reject non-contiguous tensors before reaching these updated functions. py/torch_tensorrt/_compile.py:340,
>   py/torch_tensorrt/_compile.py:438, and py/torch_tensorrt/_compile.py:564 call prepare_inputs(...) without disable_memory_format_check=True. Thus
>   torch_tensorrt.compile(..., ir="dynamo"), top-level cross-compilation, and convert_method_to_trt_engine still raise ValueError for the same transposed
>   inputs, so the change only fixes direct torch_tensorrt.dynamo.* APIs. The test covers only the direct compile path.

Fixed the other pathways

PR #4511, opened 2026-08-18
Fix cumsum dtype

jloftin-nv:
# Description

The cumsum converter drops the keyword-only dtype argument and always seeds the TRT loop accumulator as float32, so integer / typed cumsums disagree with PyTorch.

Forward dtype from the aten node and use a typed zero for the recurrence seed.

## Type of change

# Checklist:

- [X ] I have commented my code, particularly in hard-to-understand areas and hacks

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
@@ -371,7 +371,8 @@ def cumsum(
     dim: int,
     dtype: Optional[torch.dtype] = None,
 ) -> TRTTensor:
-    # aten widens integer cumsum to int64 unless dtype is set; floats keep their type.
+    # aten.cumsum accumulates bool and integer inputs in int64 and floats in
+    # their own dtype; an explicit dtype wins over both
     input_dtype = _enums.dtype._from(input.dtype).to(torch.dtype)
     if dtype is not None:
         acc_dtype = dtype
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
```diff
@@ -330,7 +333,21 @@ def cumsum(
...
     input: TRTTensor,
     dim: int,
+    dtype: Optional[torch.dtype] = None,
 ) -> TRTTensor:
+    # aten widens integer cumsum to int64 unless dtype is set; floats keep their type.
+    input_dtype = _enums.dtype._from(input.dtype).to(torch.dtype)
+    if dtype is not None:
+        acc_dtype = dtype
+    elif not input_dtype.is_floating_point:
+        acc_dtype = torch.int64
```
cehongwang, line 380: Why this is int64? Can it be int32 or other non-float dtype?

cehongwang:
Can you also add a bf 16 test case?

py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
jloftin-nv: You can see [here](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/ReduceOps.cpp#L261-L266) it defaults to int64 in PyTorch

jloftin-nv:
> Can you also add a bf 16 test case?

Added

PR #4512, opened 2026-08-18
Fix slice_scatter open end

jloftin-nv:
# Description

Export encodes open-ended slice ends as 2**63-1. The converter feeds that into np.arange / broadcast_to without clamping to the dim size, causing “array is too big” or broadcast errors.

Clamp end (and related bounds) to the input length along dim before building NumPy indices, matching aten runtime behavior.

## Type of change

# Checklist:

- [X ] I have made corresponding changes to the documentation

### DIFF
--- a/tests/py/dynamo/conversion/test_slice_scatter_aten.py
+++ b/tests/py/dynamo/conversion/test_slice_scatter_aten.py
@@ -23,14 +23,19 @@ from types import SimpleNamespace
 
 import numpy as np
 import torch
+import torch_tensorrt
 from parameterized import parameterized
-from torch.testing._internal.common_utils import run_tests
+from torch.testing._internal.common_utils import TestCase, run_tests
 from torch_tensorrt import Input
+from torch_tensorrt.dynamo.conversion.aten_ops_converters import slice_scatter_validator
 from torch_tensorrt.dynamo.conversion.impl.slice_scatter import (
     slice_scatter as slice_scatter_impl,
 )
 
 from .harness import DispatchTestCase
+
+# What torch.export writes in place of the dim size for an open-ended slice.
+OPEN_END = 2**63 - 1
 
 
 class _SliceScatterNotInputModule(torch.nn.Module):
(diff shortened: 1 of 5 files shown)


tests/py/dynamo/conversion/test_slice_scatter_aten.py
```diff
@@ -95,6 +95,22 @@ def test_fallback_dynamic_shape(self):
...
+
+    def test_fallback_open_end_interior_start(self):
+        # Mode B: interior start with open end.
+        open_end = 2**63 - 1
+        module = _SliceScatterNotInputModule(2, 3, open_end, step=1)
+        cache = torch.randn(2, 4, 16, 8)
+        update = torch.randn(2, 4, 13, 8)
+        self.run_test(module, [cache, update])
+
     def test_full_overwrite_is_identity(self):
```
apbose, line 147: Dynamic sliced dimensions remain broken. The clamp is skipped when dim_size == DYNAMIC_DIM The code then reaches np.arange(0, 2**63-1), causing compilation failure or a huge-allocation error. The existing dynamic test varies batch and other dimensions while keeping the sliced dimension fixed at 64, so it does not cover this. We should include this test

jloftin-nv: Should we land dynamic support here or keep in torch and follow up?

apbose: yeah we can have it another PR. Just that we should have a validator then

apbose:
_needs_dim_size() misses a case when the dim_is_static=False .  A positive literal bound slips through unchecked.

resolve_slice_scatter_write does correctly attempt to guard the dynamic-dim case:
if not dim_is_static and (_needs_dim_size(start) or _needs_dim_size(end)):
    return None, None, None, KVWriteStatus.DYNAMIC_DIM_SIZE
but _needs_dim_size() only flags three of the four bound shapes that actually depend on the dim's unknown size:
repro: slice_scatter(cache, update, dim=2, start=0, end=40, step=1) where dim 2's declared dynamic range is 8 to 32
| bound | `_needs_dim_size(bound)` | caught? |
|---|---|---|
| `None` | `True` | yes |
| negative (`-4`) | `True` |  yes |
| open-end sentinel | `True` |  yes |
| positive literal (`40`) | `False` | no |

I think we should widen the _needs_dim_size so concrete bound on dim isnt considered safe

jloftin-nv:
@apbose Should all be addressed, LMK

PR #4513, opened 2026-08-18
Fix complex rewrite device mismatch

jloftin-nv:
# Description

After complex-graph rewrite, FakeTensorProp can mix CUDA activations with CPU _frozen_param* weights and raise FakeTensorDeviceMismatchError, aborting builds that dry-ran cleanly.

Align parameter devices before prop (and keep the non-tensor meta["val"] guard) so rewritten graphs re-propagate on a single device.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
+++ b/py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
@@ -1833,7 +1833,36 @@ class ComplexGraphRewriter:
             if existing_fake_mode is not None
             else FakeTensorMode(allow_non_fake_inputs=True)
         )
-        FakeTensorProp(self.gm, mode=prop_mode).propagate(*fake_inputs)
+
+        # offload_module_to_cpu leaves a folded constant on CPU while its node
+        # metadata still records cuda, which FakeTensorProp would then mix with
+        # fake cuda activations.  Realign each attr to the device its own meta
+        # claims, so genuinely-CPU attrs are left alone.
+        attr_devices: Dict[str, torch.device] = {}
+        for node in self.gm.graph.nodes:
+            if node.op != "get_attr":
+                continue
+            attr_val = node.meta.get("val", None)
+            if isinstance(attr_val, torch.Tensor):
+                attr_devices[str(node.target)] = attr_val.device
+
+        if not attr_devices:
+            FakeTensorProp(self.gm, mode=prop_mode).propagate(*fake_inputs)
+            return
+
+        class _DeviceAligningFakeTensorProp(FakeTensorProp):  # type: ignore[misc]
+            def fetch_attr(self, target: str) -> Any:
+                attr = super().fetch_attr(target)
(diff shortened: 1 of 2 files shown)


tests/py/dynamo/lowering/test_complex_rewrite.py
```diff
@@ -1248,3 +1248,33 @@ def forward(self, z, opt):
...
+    class RotaryComplex(nn.Module):
+        def forward(self, xq, freqs_cis, scale):
+            xq_ = torch.view_as_complex(xq.float().reshape(*xq.shape[:-1], -1, 2))
+            out = torch.view_as_real(xq_ * freqs_cis).flatten(3).type_as(xq)
+            return out * scale
+
+    xq = torch.randn(1, 2, 4, 8)
+    freqs = torch.polar(torch.ones(1, 2, 4, 4), torch.randn(1, 2, 4, 4))
+    _export_and_lower(RotaryComplex(), (xq, freqs, scale))
+
```
apbose, line 1326: this test would pass even without the fix. 0 dim would pass in eager mode and fake tensor mode,but then if you do torch.tensor([2.0]) triggers a mismatch too early FakeTensorDeviceMismatchError—in eager/export—rather than specifically during post-rewrite metadata propagation.

py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
```diff
@@ -1835,7 +1834,33 @@ def propagate_metadata(
...
+            def __init__(self, module, mode, device):
+                super().__init__(module, mode=mode)
+                self._align_device = device
+
+            def fetch_attr(self, target: str):
+                attr = super().fetch_attr(target)
+                if isinstance(attr, torch.Tensor) and attr.device != self._align_device:
+                    return attr.to(self._align_device)
+                return attr
+
```
apbose, line 1864: Every tensor attribute is moved to the first CUDA input’s device.
     The new fetch_attr changes all tensor attributes whenever any CUDA placeholder exists. That includes intentionally CPU attributes or an independent CPU branch, producing incorrect device metadata. Something like below

```
 class M(torch.nn.Module):
      def __init__(self):
          super().__init__()
          self.register_buffer("cpu_table", torch.arange(4))  # intentionally CPU

      def forward(self, z):
          gpu_result = z * z                 # z is CUDA complex
          cpu_result = self.cpu_table + 1    # independent CPU computation
          return gpu_result, cpu_result
```
Maybe we should add
  - A real _frozen_param mismatch test proving the target attribute gets aligned.
  - A mixed-device test like the example above proving an unrelated CPU attribute remains CPU.

apbose:
LGTM. Waiting on CI.

apbose:
Looks like CI is failing on `TestComplexSubgraph::test_complex_subgraph`. Needed a rebase. Rebased

apbose (approved):

apbose:
approved. waiting on the RTX runs

PR #4516, opened 2026-08-18
complex _reshape_copy / _to_copy

jloftin-nv:
# Description

The complex rewriter has no rules for aten._reshape_copy and aten._to_copy, so it inserts view_as_complex / view_as_real wrappers and forces an avoidable graph break. Dry-run then looks like the model uses unsupported complex views.

Register _reshape_copy with the existing reshape path and add a small _to_copy rewrite so those ops stay in the real-valued complex layout.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
+++ b/py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
@@ -1284,21 +1284,40 @@ class ComplexGraphRewriter:
 
     @_complex_unpacker(torch.ops.aten._to_copy.default)
     def _rewrite_to_copy(self, node: Node) -> bool:
-        # complex target: remap dtype, [..., 2] layout unchanged
-        # real target: the cast discards the imaginary part, so select re
         kwargs = dict(node.kwargs)
         dtype = kwargs.get("dtype")
         inp = node.args[0]
-        to_real = dtype is not None and dtype not in COMPLEX_DTYPES
-        if dtype is not None and not to_real:
+        from_complex = self._is_complex_layout_node(inp)
+        to_complex = dtype is None or dtype in COMPLEX_DTYPES
+        if dtype is not None and to_complex:
             kwargs["dtype"] = COMPLEX_TO_REAL_DTYPE[dtype]
+
         with SubgraphBuilder(self.gm.graph, node) as b:
-            if to_real:
-                inp = b(torch.ops.aten.select.int, inp, -1, 0)
-            out = b(torch.ops.aten._to_copy.default, inp)
-            out.kwargs = kwargs
-            if not to_real:
+            if to_complex and from_complex:
+                # remap dtype, [..., 2] layout unchanged
+                out = b(torch.ops.aten._to_copy.default, inp)
+                out.kwargs = kwargs
                 out.meta["is_complex_layout"] = True
+            elif to_complex:
+                # a real input needs a zero imaginary half, so 1 -> [1, 0]
+                re = b(torch.ops.aten._to_copy.default, inp)
+                re.kwargs = kwargs
+                im = b(torch.ops.aten.zeros_like.default, re)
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
```diff
@@ -1282,6 +1282,25 @@ def _rewrite_scalar_tensor(self, node: Node) -> bool:
...
+        if dtype is not None:
+            kwargs["dtype"] = COMPLEX_TO_REAL_DTYPE[dtype]
+        with SubgraphBuilder(self.gm.graph, node) as b:
+            out = b(torch.ops.aten._to_copy.default, node.args[0])
+            out.kwargs = kwargs
+            out.meta["is_complex_layout"] = True
+            node.replace_all_uses_with(out)
+            self.gm.graph.erase_node(node)
+            return True
+
```
apbose, line 1324: this seems incorrect. 
Two parts to this
1.  _to_copy complex to real dtype cast produces silently wrong output
   _rewrite_to_copy returns False for a real target dtype, apparently assuming this triggers the dispatcher's generic fallback (view_as_complex/view_as_real wrapping). It doesn't since the dispatcher only runs that fallback when no handler is registered for the op; since _to_copy.default is registered (via @_complex_unpacker), a registered handler returning False just leaves the node completely unmodified.

  2.  z.to(torch.float32) should discard the imaginary part and return the original (unpacked) shape. But the lowered graph leaves the  node untouched on the [..., 2] layout, so both components (and the extra trailing dim) survive. The test at present complex128 wont catch this since  COMPLEX_TO_REAL_DTYPE[torch.complex128] = torch.float64, but the pre-existing to_copy_dtype_validator (aten_ops_converters.py) only allows {torch.float, torch.int32, torch.int64, torch.bool, torch.int8, torch.float16, torch.bfloat16} , float64 isn't in that set, so any _to_copy targeting it gets rejected by TRT and falls back to PyTorch anyway

py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
```diff
@@ -1282,6 +1282,27 @@ def _rewrite_scalar_tensor(self, node: Node) -> bool:
...
 
+    @_complex_unpacker(torch.ops.aten._to_copy.default)
+    def _rewrite_to_copy(self, node: Node) -> bool:
+        # complex target: remap dtype, [..., 2] layout unchanged
+        # real target: the cast discards the imaginary part, so select re
+        kwargs = dict(node.kwargs)
+        dtype = kwargs.get("dtype")
+        inp = node.args[0]
+        to_real = dtype is not None and dtype not in COMPLEX_DTYPES
+        if dtype is not None and not to_real:
```
apbose: the rewrite assumes that any tensor being converted to a complex dtype is already stored in the internal [..., 2] real/imaginary layout. For a real tensor converted to complex, it does not create the required zero imaginary component before marking the result as complex. For example, [1, 2] should become [[1, 0], [2, 0]], but the rewrite leaves it as [1, 2]. so real to complex would fail

py/torch_tensorrt/dynamo/lowering/passes/complex_graph_rewrite.py
```diff
@@ -1282,6 +1282,27 @@ def _rewrite_scalar_tensor(self, node: Node) -> bool:
...
+            kwargs["dtype"] = COMPLEX_TO_REAL_DTYPE[dtype]
+        with SubgraphBuilder(self.gm.graph, node) as b:
+            if to_real:
+                inp = b(torch.ops.aten.select.int, inp, -1, 0)
+            out = b(torch.ops.aten._to_copy.default, inp)
```
apbose: This always selects only the real component before casting. For boolean conversion eg: 0+1j, it selects 0, which becomes False while PyTorch returns True because the imaginary component is nonzero.

PR #4518, opened 2026-08-18
associative_scan lowering

jloftin-nv:
# Description

higher_order.associative_scan (CUDA combine_mode="pointwise") survives export and has no converter, breaking the graph on classical selective-scan SSM layers.

Add a post-lowering pass that matches the Mamba affine combine (a_l*a_r, a_r*b_l+b_r) and, for static scan length, replaces the HOP with a Hillis–Steele scan of existing aten ops. Unrecognized combines and dynamic S are left alone.

Fixes # (issue)

## Type of change

# Checklist:

- [X ] New and existing unit tests pass locally with my changes

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/passes/lower_associative_scan.py
+++ b/py/torch_tensorrt/dynamo/lowering/passes/lower_associative_scan.py
@@ -8,9 +8,9 @@ combine pattern this pass accepts is Mamba's first-order linear recurrence::
 which is ``h_t = a_t * h_{t-1} + b_t``. Anything else is left alone.
 
 For a static scan length the HOP is replaced with a Hillis–Steele inclusive
-scan built from ``slice`` / ``cat`` / ``mul`` / ``add`` / ``ones_like`` /
-``zeros_like`` — all of which already have converters. Dynamic sequence
-lengths are declined so today's PyTorch fallback behaviour is preserved.
+scan built from ``slice`` / ``cat`` / ``mul`` / ``add`` — all of which already
+have converters. Dynamic sequence lengths are declined so today's PyTorch
+fallback behaviour is preserved.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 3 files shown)


py/torch_tensorrt/dynamo/lowering/passes/lower_associative_scan.py
```diff
@@ -0,0 +1,251 @@
...
+            b_prefix = gm.graph.call_function(
+                torch.ops.aten.slice.Tensor, (b, 0, 0, scan_len - step, 1)
+            )
+            a_head = gm.graph.call_function(
+                torch.ops.aten.slice.Tensor, (a, 0, 0, step, 1)
+            )
+            b_head = gm.graph.call_function(
+                torch.ops.aten.slice.Tensor, (b, 0, 0, step, 1)
+            )
+            ones = gm.graph.call_function(torch.ops.aten.ones_like.default, (a_head,))
```
micwill755: ones_like and zeros_like are inserted after run_decompositions, and Dynamo has no converter for them. The identity pad can run in PyTorch even if the mul/add scan converts, so tests can pass without the scan in TRT. 

You already slice a_head/b_head and a_prefix/b_prefix. Use those directly instead of padding with dummy 1s and 0s:

  keep the head (a_head, b_head)
  combine the tail (prefix with a/b from step onward)
  cat(head, combined_tail)

If you prefer to keep the pad, emit aten.full.default rather than ones_like / zeros_like.

tests/py/dynamo/lowering/test_lower_associative_scan.py
```diff
@@ -0,0 +1,185 @@
...
+            f"associative_scan still present after lowering:\n{gm.graph}",
+        )
+        # Parallel scan stages should introduce doubling slices/cats.
+        targets = {n.target for n in gm.graph.nodes if n.op == "call_function"}
+        self.assertIn(torch.ops.aten.slice.Tensor, targets)
+        self.assertIn(torch.ops.aten.cat.default, targets)
+        self.assertIn(torch.ops.aten.mul.Tensor, targets)
+        self.assertIn(torch.ops.aten.add.Tensor, targets)
+
+    def test_pointwise_scan_numerics_match_eager(self):
```
micwill755, line 107: This can still pass with the PyTorch fallback from the ones_like / zeros_like comment: min_block_size=1 will compile the convertible pieces and leave those ops in PyTorch. I would add an assert to check the lowered graph has no HOP and no ones_like / zeros_like, and TRT-compile a non-power-of-two S.

micwill755 (changes requested):
Requesting changes on the identity pad (ones_like / zeros_like after decompositions won’t convert) and on the TRT numeric test, which can still pass via PyTorch fallback.

jloftin-nv:
@micwill755 Should be addressed, LMK

PR #4519, opened 2026-08-18
dryrun stats observable

jloftin-nv:
# Description

dryrun=True builds a full DryRunTracker (coverage, unsupported ops, engine count) and only prints it. There is no supported way to read those stats for CI or sweeps without monkeypatching.

Mark dryrun_stats_display with @observable() so callers can register on observers.pre / post and capture the tracker without a new public API.

Fixes # (issue)

## Type of change

# Checklist:

- [X ] I have made corresponding changes to the documentation

### DIFF
--- a/tests/py/dynamo/runtime/test_dryrun_stats_observable.py
+++ b/tests/py/dynamo/runtime/test_dryrun_stats_observable.py
@@ -1,0 +1,48 @@
+# type: ignore
+
+import unittest
+
+import torch
+import torch_tensorrt
+from torch.testing._internal.common_utils import TestCase, run_tests
+from torch_tensorrt.dynamo._DryRunTracker import DryRunTracker, dryrun_stats_display
+from torch_tensorrt.dynamo.observer import ObserveContext
+
+
+@unittest.skipIf(not torch.cuda.is_available(), "CUDA required")
+class TestDryRunStatsObservable(TestCase):
+    def test_observer_captures_tracker(self):
+        class Add(torch.nn.Module):
+            def forward(self, x):
+                return x + x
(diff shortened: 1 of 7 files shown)


tests/py/dynamo/runtime/test_dryrun_stats_observable.py
```diff
@@ -0,0 +1,48 @@
...
+
+        model = Add().cuda().eval()
+        x = torch.randn(2, 3, device="cuda")
+        trackers = []
+
+        def capture(ctx: ObserveContext) -> None:
+            trackers.append(ctx.args[0])
+
+        with dryrun_stats_display.observers.pre.add(capture):
+            torch_tensorrt.dynamo.compile(
```
micwill755, line 27: dynamo.compile() needs an ExportedProgram, not an nn.Module. Export first with torch.export.export(model, (x,)), then pass that into compile. 

Also drop enabled_precisions={torch.float32}, enabled_precisions is deprecated as of TRT 11. Dynamo uses dtypes on the model and input so this FP32 test needs no precision argument.

micwill755 (changes requested):
Request changes, for these:

1) Docs missing. Checklist says docs were updated; they were not. Add the callback recipe to dryrun docs / observer docs.

2) Hook does not always run. dryrun_stats_display is skipped on some compile_module early returns including num_supported_ops == 0 even with dryrun=True. Coverage sweeps of unsupported models would get no tracker. Call display or emit the tracker on those paths too.

jloftin-nv:
@micwill755 All should be addressed. LMK if it looks good

PR #4520, opened 2026-08-19
Fix/converter dtype promotion

dongwoonhyun:
# Description

This PR fixes three dtype-related bugs in converters.

#### scatter: fix scalar float64 promotion and src dtype mismatch

The `scatter` converter has 2 bugs that arise under use_explicit_typing:

1. scalars get unconditionally promoted to float64 because of a bare `np.ones`, which defaults to `dtype=np.float64`, raising ValueError: TensorRT does not support float64 (double) precision. **Fix:** replace `np.ones` with `np.full` to get the right dtype.
2. Certain TRT ops (e.g., `argmax`, `topk`) break the dtype contract and emit INT32 ITensors; downstream consumers like `scatter.src` then crash because they are promised INT64 inputs but receive INT32. **Fix:** add a conditional cast in case of dtype mismatch.

#### preserve bool dtype in promote_trt_tensors_to_same_dtype

`promote_trt_tensors_to_same_dtype` erroneously promotes operations between (bool, bool) dtypes to int32 because it unconditionally promotes non-floating-point types to int32. This leads to problems when using use_explicit_typing. **Fix:** explicitly handle the (bool, bool) case to output bool, matching PyTorch's behavior.

#### scalar coercion: preserve float scalar dtype in integer-typed elementwise ops
When `convert_binary_elementwise` is called with one Python scalar input and one tensor input, the scalar's dtype is always cast to the tensor's dtype. As a result, int64_tensor * float32_scalar -> int64, which conflicts with PyTorch's type promotion rules, which would be float32. This poses a problem under use_explicit_typing. **Fix:** Rather than unconditionally casting the scalar to the tensor's dtype, cast it to the dtype that PyTorch's scalar-tensor promotion rules dictate, via `torch.result_type`.

## Type of change

# Checklist:


apbose:
scatter.value still crashes on a float scalar. The fix replaces np.ones(...) with np.full(index_shape_list, src), but np.full still infers dtype purely from the Python scalar's type True to bool, 1 to int64, 2.5 tof loat64  identical to the problem being fixed. Please add a float-scalar test case (e.g. scatter.value(..., 2.5)) that actually exercises this path.
This needs an explicit dtype= argument to np.full, derived from input.dtype

lanluo-nvidia:
@dongwoonhyun  could you please take a look and resolve the comments

apbose:
The CI fails on the new test added -  test_scatter_src_dtype_mismatch_after_argmax(self). you need to have torch.ops.aten.scatter.value or use the arg use_dynamo_tracer=true

apbose:
@dongwoonhyun are you fine with this going to main, or do you want this in 
pytorch:release/2.11? I will rebase the PR to main as of now?

dongwoonhyun:
main should be fine. Thanks!

py/torch_tensorrt/dynamo/conversion/impl/select.py
```diff
@@ -554,17 +554,22 @@ def scatter(
                 input.dtype,
             )
         else:
-            # Static shape: use numpy to create the filled tensor
+            # Static shape: use numpy to create the filled tensor.
+            input_np_dtype = unified_dtype_converter(input.dtype, Frameworks.NUMPY)
             src_tensor = get_trt_tensor(
```
apbose, line 557: Just one comment here. 
unified_dtype_converter(input.dtype, Frameworks.NUMPY) has no mapping for trt.bfloat16, so static BF16 scatter.value will raise TypeError. Can you change it to this- 
src_tensor = get_trt_tensor(
   ctx, np.full(index_shape_list, src),name + "_value_tensor", dtype=input.dtype,
 ). torch.full instead of np.full would fail in faketensormode, so the above should be correct

apbose (approved):

apbose:
approved. waiting on RTX runs to merge

PR #4522, opened 2026-08-19
fix: carry aliased_io through the engine cache

### LINKED ISSUE #4521: 🐛 [Bug] aliased_io is lost on an engine cache hit, changing forward() output arity
## Bug Description

`aliased_io` is not round-tripped through the engine cache. A fresh build records it, but a
cache hit reconstructs `SerializedInterpreterResult` without it, so it silently falls back
to the `{}` default.

The visible effect is that **the same model returns a different number of outputs depending
on whether the engine came from cache**. For an aliased/KV-cache engine, the interpreter
appends a side-effect output to satisfy the network-output requirement of
`IKVCacheUpdateLayer`. `TorchTensorRTModule.forward` truncates that output using
`aliased_io`:

```python
# py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
if self.aliased_io:
    n = user_output_count(self.output_binding_names, self.aliased_io)
    if n < len(outputs):
        outputs = outputs[:n]
```

With an empty map the guard is skipped entirely, no truncation happens, and the internal
side-effect tensor leaks into the caller's return value.

Using `user_output_count` from main unmodified, with `output_binding_names = ["output0",
"output1"]` where `output1` is the appended side-effect output:

```
FRESH build  aliased_io={'output1': ('buf_cache', 'kv_cache_update')}
   user_output_count = 1  -> forward() returns 1 tensor

CACHE HIT    aliased_io={}
   user_output_count = 2  -> forward() returns 2 tensors
```

Same engine, different Python-visible output arity. Code that does `logits = model(x)` gets
a tuple instead of a tensor, or unpacking fails, only on the second run.

## Where it is lost

`py/torch_tensorrt/dynamo/conversion/_conversion.py`

The cache insert tuple omits it (7 fields, no `aliased_io`):

```python
engine_cache.insert(
    hash_val,
    (
        serialized_engine_in_cache,
        interpreter_result.input_names,
        interpreter_result.output_names,
        inputs,
        settings,
        interpreter_result.requires_output_allocator,
        interpreter_result.requires_native_multidevice,
    ),
)
```

And `pull_cached_engine` rebuilds the result without it, so the `{}` default applies:

```python
return SerializedInterpreterResult(
    serialized_engine=serialized_engine,
    input_names=input_names,
    output_names=output_names,
    requires_output_allocator=requires_output_allocator,
    requires_native_multidevice=requires_native_multidevice,
    symbolic_shape_expressions=symbolic_shape_expressions,
    # aliased_io missing -> defaults to {}
)
```

The fresh path does set it (`aliased_io=interpreter_result.aliased_io`), which is the
asymmetry.

This looks like an oversight in #4251, which introduced `aliased_io` and wired the fresh
build path and the `TorchTensorRTModule` consumer, but not the cache insert/pull. No later
commit to `_conversion.py` addresses it.

## Scope, and what is NOT affected

Engine execution itself is fine. Both runtimes reconcile against
`ICudaEngine::getAliasedInputTensor`, which is the source of truth for TRT-enforced
aliasing, and recover a `kv_cache_update` alias the serialized map lacks:

- `py/torch_tensorrt/dynamo/runtime/_TRTEngine.py` `_reconcile_aliased_io`
- `core/runtime/TRTEngine.cpp` (same logic)

So the in-place write still happens. What is *not* recovered is the Python wrapper's own
`self.aliased_io`, which is set only from the constructor argument and is what
`forward`'s truncation and `_pack_engine_info`'s `serialize_aliased_io` read.

Two consequences follow:

1. **Output arity differs on a cache hit** (shown above).
2. **`kind="user"` aliases would be lost outright.** TRT does not track user-declared
   aliases, so reconciliation cannot recover them by design. On current main only
   `AliasKind.KV_CACHE_UPDATE` is ever produced (`conversion/impl/slice_scatter.py`), so
   this part is latent rather than active, but it makes the cache path unsafe for the
   user-declared aliasing the type already supports.

## Expected behavior

A cache hit should produce the same `SerializedInterpreterResult` as a fresh build, so
`aliased_io` should be stored in the cached tuple and restored on retrieval. Note this
changes the cached tuple layout, so previously cached entries would unpack with the wrong
arity. I did not find an existing cache-version constant in `_engine_cache.py`, so some
form of backward compatibility (a length check on `cached_data`, or invalidating old
entries) probably needs deciding as part of the fix.

## Environment

- Torch-TensorRT: main at `dd0afdb7e`
- Verified by reading `origin/main` and executing main's own `user_output_count`

## Additional context

Found while reviewing #4470, which adds a `lift_mutated_buffers` prediction cross-check
that reads the wrapper's `aliased_io`. That cross-check is not on main, so this report is
deliberately scoped to what main itself does. Filing separately because the cache gap is
independent of that PR and predates it.

I have not reproduced this on a GPU with a real cache hit (no GPU/TensorRT in my
environment); the arity result above comes from running main's `user_output_count` directly
against the two map states, plus tracing the insert/pull paths. Happy to send a PR.

shoumikhin:
## What is broken

If you compile a model that updates one of its own buffers in place, such as a KV cache (a
key/value cache: the tensor a transformer writes each step so later steps can look back at
earlier ones), and you have the engine cache turned on, then **the model returns a different
number of outputs the second time you compile it**.

The first compile builds the engine and returns one tensor. The second compile loads the same
engine from the cache and returns two. So this breaks, but only on the second run:

```python
logits = model(tokens, pos)   # first compile:  a tensor
logits = model(tokens, pos)   # from cache:     a 2-tuple, so this is now the wrong shape
```

Nothing errors during compilation. The extra value simply appears, and whatever consumes the
output either unpacks wrongly or fails somewhere further along.

## Why it happens

When TensorRT updates a buffer in place, the converter has to add that buffer as an extra
**network output** (an output of the compiled engine). That is a requirement of the layer
doing the in-place write, `IKVCacheUpdateLayer`, whose output must be a network output.

That extra output is internal bookkeeping, not something the caller asked for. The map named
`aliased_io` records which outputs are bookkeeping, and `forward` uses it to hide them again:

```python
# py/torch_tensorrt/dynamo/runtime/_TorchTensorRTModule.py
if self.aliased_io:
    n = user_output_count(self.output_binding_names, self.aliased_io)
    if n < len(outputs):
        outputs = outputs[:n]
```

The engine cache never saved that map:

- `pack` did not write it into the cached blob,
- so `unpack` could not return it,
- so `pull_cached_engine` rebuilt its result without it, and the field fell back to its empty
  default.

With an empty map, the `if self.aliased_io:` check above is false, the hiding step is skipped
entirely, and the bookkeeping tensor is handed to the caller.

The fresh build path did save the map. Only the cache path lost it, and that disagreement
between the two paths is the bug.

## The fix

Save `aliased_io` in the cached blob and read it back, so a cache hit rebuilds exactly what a
fresh build produced.

- `py/torch_tensorrt/dynamo/_engine_cache.py`: add the field to `UnpackedCacheHit`, to
  `pack`, and to `unpack`.
- `py/torch_tensorrt/dynamo/conversion/_conversion.py`: pass it when inserting into the cache,
  and read it back in `pull_cached_engine`.

That is 8 added lines and no deletions.

**Cache entries written by older versions keep working.** `unpack` reads the new field with
`unpacked.get("aliased_io", {})`, matching how `requires_native_multidevice` is already read.
An older blob has no such key, so it unpacks to an empty map rather than raising `KeyError`.
The blob is a pickled dictionary rather than a fixed-length tuple, which is what makes that
possible without invalidating the cache.

## What this does not touch

Engine execution was already correct, and this change does not alter it. Both runtimes compare
the map against `ICudaEngine::getAliasedInputTensor`, the TensorRT API that reports which
output aliases which input, and recover an entry the saved map is missing:

- `py/torch_tensorrt/dynamo/runtime/_TRTEngine.py`, in `_reconcile_aliased_io`
- `core/runtime/TRTEngine.cpp`, the same logic in the constructor

So the in-place write still happened on a cache hit. What was lost is the Python wrapper's own
copy of the map, which is set only from its constructor argument, and which is what the output
hiding above and `_pack_engine_info` both read.

One detail for reviewers: that recovery cannot help an alias of kind `"user"`, because
TensorRT does not track aliases that Torch-TensorRT declares itself. Today only
`AliasKind.KV_CACHE_UPDATE` is ever produced, in `conversion/impl/slice_scatter.py`, so no user
hits that case yet. It does mean the cache path was unsafe for the user-declared aliasing the
type already allows.

## Test plan

Two tests in a new file, `tests/py/core/test_engine_cache_aliased_io.py`:

1. `test_pack_unpack_round_trips_aliased_io` puts a map through `pack` and `unpack` and
   checks it comes back unchanged.
2. `test_unpack_tolerates_a_blob_without_aliased_io` builds a blob with no `aliased_io` key,
   the way an older version wrote it, and checks it unpacks to an empty map instead of
   raising.

These are pure blob tests with no GPU and no engine, and they live under `tests/py/core` on
purpose: the `py-core` suite runs in the fast lane, so they gate every pull request. The
`models` suite that owns the rest of the engine-cache tests runs only in the full and nightly
lanes and additionally filters on a `critical` marker, so a test placed there would not have
gated this change.

One test in `tests/py/dynamo/models/test_engine_cache.py`:

3. `test_cache_hit_keeps_aliased_io_and_output_arity` compiles an in-place-write model twice
   against one cache directory, and checks the cache hit matches the fresh build in both the
   map and the number of returned outputs.

What I ran, and what I did not:

- Tests 1 and 2 pass on this change and both fail against current `main` with
  `IndexError: tuple index out of range`, so they detect the bug rather than only describing
  it.
- Test 3 needs a GPU and TensorRT. **I have not run it.** That is the one thing worth
  confirming before merging.
- `black --check .` passes across the repository. `isort` and `ruff` each report one finding
  in the touched files, and both reproduce on unmodified `main`, so neither comes from this
  change. `mypy` reports no errors on either changed source file, before or after.

The output-count claim above is not hand-waving: running the existing `user_output_count`
against the two map states gives 1 output when the map is present and 2 when it is empty, for
the same engine.

Fixes #4521

### DIFF
--- a/py/torch_tensorrt/dynamo/_engine_cache.py
+++ b/py/torch_tensorrt/dynamo/_engine_cache.py
@@ -162,9 +162,18 @@ class BaseEngineCache(ABC):
             packed_obj (bytes): packed blob
 
         Returns:
-            Tuple[bytes, List[str], List[str], Sequence[Input], CompilationSettings, bool, bool]: serialized engine, input names, output names, input specs, CompilationSettings, requires_output_allocator, requires_native_multidevice
+            Tuple[bytes, List[str], List[str], Sequence[Input], CompilationSettings, bool, bool, Dict[str, Tuple[str, str]]]: serialized engine, input names, output names, input specs, CompilationSettings, requires_output_allocator, requires_native_multidevice, aliased_io
         """
-        unpacked = pickle.loads(packed_obj)
+        return BaseEngineCache._fields_from(pickle.loads(packed_obj))
+
+    @staticmethod
+    def _fields_from(unpacked: Dict[str, Any]) -> UnpackedCacheHit:
+        """Project an already-deserialized blob onto the cache-hit tuple.
+
+        Split out so ``check`` can inspect the blob and build the tuple from a single
+        deserialization. The blob carries the serialized engine, so unpickling it
+        twice would double the peak memory of a cache hit.
+        """
         return (
             unpacked["serialized_engine"],
             unpacked["input_names"],
@@ -207,11 +216,27 @@ class BaseEngineCache(ABC):
             Optional[Tuple[bytes, List[str], List[str], CompilationSettings, Optional[Dict[Any, Any]]]]: The unpacked cache entry if found, None otherwise.
         """
         packed_cache_info = self.load(hash, *args, **kwargs)
-
-        if packed_cache_info:
-            return BaseEngineCache.unpack(packed_cache_info)
-        else:
+        if not packed_cache_info:
             return None
+
+        unpacked = pickle.loads(packed_cache_info)
+
+        # A blob written before aliased_io was cached is indistinguishable from one
+        # for a model that genuinely has no aliased IO, and guessing "none" is not
+        # safe: for a model that does have it, the engine's internal plumbing tensor
+        # is then returned to the caller and output arity changes on a cache hit.
+        # Treat such a blob as a miss so the engine is rebuilt and re-cached once,
+        # rather than serving a silently wrong result. Only entries written before
+        # this change are affected; every new blob carries the key.
+        if "aliased_io" not in unpacked:
+            _LOGGER.info(
+                "Ignoring engine cache entry %s: written before aliased_io was "
+                "cached, so it cannot be reused safely. It will be rebuilt.",
+                hash,
+            )
+            return None
+
+        return BaseEngineCache._fields_from(unpacked)
 
     @abstractmethod
     def save(self, hash: str, blob: bytes, *args: Any, **kwargs: Any) -> None:
--- a/py/torch_tensorrt/dynamo/_engine_cache.py
+++ b/py/torch_tensorrt/dynamo/_engine_cache.py
@@ -26,6 +26,7 @@ UnpackedCacheHit = Tuple[
     CompilationSettings,
     bool,
     bool,
+    Dict[str, Tuple[str, str]],
 ]
 
 
@@ -122,6 +123,7 @@ class BaseEngineCache(ABC):
         compilation_settings: CompilationSettings,
         requires_output_allocator: bool,
         requires_native_multidevice: bool,
+        aliased_io: Dict[str, Tuple[str, str]],
     ) -> bytes:
         """Pack serialized engine, input names, and output names into a single blob
 
@@ -133,6 +135,7 @@ class BaseEngineCache(ABC):
             compilation_settings (CompilationSettings): compilation settings of TRT engine
             requires_output_allocator (bool): Boolean flag indicating if the converter creates operators which require an Output Allocator to run (e.g. data dependent operators)
             requires_native_multidevice (bool): Boolean flag indicating if the converter creates operators which require multiple devices to run (e.g. multi-device collective operations)
+            aliased_io (Dict[str, Tuple[str, str]]): Map of engine output binding name -> (input binding name, kind_str) for outputs the engine writes through in place
         Returns:
             bytes: packed blob
         """
@@ -147,6 +150,7 @@ class BaseEngineCache(ABC):
                 "compilation_settings": settings,
                 "requires_output_allocator": requires_output_allocator,
                 "requires_native_multidevice": requires_native_multidevice,
+                "aliased_io": aliased_io,
             }
         )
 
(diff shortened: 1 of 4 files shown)


lanluo-nvidia (approved):
lgtm

PR #4525, opened 2026-08-19
fix: stop forcing libstdc++.a into extensions the toolchain already links

shoumikhin:
## The problem

The ExecuTorch runtime wheel does not link. This has always been true and nobody has
seen it: every build died earlier during CMake configure, so the link step was never
reached. With that fixed in #4523 the build gets to 86% and stops here:

```
ld.gold: error: .../gcc-toolset-13/.../13/libstdc++.a(functexcept.o):
  multiple definition of 'std::__throw_bad_array_new_length()'
ld.gold: .../gcc-toolset-13/.../13/libstdc++_nonshared.a(functexcept80.o):
  previous definition here
```

## Why

Two mechanisms each put a full C++ runtime on the link line.

**Explicit.** Both extensions whole-archive `libstdc++.a`, which forces every member
in whether anything references it or not.

**Implicit, and easy to miss.** On a Red Hat gcc-toolset, `libstdc++.so` is not an ELF
object at all. It is a text linker script. Read from the release image,
`docker.io/pytorch/manylinux2_28-builder:cuda13.0`:

```
$ file .../gcc-toolset-13/root/usr/lib/gcc/x86_64-redhat-linux/13/libstdc++.so
ASCII text

$ cat .../libstdc++.so
/* GNU ld script
   Use the shared library, but some functions are only in
   the static library, so try that secondarily.  */
OUTPUT_FORMAT(elf64-x86-64)
INPUT ( /usr/lib64/libstdc++.so.6 -lstdc++_nonshared )
```

So any `-lstdc++` reaching the link pulls `libstdc++_nonshared.a` in as a separate
input. That archive carries the newer-ABI pieces the base `libstdc++.so.6` lacks, and
it is a **strict subset** of `libstdc++.a`. Measured in the same image:

```
global defs in libstdc++.a            5884
global defs in libstdc++_nonshared.a  1003
defined in both                       1003
unique to nonshared                      0
```

Whole-archiving `libstdc++.a` therefore guarantees 1003 duplicate definitions.
`ld.gold` reports the first one it reaches; there are 1002 behind it.

## The change

Drop the whole-archive, and the explicit archive with it. The driver flags already
select the static runtime, and `LINKER_LANGUAGE CXX`, set in the same loop, is what
makes them take effect.

This matches upstream ExecuTorch, which whole-archives its *delegate* but uses plain
`-static-libstdc++ -static-libgcc` for the C++ runtime and never whole-archives
libstdc++ anywhere:

```cmake
# executorch/CMakeLists.txt:1478
target_link_options(executor_runner PRIVATE -static-libstdc++ -static-libgcc)
```

The comment above the block claimed the driver option alone was insufficient, which
reads as though the whole-archive is what fixes it. The symbol it names,
`exception_ptr::_M_addref`, is defined by `libstdc++_nonshared.a` as well as
`libstdc++.a`, so the toolchain already supplies it. The real fix for that bug was
the linker language, and the comment now says so.

The unrelated whole-archive on `torch_tensorrt_executorch_backend` stays. That one is
deliberate: it is how the backend's static registration survives into the module.

## The guard

Because this removes what used to force the static runtime in unconditionally, a
`POST_BUILD` check now asserts the two properties on the real artifacts: no dynamic
`libstdc++` dependency, and no undefined `_M_addref`. This wheel has no auditwheel
step behind it to catch a regression.

Verified in the release image with the Make generator, which is what
`rules_foreign_cc` uses:

```
static runtime, real readelf     build passes
dynamic runtime, real readelf    build fails: "has a dynamic libstdc++ dependency"
static runtime, broken readelf   build fails: "could not inspect"
guard not configured             build passes
default build runs the check     yes, once per extension
```

The predicate captures readelf's output and checks its status before inspecting it,
so an unreadable artifact fails the build rather than passing quietly.

## Limits

The collision itself needs the full ExecuTorch link, so it was not reproduced outside
CI. What is measured directly in the release image is the linker script, the symbol
overlap, the fact that nonshared already provides `_M_addref`, and that the guard
fires correctly in every direction. CI on this branch is the confirmation.

## Relationship to the other changes

Sequenced behind #4523, which fixes the configure so the build reaches the link at
all. Both are needed before `executorch-runtime-build` can pass and
`executorch-runtime-test` stops being skipped. #4524 is what made this error visible,
since bazel was truncating it.

### DIFF
--- a/py/torch-tensorrt-executorch-runtime/native/CMakeLists.txt
+++ b/py/torch-tensorrt-executorch-runtime/native/CMakeLists.txt
@@ -142,11 +142,16 @@ target_link_libraries(portable_lib PRIVATE
   CUDA::cudart TensorRT::nvinfer Threads::Threads)
 
 # The build and validation containers do not necessarily provide the same
-# libstdc++. Link the C++ runtime archives explicitly into both Python
-# extensions. In particular, data_loader is an ExecuTorch target and merely
-# passing -static-libstdc++ as a driver option was not sufficient to prevent
-# its shared object from retaining an unresolved
-# exception_ptr::_M_addref reference on the older CI runtime.
+# libstdc++, so both Python extensions link the C++ runtime statically. Setting
+# LINKER_LANGUAGE CXX below is what makes -static-libstdc++ take effect: without
+# it, data_loader is linked by the C driver and its shared object retained an
+# unresolved exception_ptr::_M_addref reference on the older CI runtime. That
+# symbol is supplied by libstdc++_nonshared.a, which the toolchain links for us,
+# so the driver flags plus the correct linker language are sufficient.
+# Probed as a precondition, not as a link input. -static-libstdc++ below makes the
+# driver resolve -lstdc++ to exactly this file, so if it is absent the link fails
+# with a less obvious error. Nothing names it on the link line: doing that is what
+# caused duplicate definitions against libstdc++_nonshared.a.
 execute_process(
   COMMAND "${CMAKE_CXX_COMPILER}" -print-file-name=libstdc++.a
   OUTPUT_VARIABLE TORCH_TENSORRT_STATIC_LIBSTDCXX
@@ -161,9 +166,9 @@ if(NOT EXISTS "${TORCH_TENSORRT_STATIC_LIBSTDCXX}" OR
 endif()
 
 # rules_foreign_cc injects an explicit dynamic -lstdc++ into module and shared
-# linker flags. Remove only that scoped fragment: the archives below provide
-# the C++ runtime, and retaining this injected library makes the wheel depend
-# on the build host's CXXABI version.
+# linker flags. Remove only that scoped fragment: the driver flags below select the
+# static runtime, and retaining this injected library makes the wheel depend on the
+# build host's CXXABI version.
 foreach(_torch_tensorrt_linker_flags CMAKE_MODULE_LINKER_FLAGS CMAKE_SHARED_LINKER_FLAGS)
   string(REPLACE "-Wl,--push-state,-as-needed -lstdc++ -Wl,--pop-state" ""
     ${_torch_tensorrt_linker_flags} "${${_torch_tensorrt_linker_flags}}")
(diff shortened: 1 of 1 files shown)


shoumikhin:
Restacked on top of #4523 so this can prove itself in CI.

Branched off `main`, this PR could never demonstrate its own fix: the build died at
the configure error long before reaching the link step it repairs. The branch now
carries #4523's commit underneath, so CI here exercises configure **and** link in one
run.

```
98ec226  fix: stop forcing libstdc++.a into extensions the toolchain already links   <- this PR
7d01513  fix: unblock the ExecuTorch runtime wheel configure on manylinux            <- #4523
```

Review only the top commit. The lower one is #4523 and should land first; once it
does, this diff reduces to that single commit.

Note this is the first time the wheel's link step will run on any commit, so it may
surface a further problem behind it. `data_loader.so` in particular links after
`_portable_lib.so` and has never been reached.

shoumikhin:
Both nits addressed, with one partial disagreement written down rather than left
silent.

**Symbol counts: agreed, method now stated.** The comment said 1003 without saying how
it was counted. It is `nm --defined-only -g` counting `T/W/B/D/R`. Counting all global
types gives 1141. The comment now names the method and says explicitly that "none
unique to nonshared" holds under either variant, which is the part the argument rests
on.

**The probed path: agreed it is no longer a link input, but the check is not dead.**
Confirmed the variable is referenced only by the existence check and never on a link
line:

```
157:  OUTPUT_VARIABLE TORCH_TENSORRT_STATIC_LIBSTDCXX
163:  if(NOT EXISTS "${TORCH_TENSORRT_STATIC_LIBSTDCXX}" OR ...
```

Where I disagree is "a missing file it no longer uses". The file is still used, just
by the compiler driver rather than by us: `-static-libstdc++` makes the driver resolve
`-lstdc++` to exactly the path `-print-file-name=libstdc++.a` reports. If it is
absent, the link fails anyway, with a less obvious error than this check gives. So the
precondition still holds and I kept it, but the code read as though it were a leftover.
It now says what it is for, and that nothing names it on the link line because doing
so is what caused the duplicate definitions.

Your nit did surface a second stale comment, which I would not have found otherwise:
the block above said "the archives below provide the C++ runtime". That was true when
libstdc++.a was listed; now only libgcc is, and libstdc++ comes from the driver flag.
Corrected.

lanluo-nvidia (approved):
lgtm

PR #4526, opened 2026-08-19

micwill755:
## Summary

## Test plan


micwill755:
Reopening with branch name `partition-short-circuit` (dropped the 3471- prefix).

PR #4527, opened 2026-08-19

micwill755:
## Summary

This is **PR 1/3** in a compile-startup stack:
1. This PR
2. https://github.com/pytorch/TensorRT/pull/4528
3. https://github.com/pytorch/TensorRT/pull/4529

## Test plan


micwill755:
Superseded by the real GitHub stack (stack #4533): #4530 -> #4531 -> #4532. These fork branches could not be stacked because GitHub stacked PRs do not support pull requests from forks.

PR #4528, opened 2026-08-19
Skip installing large aliased constant-folded weights

micwill755:
## Summary
- Skip installing large constant-folded tensors that only alias existing module weights, avoiding expensive `cpu().contiguous()` copies under `offload_module_to_cpu`.

This is **PR 2/3** in a stack. It includes the partition short-circuit PR plus this commit. Review only the latest commit, or this compare:
https://github.com/micwill755/TensorRT/compare/partition-short-circuit...skip-aliased-fold-install

## Test plan
- [ ] Flux NVFP4 cold compile still succeeds with `offload_module_to_cpu=True`
- [ ] Large weight permutes remain in-graph (not installed as frozen params)
- [ ] VLA still folds materialized embedding constants and compiles (int32 indices)

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -5,7 +5,6 @@ import logging
 import os
 import platform
 import warnings
-
 from typing import Any, Collection, Dict, List, Optional, Sequence, Tuple, Union
 
 import sympy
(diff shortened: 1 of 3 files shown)


micwill755:
Superseded by the real GitHub stack (stack #4533): #4530 -> #4531 -> #4532. These fork branches could not be stacked because GitHub stacked PRs do not support pull requests from forks.

PR #4530, opened 2026-08-19

micwill755:
## Summary

PR 1 of 3 in a compile-startup stack.

## Test plan


narendasan (commented):
Why do we enter partitioning at all if the graph is fully supported, global or adjacency?

cehongwang:
Previously, we needed the partitioner to tag and name the subgraphs, so we still need to enter the partitioner and it will output only one subgraph. But I didn't realize that can cause any overhead.

> Why do we enter partitioning at all if the graph is fully supported, global or adjacency?

cehongwang (approved):
LGTM. Before merging, can you share the benchmark result of speed improvement?

micwill755:
> Why do we enter partitioning at all if the graph is fully supported, global or adjacency?

On Flux, partition went from 0.354s to 0.170s (0.18s saved); on VLA vision, 0.296s to 0.047s (0.25s saved); on VLA diffusion, 0.179s to 0.081s (0.10s saved).

micwill755:
> Why do we enter partitioning at all if the graph is fully supported, global or adjacency?

@narendasan I added a full skip of partitioning so a fully supported graph goes straight to conversion. Total partition time is now Flux 0.123s (0.23s saved), VLA vision 0.022s (0.27s saved), VLA diffusion 0.060s (0.12s saved).

PR #4531, opened 2026-08-19
Skip installing large aliased constant-folded weights

micwill755:
## Summary
- Skip installing large constant-folded tensors that only alias existing module weights, avoiding expensive `cpu().contiguous()` copies under `offload_module_to_cpu`.

PR 2 of 3 in a compile-startup stack; based on #4530.

## Test plan
- [ ] Flux NVFP4 cold compile still succeeds with `offload_module_to_cpu=True`
- [ ] Large weight permutes remain in-graph (not installed as frozen params)
- [ ] VLA still folds materialized embedding constants and compiles (int32 indices)


narendasan:
Can you explain more about what is happening here? What situation are we avoiding by skipping constant folding?

micwill755:
> Can you explain more about what is happening here? What situation are we avoiding by skipping constant folding?

We are not skipping constant folding, ConstantFolder still runs and we still fold, but we skip install when the result is >1 MiB and still aliases a module weight, e.g. a Flux aten.permute on self.weight. That avoids copying hundreds of MiB of weights we already have, e.g on Klein ~372 MiB / ~0.34s which leaves the permute in FX so TRT can fold it as a shuffle at engine build.

PR #4532, opened 2026-08-19
Batch cheap FX cleanups and defer post_lowering recompile

micwill755:
## Summary
- Batch non-conflicting FX cleanups and defer DCE/lint/recompile to one flush at the end of `post_lowering`.
- Keeps `fuse_pad_into_convolution` in the pass list.

PR 3 of 3 in a compile-startup stack; based on #4531.

## Test plan
- [ ] Flux / VLA compile still succeeds
- [ ] `post_lowering` still applies assert removal, dead-user removal, and related cleanups
- [ ] Pad-into-convolution fusion still runs on graphs that need it


narendasan (commented):
Can we add tests here to stress test correctness in edge cases?

micwill755:
> Can we add tests here to stress test correctness in edge cases?

I've added nine test cases that cover applying all four repairs in one pass, matching the old standalone order, assert then orphan in the same batch, defer skipping DCE until flush, immediate cleanup with defer off, no op flush, skipping cleanup when nothing changed, and clearing the defer flag if a pass raises. Let me know if you want more.

narendasan (approved):
LGTM

PR #4534, opened 2026-08-19
fix: make the ExecuTorch runtime wheel build and its reference runner work

shoumikhin:
## Summary

The ExecuTorch runtime wheel build fails on `main`, on every row of the matrix:

```
[86%] Linking CXX shared library _portable_lib.so
checking _portable_lib.so links the C++ runtime statically
FATAL: _portable_lib.so has a dynamic libstdc++ dependency
```

The wheel is meant to carry its own copy of the C++ runtime, because the machine that builds it and the machine that runs it do not necessarily ship the same `libstdc++`. It asks for that with `-static-libstdc++` and then checks the result. The check is right and the request was not being honoured.

With this change the build passes and produces `torch_tensorrt_executorch_runtime-*.whl`, the first successful run of that job since 2026-08-14.

## Why it was failing

Three separate things, each hidden behind the last.

**1. The removal ran too late.** The extensions are created inside ExecuTorch's directory by `add_subdirectory`, and take their copy of the linker-flag variables at that moment. The removal sat further down the file, so it edited a value nothing reads. The build printed both, side by side:

```
CMAKE_SHARED_LINKER_FLAGS = [... --gc-sections  --push-state,-as-needed -lm ...]   <- fragment gone
link command              = [... --gc-sections --push-state,-as-needed -lstdc++ ...] <- fragment present
```

**2. Position does not save us, because of the linker.** The fragment arrives wrapped in `--as-needed` ahead of the objects, which the default linker drops. This build passes `-fuse-ld=gold`, and gold keeps it. Measured in the release image on a shared object using `std::string` and exceptions, all with `-static-libstdc++ -static-libgcc`:

| linker | `-lstdc++` present | libstdc++ NEEDED |
| --- | --- | --- |
| bfd | yes | no |
| gold | yes | **yes** |
| gold + LTO | yes | **yes** |
| gold + LTO | no | no |

**3. Removing it removes the only C++ runtime on the line.** The Bazel toolchain hands CMake the **C** driver as `CMAKE_CXX_COMPILER`:

```
/opt/rh/gcc-toolset-13/root/usr/bin/gcc -fPIC ...
```

`gcc` links no C++ runtime and treats `-static-libstdc++` as a no-op, which is exactly why the toolchain injected an explicit `-lstdc++`. Taking it away left `exception_ptr::_M_addref` undefined.

## The fix

Remove the injected fragment **before** anything that links is defined, and put the static archive where an archive can actually resolve something. An archive only pulls members that resolve symbols already undefined when it is scanned, so ahead of the objects it contributes nothing. `CMAKE_CXX_STANDARD_LIBRARIES` is appended after the objects, which is the placement rules_foreign_cc documents for this exact problem.

| link line (gcc driver, gold, LTO) | UND `_M_addref` | libstdc++ NEEDED |
| --- | --- | --- |
| dynamic `-lstdc++` before objects | 2 | 1 |
| nothing at all | 2 | 0 |
| static archive before objects | 2 | 0 |
| **static archive after objects** | **0** | **0** |

## Changes

* Remove the injected `-lstdc++` before `add_subdirectory`, from the three `CMAKE_{SHARED,MODULE,EXE}_LINKER_FLAGS`. Those are the only variables that can carry it.
* Append the static `libstdc++.a`, named by the path the compiler reports, to `CMAKE_CXX_STANDARD_LIBRARIES` so it lands after the objects.
* Drop the pre-`project()` attempt on the `*_INIT` variables. The generated toolchain file re-sets those while `project()` runs, so it never took effect.
* Add `CMAKE_CXX_STANDARD_LIBRARIES` to `CMAKE_TRY_COMPILE_PLATFORM_VARIABLES`, so configure probes keep the runtime.
* Match the push/pop group as a unit, and the bare `-lstdc++` with a token boundary, so `-lstdc++fs` and `-lstdc++_nonshared` survive and the group is never left unbalanced.
* Move the check into `check_static_cxx_runtime.sh`, print the `NEEDED` entries and the link command on failure, and widen it past the single `_M_addref` symbol.
* Print the resolved variables at configure time.

## Testing

The wheel build passes in CI on this PR and produces the wheel.

Everything else measured in the release container image (`manylinux2_28-builder`, gcc 13.3.1), on a replica with both targets defined in a subdirectory, the gcc driver, gold, LTO and the fragment injected:

```
try_compile HAVE_CXX_RT=[1]                       (was [] without the probe fix)
portable_lib  NEEDED libstdc++: 0 | guard: PASS
data_loader   NEEDED libstdc++: 0 | guard: PASS
```

The widened check was verified both ways. A healthy object passes and loads; one linked without the runtime fails the check and fails to load with `undefined symbol: _ZTISt9exception`. The narrow `_M_addref` test passed that same broken object.

## The next thing behind it, fixed here too

With the build green, `executorch-runtime-test` ran for the first time and failed further along, in the standalone C++ reference runner:

```
op__device_copy.cpp] Check failed (allocator != nullptr):
  _h2d_copy: no device allocator registered for device_type=1
method.cpp] KernelCall failed at instruction 0:0 in et_copy::_h2d_copy.out: 0x20
```

A Torch-TensorRT export marks its delegate inputs and outputs as CUDA memory, so ExecuTorch places device copies around the delegate. Those kernels ask for the allocator registered for the CUDA device type, and in ExecuTorch 1.4.1 exactly one thing registers it, a static initializer in `backends/cuda/runtime/cuda_backend.cpp`. The runner was built without that backend, so it loaded the program, initialized the TensorRT engine, and stopped on the first instruction.

The last commit turns on `EXECUTORCH_BUILD_CUDA` for the example, plus the `EXECUTORCH_BUILD_EXTENSION_TENSOR` companion ExecuTorch's preset checker requires. Linking needs no change: the runner already links `executorch::backends`, and `backends/cuda/CMakeLists.txt` calls `executorch_target_link_options_shared_lib(aoti_cuda_backend)` so the static initializer is not dropped. That matters, measured on this toolchain:

| link | result |
| --- | --- |
| `g++ main.o libbackend.a` | initializer dropped |
| `gold`, `--gc-sections` | initializer dropped |
| `gold`, `--gc-sections`, `--whole-archive` | initializer ran |

Registering our own allocator instead is not an option: `DeviceAllocatorRegistry::register_allocator` aborts on a second registration for the same device type, which would break any process that also loads ExecuTorch's CUDA backend.

It is in this PR rather than its own because it cannot be tested anywhere else. Until the wheel build succeeds, the job that runs the reference runner never starts, so a separate PR for it could not be green or even exercised.

For reviewers: the CUDA toolkit is now required to build that example, and the build is larger. It adds no libtorch dependency, which `verify-executorch-reference-runner.sh` already asserts.


lanluo-nvidia (approved):
Thanks for fixing this issue, I was stuck on it for a while.

PR #4539, opened 2026-08-20
fix: stop nm symbol checks failing on SIGPIPE in the reference runner gate

shoumikhin:
## Problem

The ExecuTorch reference runner CI gate fails at random on builds that are perfectly fine, and it blames a different missing symbol each time.

The symbol checks look like this:

```bash
if ! nm -D --undefined-only "${runner_path}" 2>/dev/null | grep -q "${_symbol}"; then
```

`grep -q` stops at its first match and closes the pipe. `nm` is then killed by SIGPIPE and exits 141. The script runs under `set -o pipefail`, so the pipeline reports 141, and `if !` flips that into a false "does not import" followed by `exit 1`.

Whether it fires depends on whether `nm` still has output buffered when `grep` leaves, so it follows where the symbol sits in the symbol table:

| case | false failures |
| --- | --- |
| symbol early in a large symbol table | 25 / 25 |
| symbol on the very last line | 0 / 25 |

That is why the symbol named in the error changes from run to run.

## One check fails the other way, silently

The sweep for duplicate caller-stream definitions is not negated:

```bash
if nm --defined-only --dynamic "${_so}" 2>/dev/null | grep -qE "getCallerStream|CallerStreamGuard"; then
```

Here a 141 reads as "no duplicate found", so the guard passes without really checking. Against a real shared library that returned a false negative 40 times out of 40, meaning this guard has not been protecting anything.

## Fix

Read the tool output into a variable, then match against the variable, so there is no pipe to close. A small helper does it once for all five call sites:

```bash
nm_matches() {
  local _pattern="$1"
  shift
  local _symbols
  _symbols="$(nm "$@" 2>/dev/null)"
  grep -qE "${_pattern}" <<<"${_symbols}"
}
```

Two `head -n1` pipelines had the same hazard and now trim the first line with a parameter expansion instead.

## Testing

- Reproduced the mechanism and showed which stage dies:

  ```
  $ nm -D --undefined-only "$(command -v python3)" | grep -q malloc
  status: 141   PIPESTATUS: 141 0
  ```

- 40 runs of each form against a real binary: old form 40/40 false failures, new form 0/40.
- Same contrast on the duplicate-definition sweep against a real shared library: old form 40/40 false negatives, new form correct every time.
- Confirmed the checks still fail when a symbol is genuinely absent, so they did not become vacuous.
- Re-audited the script for any remaining pipeline whose last stage exits early. None remain.
- `bash -n` passes.

Only the CI verification script changes. No runtime or build code is touched.

### DIFF
--- a/.github/scripts/verify-executorch-reference-runner.sh
+++ b/.github/scripts/verify-executorch-reference-runner.sh
@@ -311,6 +311,17 @@ for _tool in ldd readelf nm; do
   fi
 done
 
+# Capture nm output instead of piping it into `grep -q`. grep exits at its first
+# match and closes the pipe, so nm dies of SIGPIPE and pipefail reports 141, which
+# reads as a false verdict in either direction.
+nm_matches() {
+  local _pattern="$1"
+  shift
+  local _symbols
+  _symbols="$(nm "$@" 2>/dev/null)"
+  grep -qE "${_pattern}" <<<"${_symbols}"
+}
+
 # The runner must not pull in libtorch: this native path is libtorch-free.
 if ldd "${runner_path}" |
     grep -E "libtorch|libtorch_cpu|libtorch_cuda|libc10" >&2; then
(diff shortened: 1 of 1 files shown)


lanluo-nvidia (approved):
lgtm

PR #4540, opened 2026-08-20
Cache post-lowering ATen graphs to skip decomp on warm compiles.

micwill755:
## Summary
- Add a disk cache for the post-lowering ATen `GraphModule`, keyed from the exported program before decomposition.
- On a hit, skip `run_decompositions` and `post_lowering` and rebuild the FX graph from a node list + `state_dict` (`torch.save`), not `torch.export.save` or FX `__reduce__` retrace.
- Conservative: full compilation + fast partition only; bypass if the graph cannot be encoded.

## Test plan
- [x] `tests/py/dynamo/models/test_lowering_cache.py`
- [x] Flux2 Klein 4B NVFP4 warm compile + inference (engine cache hits after the engine is built from the reconstructed graph)


narendasan (commented):
Here can we use the same caching interface as engine caching? In theory users could provide their own implementation for things like remote cache or mem cache. We should use the same cache object for both engine and graph cache if possible

micwill755:
> Here can we use the same caching interface as engine caching? In theory users could provide their own implementation for things like remote cache or mem cache. We should use the same cache object for both engine and graph cache if possible

Done, lowered graphs now use BaseEngineCache.save/load as bytes, and custom_engine_cache is reused for both engines and graphs. Let me know if you want any further changes.

PR #4541, opened 2026-08-21
tests: skip TestPythonRuntimeAliasedIO on TensorRT-RTX with the C++ runtime

tp5uiuc:
# Description

## What

Skip `TestPythonRuntimeAliasedIO` on TensorRT-RTX builds that also ship the C++
runtime, where it currently fails.

## Why

The runtimes are expected to be either/or: the flavor is picked from whether
`libtorchtrt.so` is loaded, and every in-library `TRTEngine` construction site is
gated on that same flag. This class hand-builds a Python `TRTEngine` from packed
engine info regardless, producing a combination the library never does — a
torchbind-backed `RuntimeCache` asked to serve a Python engine. `ensure_cache()`
returns `None` there and `set_runtime_cache` rejects an explicitly-passed `None`,
so both tests in the class fail.

## How

- `test_aliased_io.py`: `skipIf(torch_tensorrt_runtime and tensorrt_rtx)`.
  Deliberately narrow — on standard TensorRT the runtime-config path is
  short-circuited in `ensure_initialized`, so a hand-built engine is harmless and
  the class stays live; it also stays live on python-only builds, where the
  Python `TRTEngine` is the real runtime and the test is the point.

## Testing

| Build | Result |
| --- | --- |
| C++ runtime, `tests/py/dynamo/runtime/` | 170 passed, 46 skipped |

The same suite on `main` reports 2 failed.

## Type of change

# Checklist:

### DIFF
--- a/tests/py/dynamo/runtime/test_aliased_io.py
+++ b/tests/py/dynamo/runtime/test_aliased_io.py
@@ -26,10 +26,13 @@ also honors aliasing by binding the aliased output to the source input's
 storage; :class:`TestPythonRuntimeAliasedIO` exercises that path directly.
 """
 
+import unittest
+
 import torch
 import torch_tensorrt
 from torch.export import export
 from torch.testing._internal.common_utils import TestCase, run_tests
+from torch_tensorrt._features import ENABLED_FEATURES
 from torch_tensorrt.dynamo.runtime._TRTEngine import TRTEngine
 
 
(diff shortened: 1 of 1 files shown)


tp5uiuc:
[by Claude Code] Closing in favour of #4482, which fixes these tests rather than skipping them.

`TestPythonRuntimeAliasedIO` hand-builds a Python `TRTEngine`. That engine came
up with the default `RuntimeSettings`, whose `runtime_cache` is a path string;
`RuntimeCache` then picked its backing handle from whether the C++ runtime was
loaded, and on a C++ build `ensure_cache()` returned `None`, so
`set_runtime_cache(None)` raised. #4482 defaults engines to `runtime_cache=None`
— the module owns the implicit cache — so no attach happens and the failure goes
away at the source.

Verified on TensorRT-RTX/A100, `runtime/test_aliased_io.py`: 2 failed → 10
passed, on both the Python and C++ runtimes.

PR #4547, opened 2026-08-21
Ship the constant_fold_exclusions package in the wheel

shoumikhin:
## The problem

Importing `torch_tensorrt` from an installed wheel fails outright:

```
ModuleNotFoundError: No module named
  'torch_tensorrt.dynamo.lowering.constant_fold_exclusions'
```

This is currently failing **331 jobs on `main`**, across `standard`, `rtx`, `python-only`, `python-only-rtx`, `executorch-runtime-build` and `build-docs`. I sampled jobs from each of those families and they all report this same error.

## Why it happens

`setup.py` enumerates packages explicitly rather than calling `find_packages()`, so every new subpackage has to be added to two lists by hand:

```
dynamo_packages       lists "torch_tensorrt.dynamo.lowering", "...lowering.passes"
dynamo_package_dir    maps the same names to their source paths
```

`constant_fold_exclusions` was added to the source tree without being added to either list, so it never gets copied into the wheel.

That would be harmless if nothing imported it eagerly, but the import runs as soon as the package is imported:

```
dynamo/lowering/__init__.py:7              from .passes import *
passes/_aten_lowering_pass.py:9            from ...passes.mark_constant_fold_exclusions import ...
passes/mark_constant_fold_exclusions.py:4  from ...lowering.constant_fold_exclusions._core import ...
```

**So it works from a source checkout and fails only in an installed wheel.** That is why it passed pull request CI and only broke after landing: the tests that would have caught it run against the built wheel.

## The fix

Two lines, one in each list, alongside the existing `lowering.passes` entries.

## Testing

Compared every package on disk against both lists. Before this change exactly one was missing, after it none are:

```
before:  torch_tensorrt.dynamo.lowering.constant_fold_exclusions
after :  NONE
```

Then staged a tree containing only the packages `setup.py` lists, which is what setuptools copies into the wheel, and confirmed the directory arrives with both of its modules:

```
staged wheel contains constant_fold_exclusions: True
its files: ['__init__.py', '_core.py']
```

Finally repeated that staging with these two lines removed, and confirmed the package is then absent. The check reports a different result in the two states rather than passing either way.

## Note on scope

This only makes the existing subpackage reachable from the wheel. It does not change any behaviour of the constant-fold exclusion pass itself.

If the intent was for this subpackage to stay internal rather than ship, then the alternative fix is to restructure the import so it is not reached at package import time. I went with shipping it because the import chain above is unconditional, but happy to switch if the other shape is preferred.


lanluo-nvidia (approved):
lgtm

PR #4548, opened 2026-08-21
Detect ExecuTorch pin drift across the repository

shoumikhin:
## The problem

The ExecuTorch this repository builds against is named in **eleven tracked files**, and one had drifted:

```
.github/workflows/executorch-build-linux.yml:122  executorch>=1.4.0    <- stale floor, no ceiling
```

## Where the pin lives

In `dev_dep_versions.yml`, next to the CUDA and TensorRT pins that `setup.py:75` and `versions.py:106` already read:

```yaml
__executorch_version__: "1.4.1"
__executorch_commit__: "e4d02f41f7909e8ed5bf4a14ffc520d733453d9f"
```

Thanks @lanluo-nvidia for catching that a new file would have been a second convention.

## Two spellings, chosen by role

Both spellings in the repo are correct, and `setup.py:205-208` explains why: install-time metadata takes the range so a patch release off the same branch stays installable, while a build input pins exactly because the artifact is compiled against one wheel.

| Role | Spelling | Sites |
|---|---|---|
| Metadata resolved at install time | `>=1.4.1,<1.5` | `setup.py:209`, `justfile:88`, `tests/ci/runner.py:122` |
| Build input, source sha, build recipe | `==1.4.1` | runtime `pyproject.toml:9`, 2 workflow steps, 4 `MODULE.bazel` variants, runtime `README.md` |

The test enforces the spelling each role requires rather than accepting either anywhere. Both spellings resolve to the pinned version, so a permissive check stays green while a site quietly uses the wrong kind of constraint.

**Role follows the line, not the file.** `executorch-build-linux.yml` pins a build input at line 82 and reproduces a user install at line 122, which already carries a `# this is to verify the end user's workflow` comment. So line 122 is a range and the check keys off that comment. Its floor was stale, but its spelling was right; an earlier revision of this PR wrongly hardened it to match its neighbour.

Anything that does not produce installable metadata is expected to pin exactly, so an unclassified new site is held to the stricter rule instead of going unchecked.

## Discovery, not a list of paths

Sites are found with `git grep`. A hand-written list would drift the same way the pins it checks drifted. Discovery found **five sites a list would have missed**, including a real build requirement at `py/torch-tensorrt-executorch-runtime/pyproject.toml:9` and the source commit repeated in three more `MODULE.bazel` variants.

## What this deliberately does not do

The requirement sites are not rewritten to read the yaml. That means a shell substitution inside two workflows and an import in the justfile, a behaviour change worth reviewing separately from recording the pin.

## Testing

Ran the test before fixing line 122; it failed naming the file, line, found and expected. Fixed the line and both checks pass.

Two requirements in the same workflow get opposite rules:

```
executorch-build-linux.yml:82   -> expects executorch==1.4.1
executorch-build-linux.yml:122  -> expects executorch>=1.4.1,<1.5
```

Six wrong states injected one at a time, restoring between each:

```
line 122 hardened to exact                   caught
line 82 widened to a range                   caught
setup.py exact where a range belongs         caught
runtime pyproject range where exact belongs  caught
a range that lost its upper bound            caught
the yaml bumped, sites left behind           caught
```

The middle four are cases a single permissive rule cannot see.

Both checks assert they found something before asserting agreement, so an expression that stops matching fails loudly instead of passing empty. That guard caught a real bug while writing this: `git grep -E` does not support `\s`, so the bazel pattern matched nothing and the check reported it rather than passing.

## Open question

I don't think CI collects this test. The suites in `tests/ci/suites.py` name specific directories (`tests/py/dynamo`, `tests/py/core`, `tests/py/ts`, `tests/py/kernels`) and none is `tests/py` itself, so a hygiene test sitting directly in `tests/py/` may never run. Happy to move it or add a suite entry.


lanluo-nvidia (commented):
executorch_pin.toml introduces a second dependency-version convention. This repo already has dev_dep_versions.yml, which setup.py reads, so I’d put ExecuTorch there instead.
also there are other places which has "executorch>=1.4.1,<1.5", maybe change everywhere to use the version from  dev_dep_versions.yml

shoumikhin:
Good call on both points, changed.

**`dev_dep_versions.yml` instead of a new file.** You're right that a second convention was the wrong move. The version and the source commit now live there next to the CUDA and TensorRT pins, and `executorch_pin.toml` is gone. The test reads the yaml with a regex rather than importing `yaml`, matching the shape `setup.py:75` and `versions.py:106` already use.

**On making every site use one spelling** — I dug into this and I don't think they should all agree, because the repo already deliberately splits them and explains why at `setup.py:205-208`:

> The upper bound is the load-bearing half [...] Patch releases stay allowed because they come off the same release branch; the exact pin belongs in the runtime package, which does derive it.

So there are two roles:

| Role | Spelling | Sites |
|---|---|---|
| Metadata resolved at install time | `>=1.4.1,<1.5` | `setup.py:209`, `justfile:88`, `tests/ci/runner.py:122` |
| Build input, source sha, build recipe | `==1.4.1` | runtime `pyproject.toml:9`, 2 workflow steps, 4 `MODULE.bazel` variants, runtime `README.md` |

Collapsing these to one spelling would break something either way: a range on `[build-system] requires` can resolve an ExecuTorch the artifact wasn't compiled against, and an exact pin in `setup.py` would make the shipped wheel reject a compatible patch release.

So the test now enforces *which* spelling each site's role requires, instead of accepting either one anywhere. That matters because both spellings resolve to the pinned version, so a permissive check stays green while a site quietly uses the wrong kind of constraint.

**One thing your comment made me find.** Role follows the line, not the file. `executorch-build-linux.yml` pins a build input at line 82 but reproduces a user install at line 122, which already has a `# this is to verify the end user's workflow` comment above it. Line 122 should therefore be a range. Its floor was stale at `>=1.4.0`, but its *spelling* was right, and I had wrongly hardened it to `==1.4.1` to match its neighbour. It is now `>=1.4.1,<1.5`, and the check keys off that existing comment. Verified the two adjacent lines get opposite expectations.

Six wrong states injected and each is caught with file, line, found and expected, including both directions of the wrong-spelling case.

**Still open, and I'd like your read:** I don't think CI actually collects this test. The suites in `tests/ci/suites.py` name specific directories (`tests/py/dynamo`, `tests/py/core`, `tests/py/ts`, `tests/py/kernels`) and none of them is `tests/py` itself, so a repo-hygiene test sitting directly in `tests/py/` may never run. Where would you want it, and does it need a suite entry?

lanluo-nvidia (approved):
lgtm

shoumikhin:
Thanks, and sorry for the delay on this second point. You asked:

> there are other places which has `executorch>=1.4.1,<1.5`, maybe change everywhere to use the version from `dev_dep_versions.yml`

You are right that detection alone is weaker than derivation, and I agree with the direction. I went through all 11 sites to see which can actually derive the value, and the answer splits cleanly in two.

**Sites that CAN derive (Python, evaluated at build/run time):**

```
setup.py:209                 EXECUTORCH_REQUIREMENT = "executorch>=1.4.1,<1.5"
tests/ci/runner.py:122       "executorch>=1.4.1,<1.5"
```

`setup.py` already opens `dev_dep_versions.yml` at line 75 for the CUDA and TensorRT versions, so reading `__executorch_version__` there is a small, natural change with no new machinery.

**Sites that CANNOT derive, for a mechanical reason:**

```
MODULE.bazel                             no Python at module evaluation time
docker/MODULE.bazel.docker               same
docker/MODULE.bazel.ngc                  same
toolchains/ci_workspaces/MODULE.bazel.tmpl   a template, rendered by envsubst
py/torch-tensorrt-executorch-runtime/pyproject.toml:9   PEP 518 `requires` is static TOML
```

`pyproject.toml` is the clearest case: `[build-system] requires` is parsed before any build backend exists, so there is nowhere to run a lookup. The MODULE.bazel files are evaluated by Bazel's own Starlark, which cannot read a YAML file at that point either.

For those five, the value has to be written in by something. That means the real fix is a generator, and the natural owner of a generator is the bump job that already has to rewrite every site when the pin moves. Doing it in this change would mean adding a code generator plus a "generated files are current" check, which is a different change from "detect drift".

**So, concretely:**

- This change stays as detection, which is what makes any later derivation safe to land: without the drift test, a generator that misses a site fails silently.
- I will send the `setup.py` and `tests/ci/runner.py` derivation as its own small follow-up, since those two are genuinely easy and remove two of the eleven literals.
- The remaining five get written by the bump job, so the pin has exactly one human-edited home.

If you would rather I fold the two Python sites into this change instead of a follow-up, I am happy to do that. I kept them separate so this one stays reviewable as a single idea, but it is your call.

Issue #4438, opened 2026-07-28
Bug Report: Hard-coded CUDA 12 check blocks CUDA 13 support in Torch-TensorRT


joe0731:
# Bug Report: Hard-coded CUDA 12 check blocks CUDA 13 support in Torch-TensorRT

## Bug Description
TensorRT-LLM has officially supported CUDA 13, but `torch_tensorrt/_utils.py` still contains a hard-coded version check that strictly restricts CUDA major version to 12.

The current logic blocks all CUDA 13 environments with the following error:
```python
major, minor = map(int, cuda_version.split("."))
if major != 12:
    logger.error(
        "CUDA 13 is not currently supported for TRT-LLM plugins. Please install pytorch with CUDA 12.x support"
    )
    return False
```

This check is outdated and inconsistent with the latest TensorRT-LLM compatibility, preventing normal usage on CUDA 13 stacks.

## To Reproduce
1. Set up a full CUDA 13 environment
2. Install PyTorch, TensorRT and TensorRT-LLM built for CUDA 13
3. Initialize or run any workflow using Torch-TensorRT + TRT-LLM plugins
4. The hard-coded version check fails and throws the CUDA 13 unsupported error

## Expected Behavior
The version validation logic should be updated to recognize and allow CUDA 13. The error message should also be revised to match the latest supported CUDA versions. No functional restriction should exist if TensorRT-LLM natively supports CUDA 13.

## Environment
- Torch-TensorRT Version: N/A
- PyTorch Version: N/A
- CPU Architecture: x86_64
- OS: Linux
- PyTorch Installation Method: pip / conda / source
- Build command: N/A
- Source type: prebuilt / local source
- Python version: N/A
- CUDA version: 13.x
- GPU models: N/A

## Additional Context
This is purely a hard-coded version check issue rather than a real compatibility problem. TensorRT-LLM has added official CUDA 13 support, so the Torch-TensorRT validation layer needs to be synced and updated.

narendasan:
@joe0731 is this with main or a particular commit/release? Also there is now native support for the NCCL collectives in TRT, are the plugins from TRT-LLM still necessary for your usecase?

joe0731:
the plugins from TRT-LLM still necessary @narendasan

PR #4549, opened 2026-08-22
fix: add CUDA 13 support for TRT-LLM plugins

apbose:
Addresses #4438
 Fixes hardcoded CUDA 12 check blocking CUDA 13 in TRT-LLM plugin loading


narendasan (approved):
Lgtm

PR #4555, opened 2026-08-22
Derive the ExecuTorch requirement from the version pin

shoumikhin:
## Problem

The ExecuTorch version is written out by hand in every place that needs it.
Moving the pin means editing each copy and hoping none was missed.
`dev_dep_versions.yml` already holds the value, so the places that run Python
and can read a file should read it instead of repeating it.

#4548 added a test that fails when the copies disagree. That test catches a
missed edit, but somebody still has to make every edit. This removes two of the
copies, so there is less to edit next time.

## Solution

Two sites now build the requirement from `__executorch_version__`:

- `setup.py`. `load_dep_info()` already opens `dev_dep_versions.yml` for the
  CUDA and TensorRT versions, so this reads `__executorch_version__` next to
  them.
- `tests/ci/runner.py`, the `executorch` setup step. This file declares no
  runtime dependencies of its own, so it reads the pin with a regular
  expression rather than pulling in `yaml`. That is how the drift test already
  reads the same file.

Both produce the shape the comment above `EXECUTORCH_REQUIREMENT` describes:
the floor is the pinned version, the ceiling is the next minor. With the
current pin that is `executorch>=1.4.1,<1.5`, byte for byte the literal each
one replaces.

Counting every requirement the drift test scans, and leaving out
`dev_dep_versions.yml` itself, the repository goes from 13 lines in 11 files to
11 lines in 9 files.

## What stays literal, and why

Four of the remaining lines are comments sitting next to the Bazel `commit =`
they label, in `MODULE.bazel`, `docker/MODULE.bazel.docker`,
`docker/MODULE.bazel.ngc` and `toolchains/ci_workspaces/MODULE.bazel.tmpl`.
Nothing evaluates them.

Two are in `py/torch-tensorrt-executorch-runtime/README.md`: one command a
reader copies and one sentence of prose.

One is `[build-system] requires` in
`py/torch-tensorrt-executorch-runtime/pyproject.toml`. That is static TOML,
read before any build backend exists, so there is nowhere to run a lookup.

The last four could read the YAML but are left alone on purpose: three shell
lines in the workflows (`executorch-build-linux.yml` twice and
`executorch-test-linux.yml`) and one in the `justfile`. Reading YAML from shell
means either a `python -c` line in front of every install or a `yq`
dependency, and `test_every_requirement_matches_the_pin` already blocks a wrong
value from landing. Trading a readable one-line install for a parsing step buys
nothing here.

## RANGE_SITES

The drift test decides per site whether the requirement should be an exact pin
or a range, and `setup.py` and `tests/ci/runner.py` were listed as range sites.
Its search can no longer find a literal in either file, so those two entries
became unreachable and are removed. `justfile` stays, because its line is still
literal.

## Tests

Two tests are added, next to the drift test:

- `test_derived_requirements_match_the_pin` checks both derived strings against
  what the pin implies. Without it the two sites would have no coverage at all,
  since the search-based test can no longer see them.
- `test_derived_requirements_roll_the_minor_over` feeds in `1.9.0` and checks
  the ceiling comes out as `1.10` and not `1.1`. The bound is a version, not a
  decimal number.

Reading the values back is awkward on both sides, and the helpers carry a
comment saying why. `setup.py` cannot be imported here, importing it starts a
build, so the test parses the file and evaluates only the assignments that
derive the requirement. The runner is read through a subprocess, so pointing it
at a temporary tree cannot leave a reloaded module behind for whatever runs
next.

One wrinkle worth knowing if you edit these tests: the drift test scans this
file too, so a requirement written out in full inside the rollover test reads
as a site that drifted from the pin. The expected string is built from a
variable for that reason.

## Test plan

Run locally on Python 3.11:

```
pytest tests/py/dynamo/executorch/test_executorch_pin.py
4 passed
```

That is the whole file: the two drift tests that were already there and the two
new ones. Both drift tests still pass, which also confirms the shortened
`RANGE_SITES` did not reclassify any site that is still literal.

`black --check` and `isort --check-only` are clean on the three changed files.
`ruff check` reports the same nine findings before and after this change, all
of them pre-existing in `setup.py` and `tests/ci/runner.py`, so none are new.

Not exercised: nothing here touches a build or a GPU path, so no ExecuTorch
suite was run.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,6 +72,7 @@ def load_dep_info():
     global __tensorrt_version__
     global __tensorrt_rtx_version__
     global __tensorrt_llm_version__
+    global __executorch_version__
     with open("dev_dep_versions.yml", "r") as stream:
         versions = yaml.safe_load(stream)
         if (gpu_arch_version := os.environ.get("CU_VERSION")) is not None:
@@ -83,6 +84,7 @@ def load_dep_info():
         __tensorrt_version__ = versions["__tensorrt_version__"]
         __tensorrt_rtx_version__ = versions["__tensorrt_rtx_version__"]
         __tensorrt_llm_version__ = versions["__tensorrt_llm_version__"]
+        __executorch_version__ = versions["__executorch_version__"]
 
 
 load_dep_info()
@@ -206,7 +208,11 @@ else:
 # runtime API is not stable across minor releases, and an unbounded floor would resolve a future
 # minor against a backend built for this one. Patch releases stay allowed because they come off the
 # same release branch; the exact pin belongs in the runtime package, which does derive it.
-EXECUTORCH_REQUIREMENT = "executorch>=1.4.1,<1.5"
+_executorch_major, _executorch_minor = __executorch_version__.split(".")[:2]
+EXECUTORCH_REQUIREMENT = (
+    f"executorch>={__executorch_version__},"
+    f"<{_executorch_major}.{int(_executorch_minor) + 1}"
+)
 # TODO: Enable this once the runtime wheel is published to the PyTorch index.
 # EXECUTORCH_RUNTIME_REQUIREMENT = (
 #     f"torch-tensorrt-executorch-runtime=={__version__}; " "platform_system == 'Linux'"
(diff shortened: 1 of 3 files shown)


lanluo-nvidia (approved):
lgtm

PR #4566, opened 2026-08-23
Track the ExecuTorch pin and move it to new nightlies automatically

shoumikhin:
## Problem

ExecuTorch's package version, its source revision, the development lock, and the installation
instructions all have to agree. When they are updated separately, a build can compile or test
against a different runtime than the one the installed wheel provides, and nothing catches it.

## Change

Pin ExecuTorch to a nightly and to the source revision recorded inside that same published wheel,
so the two cannot disagree.

Add a daily job that proposes the next pin. It adopts a version only when every accepted CUDA
channel publishes it, refuses a version that sorts below the current pin unless a downgrade is
asked for explicitly, and refreshes the development lock so the lock and the pin move together.
A scheduled run reaches only the default branch, so this never re-pins a release branch on its own.
Re-pinning a cut release is a manual run on the stable line, and it refuses once that release has a
final tag, because a shipped pin is history.

Decide which CUDA rows the delegate supports from the CUDA major rather than a hardcoded list. The
delegate links CUDA 13 libraries, so the major is the real requirement and the minor is not. The
matrix filter, the test runner, the install helper and the pin updater all derive it from one
place, so the delegate follows whichever CUDA rows the wheel matrix builds instead of keeping its
own copy.

Naming a channel is not the same as that channel existing. PyTorch publishes a new CUDA minor
before ExecuTorch does, so the install guidance states which CUDA major is required rather than
implying every minor is ready.

## Test plan

Ran the ExecuTorch test directory on Linux.

Confirmed the pinned version and the pinned revision match what the published wheel reports about
itself, and that the development lock names that version with nothing older left behind.

Checked which install channel each CUDA version resolves to, including that a CUDA major the
delegate does not support, an empty value and a malformed minor all resolve to no channel.

Confirmed each new check fails when the thing it protects is removed, including making any single
declaration of the supported CUDA major disagree with the others.

Compared frozen installer selections across the supported Python versions and dependency groups.
Linux passed with no unrelated package changes. Windows kept a pre-existing unrelated failure, so
Windows lock qualification is not complete.


shoumikhin:
Follow-up that depends on this: #4567 (ships only the TensorRT delegate in the runtime wheel). That change needs the prebuilt ExecuTorch runtime this pin brings in, so it should land after this one.

PR #4567, opened 2026-08-23
Ship only the TensorRT delegate in the ExecuTorch runtime wheel

shoumikhin:
## The problem

Shipping the TensorRT delegate for ExecuTorch used to mean shipping ExecuTorch with it. The wheel
carried its own full copy of the ExecuTorch runtime, about 27 MB of it, built for one Python version.

That copy is the problem, not just the size. A user who already has ExecuTorch installed now has two
runtimes in one process. They can disagree, and when they do the failure is confusing: a program
loads, registers a backend against one runtime, and then runs against the other.

## The change

The wheel now builds against the ExecuTorch you already installed, and ships three things: the
delegate library, a small Python package that registers it, and a CMake package so a C++ application
can link the prebuilt library.

The payload drops from about 27 MB to about 130 kB, and one wheel works on every supported Python
instead of one.

## Getting it

One command. The `executorch` extra brings the delegate, ExecuTorch, Torch-TensorRT and PyTorch:

```bash
python -m pip install --pre "torch-tensorrt[executorch]" \
  --index-url https://download.pytorch.org/whl/nightly/cu132 \
  --extra-index-url https://pypi.org/simple \
  --extra-index-url https://pypi.nvidia.com
```

Swap `cu132` for the CUDA version you run. Both extra indexes are needed, and so is `--pre`.
Measured in an empty environment:

| package | version |
| --- | --- |
| torch-tensorrt | 2.15.0.dev20260920+cu132 |
| torch | 2.15.0.dev20260920+cu132 |
| torch-tensorrt-executorch-runtime | 0.2.0.dev20260920+cu132 |
| executorch | 1.6.0.dev20260915+cu132 |
| tensorrt | 11.3.0.99 |
| nvidia-cuda-runtime | 13.2.86 |

ExecuTorch is older than the rest because the delegate pins the exact build it was compiled against,
which is what stops a silent mismatch.

On Arm the delegate needs glibc 2.35 where the others need 2.28, because its build container has no
devtoolset and the tag states that rather than claiming compatibility it does not have.

## Using it from Python

Importing the package registers the delegate. There is nothing else to call:

```python
from pathlib import Path

import torch
import torch_tensorrt_executorch_runtime  # noqa: F401
from executorch.runtime import Runtime

program = Runtime.get().load_program(Path("model.pte"))
forward = program.load_method("forward")
outputs = forward.execute((torch.ones((2, 3, 4, 4)),))
```

## Using it from C++

Find the package by name and link it, the way ExecuTorch names its own backends:

```cmake
find_package(executorch REQUIRED COMPONENTS backend_cuda kernels_optimized)
find_package(executorch_backend_tensorrt REQUIRED)

target_link_libraries(my_app PRIVATE
  executorch::runtime
  executorch::backend_cuda
  executorch::backend_tensorrt
  executorch::kernels_optimized
)
```

No header to include: the delegate registers itself when the library loads, and the rest is
ExecuTorch's own API.

```cpp
using namespace executorch::extension;

Module module("model.pte");
auto input = make_tensor_ptr({2, 3, 4, 4}, data.data());
const auto outputs = module.forward(input);
```

## Two behaviour changes worth knowing

Execution always waits. The runtime this plugs into has no asynchronous execute, so there was an
option to return early that could not be honoured safely, and it is gone.

A buffer on the wrong GPU is refused rather than written to. It used to be bound and then fail inside
TensorRT as an invalid program, which sends you to check a model that was fine.


lanluo-nvidia (approved):
please fix the failed executorch test cases, others lgtm

PR #4572, opened 2026-08-24
ci(executorch): run a coalesced TensorRT + CUDA program in the reference runner gate

shoumikhin:
# Description

The ExecuTorch gate exports one program, `x + 1`, and TensorRT takes it whole. So
nothing in CI ever runs a program where TensorRT and ExecuTorch's own CUDA backend
each own part of the same graph. That coalesced case is the whole point of
combining the two backends, and it is not covered end to end today. There is a
composition test that checks both delegates land in the file, but it never loads
or runs the program.

This adds the missing run.

A new example, `examples/torchtrt_executorch_example/export_coalesced.py`, exports
`cos(erfinv(tanh(x)))`. TensorRT has no converter for `erfinv`, so a
`CudaPartitioner` catch-all gives that operator to the CUDA backend while TensorRT
keeps the rest. The script fails if the saved `.pte` does not carry both a
`TensorRTBackend` and a `CudaBackend` delegate, so a partitioning change cannot
quietly turn this into a TensorRT-only run that still passes.

The script also writes `<model>.expected` next to the `.pte`, holding the output
shape and the eager reference value for an all-ones input. Both reference runners
fill inputs with 1.0 and this model is elementwise, so one number describes the
whole expected output. Reading it from a file, instead of hard-coding a number in
the shell script, keeps the expectation tied to the model.

`verify-executorch-reference-runner.sh` now takes an optional third argument, the
coalesced `.pte`. When given, it runs the runner built from source against it. Only
runner on it and compares every printed value against that reference. TensorRT,
AOTInductor and eager PyTorch use different kernels for the same math, so the
comparison uses a tolerance of 0.001 rather than matching printed digits.

The existing `x + 1` assertions keep the same strength. They now go through the
same helper with a zero tolerance, because `x + 1` on ones is exact in float32.

Usage:

```bash
python examples/torchtrt_executorch_example/export_coalesced.py \
  --model_path=coalesced.pte

.github/scripts/verify-executorch-reference-runner.sh \
  model.pte kv_cache_decode.pte coalesced.pte
```

## Type of change

# Test plan

On a Linux x86_64 host with an NVIDIA A100 GPU:

- Ran `export_coalesced.py`. It reported delegates
  `['TensorRTBackend', 'CudaBackend', 'TensorRTBackend']` and wrote `[64,64]` and
  `0.6722` into the `.expected` file.
- Ran the resulting `.pte` through the reference runner. It printed
  `output[0] shape=[64,64]` and first 8 values of `0.6722`, an exact match to the
  eager result.
- Deleted the `aoti_cuda_blob.ptd` that the CUDA backend writes and ran again.
  Same output, so this model needs no external weight file.
- Exercised the new shell assertion helper against captured runner output:
  correct output passes; one wrong value fails; a wrong shape fails; a missing
  values line fails; a value inside the tolerance passes and one outside it fails.
- `shellcheck`, `bash -n`, `black` and `isort` are clean on the changed files.

Observed in CI: an earlier push of this branch ran the coalesced case green and
printed `output[0] shape=[64,64]` with `0.6722` repeated eight times, which is the
expected reference value. That is the strongest evidence here.

At the current head the runtime build job is green and did exercise the changed
script. The runtime test job is not skipped: it runs and fails while loading
`libcurand.so.10`, before any line this change touches. That failure tracks the
torch nightly rather than this change, and the merge base is green on every
ExecuTorch build and test row, so this row should not be recorded as a known red
one. A green run at the current head is still wanted before merging.

# Checklist:

### DIFF
--- a/.github/scripts/verify-executorch-reference-runner.sh
+++ b/.github/scripts/verify-executorch-reference-runner.sh
@@ -18,6 +18,13 @@ set +x
 #   examples/torchtrt_executorch_example/export_kv_cache_decode.py). When given,
 #   kv_cache_decode_check is built and run against it as well.
 #
+# Optional third argument: path to a coalesced TensorRT + CUDA .pte (see
+#   examples/torchtrt_executorch_example/export_coalesced.py). When given, the
+#   runner built from source here is run against it and its output is compared to
+#   the eager reference that export script wrote next to the model. Only that
+#   runner: the packaged binary links the TensorRT delegate alone, so it has no
+#   CUDA backend for the partition a coalesced program hands to one.
+#
 # Optional:
 #   TensorRT_ROOT=/path/to/extracted/TensorRT
 #     If unset, the script reuses Bazel's fetched TensorRT SDK when available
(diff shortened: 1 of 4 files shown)


lanluo-nvidia (approved):
lgtm

PR #4575, opened 2026-08-25
fix(dynamo): gate refittable cumsum on RTX version

SandSnip3r:
# Description

Guard against broken TRT-RTX 1.6 behavior (static-dim cumsum) with a Validator.


lanluo-nvidia (approved):
lgtm

PR #4579, opened 2026-08-26
fix(dynamo): keep `x & False` and `x | True` out of TensorRT

shoumikhin:
## What is broken

Four converter tests are red on `main`:

- `TestBitwiseAndConverter::test_bitwise_and_scalar_3d`
- `TestBitwiseAndConverter::test_bitwise_and_scalar_tensor_3d`
- `TestBitwiseOrConverter::test_bitwise_or_scalar_2d`
- `TestBitwiseOrConverter::test_bitwise_or_scalar_tensor_2d`

They fail on the value, not with an error: `Mismatched elements: 5 / 30`. The AND cases are the ones parameterized with `False` and the OR cases are the ones parameterized with `True`. The other scalar value passes in both cases. This started with the upgrade from TensorRT 11.1 to 11.2.

## Why

`x & False` is `False` whatever `x` is, and `x | True` is `True` whatever `x` is, so TensorRT folds the layer down to a constant. It gets that fold wrong when the constant operand is smaller than the output and has to broadcast. A Python scalar always reaches the network as a rank 0 constant, so it always has to broadcast, so those two combinations always hit it.

Checked directly against the TensorRT API on 11.2.1.2, outside torch-tensorrt: a bool constant of shape `(1, 1, 1)` against a `(5, 3, 2)` bool tensor aborts the build with an internal error for AND with `False` and for OR with `True`, and is correct for every other combination, including both XOR cases. Give the same constant the full output shape, so nothing has to broadcast, and all cases are correct. Inside torch-tensorrt the build succeeds and the engine returns garbage.

So this is a TensorRT issue, and the two combinations that hit it are exactly the two that are red.

## Fix

The capability validator already rejects the `Tensor` overload when `other` is rank 0, for the same underlying reason. Extend it to the two scalar combinations that are broken, and only those, so the partitioner keeps them in PyTorch. Everything else still goes to TensorRT.

## Tests

The validator unit tests now cover all eight scalar combinations of AND, OR and XOR across both scalar overloads. The scalar converter tests keep the values TensorRT gets right, which is what shows the fallback is not wider than the problem. The two broken values move to new tests that compile through the full pipeline, because the converter harness bypasses the partitioner and never consults the validator; they assert that no engine is built and that the result matches eager.

All six new tests were checked to fail when the validator change is removed, and the five bitwise test files pass together with it (51 tests).

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -2821,6 +2821,31 @@ def aten_ops_logical_xor(
     )
 
 
+# `x & False` is False whatever x is, and `x | True` is True whatever x is. For
+# those two, TensorRT 11.2 folds the layer down to a constant, and it gets that
+# wrong when the constant operand is smaller than the output and has to
+# broadcast: the build either aborts with an internal error or the engine
+# returns garbage. A Python scalar always reaches the network as a rank-0
+# constant, so it always has to broadcast, so these two combinations always hit
+# it. Nothing else does, including both XOR cases, since XOR has no value that
+# fixes its result.
+_ABSORBING_BITWISE_SCALAR = {
+    torch.ops.aten.bitwise_and.Scalar: False,
+    torch.ops.aten.bitwise_and.Scalar_Tensor: False,
+    torch.ops.aten.bitwise_or.Scalar: True,
+    torch.ops.aten.bitwise_or.Scalar_Tensor: True,
+}
+
+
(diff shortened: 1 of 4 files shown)


py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
lanluo-nvidia, line 2880: @shoumikhin  
I thought this section should be able to pick up True or False scalar value and fall back to use the torch ops instead of tensorrt since it is a tensorrt bug.

lanluo-nvidia (approved):
lgtm

PR #4580, opened 2026-08-26
fix(dynamo): keep the complex decomposition inside the TRT op set

shoumikhin:
## What is broken

`tests/py/dynamo/lowering/test_aten_lowering_passes.py::TestComplexSubgraph::test_complex_subgraph` is red on `main`. The test compiles a rotary attention module and expects one TensorRT engine; it gets three, with PyTorch blocks in between.

## Why

The complex lowering pass hands the interior expansion to PyTorch's `decompose_complex_in_graph`, which re-traces the module with `make_fx`. The module that comes back is therefore built from whatever the complex tensor subclass dispatched to, not from the ops that survived the earlier stages of the same pass list. Three kinds of leftovers reach the partitioner:

- `aten.stack` and `aten.alias`, which the decomposition table run at export time expands and drops, are back;
- constants that constant folding had already reduced to one frozen tensor are live chains of ops again, and the converters for those ops want a TensorRT tensor rather than a weight;
- `aten.view_as_real` and `aten.view_as_complex` are left at the seams.

The last one only shows up when the complex region is interior, meaning the graph takes real tensors and returns real tensors, which is what a rotary embedding does. The boundary normalizer only knew about `aten.complex`, `aten.real` and `aten.imag`, which is what a graph with complex inputs or outputs meets at its edges.

None of these have a TensorRT converter, so the partitioner cuts the graph around them.

## Fix

Three parts, all in the lowering pass:

1. pass the compilation settings' decomposition table to the retrace, so the retraced graph lands back in the same op set the rest of the pass list works in;
2. re-run the constant fold exclusion marking and the constant fold after the retrace, so the constants that were frozen before are frozen again;
3. teach the boundary normalizer the two `view_as_real` seams: `view_as_real(view_as_complex(x))` folds to `x`, and `view_as_real(complex(re, im))` folds to the interleaved real layout the rest of the flow expects.

## Tests

The failing test passes again: the rotary attention module builds a single engine and matches eager output.

Full runs: `tests/py/dynamo/lowering/` 326 passed, and the complex, rope and graph break suites 189 passed.

Three new unit tests cover the added folds and the post-retrace constant folding. All three fail without this change and pass with it.

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/passes/complex_decomposition_adapter.py
+++ b/py/torch_tensorrt/dynamo/lowering/passes/complex_decomposition_adapter.py
@@ -8,9 +8,13 @@ interior complex -> real expansion to PyTorch's ``decompose_complex_in_graph``
   1. capture the complex I/O signature before rewriting (drives the boundary
      adapters in ``_compiler._insert_complex_io_adapters``);
   2. call upstream to expand every complex op into real ops on separate re/im;
-  3. normalize the SoA seams (``aten.complex`` / ``aten.real`` / ``aten.imag``)
-     into the interleaved ``[..., 2]`` layout the rest of the TRT flow expects,
-     so the engine only ever sees real tensors (Option A in the RFC).
+  3. put back what the retrace undid: upstream returns a module rebuilt from
+     scratch, so the decomposition and constant-folding stages that ran earlier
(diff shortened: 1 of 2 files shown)


lanluo-nvidia (approved):
lgtm

PR #4581, opened 2026-08-26
fix(executorch): register device copy kernels in the packaged runner

shoumikhin:
## What is broken

The `executorch-runtime-build` jobs are red on `main`, 10 of them. The example runner builds fine and then fails at run time:

```
kernel 'et_copy::_h2d_copy.out' not found
load_method('forward') failed with error 0x14
```

## Why

An exported program that is fully delegated to TensorRT still runs two operators outside the delegate: the host to device copy of its inputs and the device to host copy of its outputs. The device placement pass inserts them at export time.

The Bazel build of the example runner links the core ExecuTorch runtime only, and the core runtime carries no kernels, so those two operators have nothing to dispatch to and loading any exported program fails. The CMake build does not hit this because it links ExecuTorch's own kernel library.

## Fix

Compile the two kernel implementations from the pinned ExecuTorch tree and register them from a small registration-only library that executables depend on directly.

Two deliberate choices:

- only these two kernels are registered, not the whole portable kernel set, because a fully delegated program never calls the rest;
- registration lives in its own library rather than in the backend library, because the kernel registry aborts when the same kernel is registered twice and the backend can be linked into both a shared library and an executable. It is also kept out of the CMake source list, since that build already links a kernel library and would hit exactly that double registration.

## Verification

No GPU was available for an end to end run, so this was checked at the symbol level: both translation units compile against the pinned ExecuTorch headers, the registration object's undefined symbols match the kernel object's defined symbols, and the registered names are exactly `et_copy::_h2d_copy.out` and `et_copy::_d2h_copy.out`.

The end to end path is covered by the CI job that runs the example runner, which is the job that is currently failing.

### DIFF
--- a/cpp/BUILD
+++ b/cpp/BUILD
@@ -172,6 +172,25 @@ cc_library(
     }),
 )
 
+# Registration only, for the same reason as the allocator above: the kernel
+# registry aborts when the same kernel is registered twice, so this cannot ride
+# along with the backend. Executables that run a TensorRT delegated program
+# depend on this directly.
+#
+# Not part of :executorch_backend_source_files. The CMake build of the reference
+# runner links executorch::kernels, which already registers these.
+cc_library(
+    name = "tensorrt_executorch_device_copy_kernels",
+    srcs = [
+        "src/torch_tensorrt/executorch/RegisterDeviceCopyKernels.cpp",
+    ],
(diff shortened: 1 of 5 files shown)


lanluo-nvidia (approved):
lgtm

PR #4582, opened 2026-08-26
style: format test_cumsum_aten.py with black

shoumikhin:
## What is broken

The `Python Linting` job runs `black --check .` over the whole repository. One
file is not formatted, so the job fails, and it fails the same way on every open
pull request no matter what that pull request changes:

```
would reformat tests/py/dynamo/conversion/test_cumsum_aten.py
1 file would be reformatted, 602 files would be left unchanged.
```

## Fix

Run black on that file. Two call sites are split across several lines but fit on
one, so black joins them. Nothing else changes, and there is no behavior change.

## Tested

`black --check .` no longer reports this file.

### DIFF
--- a/tests/py/dynamo/conversion/test_cumsum_aten.py
+++ b/tests/py/dynamo/conversion/test_cumsum_aten.py
@@ -33,9 +33,7 @@ class TestCumsumConverter(DispatchTestCase):
                 )
             return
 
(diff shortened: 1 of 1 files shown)


lanluo-nvidia (approved):
lgtm

PR #4583, opened 2026-08-26
fix(dynamo): stop unpacking complex inputs before the refit output check

shoumikhin:
## What is broken

Three tests are red on `main`:

- `test_complex_buffer_refit`
- `test_dual_complex_buffer_refit`
- `test_complex_buffer_with_real_param_refit`

```
TypeError: view_as_real is only supported for complex tensors
```

## Why

After refitting, `refit_module_weights(verify_output=True)` runs the lowered
PyTorch module and the compiled module on the same inputs and compares the
results. When an input is complex it first converted that input to a real tensor
of shape `(..., 2)`, because the lowering pass used to rewrite a complex
placeholder into a real one.

That is no longer what the pass does. The placeholder stays complex and the
graph does the unpacking itself:

```
placeholder      arg0_1
call_function    aten.view_as_real.default   (arg0_1,)
call_function    aten.select.int             (view_as_real, 2, 0)
...
```

So the reference module got a real tensor and was then asked to unpack it a
second time, which is what raises.

## Fix

Give both modules the inputs the caller passed. That is what the non-complex
path already did, so the special case goes away and the two paths become one.

## Tested

On an H100 with the three tests above: they fail before this change and pass
after it. The whole file passes, 7 passed and 12 skipped, the skips being the
ones `main` already skips in this environment.

### DIFF
--- a/py/torch_tensorrt/dynamo/_refit.py
+++ b/py/torch_tensorrt/dynamo/_refit.py
@@ -37,7 +37,6 @@ from torch_tensorrt.dynamo.runtime._TorchTensorRTModule import TorchTensorRTModu
 from torch_tensorrt.dynamo.runtime._TRTEngine import TRTEngine
 from torch_tensorrt.dynamo.utils import (
     check_module_output,
-    check_output_equal,
     get_torch_inputs,
     to_torch_device,
     to_torch_tensorrt_device,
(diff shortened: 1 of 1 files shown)


lanluo-nvidia (approved):
lgtm

PR #4585, opened 2026-08-26
fix(runtime): back a python engine runtime cache with the python handle

shoumikhin:
## What is broken

Two tests are red on `main`, RTX only:

- `runtime/test_aliased_io.py::TestPythonRuntimeAliasedIO::test_in_place_write_through`
- `runtime/test_aliased_io.py::TestPythonRuntimeAliasedIO::test_streaming_state_accumulates`

```
TypeError: set_runtime_cache(): incompatible function arguments.
  1. (self: IRuntimeConfig, cache: nvinfer1::IRuntimeCache = None) -> bool
Invoked with: <IRuntimeConfig object at 0x...>, None
```

## Why

`RuntimeCache` chooses its backing from whether the C++ library is loaded:

```python
if ENABLED_FEATURES.torch_tensorrt_runtime:
    self._handle = torch.classes.tensorrt.RuntimeCacheHandle(path)
else:
    self._handle = _RuntimeCacheHandle(path=path)
```

A loaded library does not mean a C++ engine is the consumer. These tests build
a python `TRTEngine` straight from a packed engine tuple, which is a documented
way to use it. Its default settings carry a cache path string, so
`TRTRuntimeConfig._apply_settings` wraps that path, gets the torchbind handle,
and asks it for a live cache. The torchbind handle materializes its cache on
the C++ side and cannot pass an `IRuntimeCache` back to python, so
`ensure_cache` returns `None`, and `None` is not a cache.

## Fix

Let the caller say which runtime will consume the cache. `_apply_settings`
only ever runs for a python engine, so it asks for the python backing. Nothing
else changes: every existing caller keeps the old default.

One related detail: `IRuntimeConfig` does not own the cache attached to it, so
the wrapper now lives on the config instead of being dropped at the end of the
call. Before this change the wrapper was released immediately, which was
harmless only because the attach never happened.

## Tested

The RTX path cannot run on this machine. What was checked here, with the C++
library loaded:

- default backing: `ensure_cache` returns `None`, which is the exact input the
  error message reports
- python backing: `ensure_cache` returns the live cache

`test_000_runtime_cache.py`, `test_004_runtime_settings.py` and
`test_aliased_io.py` give the same result before and after the change on
non-RTX TensorRT, 28 passed and 27 skipped.

### DIFF
--- a/py/torch_tensorrt/runtime/_runtime_cache.py
+++ b/py/torch_tensorrt/runtime/_runtime_cache.py
@@ -222,7 +222,6 @@ class RuntimeCache:
     ``ENABLED_FEATURES.torch_tensorrt_runtime`` is set, else the
     pure-Python :class:`_RuntimeCacheHandle`; both satisfy
     :class:`_RuntimeCacheHandleProtocol` interface which this class uses.
-
     **Identity-based equality.** Two handles wrap distinct underlying
     ``IRuntimeCache`` instances even when they share a path, and the
     runtime treats them as such.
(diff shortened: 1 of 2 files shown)


lanluo-nvidia:
@shoumikhin 
since the fix from rtx folks is already in, do you think you can close this PR.
https://github.com/pytorch/TensorRT/commit/6d26eabec2a5f786a60edd6621bbffd314545af3

PR #4600, opened 2026-08-26
perf(executorch): opt-in shared per-device activation-scratch pool

Conarnar:
# Description

A multi-layer model lowered to the ExecuTorch TensorRT delegate becomes N separate single-layer engines, and by default every execution context allocates its own activation scratch and holds it for as long as the context lives. Device memory scales with the layer count, and multi-layer models OOM at runtime on the layer count alone.

This adds an **opt-in** shared per-device pool: every context created while the option is on takes its activation scratch from one buffer, grown to the largest figure any call on that device has asked for. It **ships disabled** — with `use_shared_activation_scratch` unset each context keeps its private `kSTATIC` scratch, no extra log output is emitted, and the only added work is one relaxed atomic load per engine init and, per `execute()`, four tests of the handle's `claims_pooled_scratch` — three `if` branches and one in a scope guard's destructor — plus an empty `SharedScratchClaim` released and destroyed without ever having held anything. A context whose engine reports needing no activation scratch under any shape stays outside the pool entirely.

Whatever `updateDeviceMemorySizeForShapes()` answers is binding rather than advisory: `setDeviceMemoryV2` refuses a smaller buffer, and an engine backed by less than it asked for writes past the end. Whether it answers what the shapes just bound need or the engine's profile maximum is settled when the engine is built — `PreviewFeature::kRUNTIME_ACTIVATION_RESIZE_10_10` produces the former, without it either can happen — so the pool can settle well above the live data and no runtime can change that.

**Every figure here comes from local runs on one A100 host**, TensorRT 11.2.1.2 and CUDA 13. The CI lane that would build and run these tests has not executed them at this head, so nothing here has a CI run behind it.

Memory from `cudaMemGetInfo` after a `cudaFree(0)` baseline on one 80GB A100. N per-engine copies collapse to one, so what is reclaimed is the sum of the N per-engine requirements less the largest of them:

- four execution contexts of one engine holding two fp32 8-head attention blocks over `[1,2048,512]`: **1188MB → 372MB**. Uniform engines, so that sum-less-largest is 3 × 272MB.
- one Method holding six one-block engines of the same shape interleaved with six CUDA delegates: **1656MB → 316MB**. Also uniform: 5 × 268MB.
- the same shape of Method with its six engines at differing sizes: **792MB → 316MB**. Sum 780,140,544 B less the largest 281,018,368 B is 476.0MB, which is what the A/B reads.

Outputs are identical between the two modes in every case.

The default path was A/B'd against a binary built from the commit this PR branches from, over nine runs (three models, three repetitions each): canonicalised stderr identical every time, at 45 / 45 / 17 lines. That stderr is the whole of the evidence — this harness writes nothing to stdout — and the canonicaliser normalises timestamps, pointers, source line numbers and the harness's own device-wide memory reading, which moves by a tenth of a megabyte between runs of one binary. The control, the same binary with the option on, differs at 47 lines against 45, so the comparison can fail.

Once the option is on, the pool never shrinks — a growth disposes of the buffer it replaces, but the high-water mark never comes down — so a device keeps the largest scratch it was ever asked for until the process exits, where per-context `kSTATIC` scratch is released with its context. It also keeps a handoff `cudaEvent_t` for every device a pooled call was ever made on, and, for every device that ever grew, the CUDA stream its growths queue their frees on — neither ever destroyed, for the same teardown-order reason as the buffer: nothing here runs a CUDA call at process exit. A growth allocates the new buffer before disposing of the old, so both are resident for that moment, which is what leaves the existing buffer usable when an allocation fails. Every growth then disposes of the old buffer the same way, and where the buffer it retires has an enqueue recorded against it the first of that disposal's three steps is a host wait for a whole inference with no upper bound. First known gap.

No dependencies; this does not stack on anything. Orthogonal to weight streaming (#4336), which targets engine *weight* memory rather than activation scratch. It composes with the zero-copy KV work if that lands too — measured together on the same model, byte-identical generated ids — but the two branches touch one hunk in common: this branch computes `must_sync` before the scratch claim rather than after the enqueue, and the zero-copy branch rewrites the comment at the old site, so `TensorRTBackend.cpp` conflicts, and resolves by keeping the hoisted computation with the zero-copy comment moved up to it.

### Where I would spend review attention

**The enqueue handoff, not the allocation.** Contexts share one buffer while their enqueues can still be in flight, so each device slot carries a pool-owned `cudaEvent_t`: wait on it before enqueueing, record after. An earlier revision tracked the last *stream* and was wrong three ways — synchronizing a destroyed stream segfaults rather than returning an error, CUDA recycles stream handle values so two distinct streams compare equal, and the NULL stream is a legal caller stream indistinguishable from "no previous user". All three are structural with an event, and `~EngineHandle` in the same file had already made this choice.

**A reported zero is ambiguous.** `updateDeviceMemorySizeForShapes()` answers zero for three things: a failed query, bound shapes that need no scratch (a valid empty input on a dynamic profile), and an engine needing none under *any* shape. Only the last can be told apart before any call is made, and it is, once at init: each engine loaded with the option on records whether `ICudaEngine::getDeviceMemorySizeV2()` is non-zero — `claims_pooled_scratch`, false on a handle loaded with the option off because nothing was ever asked there — and one reporting zero is left out of the pool entirely, so it claims no per-device lock and does not serialize against the engines that do.

The other two reach `execute()` indistinguishable, and neither can be handed nothing. `setDeviceMemoryV2(nullptr, 0)` is refused (`condition: memory != nullptr || expectedSize == 0`), so a context handed a zero keeps whatever buffer it last held, which a growth may already have freed — forcing that path reproduces an illegal memory access. And `enqueueV3` refuses a `kUSER_MANAGED` context with no device memory installed as soon as the engine needs some under *any* shape, however little the bound shapes need: `Parameter check failed, condition: noDeviceMemory`. So neither sizes the pool: a zero is handed whatever the pool already holds at its current size, growing nothing; only against an empty pool does it allocate, and then one byte, leaving the first real requirement to size it. Standing the engine's own `getDeviceMemorySizeV2()` in for the zero instead satisfies `enqueueV3` but covers the whole profile: on the dynamic engine the tests build that is 4 MiB against the 128 KiB the next call needs, and since the pool never shrinks a single empty batch would pin it at 32x for the process lifetime, on the path the pool exists to keep small.

**Only one of the two is safe with a buffer that small, and the install separates them.** Where the shapes genuinely need nothing the engine expects nothing and one byte is accepted; where the query *failed* the engine still expects what it always did and `setDeviceMemoryV2` refuses — but it returns `void`, so a caller that does not ask cannot tell them apart. Measured on TensorRT 11.2.1.2 against this branch's two-softmax net (expecting 8388608 bytes): install one byte and `enqueueV3` returns true, `cudaStreamSynchronize` reports no error, output matches byte for byte, and the engine writes the buffer the context was last given; free that buffer, let the next `cudaMalloc` take the address, and 4090 of 4096 bytes of an unrelated allocation are overwritten with nothing reporting an error anywhere. So the install is read back through a TensorRT `IErrorRecorder` attached for that one call: a correct install records nothing, an undersized one records `kINVALID_ARGUMENT` naming both sizes, and the genuinely-empty case (expected size zero, one byte installed) records nothing and keeps working. A refusal ends the call with `Error::InvalidState` and nothing reaches `enqueueV3`. Because the refusal is observable, the profile-wide figure never has to be installed.

**What is installed is the call's own requirement, not the pool's capacity.** They differ after any growth, and the pool never clears the buffer between users, so the larger figure would tell TensorRT it owns bytes still holding another engine's activations and would hide an engine overrunning its own requirement. It would also blunt the check above: after a growth the capacity can cover exactly what a failed query concealed. The choice is not observable from a test — TensorRT accepts an oversized install in silence and offers no getter for the installed size — so what the suite sees is the other direction, in `ALargerEngineGrowsThePoolAndFreesTheBufferItReplaces`, where the smaller engine runs again after the growth and installs less than the pool holds.

**Two hazards the pool does not absorb.** CUDA graph capture is not supported around this delegate, with the option on or off, and the shipped documentation says exactly that rather than pointing a caller anywhere. With the option on, a pooled `execute()` whose selected stream is capturing is refused with `Error::NotSupported`. A query that cannot answer is told apart from a capture: `cudaStreamIsCapturing` also hands back a sticky fault left by earlier work on the device, and a call that meets one fails with that fault's own message and `Error::InvalidProgram` rather than being told it is capturing. Everything here was measured directly against CUDA 13.0 on this device, one process per cell:

- the handoff wait is on an event recorded outside the capture, so `cudaStreamWaitEvent` fails it with `cudaErrorStreamCaptureIsolation` and invalidates the capture under *every* mode, `Relaxed` included — the caller learns of that only from `cudaStreamEndCapture` returning 901 and a null graph, far from the cause. That wait, not the allocation, produced the 901 when the guard was removed and measured: the pool was already large enough, so no growth ran.
- a growth's `cudaMalloc` returns `cudaErrorStreamCaptureUnsupported` and invalidates under Global and ThreadLocal, while Relaxed permits it and runs it uncaptured, leaving a replay pointed at a buffer the pool may since have freed.
- a growth's disposal is prohibited at its first step, the host wait on the handoff event, and the `cudaFree` it falls back to behaves like the `cudaMalloc` above. Queuing the free on the pool's own stream is no way round that, because the wait ahead of it has already happened.

The guard is one `cudaStreamIsCapturing` on the selected stream, placed ahead of everything `execute()` does that a capture cannot take, not merely ahead of the pool's own calls: the host wait on a previous enqueue and the `cudaMalloc` that grows a host-input staging buffer come earlier and, outside `Relaxed`, would invalidate the capture first. Only the device query and the device switch run before it, and a capture takes both.

Loading with the option off is not a way round any of this. Nothing refuses there — the guard sits inside the pooled branch — but the delegate does not support capture that way either, and this PR does not change what the option-off path does with a capturing stream: the completion event `execute()` records would become a node of the graph rather than an event the host can wait on. Making that path work means reworking the completion event, which the caller-stream contract's asynchronous return rests on; moving the guard out of the pooled branch instead would refuse calls for contexts that never touch the pool.

The check cannot cover a capture live on some *other* stream: under Global, and under ThreadLocal from the calling thread, the same calls invalidate it and the call is not refused. `cudaStreamIsCapturing(cudaStreamLegacy)` does not close the gap — measured, it reports another stream's capture only when that stream is a blocking one, and then reports it under Relaxed and under another thread's ThreadLocal too, where the pool's calls are permitted and refusing would be wrong. CUDA offers no "is a capture live in this process" query, so the rule is the caller's to keep and the installed header and the README say so. `cudaDeviceReset()` is documented in both rather than guarded: the pool holds its buffer, its handoff event and its disposal stream for the process lifetime and a reset destroys all three, and catching that would mean revalidating all three on every call, which costs as much as the work it protects.

**The lock scope.** The registry that finds a device's entry has one lock, held for a single find-or-insert with no CUDA call under it. The device's own lock is held from the claim through `setDeviceMemoryV2`, the enqueue, and the record of that enqueue on the handoff event — releasing earlier leaves an enqueue live in a window the event does not yet cover, and a second claimant entering that window is handed the same buffer with nothing ordering the two. That produced silently wrong output on every trial before the fix. The lock nests inside the per-handle `EngineHandle::mu`, which already spans the enqueue, and is never taken in the other order. Entries are never erased and `std::unordered_map` keeps references valid across rehashing, which is what lets the registry lock be dropped before the entry is used.

**A failing pooled call no longer returns with a copy of the caller's memory still running.** A host-resident input is staged with a `cudaMemcpyAsync` reading the caller's buffer, which the caller owns again the moment `execute()` returns. The success path synchronizes whenever anything was staged; the error returns between the copy and that point did not. From pinned memory that is live, not theoretical: measured on a 16 MiB copy on a gated stream, the copy returned in 0.0 ms having read nothing, and 4194304 of 4194304 elements the device received were what the caller wrote after the return; the same copy from pageable memory blocked for 1005.8 ms and took the bytes it was called with. A scope guard declared with the staging flag and the call's pooled flag now synchronizes on destruction unless the call reached the block that was already going to. It is scoped by the call rather than by the return: a pooled call is covered at *every* return in between, and an unpooled one at none of them. That scoping is what leaves an engine loaded with the option off waiting nowhere it did not already wait, and it is the same scoping the capture refusal has. The hazard is reachable on an unpooled call too and is not closed here: closing it would change what a default-off call does on every error return it makes past that copy, which is a change of its own to make. Of the six error returns *after* the enqueue, the four that are not already behind a synchronize drain the stream through one helper that clears the flag too, so they synchronize once rather than twice, and no path that did not already synchronize gains one. `AFailedPooledCallDrainsTheHostInputCopyItQueued` drives the one early return in the pooled scratch block that is reachable in-process — the pool's allocation, with the device full — and reads back what the device ended up holding; the failing call gets a different input from the successful one before it, so "the copy never ran at all" cannot pass either. `AFailedUnpooledCallLeavesTheHostInputCopyItQueuedAlone` pins the other half, on the one failure this suite can force past an unpooled call's staging copy — the output staging buffer's `cudaMalloc`, with the device full. Neither is redundant: un-gating the guard fails the unpooled case and leaves the pooled one green, and disabling it outright does the reverse.

**The per-handle capture.** A context's allocation strategy is fixed at creation, so each `EngineHandle` records the setting in effect at its own `init()` and `execute()` consults that, never the global. That is what lets a later `set_option` govern only subsequent engines and lets pooled and private-scratch contexts coexist, with no freeze and no rejected calls.

**What the release package exposes.** `SharedScratchPool.h` was in `executorch_api_headers` to deliver one string constant — `TensorRTBackend.h` never included it and named the option key only in comments. The key now sits beside `set_option`, and the header moved to `cpp/src/torch_tensorrt/executorch/`; `install_pooled_scratch` moved with it, into `PooledScratchInstall.h` — a separate header rather than the pool's, because it names `nvinfer1::IExecutionContext` and the pool header is kept buildable against the CUDA headers alone. Built `//:include_executorch`, the header half of the release tarball, and read the tar: five headers, none mentioning it. Dropping either header from the package was not open — `libtorchtrt.tar.gz` ships `TensorRTBackend.cpp` as buildable source and that source includes both — so they move rather than leave, and every header remaining in `cpp/include/` is installed, the invariant the original placement broke. The symbol is still a global `T` in the release object, because the backend's test links against it to cover a refusal `execute()` cannot be made to produce. The pool's two test-only entry points — one of which frees the live pool with no wait for work in flight — were defined in `TensorRTBackend.cpp` and so exported from the archive the README tells a consumer to build. They now live in a `SharedScratchPoolTestHooks` translation unit only a `testonly` Bazel target compiles, and `nm`/`readelf` on the release object shows neither symbol. The reset's *body* moved likewise, to a `testonly` `SharedScratchPoolReset.h` in neither the packaged include tree nor the packaged source set; the shipped header keeps the friend declaration alone. A three-line consumer built against only the packaged sources no longer compiles.

**`third_party/cuda/BUILD` gains a target.** The pool header needs the `cudaEvent_t` typedef — a compile-time dependency, not a runtime one — and the repo had no headers-only CUDA target. Depending on `cudart` instead put `libcudart` in the `DT_NEEDED` of a host-side test that makes no CUDA call, and broke it with exit 127. The new target is also the only place the toolkit header glob is now written: the three targets beside it that repeated it depend on it instead.

### Known gaps

- **A growth waits out an inference on the host, and that is documented rather than fixed.** A growth must dispose of the buffer it replaces before `execute()` returns, or the bytes of every size the pool ever grew to stay resident at once. Every growth disposes of it the same way: wait on the host for the enqueue the handoff event names, queue a `cudaFreeAsync` on a stream the pool owns for that device, and synchronize that stream, which is what returns the bytes. The host wait is the cost wherever the retired buffer has an enqueue recorded against it — where it has none, which is what a pooled call that returned between taking the buffer and recording its enqueue leaves behind, the handoff carries no recording and the disposal goes straight to the free. Where it is paid it is a whole inference: a call that reached its own enqueue has recorded that enqueue on the handoff event before releasing the claim, so the wait covers this call's inference as well as the one before it and such a growth returns with its own engine work finished rather than in flight. It has no upper bound, and a caller can make it permanent — park a host function ahead of the growing engine's enqueue, release it only once `execute()` has returned, and the growing call never comes back, the thread that would unblock it being inside the wait. What a growth does *not* wait for is the rest of the device, which is what the pool's own stream buys: measured on CUDA 13.0 with driver 13.0, freeing a 512 MiB buffer behind a 10 ms enqueue with a host function parked for 1500 ms on a stream nothing else in the measurement submitted to, 1500.8–1501.4 ms device-wide against 10.6–11.7 ms on the pool's stream, both with every byte back by the time the disposal returned. That device-wide wait is not hypothetical: the backend's own tests deadlocked on it once, when a change of test order made a case that parks a host function on its own stream the one that grew the pool. The stream has to be the pool's rather than the caller's, because a `cudaFreeAsync` of a `cudaMalloc`'d pointer — which is what this pool holds — hands the bytes back at the next `cudaStreamSynchronize` of the stream it was queued on and at no point before it — measured, a stream `cudaStreamQuery` reports as drained still holds the bytes, `cudaMemGetInfo` does not move, and a `cudaMalloc` of the same size fails with out of memory until the synchronize — so whoever queues the free has to be able to promise that synchronize, and only the pool can. On the caller's stream the disposal costs 0.0 ms and leaves the whole 512 MiB resident on return; where the caller does synchronize, the call reaches the same point at the same time (10.4–11.8 ms against the 10.6–11.7 above), and where it does not — the caller-stream, device-resident path this option exists for — the bytes never come back: measured, replaying the pool's growth sequence on such a stream — 8, 16, 24 and 32 GiB with no synchronize between them — turns four growths that all succeed when the disposal returns the bytes into an out-of-memory on the fourth, free stuck at 30.83 GiB on a device that started with 78.83 GiB of it. That last is a property of the pointer rather than of stream-ordered freeing in general: replayed with `cudaMallocAsync` on both sides of the same never-synchronized stream all four growths succeed, the free moving 54.83 to 38.83 GiB for the 24 GiB request as the retired 16 GiB is reused with nothing having synchronized anything, because those bytes go back to the device's memory pool rather than to the driver. The pool allocates with `cudaMalloc`, so it is in the first shape, and the pool-owned disposal stream is what that shape costs. Telling those two kinds of call apart is a claim about what the caller has still to do, and three defects on this path in a row were a disposal getting that claim wrong, the last of them a call that failed after the growth and so had no synchronization left to make; so the disposal takes no argument and the claim records nothing about the caller. A device-wide `cudaFree`, and the wait for the whole device with it, is still the fallback — where the device has no stream-ordered allocator (`cudaFreeAsync` reports `cudaErrorNotSupported` where `cudaDevAttrMemoryPoolsSupported` is absent), where it reports anything else, since a queued free that did not happen is not a free, and where the pool's stream could not be created; `cudaFreeAsync` takes the pointer `cudaMalloc` returned, so nothing here moves the pool onto the stream-ordered allocator, it only needs the device to have one. The disposal runs with the device lock dropped, so it does not hold the next claimant off — two pooled calls still serialize at submission, as the caller-stream contract promises, and a growth is no exception. Dropping the lock leaves the slot looking idle for the length of the disposal, which matters to the one observer that destroys what a slot holds rather than reading it: a test-only reset run while a disposal is still using the handoff event and the disposal stream destroys both under it — a harness over the real reset with real CUDA handles had the disposal's next `cudaStreamSynchronize` report `cudaErrorDeviceUninitialized`. Both of the reset's headers now carry the precondition that rules that out: safe only with no claim outstanding, which covers a claim whose release is still disposing. One further consequence for a growing call: its aliased reflects and D2H copies are queued after its enqueue has completed rather than while it runs, because the disposal's wait sits between them. That is in the noise end to end, but it is a real change in what a growing call does and the README states it. Seven cases cover the disposal itself. `ALargerEngineGrowsThePoolAndFreesTheBufferItReplaces` and `AGrowthOnASynchronizedCallFreesTheBufferItReplaces` take each kind of call and require the retired bytes back the moment `execute()` returns, reading before the test synchronizes anything, since a synchronize is what would release a free the disposal had only queued. `AGrowthDoesNotWaitForUnrelatedWorkQueuedOnTheDevice` and `AGrowthOnACallerStreamDoesNotWaitForUnrelatedWorkQueuedOnTheDevice` pin that neither kind waits for work parked on a stream it never submitted to: forcing the disposal device-wide fails both at the gate's 60 s watchdog, and only the caller-stream one could have failed before the pool owned the stream. `AGrowthOnACallerStreamWaitsForTheEnqueueOnTheBufferItRetires`: skipping the wait for the enqueue against the buffer being retired fails through `cudaErrorIllegalAddress`, the engine reading a buffer freed under it. `AGrowthOnACallerStreamDisposesWithTheDeviceLockDropped` takes the device's pool lock while the growth is still inside `execute()`, and asserts it was free. `AGrowthsDisposalDoesNotSynchronizeTheCallersStream` reaches for the same lock in the same window, as the precondition that says the interval it times is the disposal, so it depends on the unlock too. Measured: with the unlock moved below the disposal exactly those two fail — the first on its own assertion, the second on that precondition, both after their 5 s try-lock deadline expires — and the other five, which run their growth with nothing contending for that lock, stay green, 94 of 96 cases passing. `AGrowthsDisposalDoesNotSynchronizeTheCallersStream`: moving the free and its synchronize onto the caller's stream leaves every other case in the file green, this one included until it was written. **Nothing covers the device-wide fallback.** Reaching it needs `cudaFreeAsync` to fail a call that is correct as made, or the pool's stream creation to fail, and neither can be induced from inside the test process. That is a real loss against an earlier revision of this branch, where the caller-stream path's normal disposal *was* the device-wide free and every caller-stream growth on this hardware ran it; the skip in the two parked-work cases says the fallback is uncovered rather than pointing at coverage that no longer exists. How often a growth happens depends on the engine: `execute()` re-queries the requirement on every call once the shapes are bound, so an engine whose answer does not vary with the shapes grows the pool at most once, on its first run, and a program built only from those settles after its largest engine has run; one whose answer does vary can grow it on any call needing more than every call before it, and with one of those in the program nothing bounds the number of allocations. Run order, not load order: loading an engine allocates nothing.
- **Two growths at once can hold both retired buffers, and nothing bounds that.** The disposal runs outside the device lock, so a second grower can take the lock and allocate while the first grower's retired buffer is still resident: three staggered growths of 64, 128 and 192 MiB peak at 384 MiB rather than 192. What the peak is bounded by is the sum of the distinct scratch sizes, which is what the unpooled configuration holds permanently and for the whole process — 384 MiB transient against 384 MiB steady-state, collapsing to 192 once the disposals land — so the pool is never worse than no pool, and this is a startup transient rather than a regression the pool introduces. Bounding it would mean a grower waiting on how much another grower's disposal still holds, and the reason not to is not a missing mechanism — the bookkeeping was never the hard part: the only sound place to consult such a signal is under the device lock, where the only action available is to wait out a whole inference. That is the coupling the disposal deliberately does not have: adding the wait back would close a cycle — a claim waiting on a disposal that waits on an enqueue only the claiming thread will release.
- **Sharing one buffer has an unbounded wait of its own, growth or no growth.** A call that synchronizes the stream waits, through the handoff, for every pooled enqueue submitted before it on that device — what one buffer per device costs, and the premise of the feature rather than a defect in it. Park a host function ahead of one pooled call's enqueue and release it only once a later pooled call has returned, and the later call is the one that never returns. The installed `execute()` contract and the README both say so.
- **The option cannot be enabled from Python or from a `.pte`.** It arrives only through the C++ `set_option`; there is no load-time runtime spec and no compile-spec fallback of the kind `weight_streaming_budget` has. Deferred rather than built here.
- **The backend-linked test needs a real GPU, and the job that would give it one rarely runs.** `tests/cpp/executorch/test_shared_scratch_backend.cpp` is the only target in that package that links the delegate, and it covers `set_option`, the per-engine capture, the pooled path, the install check, the capture refusal, the two-thread concurrency regression and the one place this branch could have moved the option-off path without meaning to, its error returns. Without a device it skips every case and exits zero, which Bazel reports as a passing target: measured when the target carried 26 cases, 26 of 26 skipped and exit 0, behind a banner saying the target passing means nothing. `TORCHTRT_EXECUTORCH_REQUIRE_CUDA=1` turns a skip into a failure — same binary, same empty device list, exit 1 with 26 failures. It now carries 27. That variable covers the missing device and nothing else: eight cases carry skips it does not reach — three need the device full, two need its memory quiet, three need the device's stream-ordered allocator. The cases are counted and listed by name and reason at teardown whether or not the variable is set. That list is in the test log, which `--test_output=errors` prints nothing of for a passing target, so the CI invocation also passes `--test_summary=detailed`, which names every skipped case in Bazel's own summary. Three of them take the device to essentially zero free bytes while they hold it, measured at 16 MiB free of 81151 MiB with a neighbouring process getting out-of-memory on 256 MiB allocations throughout the window; Bazel's `exclusive` tag keeps other actions in the same build off the device but cannot keep anything else off it, and the file carries that warning. The CI invocation now passes the required-CUDA variable, and the job it sits in already asks for a GPU runner and starts its container with `--gpus all`. The job's own lane gate is not what keeps these cases off an ordinary pull-request push — it sits out only when the lane is `skip` or the backend is RTX, and `_decide.yml` resolves an unlabelled `pull_request` to `lane=fast`, `backend=standard`. What drops the job is the cancelled standard channel, below.
- **The zero-answer path is covered, but not by inducing a real query failure.** `AnEmptyInputRunsWithThePoolEnabled` reaches it through a legitimate empty input on a dynamic profile and reads the pool's capacity at three points, pinning the first to the one-byte minimum exactly — which kills a zero handed no buffer at all, a zero sized from the engine's profile-wide figure, and a minimum raised to anything the next call's requirement would still bound. `AnUndersizedScratchInstallIsSeenAsRefused` drives the same TensorRT call `execute()` makes over a live `kUSER_MANAGED` context: correct install accepted, one-byte install for an engine expecting more refused, empty-batch context still accepting one byte. A genuine *failed* `updateDeviceMemorySizeForShapes()` cannot be induced in-process — what `execute()` installs is the figure that query just returned, over a buffer grown to at least that, so the installed size is never short of what the context expects — so what is pinned end to end is the check that catches the consequence, not the cause. That `execute()` still goes through the checked install is pinned by `ExecuteInstallsPooledScratchThroughTheCheckedHelper`: a counting `IErrorRecorder` on the delegate's own context, asserted replaced and restored exactly once per pooled run, and replacing the checked call with a bare `setDeviceMemoryV2` takes both counts to zero with no other case noticing. `AnInstallKeepsAReferenceOnTheRecorderItSwapsOut` covers the swap itself: TensorRT drops a reference on the recorder it replaces and takes one again on the restore, so the install holds one of its own across the swap and the case fails if it stops. What no test pins is the refusal reaching a caller as `Error::InvalidState`, because the refusal cannot be reached.
- **The disposal's failure handling is not tested, and neither is the bail-out that also reaches it.** A failed wait on the handoff event leaks the retired buffer rather than freeing it under a possibly-live enqueue, drains the stream this call already submitted to, and returns `Error::InvalidProgram` — a device that fails a wait on device work is already faulted. The disposal does run on this hardware, on every growth, but none of a failed event wait, a `cudaFreeAsync` that refuses a call correct as made, or a failed synchronize of the pool's stream can be induced from inside the suite without corrupting the device for every case after it. Nor does any case reach a growth whose call then fails: reaching either of the two returns that could would need TensorRT or CUDA to fail a call correct as made, and every in-process injection tried is worse than the coverage — a fabricated or destroyed `cudaEvent_t` segfaults rather than returning `cudaErrorInvalidResourceHandle`, a misaligned device binding gets past `setTensorAddress` and `enqueueV3` and then raises a *sticky* asynchronous fault that poisons every teardown after it, and a foreign-device stream or event needs a second GPU and so a conditional skip. What holds that path is the shape of the code rather than a case: nothing is recorded at the claim for a bail-out to falsify, so a return between the claim and the release makes exactly the disposal every other path makes. A named case would need a test target that `#include`s `TensorRTBackend.cpp`, and a Bazel split so it does not also link the backend library; not built here.
- **`cudaDeviceReset()` is documented, not tested.** Reproducing it would destroy the primary context for every test that ran after it in the same process. The stream-capture refusal beside it *is* tested, twice: `APooledEngineRefusesACaptureBeforeAnythingCanInvalidateIt` under the default `Global`, which is the one that can see *where* in `execute()` the refusal comes, and `ARefusedPooledCallLeavesTheCallersPendingCudaErrorAlone` under `Relaxed`, which pins that the refusal does not clear a CUDA error the caller left pending. Nothing covers what a capture does with the option off, deliberately: that path is base-branch behaviour this PR does not change.
- **The ExecuTorch runtime CI job has been skipped on every run at this head**, gated on a channel that concurrency kept cancelling, and the gate ignores skipped channels — which is why every number here is local. `tests/cpp/executorch` carries 4 `cc_test` targets and 40 cases at the merge base with `main`; this PR takes it to 6 targets and 96 cases, the two new targets being `test_shared_scratch_pool` (29) and `test_shared_scratch_backend` (27). All 6 targets and all 96 cases have been compiled and run under bazel on a real A100 with `--nocache_test_results` and `TORCHTRT_EXECUTORCH_REQUIRE_CUDA=1` set, so a missing device would have been a failure; nothing skipped and all pass. The lane itself still needs one clean run. The suite also moved to the end of that job's script, so a missing or flaky device no longer takes down everything it used to sit above under `set -e`: the three example exports, `verify-executorch-reference-runner.sh` — the check that unpacks `libtorchtrt.tar.gz` and builds the packaged CMake backend out of it — and `load_model.py`. Run with a `bazel` whose `test` exits 3, the verify check never ran before the move and does after.
- **The stream-handle hazards are argued, not reproduced end to end.** The destroyed-stream crash was reproduced through the delegate; handle recycling and the NULL-stream collision were measured at the CUDA level. Corruption from a missing wait was never reproduced through a real engine, on either design.
- **Three things this PR deliberately does not do, each raised and each left.** The pool's locking rule is enforced by prose and by where the calls sit, not by a type: `SharedScratchDevice`'s state and its mutex are public members and nothing checks the rule. Wrapping them in a lock object is the right direction and a redesign of a header's public shape, the backend and the pool suite at once, so it is not folded into a change to what the disposal does. The same reasoning holds for the disposal rules, which are stated in four places — the installed contract, the README, `SharedScratchClaim::release()` and the pool header. The measurements behind them are not: they live in the README alone, and the code says what it does — the installed contract and `SharedScratchClaim::release()` pointing at the README for the rest, the pool header stating the rule without them. Picking one home per *rule* as well would restructure all four in the same change that alters the semantics they describe, so it is left, and it is the largest thing open here. And the pool is not rebuilt on `cudaMallocAsync`, which would delete the growth path, the retirement, the handoff event and most of the capture material: a different feature rather than a fix to this one.
- Not measured: an A/B of the memory saving on the ~210-engine target model — what was measured there is only how often the pool grows, once, because its largest engine runs first — and weight streaming combined with the pool end to end. The saving is linear by construction and the growth policy is exercised by the three models above.

## Type of change

# Checklist:

### DIFF
--- a/cpp/BUILD
+++ b/cpp/BUILD
@@ -191,6 +191,122 @@ cc_library(
     ],
 )
 
+# Implementation detail of :tensorrt_executorch_backend, so the header sits under
+# src/ and ships with the sources rather than in the installed API set. A caller
+# turning the pool on needs only the option key, which is in TensorRTBackend.h.
+cc_library(
+    name = "tensorrt_executorch_shared_scratch_pool",
+    hdrs = [
+        "src/torch_tensorrt/executorch/SharedScratchPool.h",
+    ],
+    strip_include_prefix = "src",
+    target_compatible_with = select({
+        ":linux_x86_64": [],
+        ":sbsa": [],
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
+    deps = select({
+        ":linux_x86_64": [
+            "@cuda//:cuda_headers",
+        ],
+        ":sbsa": [
+            "@cuda//:cuda_headers",
+        ],
+        "//conditions:default": [],
+    }),
+)
+
+# The one internal entry point of :tensorrt_executorch_backend that its test calls
+# directly. Under src/ for the same reason as the pool header above: it is an
+# implementation detail of execute(), and an installed header declaring it would
+# make it API from the next release.
+#
+# It carries the definition and not only the declaration, so the target stands on
+# its own: a header target whose symbol another target defines links only for a
+# consumer that happens to depend on that one too.
+cc_library(
+    name = "tensorrt_executorch_pooled_scratch_install",
+    srcs = [
+        "src/torch_tensorrt/executorch/PooledScratchInstall.cpp",
+    ],
+    hdrs = [
+        "src/torch_tensorrt/executorch/PooledScratchInstall.h",
+    ],
+    strip_include_prefix = "src",
+    target_compatible_with = select({
+        ":linux_x86_64": [],
+        ":sbsa": [],
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
+    deps = select({
+        ":linux_x86_64": [
+            "@executorch//:executorch_headers",
+            "@tensorrt//:nvinfer",
+        ],
+        ":sbsa": [
+            "@executorch//:executorch_headers",
+            "@tensorrt_sbsa//:nvinfer",
+        ],
+        "//conditions:default": [],
+    }),
+)
+
+# The body of the pool's test-only reset. testonly and under src/, so it is in
+# neither the packaged include tree nor executorch_backend_source_files below: a
+# release build has no definition of it anywhere. SharedScratchPool.h keeps only
+# the friend declaration that lets this reach the registry's internals.
+cc_library(
+    name = "tensorrt_executorch_shared_scratch_pool_reset",
+    testonly = True,
+    hdrs = [
+        "src/torch_tensorrt/executorch/SharedScratchPoolReset.h",
+    ],
+    strip_include_prefix = "src",
+    target_compatible_with = select({
+        ":linux_x86_64": [],
+        ":sbsa": [],
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
+    deps = [
+        ":tensorrt_executorch_shared_scratch_pool",
+    ],
+)
+
+# The pool's test-only entry points. A separate target, and testonly, so the
+# released archive -- Bazel's :tensorrt_executorch_backend and CMake's
+# executorch_trt_backend, neither of which compiles this source -- carries no
+# definition of them and exports no symbol for them. One of them frees the live
+# pool with no wait for work in flight, which is not something a release build
+# should offer a caller.
+cc_library(
+    name = "tensorrt_executorch_shared_scratch_pool_test_hooks",
+    testonly = True,
+    srcs = [
+        "src/torch_tensorrt/executorch/SharedScratchPoolTestHooks.cpp",
+    ],
+    hdrs = [
+        "src/torch_tensorrt/executorch/SharedScratchPoolTestHooks.h",
+    ],
+    strip_include_prefix = "src",
+    target_compatible_with = select({
+        ":linux_x86_64": [],
+        ":sbsa": [],
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
+    deps = [
+        ":tensorrt_executorch_shared_scratch_pool",
+        ":tensorrt_executorch_shared_scratch_pool_reset",
+    ] + select({
+        ":linux_x86_64": [
+            "@cuda//:cudart",
+        ],
+        ":sbsa": [
+            "@cuda//:cudart",
+        ],
+        "//conditions:default": [],
+    }),
+)
+
 cc_library(
     name = "tensorrt_executorch_backend",
     srcs = [
@@ -211,6 +327,8 @@ cc_library(
     deps = [
         ":tensorrt_executorch_binding_names",
         ":tensorrt_executorch_blob_header",
+        ":tensorrt_executorch_pooled_scratch_install",
+        ":tensorrt_executorch_shared_scratch_pool",
         ":tensorrt_executorch_weight_streaming_budget",
     ] + select({
         ":linux_x86_64": [
@@ -234,7 +352,10 @@ filegroup(
     name = "executorch_backend_source_files",
     srcs = [
         "src/torch_tensorrt/executorch/CMakeLists.txt",
+        "src/torch_tensorrt/executorch/PooledScratchInstall.cpp",
+        "src/torch_tensorrt/executorch/PooledScratchInstall.h",
         "src/torch_tensorrt/executorch/README.md",
+        "src/torch_tensorrt/executorch/SharedScratchPool.h",
         "src/torch_tensorrt/executorch/TensorRTBackend.cpp",
         "src/torch_tensorrt/executorch/TensorRTBlobHeader.cpp",
         "src/torch_tensorrt/executorch/WeightStreamingBudget.cpp",
--- a/cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
+++ b/cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
@@ -17,6 +17,7 @@
 
 #include <executorch/runtime/backend/interface.h>
 
+#include <cstddef>
 #include <memory>
 #include <mutex>
 #include <string>
@@ -75,6 +76,23 @@ struct EngineHandle {
   size_t num_aliased_outputs = 0;
   int device_id = 0;
   bool unified_memory = false;
+  // Whether exec_ctx was created kUSER_MANAGED, because
+  // kSharedActivationScratchKey was on at this engine's load. Such a context takes
+  // its activation scratch from the shared per-device pool on every call, unless
+  // claims_pooled_scratch below is false, in which case the engine needs none and
+  // never claims from the pool at all.
+  bool shared_scratch = false;
+  // Whether this handle takes its activation scratch from the pool. Set at init
+  // only where shared_scratch above is on, and then only where the engine reports
+  // needing activation scratch under some shape; a handle loaded with the option
+  // off keeps this false and says nothing about what its engine needs, because it
+  // has its own kSTATIC scratch either way. It is a predicate and not a size
+  // because the size is never the right amount to install: enqueueV3 refuses a
+  // kUSER_MANAGED context with no device memory once the engine needs any, however
+  // little the shapes bound to a given call need, so execute() must hand such a
+  // call some buffer -- but the engine's own figure covers every shape in the
+  // profile, and installing it would size the pool for the largest of them.
+  bool claims_pooled_scratch = false;
   std::mutex mu;
   // Makes the skip-sync fast path safe to reuse: TensorRT forbids reconfiguring or
   // destroying an execution context while one of its enqueues is in flight, so when
@@ -86,6 +104,15 @@ struct EngineHandle {
 
   ~EngineHandle();
 };
+
+// Runtime backend option that backs execution-context activation scratch with a
+// shared per-device pool instead of giving every context its own. Boolean,
+// default false. Read by TensorRTBackend::set_option below, and delivered as
+//   executorch::runtime::set_option("TensorRTBackend", options.view())
+// A context's allocation strategy is fixed when the context is created, so a
+// later call governs only the engines loaded after it, and a pooled context and
+// a private-scratch one coexist in one process.
+inline constexpr char kSharedActivationScratchKey[] = "use_shared_activation_scratch";
 
 class TensorRTBackend final : public ::executorch::runtime::BackendInterface {
  public:
@@ -102,12 +129,101 @@ class TensorRTBackend final : public ::executorch::runtime::BackendInterface {
   // past return, order any other stream against this one, and synchronize the stream
   // before reading device-resident outputs. The selected stream must be on the engine's
   // device, and calls on one handle must not overlap each other or its destruction.
-  // Note that other CUDA delegates sharing the same guard may instead synchronize before
+  // Other CUDA delegates sharing the same guard may instead synchronize before
   // returning, so do not assume results are ready on return from this one.
+  // With the shared activation scratch pool (kSharedActivationScratchKey) one
+  // buffer per device backs every context created while the option was on. Calls
+  // on two such handles on one device may overlap: a per-device lock held across
+  // the enqueue serializes them, so they do not run concurrently on the device. A
+  // handle whose context was created while the option was off keeps its own
+  // scratch and is not subject to this, nor is one whose engine reports needing no
+  // activation scratch under any shape, which is left out of the pool. For the
+  // handles that do draw on it, five further consequences:
+  //   - A call needing more scratch than the pool holds grows it, and the growth
+  //     has to get rid of the buffer it replaces before it returns, or the bytes
+  //     of every size the pool ever grew to stay resident at once. Where that
+  //     buffer has an enqueue recorded against it, the disposal begins with a host
+  //     wait for that enqueue -- a whole inference, with no upper bound, paid
+  //     whatever kind of call the growth is and whether or not it goes on to fail.
+  //     Unbounded is meant literally: if that enqueue is itself waiting on
+  //     something only this thread supplies once execute() returns -- a host
+  //     function it will release, a copy it will enqueue next -- the call does not
+  //     return, and the thread that would unblock it is the one inside the wait. A
+  //     growth that reached its own enqueue therefore comes back with that engine
+  //     work finished rather than in flight, since the event it waits on names that
+  //     enqueue too. What the growth does not wait for is the rest of the device:
+  //     the free is queued on a stream the pool owns and synchronizes itself, so
+  //     work on streams this call never submitted to is not in the wait, and the
+  //     bytes are back before it returns. Three things make the free a device-wide
+  //     cudaFree instead, and then the call does block until the device is idle: a
+  //     device with no stream-ordered allocator, where cudaFreeAsync reports
+  //     cudaErrorNotSupported rather than freeing; any other code it reports, since
+  //     a queued free that did not happen is not a free; and a disposal stream the
+  //     pool could not create at all. Which calls grow the pool is not knowable
+  //     from here; see the README.
+  //   - A call that synchronizes the stream waits, through the handoff, for every
+  //     pooled enqueue submitted before it on that device. That is what one buffer
+  //     costs, and it is another unbounded wait: park a host function ahead of a
+  //     pooled enqueue and release it only after a later execute() returns, and
+  //     the later call is the one that does not return.
+  //   - Capturing a CUDA graph around this delegate is not supported, with the
+  //     option on or off. With it on, a call whose selected stream is capturing is
+  //     refused with Error::NotSupported, ahead of every CUDA call it makes that a
+  //     capture cannot take -- only the device query and the device switch run
+  //     first. The pool's event handoff waits on an event recorded outside the
+  //     capture, which invalidates it under every capture mode, and a growth's
+  //     allocation and its disposal of the buffer it replaces invalidate it under
+  //     every mode but cudaStreamCaptureModeRelaxed. The alternative to refusing is
+  //     a capture that silently comes back invalidated. Where the query cannot be
+  //     answered the call is not told it is capturing: cudaStreamIsCapturing also
+  //     hands back a sticky fault left by earlier work on the device, and a call
+  //     that meets one fails with that fault's own message and
+  //     Error::InvalidProgram instead. Only a capture on the selected stream is
+  //     caught; one running on any other stream under cudaStreamCaptureModeGlobal,
+  //     or under cudaStreamCaptureModeThreadLocal from this thread, is invalidated
+  //     by the same calls and is not refused, because CUDA offers no query for it.
+  //     Do not run a pooled engine while capturing anywhere in the process. With
+  //     the option off nothing refuses, because the check is inside the pooled
+  //     path, but capture is unsupported there too: no call shape is built or
+  //     tested for it. The README states this in full.
+  //   - cudaDeviceReset() invalidates the pool without emptying it. The buffer,
+  //     the handoff event and the disposal stream it still holds are destroyed
+  //     with the primary context, and the next call on that device uses all three.
+  //     There is no guard: do not reset a device this backend has run a pooled
+  //     engine on.
+  //   - The pool adds failure points to a call, on top of Error::NotSupported for
+  //     the capture above. Only Error::Internal is its own: a pooled call returns
+  //     it when the device's handoff event cannot be created, and an unpooled call
+  //     never returns it at all. The rest reuse codes an unpooled call already
+  //     returns from elsewhere in execute(), so the code alone does not say the
+  //     pool was involved -- the log line does, and every one of them logs at Error
+  //     first. Error::MemoryAllocationFailed if the shared buffer cannot be
+  //     allocated or grown; Error::InvalidState if the wait ordering this call
+  //     behind the previous enqueue fails, if TensorRT refuses the installed
+  //     buffer, or if this call's enqueue cannot be recorded for the next claimant;
+  //     Error::InvalidProgram if the capture query fails for a reason that is not a
+  //     capture, or if the disposal's host wait fails -- and that last one leaks
+  //     the retired buffer rather than freeing it under an enqueue that may still
+  //     be reading it. The record and the disposal are the only two reached after
+  //     this call's enqueue is submitted, and both synchronize the stream before
+  //     returning, so neither leaves engine work in flight. Separately, a pooled
+  //     call that staged a host-resident input synchronizes the stream on any error
+  //     return it makes after that copy: the copy reads the caller's own memory and
+  //     the caller owns that memory again the moment execute() returns. That too is
+  //     a whole-stream wait, so it covers work the caller queued before calling. An
+  //     unpooled call makes no such wait -- its error returns can leave that copy
+  //     running, which is what they have always done.
   ::executorch::runtime::Error execute(
       ::executorch::runtime::BackendExecutionContext& context,
       ::executorch::runtime::DelegateHandle* handle,
       ::executorch::runtime::Span<::executorch::runtime::EValue*> args) const override;
+
+  // Applies the runtime backend options a caller passes to
+  // executorch::runtime::set_option("TensorRTBackend", ...). The only key read is
+  // kSharedActivationScratchKey, a boolean.
+  ::executorch::runtime::Error set_option(
+      ET_UNUSED ::executorch::runtime::BackendOptionContext& context,
+      const ::executorch::runtime::Span<::executorch::runtime::BackendOption>& backend_options) override;
 
   void destroy(::executorch::runtime::DelegateHandle* handle) const override;
 };
--- a/cpp/src/torch_tensorrt/executorch/README.md
+++ b/cpp/src/torch_tensorrt/executorch/README.md
@@ -68,8 +68,7 @@ type.
 The upstream `CallerStreamGuard` documents the generic contract (per-thread,
 nested scoping; the caller owns the stream for the guard's lifetime; the caller
 manages host-data lifetime for async work). The TensorRT backend adds these
-requirements, which previously lived on
-the removed `CudaStreamGuard`:
+requirements, which previously lived on the removed `CudaStreamGuard`:
 
 - The selected stream must be on the TensorRT engine's device.
 - Calls using one delegate handle must not overlap, and must not overlap with
@@ -85,6 +84,18 @@ the removed `CudaStreamGuard`:
   complete, order any cross-stream producers/consumers with their own events,
   and synchronize the stream before reading outputs on the host.
 - With no guard active, the backend falls back to `cudaStreamPerThread`.
+- With the `use_shared_activation_scratch` backend option enabled, one buffer
+  per device backs the activation scratch of every execution context created
+  while it was on, so no two enqueues against it may overlap. The backend
+  enforces this itself: it holds a per-device lock from the claim on the buffer
+  through the enqueue and the completion event recorded on it, so two
+  `execute()` calls on one device are serialized at submission and the second's
+  stream waits on the first's enqueue. They may run on one stream or on two, and
+  they may be submitted concurrently from two threads, but they will not run
+  concurrently on the device, so the pool costs the parallelism between them.
+  Contexts created while the option was off keep their own scratch and are
+  unaffected, and so is a context whose engine needs no activation scratch under
+  any shape: that one is left out of the pool and serializes against nothing.
 - The reference-runner smoke test runs inference inside a caller-stream guard on
   the discrete-GPU CI configuration, where all inputs and outputs are host-backed
   and therefore take the synchronized staging path. CI separately asserts that the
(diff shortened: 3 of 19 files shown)


cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -906,7 +1043,34 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
...
+  // 4. Back activation scratch with the shared per-device pool
+  // ------------------------------------------------------------------
+  // All input shapes are bound by now, so the exact scratch requirement for this
+  // call is known. The buffer is installed on every call, not once, because a
+  // larger engine may have grown the pool and moved it since the last one. A
+  // kSTATIC context owns its private scratch, so setDeviceMemoryV2 must not be
+  // called on one.
+  bool scratch_from_pool = false;
+  if (engine->shared_scratch) {
+    const size_t need = ctx->updateDeviceMemorySizeForShapes();
```
shoumikhin: this treats the error sentinel as a real answer, and the argument in the comment only covers the first call.

tested against 11.2.1.2:
- first call with nothing set, `enqueueV3` does refuse. so the comment is right about that case.
- `setDeviceMemoryV2(nullptr, 0)` is itself rejected ("Cannot set memory to nullptr"), and it returns void, so the failure is invisible to us.
- on a later call the context silently keeps its previous pointer and `enqueueV3` returns true.

that previous pointer can be freed memory. once another engine grows the pool you `cudaFree` the old buffer, so a spurious 0 here runs the engine against a dead allocation. i reproduced it: run once with a good buffer, free it the way the release lambda does, let an unrelated `cudaMalloc` take the address, then hit the zero path. all 4194304 bytes of the unrelated allocation got overwritten, and `enqueueV3` still returned true with a correct output.

the zero branch also skips `get_or_grow_shared_scratch` entirely, so you lose the wait and the in-flight mark in the same step.

simplest fix is to return an error on 0.

shoumikhin (commented):
took a proper look at this, most of it against 11.2.1.2 since that's the pin in MODULE.bazel. the design holds up and the event handoff is doing real work. one blocking thing on the zero path, rest is smaller stuff.

two things i chased that turned out to be fine, noting them so nobody else burns time on them: the pool does cover weight streaming scratch (`updateDeviceMemorySizeForShapes` tracks `getDeviceMemorySizeV2` exactly, scratch included, checked with the budget moved around), and a caller stream from a green context records on the per-device event without complaint and actually orders the work. no concerns on either.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -906,7 +1043,34 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   }
 
   // ------------------------------------------------------------------
-  // 4. Enqueue inference on the current CUDA stream
+  // 4. Back activation scratch with the shared per-device pool
+  // ------------------------------------------------------------------
+  // All input shapes are bound by now, so the exact scratch requirement for this
```
shoumikhin: "the exact scratch requirement for this call" is only true with `kRUNTIME_ACTIVATION_RESIZE_10_10` on, and nothing here enables it (we only set `MULTIDEVICE_RUNTIME_10_16`).

measured on 11.2.1.2: preview off, a query at batch 64 under a profile max of 256 returns exactly the profile-max size. preview on, same engine returns 8192 at batch 1 vs 33554432 at batch 4096.

not a safety issue since it oversizes, but the pool ends up sized to the profile max rather than to the call, and that's most of the savings story.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -204,6 +223,124 @@ bool is_cuda_accessible_ptr(const void* ptr) {
...
+// Sets out_ptr to a buffer of at least `need` bytes on `device_id` and out_size to
+// its capacity, with `stream` ordered after the enqueue that last used the buffer.
+// The caller must call mark_shared_scratch_in_flight once it has submitted its own
+// enqueue.
+//
+// Must be called with `device_id` already current: cudaEventCreateWithFlags,
+// cudaMalloc, cudaFree and cudaDeviceSynchronize all act on the *current* device
+// and nothing in here sets it.
+Error get_or_grow_shared_scratch(int device_id, size_t need, cudaStream_t stream, void*& out_ptr, size_t& out_size) {
+  std::lock_guard<std::mutex> lk(scratch_pool_mu);
```
shoumikhin: `scratch_pool_mu` is one mutex for all devices and it's held for the whole function, including `cudaMalloc` and the release lambda's `cudaDeviceSynchronize` + `cudaFree`.

so a growth on device 0 blocks a plain claim on device 1, which only needs its own map slot. the README says concurrent execute on different devices is fine, and that stops being true during a growth. the sync is unbounded as well, it waits on everything queued on the device, not just the scratch users.

per-device lock would fix both scopes, or move the cuda calls out from under the map lock.

cpp/src/torch_tensorrt/executorch/README.md
```diff
@@ -104,6 +112,34 @@ the removed `CudaStreamGuard`:
...
+executorch::runtime::BackendOptions<1> options;
+options.set_option("use_shared_activation_scratch", true);
+executorch::runtime::set_option("TensorRTBackend", options.view());
+```
+
+Check what `executorch::runtime::set_option` returns: `Error::NotFound` means no
+backend is registered under that name, which is what a binary that has not linked
+the backend archive gets.
+
+N per-engine copies collapse to one, so the reclaimed memory is `(N-1)` times the
```
shoumikhin: this only holds when every engine needs the same amount. the pool grows to `max(s_i)` and never shrinks, so the saving is `sum(s_i) - max(s_i)`. for engines needing 1, 2 and 4 units that's 3, not 8.

the numbers in the description used uniform engines so it wouldn't show up there. the commit message has the same claim.

cpp/include/torch_tensorrt/executorch/SharedScratchPool.h
```diff
@@ -0,0 +1,123 @@
...
+// carry it: synchronizing on a handle whose stream the caller has since
+// destroyed is a crash rather than an error return, CUDA recycles handle values
+// so a genuinely different stream can compare equal to the recorded one, and the
+// NULL stream is both a legal stream a caller can select and the only available
+// "no previous user" sentinel. An event names the work instead of the queue --
+// it stays valid after the stream that recorded it is destroyed, and waiting on
+// it from the stream that recorded it is already satisfied, so the common
+// single-stream case costs a host call and no device stall.
+template <typename CreateEvent>
+SharedScratchHandoff shared_scratch_claim_event(
```
shoumikhin: both helpers mutate the caller's map with no locking, and the header never says the caller has to serialize. the backend gets away with it by holding `scratch_pool_mu`, but this header goes out in `executorch_api_headers`, so it's API and the next caller won't know.

either document the precondition, or wrap the maps in a type that owns the lock, or keep the header private.

tests/cpp/executorch/test_shared_scratch_pool.cpp
```diff
@@ -0,0 +1,286 @@
...
+  EXPECT_EQ(failed, nullptr);
+  EXPECT_TRUE(a.released.empty());
+
+  // The pool still holds the original buffer and serves it on the next request.
+  void* again = call(pool, a, 0, 1024, out);
+  EXPECT_EQ(again, first);
+  EXPECT_EQ(out, 1024u);
+}
+
+TEST(SharedScratchPool, FirstAllocationFailureReturnsNullAndStoresNothing) {
```
shoumikhin, line 213: name says stores nothing, but `pool[device_id]` default-inserts before `alloc` runs, so there is an entry, just one with a null pointer. the test only checks the return value and the retry, so it passes either way.

the header comment ("the slot is then left untouched") says the same thing. `EventCreationFailureIsReportedAndRetried` has the identical gap on the markers map. either narrow the names or assert `pool.empty()` and don't insert until the alloc succeeds.

tests/cpp/executorch/BUILD
```diff
@@ -47,3 +48,12 @@ cc_test(
         "@googletest//:gtest_main",
     ],
 )
+
+cc_test(
```
shoumikhin, line 53: this only depends on `tensorrt_executorch_shared_scratch_pool`, never on the backend, so nothing here covers the wiring. i can revert the context to `kSTATIC`, or delete the `updateDeviceMemorySizeForShapes`/`setDeviceMemoryV2` pair, or drop the wait/record calls, and all 11 tests stay green.

the description calls out the missing set_option and concurrency tests, but not that the plain single-threaded path has no automated coverage at all.

cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
```diff
@@ -109,6 +113,13 @@ class TensorRTBackend final : public ::executorch::runtime::BackendInterface {
       ::executorch::runtime::DelegateHandle* handle,
       ::executorch::runtime::Span<::executorch::runtime::EValue*> args) const override;
 
+  // Applies the runtime backend options a caller passes to
```
shoumikhin, line 136: while you're in this header, the `execute()` comment just above still ends with "calls on one handle must not overlap each other or its destruction". the pool adds a stronger rule (no two handles on the same device may overlap) and that only lives in the README right now. this is the installed header, so it should carry it.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Fixed.
It is actually possible for the engine to require 0 activation scratch so an extra check for that case is also needed.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Fixed.

cpp/src/torch_tensorrt/executorch/README.md
Conarnar: Fixed.

cpp/include/torch_tensorrt/executorch/SharedScratchPool.h
Conarnar: Fixed, with your first two suggestions. The header could not be made private since `TensorRTBackend.h` needs it for `set_option`.

tests/cpp/executorch/test_shared_scratch_pool.cpp
Conarnar: Fixed.

tests/cpp/executorch/BUILD
Conarnar: Fixed.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Fixed

cpp/include/torch_tensorrt/executorch/TensorRTBackend.h
Conarnar: Fixed

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -204,6 +229,127 @@ bool is_cuda_accessible_ptr(const void* ptr) {
...
+// its capacity, with `stream` ordered after the enqueue that last used the buffer.
+// The caller must call mark_shared_scratch_in_flight once it has submitted its own
+// enqueue.
+//
+// Must be called with `device_id` already current: cudaEventCreateWithFlags,
+// cudaMalloc and cudaFree all act on the *current* device and nothing in here
+// sets it.
+Error get_or_grow_shared_scratch(int device_id, size_t need, cudaStream_t stream, void*& out_ptr, size_t& out_size) {
+  SharedScratchDevice& dev = scratch_pool.get(device_id);
+  std::lock_guard<std::mutex> lk(dev.mu);
```
shoumikhin: This lock ends when the function returns, so `setDeviceMemoryV2` and `enqueueV3` run without it, and the enqueue is only recorded on the event after that. Nothing else serializes two pooled engines on one device, since `engine->mu` is per handle.

So a second thread can claim inside that gap and get the same buffer back, because the reuse branch returns the existing buffer untouched whenever the capacity fits. It does call `cudaStreamWaitEvent`, but the event carries an earlier enqueue rather than the one now in flight, so the wait does not order the two against each other. Both engines then write the same scratch.

I tried this with two real engines sharing one buffer. On every trial one engine's output was wrong, with no CUDA error and no TensorRT error. Nothing grows and nothing is freed, so this is the ordinary state once the pool has settled on its largest size. The caveat in the comment above, and in the README, points at growth freeing a live buffer, which is the case that did not corrupt in my testing: `cudaFree` is implicitly synchronizing, and I measured it blocking for the whole remaining kernel. So the documented hazard is the survivable one and this one is not mentioned.

In a harness following the same call order, holding the claim through the enqueue and the event record fixed it every time. That means `execute()` owning the device lock across `setDeviceMemoryV2`, `enqueueV3` and the mark, not a local change here. It looks safe: `engine->mu` is already held across `enqueueV3` in `execute()` today, and `core/runtime/execute_engine.cpp` already puts a mutex around the enqueue and says the other context calls belong in the same scope. Worth checking the cost for a weight streaming engine, where the header says `enqueueV3` becomes synchronous.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -151,6 +156,13 @@ bool infer_binding_names(
   return true;
 }
 
+// The setting behind kSharedActivationScratchKey: whether an execution context
+// created subsequently draws its activation scratch from the shared per-device
+// pool rather than allocating its own.
+//
+// execute() must read EngineHandle::shared_scratch, never this.
+std::atomic<bool> scratch_enabled{false};
```
shoumikhin, line 164: This global is the only way in, and the only writer is the C++ `set_option`, so a program loaded from Python has no way to turn the pool on. A hand written compile spec for the key would not help either, since `init()` only looks for the weight streaming key.

`weight_streaming_budget` in this same file takes a load time runtime spec first and falls back to a compile spec baked in at export, and the comment there explains that the compile spec exists for loaders that cannot pass backend options yet. The new option has neither channel.

Not a blocker, since the C++ path works and the feature is off by default. But the users who hit the memory problem this solves are often the ones loading a `.pte`.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -204,6 +229,127 @@ bool is_cuda_accessible_ptr(const void* ptr) {
...
+            Info,
+            "TensorRTBackend::execute: shared scratch pool (device %d) %s %zu bytes",
+            device_id,
+            first_buffer ? "allocated" : "grew to",
+            bytes);
+        return p;
+      },
+      [](void* old, cudaEvent_t wait_for) {
+        if (wait_for != nullptr) {
+          cudaEventSynchronize(wait_for);
```
shoumikhin: The result is discarded and `cudaFree(old)` runs on the next line either way. If the wait fails, the free is the exact thing the wait exists to prevent, so an enqueue may still be reading that buffer. Returning early instead would leak that one replaced buffer, which is a bounded cost.

Also worth knowing that no test reaches this path: every pooled engine in the backend test asks for the same size, so the reuse branch is always taken and the pool never grows. One extra engine with a larger shape would cover both the free and the wait.

shoumikhin (commented):
The pool design reads well. Three things.

**Two pooled engines running at once on one device give wrong numbers, silently.** The device lock is released when `get_or_grow_shared_scratch` returns, so `setDeviceMemoryV2` and `enqueueV3` run without it. The new enqueue is only recorded on the event afterwards. A second thread claiming in that gap gets the same buffer. That thread does wait on the event, but the event carries an earlier enqueue, not the one now in flight. So the wait does not order the two against each other, and both engines write the same scratch.

I tried this with two real engines sharing one buffer. On every trial one engine's output was wrong, with no CUDA error and no TensorRT error. The written caveat points at growing the pool and freeing the buffer. This needs neither, so it is the normal state once the pool has settled.

Holding the device lock from the claim through the enqueue and the event record fixed it every time. It looks safe here: `engine->mu` is already held across `enqueueV3` today, and `core/runtime/execute_engine.cpp` already puts a mutex around the enqueue and says the other context calls belong in the same scope.

**The option cannot be turned on from Python or from a `.pte`.** It only arrives through the C++ `set_option`. `weight_streaming_budget` handles this in the same file with a load-time runtime spec plus a compile-spec fallback.

**The release lambda frees the old buffer even when the `cudaEventSynchronize` before it failed**, which is the one case the wait exists to prevent.

One note: the ExecuTorch CI jobs are skipped on this commit because the matrix job was cancelled, so the two new test files have not run yet.

Happy to share the harness or the measurements.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -204,6 +229,127 @@ bool is_cuda_accessible_ptr(const void* ptr) {
...
+        }
+        cudaFree(old);
```
shoumikhin: Swapping the device sync for an event wait does not remove the device-wide wait, because `cudaFree` performs one itself. On an H100 with the handoff event already complete and unrelated work queued on another stream, the event wait returned in under a millisecond and this `cudaFree` took about 1.5 seconds, matching the queued work. It also runs with the device lock held, so a claim on that device waits behind unrelated work too.

Either move the free outside the lock and keep a retire list, or say in the commit message and the README that a growth still stalls on everything queued on the device.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Fixed. Checked the weight-streaming cost you flagged: at budget 0, `enqueueV3` took 0.439 ms against a 75.7 ms inference, and a second context's enqueue issued mid-stream returned in 0.267 ms. Didn't reproduce.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Fixed. The larger engine covers the allocation and the free, but not the wait — `cudaFree` synchronizes device-wide anyway, so deleting the wait still leaves the test green. The comment says so rather than claiming it.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Fixed

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
Conarnar: Deferred. There's no Python binding for `set_option` so for this to be done in a follow-up, the only route would be through compile spec during export.

shoumikhin (commented):
Some notes inline.

cpp/src/torch_tensorrt/executorch/README.md
```diff
@@ -104,6 +115,52 @@ the removed `CudaStreamGuard`:
...
+device, so the largest scratch it was ever asked for stays allocated until the
+process exits.
+
+The buffer grows when an engine asks for more than every engine before it did,
+and a growth is not free. It frees the buffer it replaces, and `cudaFree` waits
+for everything queued on the device, not only for the enqueues that used that
+buffer, so it can stall for far longer than the event wait that precedes it. The
+backend keeps that stall out from under the per-device lock, so it does not hold
+up another engine on the device, but it does fall after the growing call's own
+enqueue, so that one `execute()` waits for its own engine work too. A growth
```
shoumikhin: The line that says a growth happens only on an engine's first run does not hold for a dynamic shape engine, and the paragraph three lines below it says so itself. The per call query answers for the shapes just bound. I built an engine with a real dynamic profile and a user managed context, no preview feature, and the same engine asked for 33554432, 67108864, 100663296 and 134217728 bytes as the batch went from 1 to 4. That is three growths from one engine, so loading the largest engine first does not bound the allocations.

This matters beyond the text. The comment above the leak on a failed pre free wait says the cost is bounded because growth is rare, and the readme tells a caller the device wide free stall is avoidable by load order. Neither is true here. Could you either scope both claims to fixed shape engines, or say plainly that a dynamic shape engine can grow on any call?

cpp/BUILD
```diff
@@ -254,6 +277,7 @@ filegroup(
 filegroup(
     name = "executorch_api_headers",
     srcs = [
+        "include/torch_tensorrt/executorch/SharedScratchPool.h",
```
shoumikhin: The reason given for keeping this header public does not hold in the code as it stands. TensorRTBackend.h does not include it. It names the option key in three comments and nowhere else, and it compiles without it. The readme example passes the string literal, and the new backend test spells the key out on purpose.

So what ships to users, for one string constant, is a mutex, a per device registry, and three functions whose only lock rule is a sentence saying to call them with the device lock held. There is no assert and no annotation, and the lock and the state are separate public members so the type cannot enforce it. Fourteen of the sixteen cases in the pool unit test call those helpers with no lock at all, which is the first thing a new reader will copy.

Moving the key constant into TensorRTBackend.h next to set_option, and taking the pool header out of the installed set, closes this without adding anything.

tests/cpp/executorch/BUILD
```diff
@@ -9,6 +9,8 @@ test_suite(
         ":test_executorch_binding_names",
         ":test_executorch_blob_header",
         ":test_executorch_weight_streaming_budget",
+        ":test_shared_scratch_backend",
```
shoumikhin, line 12: The two new tests only run inside the ExecuTorch runtime job, and that job is skipped in every run at this head. It is gated on the standard channel, and that channel was cancelled by concurrency. The required gate only fails on a real failure, so skipped channels do not block anything.

That means every test number in the description is a local number, and the first time either file gets compiled in CI would be the push to main. Could you run the full lane on this branch once before merging?

Worth pairing with the device less skip you already documented. Right now a run that skips all twelve cases and a run that never happened both look green.

cpp/src/torch_tensorrt/executorch/README.md
Conarnar: Fixed, though I was not able to replicate your findings. Growth depends on whether an engine's reported requirement varies with the bound shapes, so I scoped the README by that rather than by fixed versus dynamic shape.

cpp/BUILD
Conarnar: Fixed, though `SharedScratchPool.h` was moved to `cpp/src/torch_tensorrt/executorch/` as `libtorchtrt.tar.gz` ships `TensorRTBackend.cpp` as buildable source, which requires the header.

tests/cpp/executorch/BUILD
Conarnar: Ran `//tests/cpp/executorch:executorch_backend_tests` under bazel on a real device, at the pushed head and again with this stack applied. 6/6 both times, including the two new files.

Conarnar:
Pushed two more commits.

**The zero-scratch guard rejected a valid input.** With the pool on, a dynamic engine given an
empty input (batch 0 inside a `[0..128]` profile, all dimensions specified) has
`updateDeviceMemorySizeForShapes()` legitimately answer 0 while the engine's own requirement is
non-zero. That is the shape the guard treated as a failed query, so `execute()` returned
`Error::InvalidState` and the engine never ran, on an input the backend otherwise supports (it
allocates one-byte placeholders for empty tensors). Reproduced on TensorRT 11.1, 10.13.3 and
11.2.1.2, with and without `kRUNTIME_ACTIVATION_RESIZE_10_10`.

A zero has three causes: a query that failed, a call whose bound shapes need no scratch, and an
engine that needs none under any shape. Only the last can be told apart, and it is told apart once
at init: such an engine is left out of the pool entirely, so it takes no per-device lock and does
not serialize against the engines that do. The other two arrive at `execute()` indistinguishable,
and both are handed whatever the pool already holds, at its current size, which grows nothing;
any live buffer is large enough for a call that needs none.

Neither can be handed nothing at all. `enqueueV3` refuses a `kUSER_MANAGED` context with no device
memory installed as soon as the engine reports needing some under any shape, however little the
bound shapes need: `Parameter check failed, condition: noDeviceMemory`. So against a pool holding
nothing the call allocates a one-byte minimum and the first call with a real requirement sizes the
pool. Standing the engine's own `getDeviceMemorySizeV2()` in for the zero satisfies that check too,
but that figure covers the whole profile: on the dynamic engine the tests build it is 4 MiB against
the 128 KiB the next call needs, and since the pool never shrinks a single empty batch would pin it
at 32x for the process lifetime, on the path the pool exists to keep small.

**Capturing a CUDA graph from a pooled `execute()`'s stream is now refused** with
`Error::NotSupported`. The handoff between one enqueue and the next is a wait on an event recorded
outside the capture, and `cudaStreamWaitEvent` on such an event fails with
`cudaErrorStreamCaptureIsolation` and invalidates the capture under every capture mode,
`cudaStreamCaptureModeRelaxed` included. That wait, not the allocation, is what produced the 901
and the null graph from `cudaStreamEndCapture` when the guard was removed and measured; the pool
was already large enough that no growth ran. A growth is mode-dependent by comparison: its
`cudaMalloc`, `cudaEventSynchronize` and `cudaFree` each invalidate the capture under Global and
ThreadLocal, while Relaxed permits all three and runs them uncaptured, leaving a replay pointed at
a buffer the pool may since have freed. Measured directly on CUDA 13.0, one process per cell,
rather than read off the documentation. The check is one `cudaStreamIsCapturing` ahead of the
device lock; loading the engine with the option off still captures cleanly.

**`cudaDeviceReset()` is documented rather than guarded**, in the installed header and the README.
The pool holds its buffer and its handoff event for the process lifetime and a reset destroys the
primary context under both. Catching it would mean revalidating both on every call, on the pool's
hot path, and the check costs about what it protects. The header's `execute()` contract also now
states that a growth turns an otherwise asynchronous call into a device-wide `cudaFree` wait.

**The handoff event passes `cudaEventBlockingSync`.** Created with `cudaEventDisableTiming` alone,
`cudaEventSynchronize` busy-waits, measured at 99.5% of a core on an H100 versus 0.1% with the
flag, and that wait runs while the per-device lock is held, so one core spins and every other
engine on the device queues behind it. The engine-completion event in the same file already passed
the flag. The README example also handles `set_option`'s `Error`, which is not `ET_NODISCARD` and
so compiled while dropping it, immediately above a paragraph telling the caller to check exactly
that return.

**Two test defects, both of which weakened the suite rather than the backend.** Nothing reset the
process-wide pool between tests, so the suite passed in declaration order and in no other: running
the empty-input case before `ASecondPooledEnqueueWaitsForTheFirstOnAnotherStream` turned the
latter's first run into a growth, whose `cudaFree` waits device-wide and so waited on the blocking
host function that test parks on its own stream, from the one thread that could release it. A
deadlock ending at the 60 s watchdog. `--gtest_repeat=2` on the growth case failed on its own
second iteration, blaming the backend for a fixture artifact. A fixture now resets the pool in
setup and teardown through test-only hooks declared in `SharedScratchPool.h`, which ships under
`src/` and is not installed, so no production semantics and no installed API changed. Separately,
`device_bytes_in_use()` returned zero on a failed `cudaMemGetInfo`; both memory comparisons
subtract two of its readings, and a zero for the first of a pair made the second look like the
whole cost of what was measured between them. It now reports the failure and the call sites assert
on it.

`tests/cpp/executorch` is 6 targets and 70 cases, run on a real device: 69 at the pushed head and
70 with the last commit, the extra being a unit test for the reset hook. Two tests that existed
only to pin a now-unused engine-wide byte count were replaced by one pinning the predicate the
pooled path branches on.

PR #4626, opened 2026-08-28
fix: unify LINSPACE alpha/beta dtype in the arange converter

### LINKED ISSUE #4625: 🐛 [Bug] arange passes a rank-1 tensor to LINSPACE's rank-0 alpha input; the API error is logged, not raised
##  Bug Description

On the `torch.compile(backend="tensorrt")` path, `torch.arange(start, end, step)` where
`start` is a data-dependent (unbacked) `SymInt` builds an invalid TensorRT network. This one
does fail -- but it fails a long way from its cause, with a message that never mentions
`arange`.

`arange` in `py/torch_tensorrt/dynamo/conversion/impl/arange.py` builds an `IFillLayer` in
`LINSPACE` mode and passes the start value as `set_input(1, start_rank_0)`:

```python
# py/torch_tensorrt/dynamo/conversion/impl/arange.py :: arange
        # Convert start, end, step into TRT tensors with appropriate rank
        start_rank_0 = get_trt_tensor(ctx, start, name + "_start_rank_0", min_rank=0)
        start_rank_1 = get_trt_tensor(ctx, start, name + "_start_rank_1", min_rank=1)
        ...
        fill_layer.set_input(1, start_rank_0)  # start value
```

`min_rank` is a *lower* bound. A symbolic `start` reaches the converter as a **rank-1 shape
`ITensor`**, so `min_rank=0` is a no-op and `start_rank_0` stays rank 1. TensorRT requires
that input to be rank 0, and says so:

```
ERROR:torch_tensorrt [TensorRT Conversion Context]:ITensor::getDimensions: Error Code 4: API Usage Error ((Unnamed Layer* 16) [Fill]: FillOperation LINSPACE requires that input at index 1 have rank 0. Actual rank is 1. In lowerParams at /_src/optimizer/common/nodes/fillNode.cpp:105)
```

**That error is only logged, never raised.** Conversion continues on the poisoned layer, its
output dimensions become invalid, downstream shape inference is poisoned too, and compilation
finally dies with a bare assertion that names nothing useful:

```
    assert cuda_engine
AssertionError: 
```

Where it surfaces depends on what consumes the `arange` result. With `aten.expand`
downstream it comes out of a different converter entirely, as a `ValueError` about a
negative length:

```
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/conversion/impl/slice/ops.py", line 227, in expand
    initial_tensor_rank = len(input_t.shape)
ValueError: __len__() should return >= 0
```

A second defect sits behind the first, in the same function. TensorRT requires the LINSPACE
`alpha` (input 1) and `beta` (input 2) to share a dtype, and the converter never unifies
them. With only the rank problem fixed, a symbolic `start` that arrives as Int64 against a
literal `step` that becomes Int32 is rejected; the exact observed error and the model that
triggers it are given below. Both defects have to be addressed for the
converter to work with a symbolic `start`.

### How we found this, and why it is newly easy to reach

We were reproducing a *different* defect, since fixed: `get_positive_dim` used to raise
`GuardOnDataDependentSymNode` when clamping an omitted slice stop (`INT64_MAX`) against an
unbacked dimension. That was fixed by #4236 ("fix: support SymInt for
`slice_scatter_decomposition`"), which rewrote the clamp in terms of
`torch.sym_ite`/`torch.sym_min`. Our reproducer for that one passes on 2.14.0a0.

**That fix is what makes the bug reported here reachable.**
`slice_scatter_decomposition` now computes

```python
start = get_positive_dim(start, dim_size)   # -> torch.sym_min(start, dim_size), a SymInt
end   = get_positive_dim(end, dim_size)     # -> torch.sym_min(end, dim_size), a SymInt
```

and passes them straight into `torch.arange(start, end, step)`. So ordinary Python written
against a data-dependent tensor now routes a symbolic `start` into this converter. Anyone who
takes #4236 inherits this.

That path is worth showing, because it is how a real model hits it and because it is where the
misleading diagnostics are at their worst. Compiled exactly like the models in `repro.py`
(`backend="tensorrt"`, `pass_through_build_failures=True`, `min_block_size=1`, with
`torch._dynamo.config.capture_dynamic_output_shape_ops = True`):

```python
class SliceScatter(torch.nn.Module):
    def forward(self, x, mask):
        rows = x[mask.nonzero().squeeze(1)]   # leading dim is unbacked
        out = torch.zeros_like(rows)
        out[1:] = rows[1:] * 2.0             # -> aten.slice_scatter(out, ..., 0, 1, INT64_MAX)
        return out
```

logs the same LINSPACE rank error and then dies in a completely unrelated converter:

```
ERROR:torch_tensorrt [TensorRT Conversion Context]:ITensor::getDimensions: Error Code 4: API Usage Error ((Unnamed Layer* 10) [Fill]: FillOperation LINSPACE requires that input at index 1 have rank 0. Actual rank is 1. In lowerParams at /_src/optimizer/common/nodes/fillNode.cpp:105)
...
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/conversion/aten_ops_converters.py", line 1447, in aten_ops_expand
    return impl.slice.expand(
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/conversion/impl/slice/ops.py", line 227, in expand
    initial_tensor_rank = len(input_t.shape)
ValueError: __len__() should return >= 0

While executing %expand : [num_users=1] = call_function[target=torch.ops.aten.expand.default](args = (%_reshape_copy, [%sym_size_int_1, 8]), kwargs = {})
```

`MODEL 1` and `MODEL 2` in `repro.py` are deliberately constructed so that the only
difference between them is whether `start` is symbolic; the `SliceScatter` snippet above is
the form the problem actually takes in a model.

### The dtype defect behind the rank defect

The same model is what exposes the second defect. With **only** the rank fix applied (the
squeeze below, no dtype unification), the LINSPACE inputs for that model come out as
`alpha=Int64` (the `sym_min(...)` start) and `beta=Int32` (the literal step), and TensorRT
logs:

```
ERROR:torch_tensorrt [TensorRT Conversion Context]:ITensor::getDimensions: Error Code 4: API Usage Error ((Unnamed Layer* 11) [Fill]: IFillLayer `alpha` and `beta` must have the same type. `alpha` is of type Int64 but `beta` is of type Int32. In validateTypes at /_src/optimizer/common/nodes/fillNode.cpp:32)
```

which, again only logged, again ends as `ValueError: __len__() should return >= 0` from
`expand`. For `MODEL 1` in `repro.py` both inputs happen to arrive as Int32, so the rank fix
alone is sufficient there; the dtype unification is what the `slice_scatter` route needs.

## To Reproduce

```
```

[repro.py](https://github.com/user-attachments/files/31532512/repro.py)

**repro.py**

```
import logging
import sys

import tensorrt
import torch
import torch_tensorrt

# nonzero() must stay in the graph so the data-dependent size reaches Torch-TensorRT;
# otherwise dynamo graph-breaks on it and the size becomes static.
torch._dynamo.config.capture_dynamic_output_shape_ops = True


class SymbolicStart(torch.nn.Module):
    """arange(start, end) where BOTH bounds are data-dependent -> fails."""

    def forward(self, x, mask):
        n = mask.nonzero().size(0)
        return x.index_select(0, torch.arange(n // 2, n, 1, device=x.device))


class StaticStart(torch.nn.Module):
    """arange(start, end) where only `end` is data-dependent -> compiles."""

    def forward(self, x, mask):
        n = mask.nonzero().size(0)
        return x.index_select(0, torch.arange(1, n, 1, device=x.device))


def run(label, model, args):
    torch._dynamo.reset()
    model = model.eval().cuda()
    # pass_through_build_failures surfaces the error; the default silently falls back
    # to eager for the whole subgraph.
    optimized_model = torch.compile(
        model,
        backend="tensorrt",
        # min_block_size=1 keeps both subgraphs in TensorRT instead of falling back
        # to Torch for being too small.
    )
    try:
        out = optimized_model(*args)
    except Exception:  # noqa: BLE001
        import traceback

        traceback.print_exc()
        return False
    torch.testing.assert_close(out, model(*args))
    return True


class _TrtLogCapture(logging.Handler):
    """Collects TensorRT conversion-context records.

    The LINSPACE rank error is only logged by TensorRT, never raised, so the Python
    exception that eventually surfaces does not identify the defect. Matching on the logged
    text is what makes the verdict specific to this bug rather than to any failure.
    """

    def __init__(self) -> None:
        super().__init__()
        self.text: list[str] = []

    def emit(self, record: logging.LogRecord) -> None:
        self.text.append(record.getMessage())


LINSPACE_RANK_ERROR = "LINSPACE requires that input at index 1 have rank 0"


def main():
    print(f"torch            {torch.__version__}")
    print(f"torch_tensorrt   {torch_tensorrt.__version__}")
    print(f"tensorrt         {tensorrt.__version__}")
    print(f"cuda             {torch.version.cuda}")
    print(f"python           {sys.version.split()[0]}")
    print(f"gpu              {torch.cuda.get_device_name(0)}")

    capture = _TrtLogCapture()
    trt_logger = logging.getLogger("torch_tensorrt [TensorRT Conversion Context]")
    trt_logger.addHandler(capture)
    trt_logger.setLevel(logging.ERROR)

    x = torch.randn((16, 8)).cuda()
    mask = torch.arange(16).cuda() % 2 == 0

    ok_symbolic = run("MODEL 1: symbolic start", SymbolicStart(), (x, mask))
    ok_static = run("MODEL 2: static start", StaticStart(), (x, mask))

    saw_linspace_rank_error = any(LINSPACE_RANK_ERROR in line for line in capture.text)

    print("\n===== summary =====")
    print(f"{'MODEL 1 (symbolic start)':30s} {'compiled' if ok_symbolic else 'FAILED'}")
    print(f"{'MODEL 2 (static start)':30s} {'compiled' if ok_static else 'FAILED'}")
    print(f"{'LINSPACE rank-0 error logged':30s} {saw_linspace_rank_error}")

    # Keyed on the logged TensorRT error, not merely on MODEL 1 failing, so an unrelated
    # failure cannot be mistaken for this defect.
    reproduced = saw_linspace_rank_error and not ok_symbolic
    if not reproduced and not ok_symbolic:
        print(
            "\nMODEL 1 failed but the LINSPACE rank error was not logged, so this is a\n"
            "different defect, not the one reported here."
        )
    return 0 if reproduced else 1


if __name__ == "__main__":
    raise SystemExit(main())
```

**output**

```
torch            2.13.0a0+9186a08b2c.nv26.07
torch_tensorrt   2.14.0a0
tensorrt         11.1.0.106
cuda             13.3
python           3.12.3
gpu              NVIDIA L4

===== MODEL 1: symbolic start =====
ERROR:torch_tensorrt [TensorRT Conversion Context]:ITensor::getDimensions: Error Code 4: API Usage Error ((Unnamed Layer* 16) [Fill]: FillOperation LINSPACE requires that input at index 1 have rank 0. Actual rank is 1. In lowerParams at /_src/optimizer/common/nodes/fillNode.cpp:105)
ERROR:torch_tensorrt [TensorRT Conversion Context]:ITensor::getDimensions: Error Code 4: API Usage Error (Output shape can not be computed for node (Unnamed Layer* 16) [Fill]. In needTypeAndDimensions at /_src/optimizer/shapeof/graphShapeAnalyzer.cpp:2986)
ERROR:torch_tensorrt [TensorRT Conversion Context]:IBuilder::buildEngineWithConfig: Error Code 3: API Usage Error ([GATHER]-[aten_ops.index_select.default]-[index_select_gather_layer_default]: rank(index) must >= nbElementWiseDims for GatherMode::kDEFAULT In validate at /_src/optimizer/api/layers/gatherLayer.cpp:88)
  File "/w/repro.py", line 104, in run
    out = optimized_model(*args)
  ...
    assert cuda_engine
AssertionError: 

----- MODEL 1: symbolic start: FAILED

===== MODEL 2: static start =====
----- MODEL 2: static start: OK (7, 8), matches eager

===== summary =====
MODEL 1 (symbolic start)       FAILED
MODEL 2 (static start)         compiled
LINSPACE rank-0 error logged   True

reproduced: True
```

## Expected behavior

`torch.arange` with a symbolic `start` should convert to a valid `IFillLayer` and the engine
should produce the same result as eager. `MODEL 1` and `MODEL 2` should both compile. We have
a fix for both defects, verified to make `MODEL 1` compile and match eager, and will open a
pull request with it.

Separately, and this is a legitimate second request: **the diagnostics here are actively
misleading and deserve a fix of their own.** A TensorRT API-usage error raised while a layer
is being constructed is currently only written to the log; conversion carries on with a layer
whose output dimensions are invalid, and the Python exception that eventually surfaces comes
from somewhere else -- a bare `AssertionError` from `assert cuda_engine`, or
`ValueError: __len__() should return >= 0` from the `expand` converter. Neither names
`arange`, neither names the FX node, and the second one actively points at an innocent
converter. Conversion should abort at the point the API error is reported and name the
offending FX node. The `Fill` layer is also unnamed today, which is why the log says
`(Unnamed Layer* 16)`; naming it would make the logged error traceable back to the node on
its own.

## Environment

SrivastavaKshitij:
# Description

Issue #4625 has two parts. 
1. Part 1 was solved on main by #4251 and #4456 
2. This PR fixes the second part which relates to dtype mismatch

`arange` resolves a common `value_dtype` for the sequence and passes it to each
`get_trt_tensor` call. `get_trt_tensor` applies that dtype when it constructs a constant, and
returns a value that is already an ITensor unchanged:

```python
    elif isinstance(input_val, TRTTensor):
        return input_val
```

So a dynamic `start` keeps the dtype of its incoming ITensor, while a literal `step` becomes a
constant of `value_dtype`. TensorRT asks that the LINSPACE `alpha` (input 1) and `beta`
(input 2) have the same type, and reports:

```
IFillLayer `alpha` and `beta` must have the same type.
`alpha` is of type Int32 but `beta` is of type Int64.
```

This routes every operand through a small helper that casts after `get_trt_tensor`, so the
resolved dtype holds for all of them. `cast_trt_tensor` returns the tensor unchanged when the
dtype already matches, so paths that were already consistent are unaffected.

# Issues

Closes #4625 

## Type of change

# Testing

Three cases with a dynamic bound are covered: an int32 `start`, an int64 `start`, and both
bounds dynamic. Of those, only the int64 `start` works today, because the sequence dtype for
integer operands is already int64 and matches.

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/arange.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/arange.py
@@ -41,6 +41,28 @@ def _sequence_dtype(
         return trt.DataType.INT64
 
 
+def _operand_as(
+    ctx: ConversionContext,
+    target: Target,
+    source_ir: Optional[SourceIR],
+    name: str,
+    value: Union[int, float, TRTTensor],
+    dtype: trt.DataType,
+    min_rank: int,
+) -> TRTTensor:
+    """
+    Materialize an arange operand as an ITensor of exactly `dtype`.
+
+    ``get_trt_tensor`` applies its ``dtype`` argument when it constructs a constant, and
+    returns a value that is already an ITensor unchanged. Casting afterwards lets the
+    resolved dtype hold for every operand, which is what LINSPACE asks of ``alpha`` and
+    ``beta``. ``cast_trt_tensor`` returns the tensor unchanged when the dtype already
+    matches.
+    """
+    operand = get_trt_tensor(ctx, value, name, dtype, min_rank=min_rank)
+    return cast_trt_tensor(ctx, operand, dtype, name + "_casted", target, source_ir)
+
+
 def arange(
     ctx: ConversionContext,
     target: Target,
@@ -62,8 +84,8 @@ def arange(
     # If any argument is a TRT tensor, use dynamic arange with a Fill layer
     if any(isinstance(x, TRTTensor) for x in (start, end, step)):
         value_dtype = _sequence_dtype(dtype, start, end, step)
-        start_rank_0 = get_trt_tensor(
-            ctx, start, name + "_start_rank_0", value_dtype, min_rank=0
+        start_rank_0 = _operand_as(
+            ctx, target, source_ir, name + "_start_rank_0", start, value_dtype, 0
         )
         # LINSPACE's start input requires rank 0; if the upstream ITensor came in
         # as rank-1 (e.g. a SymInt materialized by a sym_size op), reshape it.
@@ -75,11 +97,11 @@ def arange(
             )
             start_rank_0 = squeeze_layer.get_output(0)
 
-        start_rank_1 = get_trt_tensor(
-            ctx, start, name + "_start_rank_1", value_dtype, min_rank=1
+        start_rank_1 = _operand_as(
+            ctx, target, source_ir, name + "_start_rank_1", start, value_dtype, 1
         )
-        end = get_trt_tensor(ctx, end, name + "_end", value_dtype, min_rank=1)
-        step = get_trt_tensor(ctx, step, name + "_step", value_dtype, min_rank=1)
+        end = _operand_as(ctx, target, source_ir, name + "_end", end, value_dtype, 1)
+        step = _operand_as(ctx, target, source_ir, name + "_step", step, value_dtype, 1)
 
         # The number of elements is ceil((end - start) / step), computed as
         # -floor((start - end) / step) so that the whole expression stays in the
--- a/tests/py/dynamo/conversion/test_arange_aten.py
+++ b/tests/py/dynamo/conversion/test_arange_aten.py
@@ -93,6 +93,78 @@ class TestArangeConverter(DispatchTestCase):
             use_dynamo_tracer=False,
         )
 
+    @parameterized.expand([("int32", torch.int32), ("int64", torch.int64)])
+    def test_arange_dynamic_start(self, _, dtype):
+        """A dynamic `start` reaches the Fill layer as an ITensor.
+
+        LINSPACE requires `alpha` (start) and `beta` (step) to share a dtype. A dynamic
+        `start` keeps the dtype of its incoming ITensor, while a literal `step` is
+        materialized as a constant of the sequence dtype, so the two can differ.
+
+        Three cases with a dynamic bound are covered here: an int32 `start`, an int64
+        `start`, and both bounds dynamic. Of those, only the int64 `start` works today,
+        because the sequence dtype for integer operands is already int64 and matches.
+        """
+
+        class Arange(nn.Module):
+            def forward(self, start_tensor):
+                return torch.ops.aten.arange.start_step(start_tensor, 10, 1)
+
+        pyt_input = 2
+        inputs = [
+            torch_tensorrt.Input(
+                min_shape=(0,),
+                opt_shape=(2,),
+                max_shape=(5,),
+                dtype=dtype,
+                torch_tensor=torch.tensor(pyt_input, dtype=dtype).cuda(),
+                is_shape_tensor=True,
+            )
+        ]
+        self.run_test_with_dynamic_shape(
+            Arange(),
+            inputs,
+            use_example_tensors=False,
+            check_dtype=False,
+            pyt_inputs=[pyt_input],
+            use_dynamo_tracer=False,
+        )
+
+    def test_arange_dynamic_start_and_end(self):
+        """Both bounds dynamic, so neither is a constant carrying the sequence dtype."""
+
+        class Arange(nn.Module):
+            def forward(self, start_tensor, end_tensor):
+                return torch.ops.aten.arange.start_step(start_tensor, end_tensor, 1)
+
+        pyt_inputs = [2, 9]
+        inputs = [
+            torch_tensorrt.Input(
+                min_shape=(0,),
+                opt_shape=(2,),
+                max_shape=(5,),
+                dtype=torch.int32,
+                torch_tensor=torch.tensor(pyt_inputs[0], dtype=torch.int32).cuda(),
+                is_shape_tensor=True,
+            ),
+            torch_tensorrt.Input(
+                min_shape=(6,),
+                opt_shape=(9,),
+                max_shape=(12,),
+                dtype=torch.int64,
+                torch_tensor=torch.tensor(pyt_inputs[1], dtype=torch.int64).cuda(),
+                is_shape_tensor=True,
+            ),
+        ]
+        self.run_test_with_dynamic_shape(
+            Arange(),
+            inputs,
+            use_example_tensors=False,
+            check_dtype=False,
+            pyt_inputs=pyt_inputs,
+            use_dynamo_tracer=False,
+        )
+
 
 if __name__ == "__main__":
     run_tests()


SrivastavaKshitij:
@narendasan @apbose for viz

tests/py/dynamo/conversion/test_arange_aten.py
```diff
@@ -93,6 +93,78 @@ def forward(self, end_tensor):
...
+        """A dynamic `start` reaches the Fill layer as an ITensor.
+
+        LINSPACE requires `alpha` (start) and `beta` (step) to share a dtype. A dynamic
+        `start` keeps the dtype of its incoming ITensor, while a literal `step` is
+        materialized as a constant of the sequence dtype, so the two can differ.
+
+        Three cases with a dynamic bound are covered here: an int32 `start`, an int64
+        `start`, and both bounds dynamic. Of those, only the int64 `start` works today,
+        because the sequence dtype for integer operands is already int64 and matches.
+        """
```
apbose, line 107: Minor comment: this looks to be the pre-fix behavior.  `_operand_as()` now force-casts start_rank_0/start_rank_1 to value_dtype via `cast_trt_tensor` regardless of the incoming ITensor's own dtype, so start no longer "keeps the dtype of its incoming ITensor." And the "only int64 works today" claim is directly contradicted by the test right below it: test_arange_dynamic_start[int32] passes along with [int64]. Could you please update the docstring

py/torch_tensorrt/dynamo/conversion/impl/arange.py
```diff
@@ -41,6 +41,28 @@ def _sequence_dtype(
         return trt.DataType.INT64
```
apbose, line 42: this is unrelated to your changes. ` _sequence_dtype() `(same file, lines 29-41) has a bug that affects this PR's own scenario, the return` trt.DataType.INT64 `fallback is inside the for loop, so it only ever looks at the first operand and ignores the rest. For arange(dynamic_int_start, 10.0, 1.5), it returns INT64 (wrong) instead of FLOAT, because it stops checking after start and never sees that end/step are floats. Since your _operand_as() now enforces whatever dtype this function returns on every operand, this silently produces wrong integer output instead of the correct float sequence. 
You could include the fix in this PR, with a minimal test scenario

SrivastavaKshitij:
I can close this one as #4633 has landed.

PR #4631, opened 2026-08-28
fix(dynamo): handle dynamic scatter and index updates

### LINKED ISSUE #4602: 🐛 [Bug] scatter_add decomposition requires a static extent, aborting on a data-dependent one
##  Bug Description

The bug lives in `scatter_add_decomposition` in
`py/torch_tensorrt/dynamo/lowering/_decompositions.py`. The decomposition unrolls a Python
loop over the scatter dimension of `src`:

    src_dim = src_shape[dim]
    for i in range(0, src_dim):

`range()` requires a concrete `int`, so it calls `int()` on the SymInt, and an unbacked
(data-dependent) extent has no hint to specialize from. The whole compilation dies with
`GuardOnDataDependentSymNode: Could not extract specialized integer from data-dependent
expression ...` even though there is no `scatter_add` converter, so declining the
decomposition would have fallen back to Torch.

Note the bound on the symbol is irrelevant: `range()` cannot specialize from any hintless
symbol, so a bounded unbacked extent fails identically to an unbounded one. Both unbacked
cases below are therefore expected to fail, and the script reports each separately.

Cases:
  * unbounded unbacked src dim  -- custom op whose fake kernel returns new_dynamic_size(),
    range [0, int_oo]. Expected to FAIL (this is the bug).
  * bounded unbacked src dim    -- nonzero() on a statically shaped input, range [0, 16].
    Expected to FAIL as well; recorded to show the bound is irrelevant here.
  * static src dim (control)    -- identical model with a concrete extent. Expected to
    COMPILE, which is what proves the failures above are about the extent and not the op.

## To Reproduce

```
```

[repro.py](https://github.com/user-attachments/files/31527954/repro.py)

## Expected behavior

`scatter_add` over a data-dependent extent is genuinely not expressible in TensorRT - an
unknown number of loop iterations cannot be unrolled into a static graph - so the right
outcome is for the decomposition to DECLINE the node, not to abort the compilation.
Concretely: when `src_tensor.shape[dim]` is not a concrete `int`,
`scatter_add_decomposition` should return `NotImplemented`, which is the sentinel
`torch.fx.experimental.proxy_tensor.maybe_handle_decomp` already understands. That leaves
`aten.scatter_add.default` in the graph, the partitioner finds no converter for it and runs it
in Torch, and compilation of the rest of the model succeeds.

Declining is arguably the better outcome even for large *concrete* extents. In the model where
we first hit this, the extent was an expression around 153816, so a successful unroll would
have emitted six figures' worth of `zeros`/`select`/`unsqueeze`/`scatter`/`add` groups into
the graph. A size threshold above which the decomposition declines seems worth having
independently of the data-dependent case.

## Environment

### LINKED ISSUE #4609: 🐛 [Bug] index_put casts scatter indices but not updates, so IScatterLayer rejects mismatched types
##  Bug Description

TensorRT requires an `IScatterLayer`'s `updates` tensor to have the same type as the tensor
being scattered into. `py/torch_tensorrt/dynamo/conversion/impl/select.py ::
index_put_converter` casts the *indices* explicitly and then passes the *values* straight
through:

```python
indices_cat = cast_trt_tensor(ctx, indices_cat, trt.int32, f"{name}_idx_int32")
scatter_layer = ctx.net.add_scatter(
    input_tensor,
    indices_cat,
    flattened_values,
    trt.ScatterMode.ND,
)
```

Nothing casts `flattened_values` to `input_tensor.dtype`, so the destination type and the update
type agree only by luck. They diverge whenever a converter upstream of the values legitimately
produces a different TensorRT type than the torch graph's dtype for the same node.

`argmax` is exactly such a case, and the divergence is manufactured legitimately rather than
contrived: `aten.argmax` is `int64` in torch, while `impl.topk.argmax_argmin` returns the
`ITopKLayer` index output, which is `int32` in TensorRT. Scattering an argmax result into an
int64 tensor is ordinary, valid PyTorch; TensorRT rejects the resulting layer.

**Classification note:** the converter is silent about the mismatch, but the observable outcome
is a *raised error at engine build time*, not a wrong graph that runs. Please do not read this
as a numerical-divergence report:

```
IScatterLayer `input` and `updates` must have identical types. `input` type is Int64 and `updates` type is Int32.
```

The message above is only logged, however. What reaches Python is a bare `AssertionError` from
`assert cuda_engine` in `_TRTInterpreter.run`, with no mention of scatter, dtypes or
`index_put`.

Note that a sibling function in the same file gets this right. `index_put_scatter_add_plugin`
(the `accumulate=True` path) contains:

```python
# Values must match src dtype after any cast above.
if values.dtype != input_tensor.dtype:
    values = cast_trt_tensor(ctx, values, input_tensor.dtype, f"{name}_values_cast")
```

`index_put_converter` needs the same guarantee immediately before `add_scatter`.

## To Reproduce

```
```

[repro.py](https://github.com/user-attachments/files/31530048/repro.py)

**repro.py**

```
import sys
import traceback

import torch
import torch_tensorrt  # noqa: F401  (registers the "tensorrt" torch.compile backend)


class Int64Destination(torch.nn.Module):
    """data[idx] = scores.argmax(1): int64 in torch, int32 out of the TRT topk layer."""

    def forward(self, data: torch.Tensor, idx: torch.Tensor, scores: torch.Tensor) -> torch.Tensor:
        values = torch.ops.aten.argmax.default(scores, 1)
        return torch.ops.aten.index_put.default(data, [idx], values)


class Int32Destination(torch.nn.Module):
    """Control: the same scatter where the torch dtype and the TRT type agree."""

    def forward(self, data: torch.Tensor, idx: torch.Tensor, scores: torch.Tensor) -> torch.Tensor:
        values = torch.ops.aten.argmax.default(scores, 1).to(torch.int32)
        return torch.ops.aten.index_put.default(data, [idx], values)


def run_case(label: str, model: torch.nn.Module, args: tuple[torch.Tensor, ...]) -> bool:
    """Compiles and runs `model` under the TensorRT backend. Returns True if it worked."""
    torch._dynamo.reset()
    model = model.eval().cuda()
    optimized = torch.compile(
        model,
        backend="tensorrt",
    )
    try:
        out = optimized(*args)
        torch.testing.assert_close(out, model(*args))
    except Exception:  # pylint: disable=broad-except
        traceback.print_exc()
        return False
    print(f"----- {label}: OK {out.dtype} {tuple(out.shape)}, matches eager", flush=True)
    return True


def main(argv: list[str] | tuple[str, ...] = ()) -> int:
    """Runs the int64 destination (the bug) and the int32 destination (the control)."""
    del argv
    print(f"torch          {torch.__version__}")
    print(f"torch_tensorrt {torch_tensorrt.__version__}")

    idx = torch.tensor([1, 3]).cuda()
    scores = torch.randn((2, 5)).cuda()

    bug = run_case(
        "A: int64 destination, int32 TRT updates",
        Int64Destination(),
        (torch.zeros(8, dtype=torch.int64).cuda(), idx, scores),
    )
    control = run_case(
        "control: int32 destination, int32 TRT updates",
        Int32Destination(),
        (torch.zeros(8, dtype=torch.int32).cuda(), idx, scores),
    )

    reproduced = not bug and control
    return 0 if reproduced else 1


if __name__ == "__main__":
    sys.exit(main(argv=sys.argv))

```

**output**

```
===== A: int64 destination, int32 TRT updates =====
ERROR:torch_tensorrt [TensorRT Conversion Context]:ITensor::getDimensions: Error Code 4: API Usage Error ([SCATTER]-[aten_ops.index_put.default]-[index_put_scatter]: IScatterLayer `input` and `updates` must have identical types. `input` type is Int64 and `updates` type is Int32. In validateTypes at /_src/optimizer/common/nodes/scatterNode.cpp:100)
ERROR:torch_tensorrt [TensorRT Conversion Context]:ITensor::getDimensions: Error Code 4: API Usage Error (Output shape can not be computed for node [SCATTER]-[aten_ops.index_put.default]-[index_put_scatter]. In needTypeAndDimensions at /_src/optimizer/shapeof/graphShapeAnalyzer.cpp:2986)
ERROR:torch_tensorrt [TensorRT Conversion Context]:IBuilder::buildEngineWithConfig: Error Code 4: API Usage Error ([SCATTER]-[aten_ops.index_put.default]-[index_put_scatter]: IScatterLayer `input` and `updates` must have identical types. `input` type is Int64 and `updates` type is Int32. In validateTypes at /_src/optimizer/common/nodes/scatterNode.cpp:100)
  File "/w/repro.py", line 69, in run_case
    out = optimized(*args)
          ^^^^^^^^^^^^^^^^
[...]
                         ^^^^^^^^^^^^^^^^^
    assert cuda_engine
           ^^^^^^^^^^^
----- A: int64 destination, int32 TRT updates: FAILED
AssertionError:

===== control: int32 destination, int32 TRT updates =====
----- control: int32 destination, int32 TRT updates: OK torch.int32 (8,), matches eager

reproduced: True
```

## Expected behavior

`index_put_converter` should cast `flattened_values` to `input_tensor.dtype` before calling
`ctx.net.add_scatter`, in the same way it already casts `indices_cat` to `trt.int32` on the line
immediately above, and in the same way `index_put_scatter_add_plugin` already casts its values.
`data[idx] = scores.argmax(1)` on an int64 destination should then compile and match eager, as
the int32 control in the reproducer does today.

We have not implemented or tested such a patch, so treat the above as the shape of the expected
fix rather than a verified one.

Separately: the diagnostics are poor. TensorRT states the problem precisely ("`input` type is
Int64 and `updates` type is Int32"), but that text is only logged; the exception delivered to
the caller is a bare `AssertionError` with an empty message from `assert cuda_engine` in
`_TRTInterpreter.run`. Please consider surfacing the builder error text in the raised exception,
so users are not left correlating a log line against an assertion.

## Environment

 - Pytorch ngc container : 26.07-py3

### LINKED ISSUE #4621: 🐛 [Bug] index_put with a None index materialises one index per element, failing on a runtime-known extent
##  Bug Description

A `None` entry in an `index_put` index list leaves that dimension unindexed, and Torch
broadcasts `values` across it. ScatterND has no equivalent of an unindexed dimension, so
`index_put_converter` compensates by materialising one index row per element of the free
dimensions -- `N * F_volume` rows, where `N` is the index length -- and separately reshaping
and expanding `values` to `(N,) + F_shape_values`. That construction is only correct when
those extents are known at build time. When the index length is only known at runtime, the
per-dimension broadcast check accepts a mismatch instead of rejecting it, and the emitted
expand ends up reading past the end of an axis.

`py/torch_tensorrt/dynamo/conversion/impl/select.py :: index_put_converter`

```python
    expected_shape = (N,) + tuple(F_shape_values)                       # line 970
    ...
        else:
            # Discontinuous case (K > 1 or K == 0)
            values_shape_padded = [1] * (
                len(expected_shape) - len(values.shape)
            ) + list(values.shape)
            broadcast_shape = []
            for exp_dim, val_dim in zip(expected_shape, values_shape_padded):
                if val_dim == DYNAMIC_DIM or exp_dim == DYNAMIC_DIM:     # line 1065
                    broadcast_shape.append(-1)
                elif val_dim == 1 or exp_dim == val_dim:
                    broadcast_shape.append(exp_dim)
                else:
                    raise ValueError(
                        f"Cannot broadcast {values.shape} to {expected_shape}"
                    )

            values_expanded = impl.slice.expand(
                ctx,
                target,
                source_ir,
                f"{name}_expand_values",
                values,
                expected_shape,
            )
```

For `data[:, idx] = values` on rank-2 `data` with 1-D `values`, `K == 1` and `len(F) == 1`, so
neither of the earlier branches applies and this `else` branch (labelled "Discontinuous case")
is the one taken. `expected_shape` is `(N, 32)` with `N` dynamic; `values_shape_padded` is
`[1, -1]`. The second pair, `(32, -1)`, is a real mismatch -- the static free-dim extent
against the runtime index length -- but line 1065 fires first and accepts it by appending
`-1`.

`impl.slice.expand` is then called with `expected_shape`. It prepends ones to reach the target
rank, so the rank-1 `values` of runtime extent `N` becomes `(1, N)`, and its stride loop takes
the branch commented `# No broadcasting is happening. The output should have the same size as
input at this dimension.` for the axis whose input extent is dynamic and whose target is the
static `32`. The result is an `ISliceLayer` reading 32 elements along an axis whose runtime
extent is only `N`.

Worth noting: `broadcast_shape` is computed in that branch but never used -- the `expand` call
passes `expected_shape` -- so the loop's only effect is whether it raises, and line 1065 is
what stops it raising.

**This fails while BUILDING the engine, not at execution.** TensorRT's shape analyzer catches
the contradiction during `buildEngineWithConfig`:

```
IBuilder::buildEngineWithConfig: Error Code 4: Internal Error (kOPT values for profile 0
violate shape constraints: [SLICE]-[aten_ops.index_put.default]-[index_put_expand_values]:
ISliceLayer has out of bounds access on axis 1 Condition '<' violated: 31 >= 4.
```

`31` is the last index of the free dimension of extent 32; `4` is the index length at kOPT.

## To Reproduce

```
```

[repro.py](https://github.com/user-attachments/files/31531849/repro.py)

**repro.py**

```
import sys
import traceback

import torch
import torch_tensorrt  # noqa: F401  (registers the "tensorrt" torch.compile backend)


class OneDimValues(torch.nn.Module):
    """data[:, idx] = values with 1-D values, broadcast across the free dim 0."""

    def forward(self, data: torch.Tensor, idx: torch.Tensor) -> torch.Tensor:
        values = idx.to(torch.float32)
        return torch.ops.aten.index_put.default(data, [None, idx], values)


class FullRankValues(torch.nn.Module):
    """data[:, idx] = values with values in torch's layout for data[:, idx]."""

    def forward(self, data: torch.Tensor, idx: torch.Tensor) -> torch.Tensor:
        values = idx.to(torch.float32).reshape(1, -1, 1).expand(32, -1, 128)
        return torch.ops.aten.index_put.default(data, [None, idx], values)


def run_case(
    label: str, model: torch.nn.Module, data: torch.Tensor, idx: torch.Tensor, dynamic: bool
) -> bool:
    """Compiles and runs `model`, marking the index length dynamic if asked."""
    torch._dynamo.reset()
    idx = idx.clone()
    if dynamic:
        torch._dynamo.mark_dynamic(idx, 0)
    model = model.eval().cuda()
    optimized = torch.compile(
        model,
        backend="tensorrt",
    )
    try:
        out = optimized(data, idx)
        torch.testing.assert_close(out, model(data, idx))
    except Exception:  # pylint: disable=broad-except
        traceback.print_exc()
        return False
    return True


def main(argv: list[str] | tuple[str, ...] = ()) -> int:
    """Runs the two runtime-extent formulations and the static-extent control."""
    del argv
    print(f"torch          {torch.__version__}")
    print(f"torch_tensorrt {torch_tensorrt.__version__}")
    # Severity 4 == VERBOSE. Not needed on the versions tested here - the build-time
    # error is reported at ERROR severity either way - but kept because when this
    # failure surfaces at execution instead, the engine can throw before writing its
    # outputs, in which case the first read reports `_Map_base::at` and python surfaces
    # an unrelated "a Tensor with 0 elements cannot be converted to Scalar".
    torch.ops.tensorrt.set_logging_level(4)

    idx = torch.tensor([1, 3, 5, 7]).cuda()
    data_2d = torch.zeros((32, 64)).cuda()
    data_3d = torch.zeros((32, 64, 128)).cuda()

    case_a = run_case("A: 1-D values, runtime-length index", OneDimValues(), data_2d, idx, True)
    case_b = run_case(
        "B: full-rank values, runtime-length index", FullRankValues(), data_3d, idx, True
    )
    case_c = run_case(
        "C: full-rank values, static-length index", FullRankValues(), data_3d, idx, False
    )

    reproduced = not case_a or not case_b
    print(f"\nstatic-extent control passed: {case_c}")
    print(f"reproduced: {reproduced}")
    return 0 if reproduced else 1


if __name__ == "__main__":
    sys.exit(main(argv=sys.argv))
```

**output**

```
torch          2.13.0a0+9186a08b2c.nv26.07
torch_tensorrt 2.14.0a0

===== A: 1-D values, runtime-length index =====
WARNING:torch_tensorrt.dynamo.partitioning.common:Dynamic input arg1_1 (shape: torch.Size([s95])) has no max bound for dim 0, attempting to use a sane default (max: min(1) * 2^12). Please set an upper bound using torch._dynamo.mark_dynamic or torch.export.Dim
ERROR:torch_tensorrt [TensorRT Conversion Context]:IBuilder::buildEngineWithConfig: Error Code 4: Internal Error (kOPT values for profile 0 violate shape constraints: [SLICE]-[aten_ops.index_put.default]-[index_put_expand_values]: ISliceLayer has out of bounds access on axis 1 Condition '<' violated: 31 >= 4. In operator() at /_src/runtime/core/shapeMachine.cpp:628 In evaluateShapeChecks at /_src/optimizer/common/shapeCompiler.cpp:1375)
  File "/w/repro.py", line 97, in run_case
    out = optimized(data, idx)
          ^^^^^^^^^^^^^^^^^^^^
  [...]
                         ^^^^^^^^^^^^^^^^^
    assert cuda_engine
           ^^^^^^^^^^^
AssertionError: 

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"

----- A: 1-D values, runtime-length index: FAILED

===== B: full-rank values, runtime-length index =====
WARNING:torch_tensorrt.dynamo.partitioning.common:Dynamic input arg1_1 (shape: torch.Size([s95])) has no max bound for dim 0, attempting to use a sane default (max: min(1) * 2^12). Please set an upper bound using torch._dynamo.mark_dynamic or torch.export.Dim
----- B: full-rank values, runtime-length index: OK (32, 64, 128), matches eager

===== C: full-rank values, static-length index =====
----- C: full-rank values, static-length index: OK (32, 64, 128), matches eager

static-extent control passed: True
reproduced: True
```

## Expected behavior

`data[:, idx] = values` with a rank-deficient `values` and a runtime-length index should
either be converted correctly -- broadcasting `values` across the free dimension the same way
Torch does -- or be declined.

Broadcasting across a dimension whose extent is only known at runtime is genuinely awkward to
express in ScatterND, since the whole index-materialisation strategy assumes the extents are
build-time constants. If it cannot be expressed, the right outcome is for the converter (or
its capability validator) to DECLINE the node, so the partitioner leaves `aten.index_put` in
Torch and the rest of the model still compiles -- rather than emitting a network TensorRT will
reject and taking the whole compilation down. The specific thing to fix in either case is line
1065: a pair like `(32, -1)` is a mismatch that happens to involve a dynamic extent, and
collapsing it to `-1` turns a detectable problem into a malformed network.

**Separate, smaller request: the diagnostics here are poor.** The TensorRT build error is
*logged*, but what Python actually raises is a bare `AssertionError` with an empty message,
from `assert cuda_engine` at `_TRTInterpreter.py:653`. A user who does not have the TensorRT
logger output in front of them gets `BackendCompilerFailed: backend='tensorrt' raised:
AssertionError:` and nothing else -- no layer name, no shape constraint, no mention of
`index_put`. Raising an exception that carries the builder's error text would make this class
of failure far cheaper to diagnose, independently of the broadcast bug.

## Environment

narendasan:
## Summary

- skip scatter_add decomposition when its extent is data-dependent so supported dynamic graphs remain compilable
- normalize index_put update dtype to the destination tensor before TensorRT scatter
- represent None-index expansion with dynamic arange/broadcast logic instead of one index per statically known element
- make static arange conversion safe under FakeTensorMode and clamp inferred dynamic profile minima

## Issues

Closes #4602
Closes #4609
Closes #4621

## Tests

- pytest tests/py/dynamo/conversion/test_index_put_aten.py -n0 (74 passed)
- pytest tests/py/dynamo/lowering/test_decompositions.py tests/py/dynamo/models/test_dynamic_shape_user_bounds.py -n0 (94 passed)
- original reproducers for #4602, #4609, and #4621 no longer reproduce

## Stack

5 of 7. Base: #4630. Follow-up: narendasan/fix-converter-profile-edge-cases.

### DIFF
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -717,11 +717,35 @@ FP32_ACC_ATTENTION_DECOMPOSITIONS = {
 }
 
 
+def _has_symbolic_scatter_add_extent(
+    graph_module: Optional[torch.fx.GraphModule],
+) -> bool:
+    """Return whether scatter_add would require unrolling a symbolic extent."""
+    if graph_module is None:
+        return False
+
+    for node in graph_module.graph.nodes:
+        if node.target != torch.ops.aten.scatter_add.default or len(node.args) < 4:
+            continue
+        dim = node.args[1]
+        src_node = node.args[3]
+        if not isinstance(dim, int) or not isinstance(src_node, torch.fx.Node):
+            continue
+        src_val = src_node.meta.get("val", src_node.meta.get("example_value"))
+        if not isinstance(src_val, torch.Tensor) or not src_val.ndim:
+            continue
+        if isinstance(src_val.shape[get_positive_dim(dim, src_val.ndim)], torch.SymInt):
+            return True
+
+    return False
+
+
 def get_decompositions(
     enable_experimental_decompositions: bool = False,
     decompose_attention: bool = False,
     use_distributed_mode_trace: bool = False,
     use_fp32_acc: bool = False,
+    graph_module: Optional[torch.fx.GraphModule] = None,
 ) -> Dict[OpOverload, Callable[[Any], Any]]:
     trt_decomps = (
         dict(TORCH_TRT_DECOMPOSITIONS)
@@ -749,7 +773,7 @@ def get_decompositions(
             for decomp in _core_aten_decompositions
             if decomp not in discard_decompositions
         }
-        return {**CORE_ATEN_DECOMPOSITIONS_FILTERED, **trt_decomps}
+        decompositions = {**CORE_ATEN_DECOMPOSITIONS_FILTERED, **trt_decomps}
     else:
         # changes made here due to torch2.6 changes https://github.com/pytorch/pytorch/pull/135080
         decomp_table = {}
@@ -763,8 +787,16 @@ def get_decompositions(
             and decomp not in ATTENTION_DECOMPOSITION_OPS
         }
 
-        return {
+        decompositions = {
             **ENABLED_TORCH_DECOMPOSITIONS,
             **DECOMP_TABLE_FILTERED,
             **trt_decomps,
         }
+
+    if _has_symbolic_scatter_add_extent(graph_module):
+        # The custom decomposition uses a Python range over this extent.
+        # Keeping the op lets partitioning fall back to Torch without trying
+        # to specialize an unbacked or otherwise dynamic SymInt.
+        decompositions.pop(torch.ops.aten.scatter_add.default, None)
+
+    return decompositions
--- a/py/torch_tensorrt/dynamo/lowering/_decompositions.py
+++ b/py/torch_tensorrt/dynamo/lowering/_decompositions.py
@@ -717,35 +717,11 @@ FP32_ACC_ATTENTION_DECOMPOSITIONS = {
 }
 
 
-def _has_symbolic_scatter_add_extent(
-    graph_module: Optional[torch.fx.GraphModule],
-) -> bool:
-    """Return whether scatter_add would require unrolling a symbolic extent."""
-    if graph_module is None:
-        return False
-
-    for node in graph_module.graph.nodes:
-        if node.target != torch.ops.aten.scatter_add.default or len(node.args) < 4:
-            continue
-        dim = node.args[1]
-        src_node = node.args[3]
-        if not isinstance(dim, int) or not isinstance(src_node, torch.fx.Node):
-            continue
-        src_val = src_node.meta.get("val", src_node.meta.get("example_value"))
-        if not isinstance(src_val, torch.Tensor) or not src_val.ndim:
-            continue
-        if isinstance(src_val.shape[get_positive_dim(dim, src_val.ndim)], torch.SymInt):
-            return True
-
-    return False
-
-
 def get_decompositions(
     enable_experimental_decompositions: bool = False,
     decompose_attention: bool = False,
     use_distributed_mode_trace: bool = False,
     use_fp32_acc: bool = False,
-    graph_module: Optional[torch.fx.GraphModule] = None,
 ) -> Dict[OpOverload, Callable[[Any], Any]]:
     trt_decomps = (
         dict(TORCH_TRT_DECOMPOSITIONS)
@@ -792,11 +768,5 @@ def get_decompositions(
             **DECOMP_TABLE_FILTERED,
             **trt_decomps,
         }
-
-    if _has_symbolic_scatter_add_extent(graph_module):
-        # The custom decomposition uses a Python range over this extent.
-        # Keeping the op lets partitioning fall back to Torch without trying
-        # to specialize an unbacked or otherwise dynamic SymInt.
-        decompositions.pop(torch.ops.aten.scatter_add.default, None)
 
     return decompositions
--- a/py/torch_tensorrt/_Input.py
+++ b/py/torch_tensorrt/_Input.py
@@ -196,16 +196,6 @@ class Input(object):
                 }
                 self.shape_mode = Input._ShapeMode.DYNAMIC
 
-                # Warn if min_shape has any 0 dimension (empty tensor) - TensorRT doesn't support this
-                # @apbose: Is this warning necessary?
-                if any(dim == 0 for dim in self.shape["min_shape"]):
-                    logger.warning(
-                        f"min_shape contains a 0 dimension: {self.shape['min_shape']}. "
-                        "TensorRT does not support dynamic shapes with min dimension of 0 (empty tensors). "
-                        "TensorRT will internally clamp min dimensions to 1, which may cause runtime errors "
-                        "if you try to run inference with empty tensor inputs."
-                    )
-
                 # Namedtuple shape API: field names encode per-axis dimension names.
                 # Convert to shared_dims so _tracer.py needs no changes — axes with the
                 # same name across inputs become one shared torch.export.Dim.
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -400,6 +400,7 @@ def cross_compile_for_windows(
             decompose_attention,
             use_distributed_mode_trace,
             use_fp32_acc=use_fp32_acc,
+            graph_module=exported_program.graph_module,
         )
     )
 
@@ -806,6 +807,7 @@ def compile(
             enable_experimental_decompositions,
             decompose_attention,
             use_distributed_mode_trace,
+            graph_module=exported_program.graph_module,
             use_fp32_acc=use_fp32_acc,
         )
     )
@@ -2113,6 +2115,7 @@ def convert_exported_program_to_serialized_trt_engine(
             decompose_attention,
             use_distributed_mode_trace,
             use_fp32_acc=use_fp32_acc,
+            graph_module=exported_program.graph_module,
         )
     )
 
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -40,6 +40,7 @@ from torch_tensorrt.dynamo.conversion._ConverterRegistry import (
 from torch_tensorrt.dynamo.debug._DebuggerConfig import DebuggerConfig
 from torch_tensorrt.dynamo.debug._supports_debugger import fn_supports_debugger
 from torch_tensorrt.dynamo.lowering import (
+    filter_decomposition_table,
     get_decompositions,
     post_lowering,
     pre_export_lowering,
@@ -395,12 +396,14 @@ def cross_compile_for_windows(
     logger.info("Compilation Settings: %s\n", settings)
     exported_program = pre_export_lowering(exported_program, settings)
     exported_program = exported_program.run_decompositions(
-        get_decompositions(
-            enable_experimental_decompositions,
-            decompose_attention,
-            use_distributed_mode_trace,
-            use_fp32_acc=use_fp32_acc,
-            graph_module=exported_program.graph_module,
+        filter_decomposition_table(
+            get_decompositions(
+                enable_experimental_decompositions,
+                decompose_attention,
+                use_distributed_mode_trace,
+                use_fp32_acc=use_fp32_acc,
+            ),
+            exported_program.graph_module,
         )
     )
 
@@ -803,12 +806,14 @@ def compile(
     logger.info("Compilation Settings: %s\n", settings)
     exported_program = pre_export_lowering(exported_program, settings)
     exported_program = exported_program.run_decompositions(
-        get_decompositions(
-            enable_experimental_decompositions,
-            decompose_attention,
-            use_distributed_mode_trace,
-            graph_module=exported_program.graph_module,
-            use_fp32_acc=use_fp32_acc,
+        filter_decomposition_table(
+            get_decompositions(
+                enable_experimental_decompositions,
+                decompose_attention,
+                use_distributed_mode_trace,
+                use_fp32_acc=use_fp32_acc,
+            ),
+            exported_program.graph_module,
         )
     )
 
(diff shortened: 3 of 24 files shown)


py/torch_tensorrt/dynamo/lowering/_decompositions.py
```diff
@@ -763,8 +787,16 @@ def get_decompositions(
...
             **ENABLED_TORCH_DECOMPOSITIONS,
             **DECOMP_TABLE_FILTERED,
             **trt_decomps,
         }
+
+    if _has_symbolic_scatter_add_extent(graph_module):
+        # The custom decomposition uses a Python range over this extent.
+        # Keeping the op lets partitioning fall back to Torch without trying
+        # to specialize an unbacked or otherwise dynamic SymInt.
+        decompositions.pop(torch.ops.aten.scatter_add.default, None)
```
narendasan: @apbose can you make this a real pass instead of a conditional decompoistion table? Or we should have like a filter decomp table phase or something. Dont think the graph module should be in here

sylvesterkaczmarek (commented):
`construct_dynamic_input()` now clamps a symbolic minimum of 0 to 1. For data-dependent extents such as `nonzero`, zero is a valid runtime shape, so compilation can succeed with a profile that rejects the legitimate empty result at runtime. Keep that boundary out of TensorRT (or preserve a Torch fallback) when the symbolic range includes zero, and add an all-false `nonzero` runtime regression.

PR #4632, opened 2026-08-28
fix(dynamo): harden dynamic converter boundaries

### LINKED ISSUE #4607: 🐛 [Bug] slice does not clamp a finite stop on a dynamic axis, baking the literal stop into the engine
##  Bug Description

`slice_op` in `py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py :: slice_op` clamps the
slice `stop` bound to the axis length only when the axis is static. On a dynamic axis it takes
the `else` branch, and that branch's inner condition covers only four shapes of input: a
non-`int` (ITensor) start/stop, a negative start/stop, an omitted stop (`stop_dynamic_None`),
and `stop == sys.maxsize`. A plain finite positive `int` stop matches none of them, so control
falls straight through the entire dynamic-shape block:

```python
if input.shape[dim] != -1 and isinstance(start, int) and isinstance(stop, int):
    start = get_positive_dim(start, input.shape[dim])
    stop = get_positive_dim(stop, input.shape[dim])   # static path: min(stop, dim_size)
    start_slice[dim] = start
else:
    # the start and stop or None is dynamic along dim or or start or stop is an ITensor
    if (
        not (isinstance(start, int))
        or not (isinstance(stop, int))
        or start < 0
        or stop < 0
        or stop_dynamic_None
        or stop == sys.maxsize
    ):
        ...            # builds a runtime-computed extent and returns from inside here
        return layer.get_output(0)

output_shape[dim] = math.ceil((stop - start) / step)   # reached with a FINITE stop
```

The result is a constant extent baked into the `ISliceLayer` on an axis whose real extent is
only known at runtime. Torch's own semantics are that a slice stops at the end of the tensor,
so the two agree only while the axis happens to be at least `stop` long.

**The converter is silent about this.** It neither logs nor raises, and nothing in the Python
traceback names `slice/ops.py`. What surfaces is a TensorRT shape-analysis error during engine
build naming the slice layer, and then a bare `AssertionError` from `assert cuda_engine` in
`_TRTInterpreter.run`. Identification therefore rests on the layer identity in the TensorRT
message plus two controls, both included in the reproducer below:

* the *same* finite stop on a *static* axis compiles, because `get_positive_dim` clamps there;
* an *omitted* stop on the *same* dynamic axis compiles, because `stop == sys.maxsize` is one of
  the conditions the dynamic branch does test. `x[:, 5:]` reaches the converter as
  `stop = INT64_MAX`, which equals `sys.maxsize`; the reproducer passes `sys.maxsize` explicitly
  so the control does not depend on that detail.

One further limitation of what TensorRT can catch, stated up front: only a stop *above* the
optimization profile's kOPT extent is rejected. A stop *below* the example extent (case B in
the reproducer, `stop=48` with an example extent of 64) compiles cleanly and matches eager for
that input, yet the engine still carries the fixed extent 48 and is wrong for any runtime input
shorter than 48. TensorRT cannot detect that, because the engine is valid for its profile.

## To Reproduce

```
<!-- If you have a code sample, error messages, stack traces, please provide it here as well -->
```

[repro.py](https://github.com/user-attachments/files/31529538/repro.py)

**repro.py**

```
import sys
import traceback
from typing import Final

import torch
import torch_tensorrt  # noqa: F401  (registers the "tensorrt" torch.compile backend)

# The example extent of the dynamic axis, which is also the profile's kOPT extent.
HINT_EXTENT: Final[int] = 64


class FiniteStop(torch.nn.Module):
    """x[:, :stop] on a runtime-length axis, with a plain Python int stop."""

    def __init__(self, stop: int) -> None:
        super().__init__()
        self.stop = stop

        return torch.ops.aten.slice.Tensor(x, 1, 0, self.stop) * 2.0


class OmittedStop(torch.nn.Module):
    """Control: x[:, 5:], which aten canonicalizes to stop=INT64_MAX=sys.maxsize.

    `stop == sys.maxsize` is one of the conditions the dynamic branch does test for, so
    this identical-shaped slice builds a runtime-computed extent and compiles.
    """

        return torch.ops.aten.slice.Tensor(x, 1, 5, sys.maxsize) * 2.0


def run_case(label: str, model: torch.nn.Module, x: torch.Tensor, dynamic: bool) -> bool:
    """Compiles `model` for `x`, marking dim 1 dynamic if asked. True if it worked."""
    torch._dynamo.reset()
    x = x.clone()
    if dynamic:
        torch._dynamo.mark_dynamic(x, 1)
    model = model.eval().cuda()
    optimized = torch.compile(
        model,
        backend="tensorrt",
    )
    try:
        out = optimized(x)
        torch.testing.assert_close(out, model(x))
    except Exception:  # pylint: disable=broad-except
        traceback.print_exc()
        return False
    return True


def main(argv: list[str] | tuple[str, ...] = ()) -> int:
    """Runs a finite stop above and below the example extent, plus two controls."""
    del argv
    print(f"torch          {torch.__version__}")
    print(f"torch_tensorrt {torch_tensorrt.__version__}")

    x = torch.randn((2, HINT_EXTENT)).cuda()

    above = run_case("A: stop 100 on a dynamic axis", FiniteStop(100), x, dynamic=True)
    below = run_case("B: stop 48 on a dynamic axis", FiniteStop(48), x, dynamic=True)
    control_maxsize = run_case(
        "control: omitted stop on a dynamic axis", OmittedStop(), x, dynamic=True
    )
    control_static = run_case(
        "control: stop 100 on a static axis", FiniteStop(100), x, dynamic=False
    )

    reproduced = (not above or not below) and control_maxsize and control_static
    return 0 if reproduced else 1


if __name__ == "__main__":
    sys.exit(main(argv=sys.argv))

```

**output**

```
torch          2.13.0a0+9186a08b2c.nv26.07
torch_tensorrt 2.14.0a0

===== A: stop 100 on a dynamic axis =====
ERROR:torch_tensorrt [TensorRT Conversion Context]:IBuilder::buildEngineWithConfig: Error Code 4: API Usage Error (ISliceLayer [SLICE]-[aten_ops.slice.Tensor]-[slice_1]: [SLICE]-[aten_ops.slice.Tensor]-[slice_1]: ISliceLayer has out of bounds access on axis 1 In processCheck at /_src/optimizer/shapeof/graphShapeAnalyzer.cpp:1042 In analyzeShapes at /_src/optimizer/shapeof/graphShapeAnalyzer.cpp:2785)
  File "/w/repro.py", line 79, in run_case
    out = optimized(x)
          ^^^^^^^^^^^^
[...]
                         ^^^^^^^^^^^^^^^^^
    assert cuda_engine
           ^^^^^^^^^^^
AssertionError:

----- A: stop 100 on a dynamic axis: FAILED

===== B: stop 48 on a dynamic axis =====
WARNING:torch_tensorrt.dynamo.partitioning.common:Dynamic input arg1_1 (shape: torch.Size([2, s27])) has no max bound for dim 1, attempting to use a sane default (max: min(48) * 2^12). Please set an upper bound using torch._dynamo.mark_dynamic or torch.export.Dim
----- B: stop 48 on a dynamic axis: OK (2, 48), matches eager

===== control: omitted stop on a dynamic axis =====
----- control: omitted stop on a dynamic axis: OK (2, 59), matches eager

===== control: stop 100 on a static axis =====
----- control: stop 100 on a static axis: OK (2, 64), matches eager

reproduced: True
```

## Expected behavior

A finite `stop` on a dynamic axis should be clamped to the runtime extent of that axis, the same
way `get_positive_dim` clamps it on the static path. Concretely, the dynamic branch should also
be taken for a finite positive `int` stop, and the emitted stop should be
`min(stop, dim_extent)` computed from the input's shape tensor. Then `x[:, :100]` on an axis of
runtime length 64 would produce 64 columns, as it does in Torch, instead of a slice layer that
reads out of bounds.

If clamping on the dynamic path is not something you want to add, the acceptable alternative is
for the converter to DECLINE the node (so the partitioner falls back to Torch for that slice)
rather than emit a layer that is only correct for a subset of the profile.

Separately, and independently useful: the diagnostics here are poor. The converter makes a
silent decision, the TensorRT error is only logged, and the exception that actually reaches
Python is a bare `AssertionError` with an empty message from `assert cuda_engine` in
`_TRTInterpreter.run`. Please consider raising an exception that carries the TensorRT builder
error text, so a user does not have to correlate a log line with an assertion.

## Environment

 - Pytorch ngc container : 26.07-py3

### LINKED ISSUE #4608: 🐛 [Bug] rank-8+ tensors cross the partition boundary and fail in add_input, not at their producer
##  Bug Description

`trt.Dims` holds at most `trt.Dims.MAX_DIMS` axes, which is 8 (verified in this container:
`python -c "import tensorrt as trt; print(trt.Dims.MAX_DIMS)"` prints `8`). Partitioning only
asks whether a converter exists for a node:

* `py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py :: OpSupportTester.is_node_supported`
* `py/torch_tensorrt/dynamo/partitioning/_global_partitioner.py :: TorchTensorRTOperatorSupport.is_node_supported`

Neither asks whether the tensors crossing the partition boundary are representable in
`trt.Dims`. **The crux is that declining only the producing op is not enough.** In the
reproducer, a custom op with no converter produces a rank-10 tensor, so the partitioner
correctly declines it and it runs in Torch. But its consumer (`* 2.0`) *does* have a converter,
so the consumer is accepted into a TensorRT subgraph, and that drags the rank-10 tensor across
the boundary as a network *input*. It then fails as an input, not at its producer, in

```python
# py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py :: placeholder
return self.ctx.net.add_input(
    name=target,
    shape=tuple(shape),
    dtype=trt_input_dtype,
)
```

with a pybind overload `TypeError` (the 10-element tuple cannot be converted to
`tensorrt.tensorrt.Dims`). Because the failing node is a `placeholder`, the error carries no
useful provenance - the report ends with `Original traceback: None`, and nothing points at the
op that actually produced the over-rank tensor.

## To Reproduce

```
```

[repro.py](https://github.com/user-attachments/files/31529706/repro.py)

**repro.py**

```
import traceback

import torch

import torch_tensorrt  # noqa: F401  # registers the "tensorrt" torch.compile backend

COMPILE_OPTIONS = {"pass_through_build_failures": True, "min_block_size": 1}
RANK_10_SHAPE = (1, 2, 1, 1, 1, 1, 1, 1, 4, 8)
RANK_8_SHAPE = (1, 2, 1, 1, 1, 1, 4, 8)


@torch.library.custom_op("repro::high_rank", mutates_args=())
def high_rank(x: torch.Tensor) -> torch.Tensor:
    """Reshapes to rank 10. Opaque to TensorRT, so the partitioner has to decline it."""
    return x.reshape(RANK_10_SHAPE).clone()


@high_rank.register_fake
def _high_rank_fake(x: torch.Tensor) -> torch.Tensor:
    return x.new_empty(RANK_10_SHAPE)


@torch.library.custom_op("repro::low_rank", mutates_args=())
def low_rank(x: torch.Tensor) -> torch.Tensor:
    """Control counterpart of `high_rank`: same structure, rank 8."""
    return x.reshape(RANK_8_SHAPE).clone()


@low_rank.register_fake
def _low_rank_fake(x: torch.Tensor) -> torch.Tensor:
    return x.new_empty(RANK_8_SHAPE)


class HighRankBoundary(torch.nn.Module):
    """The rank-10 producer falls back to Torch; the consumer is converted."""

        return torch.ops.repro.high_rank(x) * 2.0


class LowRankBoundary(torch.nn.Module):
    """Control: identical structure, rank 8, within trt.Dims.MAX_DIMS."""

        return torch.ops.repro.low_rank(x) * 2.0


def compile_and_run(model: torch.nn.Module, args: tuple[torch.Tensor, ...]) -> BaseException | None:
    """Compiles with the TensorRT backend and runs it. Returns the exception, or None."""
    torch._dynamo.reset()
    optimized = torch.compile(model, backend="tensorrt", options=COMPILE_OPTIONS)
    try:
        optimized(*args)
    except BaseException as exc:  # noqa: BLE001  # a bug demonstration: report, do not handle
        traceback.print_exc()
        return exc
    return None


def main() -> int:
    print(f"torch {torch.__version__}")
    print(f"torch_tensorrt {torch_tensorrt.__version__}")

    x = torch.rand(2, 4, 8, device="cuda")

    print("\n=== control: rank-8 tensor across the partition boundary ===")
    control_exc = compile_and_run(LowRankBoundary().cuda().eval(), (x,))
    control_ok = control_exc is None
    print(f"control compiled: {control_ok}")

    print("\n=== bug case: rank-10 tensor across the partition boundary ===")
    bug_exc = compile_and_run(HighRankBoundary().cuda().eval(), (x,))
    if bug_exc is None:
        print("bug case compiled cleanly -- the bug did not reproduce")
        reproduced = False
    else:
        tb = "".join(traceback.format_exception(type(bug_exc), bug_exc, bug_exc.__traceback__))
        names_add_input = "add_input" in str(bug_exc) or "add_input" in tb
        in_placeholder = "_TRTInterpreter.py" in tb and ", in placeholder" in tb
        print(f"failure names add_input                 : {names_add_input}")
        print(f"traceback names placeholder             : {in_placeholder}")
        print(f"exception: {type(bug_exc).__name__}: {str(bug_exc)[:400]}")
        reproduced = names_add_input and in_placeholder

    print(f"\ncontrol compiles: {control_ok}")
    print(f"reproduced: {reproduced}")
    return 0 if (reproduced and control_ok) else 1


if __name__ == "__main__":
    raise SystemExit(main())
```

**output**

```
torch 2.13.0a0+9186a08b2c.nv26.07
torch_tensorrt 2.14.0a0

=== control: rank-8 tensor across the partition boundary ===
control compiled: True

=== bug case: rank-10 tensor across the partition boundary ===
  File "/w/repro.py", line 88, in compile_and_run
    optimized(*args)
[...]
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py", line 384, in _construct_trt_network_def
    super().run()
  File "/usr/local/lib/python3.12/dist-packages/torch/fx/interpreter.py", line 197, in run
    self.env[node] = self.run_node(node)
                     ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py", line 686, in run_node
    trt_node: torch.fx.Node = super().run_node(n)
                              ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/fx/interpreter.py", line 294, in run_node
    return getattr(self, n.op)(n.target, args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py", line 749, in placeholder
    return self.ctx.net.add_input(
           ^^^^^^^^^^^^^^^^^^^^^^^
TypeError: add_input(): incompatible function arguments. The following argument types are supported:
    1. (self: tensorrt.tensorrt.INetworkDefinition, name: str, dtype: tensorrt.tensorrt.DataType, shape: tensorrt.tensorrt.Dims) -> tensorrt.tensorrt.ITensor

Invoked with: <tensorrt.tensorrt.INetworkDefinition object at 0x727bfad294b0>; kwargs: name='high_rank', shape=(1, 2, 1, 1, 1, 1, 1, 1, 4, 8), dtype=<DataType.FLOAT: 0>

While executing %high_rank : [num_users=1] = placeholder[target=high_rank]
Original traceback:
None

failure names add_input                 : True
traceback names placeholder             : True

control compiles: True
reproduced: True
```

## Expected behavior

The partitioner should decline any node whose output *or any of whose inputs* has rank greater
than `trt.Dims.MAX_DIMS`, so the whole affected region falls back to Torch and compilation of
the rest of the model succeeds. Aborting the entire `torch.compile` because one boundary tensor
has 10 axes is the wrong outcome: rank > 8 is simply not expressible in TensorRT, and that is
exactly the situation partitioning exists to handle.

There is already a precedent for the "consider a node's inputs too" pattern in the same place.
`TorchTensorRTOperatorSupport._has_complex_dtype` (added in #4119) is called from *both*
`is_node_supported` implementations and does exactly this:

```python
if _dtype(node) in COMPLEX_DTYPES:
    return True
for arg in node.all_input_nodes:
    if _dtype(arg) in COMPLEX_DTYPES:
        return True
return False
```

A rank check belongs alongside it - same call site, same shape of predicate, using
`trt.Dims.MAX_DIMS` instead of `COMPLEX_DTYPES`. We have not implemented or tested such a
patch, so treat that as a suggestion rather than a verified fix.

Separately: even if the decline is not added, the diagnostics deserve improvement. A pybind
overload-resolution `TypeError` listing the accepted C++ signatures, attached to a `placeholder`
with `Original traceback: None`, gives a user nothing to act on. An explicit check in
`placeholder` (or better, at partitioning time) with a message naming the tensor, its rank and
the limit would be far more actionable.

## Environment

### LINKED ISSUE #4610: 🐛 [Bug] optimization-profile opt extent is computed as min + max/2, exceeds max value in certain cases
##  Bug Description

`py/torch_tensorrt/dynamo/partitioning/common.py :: construct_dynamic_input` invents
optimization-profile extents for dynamic dimensions, and two of the arithmetic defaults it uses
are wrong.

**1. The invented `opt` extent is `min + max/2`, not `(min + max)/2`, so it can exceed `max`.**

```python
# if opt not exist, set it to the mean of min and max
if "opt" not in min_max_opt or min_max_opt["opt"] is None:
    logger.info(...)
    unwrapped_min_max_opt["opt"] = int(
        unwrapped_min_max_opt["min"] + unwrapped_min_max_opt["max"] / 2
    )
```

The comment says "the mean of min and max"; the expression is `min + max/2`. The difference
between the two is `min/2`, so they disagree for every nonzero `min`, and the computed value
exceeds `max` whenever `min > max/2`. A dimension
bounded to `[1000, 1200]` is given `opt = 1000 + 600 = 1600`, and TensorRT rejects the profile:

```
IOptimizationProfile::isValid: Error Code 4: API Usage Error (Invalid optimization profiles for input tensors dynamic_rows, please ensure that the optimization profiles satisfy MIN<=OPT<=MAX. In isValid at /_src/optimizer/api/optimizationProfile.cpp:357)
```

**2. The absent-max default is `min * 2**12`, which is 0 for an axis whose min is legitimately
0.**

```python
if "max" not in min_max_opt or min_max_opt["max"] is None:
    logger.warning(
        f"Dynamic input {name} (shape: {input_shape}) has no max bound for dim {d}, attempting to use a sane default (max: min({unwrapped_min_max_opt['min']}) * 2^12). ..."
    )
    unwrapped_min_max_opt["max"] = unwrapped_min_max_opt["min"] * (2**12)
```

A data-dependent row count has a perfectly reasonable lower bound of 0, and `0 * 2**12 == 0`,
so `min`, `opt` and `max` all become 0 and the whole dynamic axis is pinned to extent 0. The
reproducer observes `bounds [0, None] -> min=(0, 4) opt=(0, 4) max=(0, 4)`.

**Where the failure surfaces.** The TensorRT profile rejection is only written to the TensorRT
log. The exception that reaches Python is a bare `AssertionError` from `assert cuda_engine` in
`py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py :: run` - no mention of profiles, of
`opt`, or of the input whose extents are inconsistent. The reproducer installs a logging handler
purely to make the real cause visible.

## To Reproduce

```
```

[repro.py](https://github.com/user-attachments/files/31530365/repro.py)

**repro.py**

```
import logging
import sys
import traceback

import torch
import torch_tensorrt
from torch import nn
from torch_tensorrt.dynamo.partitioning.common import construct_dynamic_input


@torch.library.custom_op("repro::dynamic_rows", mutates_args=())
def dynamic_rows(x: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
    return x[mask.nonzero().squeeze(1)].clone()


@dynamic_rows.register_fake
def _(x: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
    return x.new_empty(torch.library.get_ctx().new_dynamic_size(), x.shape[1])


class Model(nn.Module):
    """Produces an unbacked row count bounded to [lower, upper]."""

    def __init__(self, lower: int, upper: int | None) -> None:
        super().__init__()
        self.lower = lower
        self.upper = upper

    def forward(self, x: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
        rows = torch.ops.repro.dynamic_rows(x, mask)
        torch._check(rows.shape[0] >= self.lower)
        if self.upper is not None:
            torch._check(rows.shape[0] <= self.upper)
        return rows * 2.0 + 1.0


class LogCapture(logging.Handler):
    """Collects every log record so the TensorRT profile rejection stays visible."""

    def __init__(self) -> None:
        super().__init__(level=logging.DEBUG)
        self.messages: list[str] = []

    def emit(self, record: logging.LogRecord) -> None:
        self.messages.append(record.getMessage())


def export_model(lower: int, upper: int | None) -> torch.export.ExportedProgram:
    x = torch.randn(2048, 4, device="cuda")
    mask = torch.zeros(2048, dtype=torch.bool, device="cuda")
    mask[:1100] = True
    return torch.export.export(Model(lower, upper).eval(), (x, mask))


def symbolic_shape(exported: torch.export.ExportedProgram) -> torch.Size:
    for node in exported.graph.nodes:
        value = node.meta.get("val", None)
        if isinstance(value, torch.Tensor) and any(
            isinstance(dim, torch.SymInt) for dim in value.shape
        ):
            return value.size()
    raise RuntimeError("no symbolic shape found in the exported program")


def profile_for(lower: int, upper: int | None) -> torch_tensorrt.Input:
    """Runs the production profile construction on a real symbolic shape."""
    shape = symbolic_shape(export_model(lower, upper))
    trt_input = construct_dynamic_input(shape, torch.float32, name="dynamic_rows")
    print(
        f"bounds [{lower}, {upper}] -> min={trt_input.shape['min_shape']} "
        f"opt={trt_input.shape['opt_shape']} max={trt_input.shape['max_shape']}"
    )
    return trt_input


def compile_model(tag: str, lower: int, upper: int | None) -> tuple[str, list[str]]:
    capture = LogCapture()
    logging.getLogger().addHandler(capture)
    logging.getLogger().setLevel(logging.DEBUG)
    exported = export_model(lower, upper)
    x, mask = exported.example_inputs[0]
    formatted = ""
    try:
        torch_tensorrt.dynamo.compile(
            exported,
            inputs=(x, mask),
            min_block_size=1,
        )
        print(f"--- {tag}: compile succeeded ---")
    except BaseException:
        formatted = traceback.format_exc()
        print(f"--- {tag}: compile raised ---\n{formatted}")
    finally:
        logging.getLogger().removeHandler(capture)
    return formatted, capture.messages


def main() -> int:
    print("torch", torch.__version__, "torch_tensorrt", torch_tensorrt.__version__)

    bug_profile = profile_for(1000, 1200)
    control_profile = profile_for(100, 1200)
    no_max_profile = profile_for(0, None)

    opt_above_max = bug_profile.shape["opt_shape"][0] > bug_profile.shape["max_shape"][0]
    control_ok = (
        control_profile.shape["min_shape"][0]
        <= control_profile.shape["opt_shape"][0]
        <= control_profile.shape["max_shape"][0]
    )
    max_pinned_to_zero = no_max_profile.shape["max_shape"][0] == 0

    bug_traceback, bug_logs = compile_model("BUG (min > max/2)", 1000, 1200)
    control_traceback, _ = compile_model("CONTROL (min < max/2)", 100, 1200)

    profile_errors = [message for message in bug_logs if "MIN<=OPT<=MAX" in message]
    for message in profile_errors:
        print(f"TensorRT: {message}")

    trt_rejected_profile = bool(profile_errors)
    print(f"invented opt exceeds max: {opt_above_max}")
    print(f"control profile is well ordered and compiles: {control_ok and not control_traceback}")
    print(f"absent-max default pins the axis to 0: {max_pinned_to_zero}")
    print(f"TensorRT rejected the profile: {trt_rejected_profile}")

    reproduced = opt_above_max and trt_rejected_profile and bool(bug_traceback)
    print(f"reproduced: {reproduced}")
    return 0 if reproduced else 1


if __name__ == "__main__":
    sys.exit(main())

```

**output**

```
Dynamic input dynamic_rows (shape: torch.Size([u0, 4])) has no max bound for dim 0, attempting to use a sane default (max: min(0) * 2^12). Please set an upper bound using torch._dynamo.mark_dynamic or torch.export.Dim
min_shape contains a 0 dimension: (0, 4). TensorRT does not support dynamic shapes with min dimension of 0 (empty tensors). TensorRT will internally clamp min dimensions to 1, which may cause runtime errors if you try to run inference with empty tensor inputs.
torch 2.13.0a0+9186a08b2c.nv26.07 torch_tensorrt 2.14.0a0
bounds [1000, 1200] -> min=(1000, 4) opt=(1600, 4) max=(1200, 4)
bounds [100, 1200] -> min=(100, 4) opt=(700, 4) max=(1200, 4)
bounds [0, None] -> min=(0, 4) opt=(0, 4) max=(0, 4)
--- BUG (min > max/2): compile raised ---
  File "/w/repro.py", line 129, in compile_model
    torch_tensorrt.dynamo.compile(
[...]
                         ^^^^^^^^^^^^^^^^^
    assert cuda_engine
           ^^^^^^^^^^^
AssertionError

--- CONTROL (min < max/2): compile succeeded ---
TensorRT: IOptimizationProfile::isValid: Error Code 4: API Usage Error (Invalid optimization profiles for input tensors dynamic_rows, please ensure that the optimization profiles satisfy MIN<=OPT<=MAX. In isValid at /_src/optimizer/api/optimizationProfile.cpp:357)
invented opt exceeds max: True
control profile is well ordered and compiles: True
absent-max default pins the axis to 0: True
TensorRT rejected the profile: True
reproduced: True
```

## Expected behavior

The invented `opt` extent should be the mean the comment describes, `(min + max) // 2`, and in
any case it must satisfy `min <= opt <= max`. With bounds `[1000, 1200]` that gives `opt = 1100`
and the profile is accepted.

The absent-max default should produce a `max` strictly greater than `min` even when `min` is 0 -
`max(min, 1) * 2**12` would do it - so that a legitimately-zero lower bound does not collapse a
dynamic axis to extent 0.

More generally, `construct_dynamic_input` should validate the triple it just built and refuse to
hand an inconsistent profile to TensorRT. Raising `ValueError("computed opt extent 1600 for
dimension 0 of 'dynamic_rows' exceeds max 1200")` from the function that computed it would point
straight at the defect; today the user gets an empty `AssertionError` from a different file.

Separately, and independently useful: please surface the TensorRT builder error text in the
exception raised from `_TRTInterpreter.run` instead of `assert cuda_engine`. In this case
TensorRT says exactly what is wrong ("please ensure that the optimization profiles satisfy
MIN<=OPT<=MAX") and that message never reaches the caller.

## Environment

 - Pytorch NGC container 26.07-py3

### LINKED ISSUE #4611: 🐛 [Bug] extract_var_range_info calls int() on unbounded range bounds (no lower guard, upper guards only int_oo)
##  Bug Description

`py/torch_tensorrt/dynamo/utils.py :: extract_var_range_info` converts a symbol's `ValueRange`
into profile bounds with a single expression that guards one sentinel on one side and nothing on
the other:

```python
min_val, max_val = (
    int(var_range.lower),                                         # no guard at all
    int(var_range.upper) if var_range.upper != int_oo else None,   # guards int_oo only
)
```

**Case A - the lower bound has no guard.** An unbacked scalar `SymInt` produced by
`Tensor.item()` is not size-like, so its `ValueRange` is `[-int_oo, int_oo]`. When that `SymInt`
becomes an input of a TensorRT subgraph, `partitioning/common.py` asks
`extract_var_range_info` for its bounds and `int(-int_oo)` blows up inside sympy. Two exceptions
are involved: an `OverflowError: cannot convert float infinity to integer` raised in sympy's
`_mag`, and, while that is being handled, an `AttributeError: 'Infinity' object has no attribute
'_mpf_'` which is what finally propagates. Compilation aborts.

**Case B - the upper bound guards only `int_oo`.** A range whose upper bound is a large finite
integer is taken as a real bound, no matter how implausible. With an upper bound of
`sys.maxsize - 1`, `construct_dynamic_input` computes an `opt` extent of `2**62` and then
`torch_tensorrt.Input.__init__` tries to allocate an example tensor at that extent:

```
RuntimeError: Storage size calculation overflowed with sizes=[4611686018427387904, 4]
```

Note where case B actually lands: the proximate crash is in
`py/torch_tensorrt/_Input.py :: Input.example_tensor` via `torch.rand(...)`, not in
`extract_var_range_info`. And the `2**62` comes from `int(min + max / 2)` in
`partitioning/common.py :: construct_dynamic_input` - a separate defect in the profile-extent
arithmetic, mentioned here only because it is what turns an unrecognized sentinel into an
allocation. The part that belongs to this issue is that `sys.maxsize - 1` was accepted as a
meaningful upper bound in the first place.

## To Reproduce

```
```

[repro.py](https://github.com/user-attachments/files/31530550/repro.py)

**repro.py**

```
import sys
import traceback

import torch
import torch_tensorrt
from torch import nn

UNBOUNDED_SENTINEL = sys.maxsize - 1


class ScalarInputModel(nn.Module):
    """Case A: an unbacked scalar SymInt reaches a TensorRT subgraph as an input."""

    def forward(self, x: torch.Tensor, k: torch.Tensor) -> torch.Tensor:
        n = k.item()
        return x * n + x


@torch.library.custom_op("repro::dynamic_rows", mutates_args=())
def dynamic_rows(x: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
    return x[mask.nonzero().squeeze(1)].clone()


@dynamic_rows.register_fake
def _(x: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
    return x.new_empty(torch.library.get_ctx().new_dynamic_size(), x.shape[1])


class SentinelBoundModel(nn.Module):
    """Case B: an unbacked dimension whose upper bound is sys.maxsize - 1."""

    def forward(self, x: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
        rows = torch.ops.repro.dynamic_rows(x, mask)
        torch._check(rows.shape[0] <= UNBOUNDED_SENTINEL)
        return rows * 2.0 + 1.0


def compile_and_report(tag: str, model: nn.Module, args: tuple[torch.Tensor, ...]) -> str:
    """Compiles through torch.export + dynamo.compile and returns the formatted traceback."""
    exported = torch.export.export(model, args)
    try:
        torch_tensorrt.dynamo.compile(
            exported,
            inputs=args,
            min_block_size=1,
        )
    except BaseException:
        formatted = traceback.format_exc()
        print(f"--- {tag}: compile raised ---\n{formatted}")
        return formatted
    print(f"--- {tag}: compile succeeded (no failure) ---")
    return ""


def main() -> int:
    print("torch", torch.__version__, "torch_tensorrt", torch_tensorrt.__version__)

    x = torch.randn(8, 8, device="cuda")
    k = torch.tensor(3, device="cuda")
    case_a = compile_and_report("case A (unguarded lower bound)", ScalarInputModel().eval(), (x, k))

    rows_in = torch.randn(16, 4, device="cuda")
    mask = torch.zeros(16, dtype=torch.bool, device="cuda")
    mask[:5] = True
    case_b = compile_and_report(
        "case B (sys.maxsize - 1 sentinel)", SentinelBoundModel().eval(), (rows_in, mask)
    )

    a_in_extract = "extract_var_range_info" in case_a and "int(var_range.lower)" in case_a
    b_overflowed = "4611686018427387904" in case_b

    print(f"case A raised inside extract_var_range_info: {a_in_extract}")
    print(f"case B opt extent overflowed at 2**62: {b_overflowed}")
    print(f"reproduced: {a_in_extract}")
    return 0 if a_in_extract else 1


if __name__ == "__main__":
    sys.exit(main())

```

**output**

```
--- case A (unguarded lower bound): compile raised ---
  File "/usr/local/lib/python3.12/dist-packages/sympy/core/expr.py", line 4066, in _mag
    mag_first_dig = int(ceil(log10(xpos)))
                        ^^^^^^^^^^^^^^^^^
OverflowError: cannot convert float infinity to integer

During handling of the above exception, another exception occurred:

  File "/w/repro.py", line 79, in compile_and_report
    torch_tensorrt.dynamo.compile(
[...]
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/partitioning/common.py", line 104, in get_input
    return construct_dynamic_input(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/partitioning/common.py", line 36, in construct_dynamic_input
    min_max_opt = extract_var_range_info(dim)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/utils.py", line 439, in extract_var_range_info
    int(var_range.lower),
    ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/sympy/core/expr.py", line 343, in __int__
    r = self.round(2)
        ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/sympy/core/expr.py", line 3887, in round
    digits_to_decimal = _mag(x)  # _mag(12) = 2, _mag(.012) = -1
                        ^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/sympy/core/expr.py", line 4068, in _mag
    mag_first_dig = int(ceil(Float(mpf_log(xpos._mpf_, 53))/log(10)))
                                           ^^^^^^^^^^
AttributeError: 'Infinity' object has no attribute '_mpf_'

--- case B (sys.maxsize - 1 sentinel): compile raised ---
  File "/w/repro.py", line 79, in compile_and_report
    torch_tensorrt.dynamo.compile(
[...]
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/partitioning/common.py", line 104, in get_input
    return construct_dynamic_input(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/dynamo/partitioning/common.py", line 73, in construct_dynamic_input
    return Input(
           ^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/_Input.py", line 213, in __init__
    self.torch_tensor = self.example_tensor("opt_shape")
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_tensorrt/_Input.py", line 463, in example_tensor
    return torch.rand(self.shape[optimization_profile_field]).to(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Storage size calculation overflowed with sizes=[4611686018427387904, 4]

case A raised inside extract_var_range_info: True
case B opt extent overflowed at 2**62: True
reproduced: True
```

## Expected behavior

`extract_var_range_info` should never raise on a range bound it does not recognize. Both ends
need the same treatment: an unrepresentable or effectively-infinite bound should be reported as
absent (`None`) so the documented defaults in `construct_dynamic_input` take over, exactly as an
`int_oo` upper bound already does. That means guarding the lower bound at all, and recognizing
more than the single `int_oo` object on the upper - `sympy.oo` and finite values of
`sys.maxsize` magnitude are not usable profile extents.

For case A specifically, an unbacked scalar with range `[-int_oo, int_oo]` carries no
information a TensorRT optimization profile can use. Falling back to the documented default
(`min: 1`, `max: min * 2**12`) is acceptable; DECLINING the node so the partitioner runs that
region in Torch would be better than aborting the entire compilation, which is what happens
today.

For case B, whatever bound is derived must be small enough to allocate. `Input.__init__` calls
`example_tensor("opt_shape")` eagerly, so any absurd extent becomes an immediate allocation
failure rather than a diagnosable profile problem.

## Environment

 - Pytorch ngc container : 26.07-py3

narendasan:
## Summary

- clamp finite slice stops against a dynamic runtime axis extent
- keep rank-8-and-higher tensors out of TensorRT partitions at both producer and boundary checks
- compute optimization-profile opt extents as the true midpoint between min and max
- handle unbounded symbolic range limits without calling int() on infinity-like sentinels

## Issues

Closes #4607
Closes #4608
Closes #4610
Closes #4611

## Tests

- pytest tests/py/dynamo/conversion/test_slice_aten.py tests/py/dynamo/models/test_dynamic_shape_user_bounds.py tests/py/dynamo/partitioning/test_rank_limit.py -n0 (56 passed)
- original reproducers for #4607, #4608, #4610, and #4611 no longer reproduce

## Stack

6 of 7. Base: #4631. Follow-up: narendasan/fix-symbolic-converter-parameters.

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
+++ b/py/torch_tensorrt/dynamo/conversion/_ConverterRegistry.py
@@ -138,40 +138,37 @@ def _has_dynamic_shapes(
         node, torch.fx.Node
     ), "Inputs to validator functions must be FX Nodes"
 
-    def _is_subnode_dynamic(subnode: torch.fx.Node) -> bool:
-        """Checks if a node itself has Dynamic properties"""
-        _has_symbolic_sizes_strides, is_shape_dynamic = False, False
-        if "val" in subnode.meta:
-            _has_symbolic_sizes_strides = getattr(
-                subnode.meta["val"], "_has_symbolic_sizes_strides", False
-            )
-            meta_val = subnode.meta["val"]
-            if isinstance(meta_val, (list, tuple)):
-                for val in meta_val:
-                    if val is None:
-                        continue
-                    if isinstance(val, (SymFloat, SymInt, SymBool)):
-                        is_shape_dynamic = True
-                        break
-                    if not hasattr(val, "size"):
-                        continue
-                    shape = val.size()
-                    if any(
-                        isinstance(dim, (SymFloat, SymInt, SymBool)) for dim in shape
-                    ):
-                        is_shape_dynamic = True
-                        break
-            elif isinstance(meta_val, (SymFloat, SymInt, SymBool)):
-                is_shape_dynamic = True
-            elif isinstance(meta_val, (int, float, bool)):
-                is_shape_dynamic = False
-            else:
-                shape = subnode.meta["val"].size()
-                is_shape_dynamic = any(
-                    isinstance(dim, (SymFloat, SymInt, SymBool)) for dim in shape
-                )
+    def _contains_dynamic_value(value: Any) -> bool:
+        """Recursively checks FX metadata for symbolic values or dimensions."""
+        if isinstance(value, (SymFloat, SymInt, SymBool)):
+            return True
+        if isinstance(value, dict):
+            return any(_contains_dynamic_value(item) for item in value.values())
+        if isinstance(value, (list, tuple)):
+            return any(_contains_dynamic_value(item) for item in value)
+        if value is None or isinstance(value, (int, float, bool)):
+            return False
+        if hasattr(value, "size"):
+            return any(_contains_dynamic_value(dim) for dim in value.size())
+        return False
 
-        return _has_symbolic_sizes_strides or is_shape_dynamic
+    def _is_subnode_dynamic(subnode: torch.fx.Node) -> bool:
+        """Checks if a node itself has Dynamic properties."""
+        meta_val = subnode.meta.get("val", subnode.meta.get("example_value"))
+        return bool(
+            getattr(meta_val, "_has_symbolic_sizes_strides", False)
+            or _contains_dynamic_value(meta_val)
+        )
+
+    def _is_argument_dynamic(argument: Argument) -> bool:
+        """Checks nodes nested inside FX container arguments."""
+        if isinstance(argument, torch.fx.Node):
+            return _is_subnode_dynamic(argument)
+        if isinstance(argument, dict):
+            return any(_is_argument_dynamic(item) for item in argument.values())
+        if isinstance(argument, (list, tuple)):
+            return any(_is_argument_dynamic(item) for item in argument)
+        return False
 
     # Check node value itself
     if arg_positions_to_check is None and _is_subnode_dynamic(node):
@@ -179,22 +176,18 @@ def _has_dynamic_shapes(
 
     # Check node arguments individually
     if arg_positions_to_check is None and any(
-        _is_subnode_dynamic(arg) for arg in node.args if isinstance(arg, torch.fx.Node)
+        _is_argument_dynamic(arg) for arg in node.args
     ):
         return True
     # Check specific arg positions if the caller has specified positions to check
     elif arg_positions_to_check is not None and any(
-        _is_subnode_dynamic(node.args[i])
-        for i in arg_positions_to_check
-        if isinstance(node.args[i], torch.fx.Node)
+        _is_argument_dynamic(node.args[i]) for i in arg_positions_to_check
     ):
         return True
 
     # Check node keyword arguments individually
     if arg_positions_to_check is None and any(
-        _is_subnode_dynamic(kwarg)
-        for kwarg in node.kwargs.values()
-        if isinstance(kwarg, torch.fx.Node)
+        _is_argument_dynamic(kwarg) for kwarg in node.kwargs.values()
     ):
         return True
 
--- a/py/torch_tensorrt/dynamo/conversion/impl/arange.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/arange.py
@@ -38,7 +38,7 @@ def _sequence_dtype(
         if isinstance(x, float):
             return trt.DataType.FLOAT
 
-        return trt.DataType.INT64
+    return trt.DataType.INT64
 
 
 def arange(
@@ -65,6 +65,9 @@ def arange(
         start_rank_0 = get_trt_tensor(
             ctx, start, name + "_start_rank_0", value_dtype, min_rank=0
         )
+        start_rank_0 = cast_trt_tensor(
+            ctx, start_rank_0, value_dtype, name + "_start_rank_0_casted"
+        )
         # LINSPACE's start input requires rank 0; if the upstream ITensor came in
         # as rank-1 (e.g. a SymInt materialized by a sym_size op), reshape it.
         if len(start_rank_0.shape) > 0:
@@ -80,6 +83,11 @@ def arange(
         )
         end = get_trt_tensor(ctx, end, name + "_end", value_dtype, min_rank=1)
         step = get_trt_tensor(ctx, step, name + "_step", value_dtype, min_rank=1)
+        start_rank_1 = cast_trt_tensor(
+            ctx, start_rank_1, value_dtype, name + "_start_rank_1_casted"
+        )
+        end = cast_trt_tensor(ctx, end, value_dtype, name + "_end_casted")
+        step = cast_trt_tensor(ctx, step, value_dtype, name + "_step_casted")
 
         # The number of elements is ceil((end - start) / step), computed as
         # -floor((start - end) / step) so that the whole expression stays in the
--- a/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
@@ -109,7 +109,7 @@ def slice_op(  # TODO: This should be slice not whatever is in base
     # Assign the initial stop tensor
     stop_slice = []
     if isinstance(stop, int) and dynamic_input_dim_equal:
-        stop_slice = input.shape
+        stop_slice = list(input.shape)
         stop_slice[dim] = stop
     else:
         # required for cases where stop is ITensor and dim != dynamic dim of input
@@ -125,6 +125,19 @@ def slice_op(  # TODO: This should be slice not whatever is in base
                 stop_slice.append(stop)
             else:
                 stop_slice.append(input.shape[i])
+
+    finite_stop_on_dynamic_dim = (
+        input.shape[dim] == DYNAMIC_DIM
+        and isinstance(stop, int)
+        and 0 <= stop < sys.maxsize
+    )
+    if finite_stop_on_dynamic_dim:
+        dim_size = get_shape(
+            ctx, target, source_ir, name + "_stop_dim_size", input, dim
+        )
+        stop_slice[dim] = impl.elementwise.min(
+            ctx, target, source_ir, name + "_stop_clamped", stop, dim_size
+        )
 
     stride_slice = [1] * len(input.shape)
     stride_slice[dim] = step
@@ -143,6 +156,7 @@ def slice_op(  # TODO: This should be slice not whatever is in base
             or stop < 0
             or stop_dynamic_None
             or stop == sys.maxsize
+            or finite_stop_on_dynamic_dim
         ):
             # special assignments for dynamic cases
             if isinstance(start, int) and start < 0:
--- a/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/slice/ops.py
@@ -135,6 +135,12 @@ def slice_op(  # TODO: This should be slice not whatever is in base
         dim_size = get_shape(
             ctx, target, source_ir, name + "_stop_dim_size", input, dim
         )
+        if isinstance(start, int) and start > 0:
+            # A runtime axis can be shorter than a positive literal start.
+            # Normalize it just like Python so the slice extent cannot go negative.
+            start_slice[dim] = impl.elementwise.min(
+                ctx, target, source_ir, name + "_start_clamped", start, dim_size
+            )
         stop_slice[dim] = impl.elementwise.min(
             ctx, target, source_ir, name + "_stop_clamped", stop, dim_size
         )
--- a/py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
+++ b/py/torch_tensorrt/dynamo/partitioning/_adjacency_partitioner.py
@@ -67,6 +67,14 @@ class OpSupportTester(ops.OperatorSupportBase):  # type: ignore
             )
             return False
 
+        if TorchTensorRTOperatorSupport._exceeds_max_tensor_rank(node):
+            # Keep unrepresentable tensors entirely in the Torch partition.
+            if not node.is_impure():
+                self.unsupported_operators[node_name] = (
+                    self.unsupported_operators.get(node_name, 0) + 1
+                )
+            return False
+
         if TorchTensorRTOperatorSupport._has_complex_dtype(node):
             # Complex-dtype tensors are not supported by TensorRT; force PyTorch fallback
             if not node.is_impure():
--- a/py/torch_tensorrt/dynamo/partitioning/_global_partitioner.py
+++ b/py/torch_tensorrt/dynamo/partitioning/_global_partitioner.py
@@ -1,6 +1,7 @@
 import logging
 from typing import Collection, Dict, List, Mapping, Optional, Sequence, Tuple
 
+import tensorrt as trt
 import torch
 from torch.fx.graph_module import GraphModule
 from torch.fx.node import Target
@@ -223,6 +224,27 @@ class TorchTensorRTOperatorSupport(OperatorSupport):  # type: ignore[misc]
         return cache[node]
 
     @staticmethod
+    def _exceeds_max_tensor_rank(node: torch.fx.Node) -> bool:
+        """Return whether a node would move a rank unsupported by TensorRT."""
+
+        def _value_exceeds_limit(value: object) -> bool:
+            if isinstance(value, torch.Tensor):
+                return value.ndim > trt.Dims.MAX_DIMS
+            if isinstance(value, (tuple, list)):
+                return any(_value_exceeds_limit(item) for item in value)
+            if isinstance(value, dict):
+                return any(_value_exceeds_limit(item) for item in value.values())
+            return False
+
+        def _node_exceeds_limit(candidate: torch.fx.Node) -> bool:
+            value = candidate.meta.get("val", candidate.meta.get("example_value"))
+            return _value_exceeds_limit(value)
+
+        return _node_exceeds_limit(node) or any(
+            _node_exceeds_limit(input_node) for input_node in node.all_input_nodes
+        )
+
+    @staticmethod
     def _requires_output_allocator(node: torch.fx.Node) -> bool:
         # True if the converter selected for this node needs a TRT output allocator,
         # i.e. the node has a data-dependent output shape (e.g. nonzero or boolean
@@ -252,6 +274,14 @@ class TorchTensorRTOperatorSupport(OperatorSupport):  # type: ignore[misc]
                 "non-target device region",
                 node_name,
             )
+            return False
+
+        if self._exceeds_max_tensor_rank(node):
+            # TensorRT network inputs and outputs are limited by trt.Dims.
+            if not node.is_impure():
+                self.unsupported_operators[node_name] = (
+                    self.unsupported_operators.get(node_name, 0) + 1
+                )
             return False
 
         if self._has_complex_dtype(node):
--- a/py/torch_tensorrt/dynamo/partitioning/common.py
+++ b/py/torch_tensorrt/dynamo/partitioning/common.py
@@ -132,10 +132,14 @@ def construct_dynamic_input(
             # if opt not exist, set it to the mean of min and max
             if "opt" not in min_max_opt or min_max_opt["opt"] is None:
                 logger.info(
-                    f"Dynamic input {name} (shape: {input_shape}) has no opt target i.e. which shape to specialize for, for dim {d}, attempting to use a sane default (opt: min({min_max_opt['min']}) + max({min_max_opt['max']}) / 2). If you want to specialized further, use torch_tensorrt.compile"
+                    f"Dynamic input {name} (shape: {input_shape}) has no opt target "
+                    f"for dim {d}; using the midpoint of min "
+                    f"({unwrapped_min_max_opt['min']}) and max "
+                    f"({unwrapped_min_max_opt['max']})."
                 )
-                unwrapped_min_max_opt["opt"] = int(
-                    unwrapped_min_max_opt["min"] + unwrapped_min_max_opt["max"] / 2
+                unwrapped_min_max_opt["opt"] = (
+                    unwrapped_min_max_opt["min"]
+                    + (unwrapped_min_max_opt["max"] - unwrapped_min_max_opt["min"]) // 2
                 )
             else:
                 unwrapped_min_max_opt["opt"] = min_max_opt["opt"]
--- a/py/torch_tensorrt/dynamo/utils.py
+++ b/py/torch_tensorrt/dynamo/utils.py
@@ -6,6 +6,7 @@ import importlib.util
 import logging
 import os
 import platform
+import sys
 import warnings
 from dataclasses import fields, replace
 from enum import Enum
@@ -458,9 +459,14 @@ def extract_var_range_info(
         if value == sympy.oo or value == -sympy.oo:
             return None
         try:
-            return int(value)
+            bound = int(value)
         except (TypeError, OverflowError, AttributeError):
             return None
+        # PyTorch uses the largest representable size as a finite-looking
+        # sentinel for an effectively unbounded symbolic dimension.
+        if abs(bound) >= sys.maxsize - 1:
+            return None
+        return bound
 
     min_val_opt = _bound_to_int_or_none(var_range.lower)
     max_val = _bound_to_int_or_none(var_range.upper)
--- a/tests/py/dynamo/conversion/test_arange_aten.py
+++ b/tests/py/dynamo/conversion/test_arange_aten.py
@@ -1,5 +1,3 @@
-import unittest
-
 import torch
 import torch.nn as nn
 import torch_tensorrt
@@ -105,6 +103,39 @@ class TestArangeConverter(DispatchTestCase):
             pyt_inputs=[pyt_input],
             use_dynamo_tracer=False,
         )
+
+    def test_arange_data_dependent_start(self):
+        class Arange(torch.nn.Module):
+            def forward(self, x, mask):
+                end = mask.nonzero().size(0)
+                start = end // 2
+                indices = torch.arange(start, end, 1, device=x.device)
+                return x.index_select(0, indices)
+
+        previous_capture_setting = torch._dynamo.config.capture_dynamic_output_shape_ops
+        try:
+            torch._dynamo.config.capture_dynamic_output_shape_ops = True
+            torch._dynamo.reset()
+
+            model = Arange().eval().cuda()
+            x = torch.randn((16, 8), device="cuda")
+            mask = torch.arange(16, device="cuda") % 2 == 0
+            expected = model(x, mask)
+
+            compiled = torch.compile(
+                model,
+                backend="tensorrt",
+                options={
+                    "pass_through_build_failures": True,
+                    "min_block_size": 1,
+                },
+            )
+            torch.testing.assert_close(compiled(x, mask), expected)
+        finally:
+            torch._dynamo.config.capture_dynamic_output_shape_ops = (
+                previous_capture_setting
+            )
+            torch._dynamo.reset()
 
 
 if __name__ == "__main__":
--- a/tests/py/dynamo/conversion/test_slice_aten.py
+++ b/tests/py/dynamo/conversion/test_slice_aten.py
@@ -73,6 +73,16 @@ class TestSliceConverterDynamicShape(DispatchTestCase):
                 2,
             ),
             (
+                "slice_dynamic_dim_finite_stop_past_opt",
+                (2, 16),
+                (2, 64),
+                (2, 128),
+                1,
+                0,
+                100,
+                1,
+            ),
+            (
                 "slice_dynamic_dim_start_stop_step_negatives",
                 (1, 10, 10),
                 (10, 10, 10),
--- a/tests/py/dynamo/conversion/test_slice_aten.py
+++ b/tests/py/dynamo/conversion/test_slice_aten.py
@@ -257,6 +257,27 @@ class TestSliceConverterDynamicShape(DispatchTestCase):
             use_example_tensors=False,
         )
 
+    def test_slice_finite_start_past_runtime_extent(self):
+        class TestModule(torch.nn.Module):
+            def forward(self, input):
+                return torch.ops.aten.slice.Tensor(input, 1, 32, 48, 1)
+
+        runtime_input = torch.randn(2, 16)
+        input_specs = [
+            Input(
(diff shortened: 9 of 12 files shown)


sylvesterkaczmarek (commented):
`finite_stop_on_dynamic_dim` still breaks slices whose runtime dimension is smaller than a positive `start`. With runtime size 16 and `[32:48]`, stop is clamped to 16 but the extent calculation becomes `-(32 - 16) = -16`, while PyTorch returns an empty slice. Clamp the computed extent to zero (or clamp start too) and add a dynamic `start > runtime_dim` regression.

Issue #4496, opened 2026-08-16
[CI] torch-compile-models-standard: test_bert_base_uncased failed


lanluo-nvidia:
##  Bug Description

https://github.com/pytorch/TensorRT/actions/runs/31944904591/job/95213434731
```
FAILED models/test_models.py::test_bert_base_uncased[dtype1] - torch._dynamo.exc.BackendCompilerFailed: backend='torch_tensorrt_backend' raised:
  ValueError: __len__() should return >= 0
  
  While executing %permute_6 : [num_users=1] = call_function[target=torch.ops.aten.permute.default](args = (%getitem_3, [0, 2, 1, 3]), kwargs = {})
  Original traceback:
  None
  Use tlparse to see full graph. (https://github.com/pytorch/tlparse?tab=readme-ov-file#tlparse-parse-structured-pt2-logs)
  
  Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"

```
## To Reproduce

1.
2.
3.

## Environment

> Build information about Torch-TensorRT can be found by turning on debug messages

 - Torch-TensorRT Version (e.g. 1.0.0):
 - PyTorch Version (e.g. 1.0):
 - CPU Architecture:
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source):
 - Build command you used (if compiling from source):
 - Are you using local sources or building from archives:
 - Python version:
 - CUDA version:
 - GPU models and configuration:
 - Any other relevant information:

PR #4636, opened 2026-08-28
Cast/normalize attn_bias in scaled_dot_product_efficient_attention

apbose:
# Description

attn_bias was assigned straight to attention_layer.mask with no dtype cast or rank normalization, unlike scaled_dot_product_attention() which already does this for attn_mask. TRT's AttentionInput layer rejects non-Float/Half/BFloat16/Bool masks (e.g. int32 padding masks), leaving a broken output tensor that crashes whatever converter reads it next.

Fixes the root cause of #4496.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/attention.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/attention.py
@@ -430,6 +430,17 @@ def scaled_dot_product_efficient_attention(
             # TRT's IAttention layer does not support passing in both attn_bias/mask and causal mask at the same time,
             # so we convert causal mask to an additive causal mask and add it to the attn_bias
             attn_bias = get_trt_tensor(ctx, attn_bias, f"{name}_attn_bias")
+            # attn_bias is combined with an additive (query.dtype) causal mask below via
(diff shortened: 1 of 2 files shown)


py/torch_tensorrt/dynamo/conversion/impl/attention.py
```diff
@@ -430,6 +430,17 @@ def scaled_dot_product_efficient_attention(
...
+            # elementwise add, so it must already be a matching-dtype additive bias.
+            if attn_bias.dtype != query.dtype:
+                attn_bias = cast_trt_tensor(
+                    ctx,
+                    attn_bias,
+                    query.dtype,
+                    f"{name}_cast_attn_bias",
+                    target,
+                    source_ir,
+                )
```
zewenli98, line 443: I think `impl.elementwise.add` below promotes the dtype inside that function. Can you verify?

apbose: yup removed the redundant one

zewenli98 (approved):
LGTM

PR #4639, opened 2026-08-28
feature: Dynamo implementation of torch_executed_modules

dongwoonhyun:
# Description

Implementation of `torch_executed_modules` for the dynamo path. The usage is consistent with the TorchScript version (matching by fully-qualified class name). The actual implementation walks through `node.meta.get("nn_module_stack")` and excludes the node if its or any of its parents' modules are in `torch_executed_modules`.

Also added tests to the fast and global partitioners that exercise the feature.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -92,7 +92,7 @@ def cross_compile_for_windows(
     require_full_compilation: bool = _defaults.REQUIRE_FULL_COMPILATION,
     min_block_size: int = _defaults.MIN_BLOCK_SIZE,
     torch_executed_ops: Optional[Collection[Target]] = None,
-    torch_executed_modules: Optional[List[str]] = None,
+    torch_executed_modules: Optional[Collection[str]] = None,
     pass_through_build_failures: bool = _defaults.PASS_THROUGH_BUILD_FAILURES,
     max_aux_streams: Optional[int] = _defaults.MAX_AUX_STREAMS,
     version_compatible: bool = _defaults.VERSION_COMPATIBLE,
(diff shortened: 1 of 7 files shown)


tests/py/dynamo/partitioning/test_000_global_partitioning.py
```diff
@@ -3,7 +3,6 @@
 import numpy as np
 import pytest
 import torch
-import torch.nn.functional as F
 import torch_tensorrt
```
apbose, line 6: why is this line deleted? I think this should still be there.

py/torch_tensorrt/dynamo/_compiler.py
apbose, line 1276: I think settings.torch_execute_models should also be passed here for the prescan

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1785,7 +1781,7 @@ def convert_exported_program_to_serialized_trt_engine(
     require_full_compilation: bool = _defaults.REQUIRE_FULL_COMPILATION,
     min_block_size: int = _defaults.MIN_BLOCK_SIZE,
     torch_executed_ops: Optional[Collection[Target]] = None,
-    torch_executed_modules: Optional[List[str]] = None,
+    torch_executed_modules: Optional[Collection[str]] = None,
     pass_through_build_failures: bool = _defaults.PASS_THROUGH_BUILD_FAILURES,
```
apbose, line 1785: I dont see this being checked in this function. It lacks the equivalent "disable converters for nodes in this module" behavior. The fast or global partitioner gors through `node_in_torch_execute_module` but this goes to `interpret_module_to_result` which does not check this arg

tests/py/dynamo/partitioning/test_000_global_partitioning.py
dongwoonhyun: Sorry, this was inadvertent. Fixed.

py/torch_tensorrt/dynamo/_compiler.py
dongwoonhyun: Added `torch_executed_modules` to `get_graph_converter_support`

py/torch_tensorrt/dynamo/_compiler.py
dongwoonhyun: Good point. I added a check for this in `TRTInterpreter`, but I'm not sure if that's the best place for it. Is the desired behavior is to raise an unsupported op exception if any op belongs to a torch_executed_module?

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -748,6 +749,17 @@ def call_module(
         assert isinstance(target, str)
         submod = self.fetch_attr(target)
         submod_type = getattr(submod, "_base_class_origin", type(submod))
+        if self._cur_node is not None and (
```
apbose, line 752: can we have tests for this?

py/torch_tensorrt/dynamo/_compiler.py
apbose, line 1310: require_full_compilation=True is currently bypassed when the exclusions leave zero supported operations. compile_module() returns the original torch graph at line 1320 before either
 partitioner reaches its existing full-compilation validation. We should avoid this early return when settings.require_full_compilation is true. Also we can have regression tests for this. Applicable to both torch_executed_ops and torch_executed_modules then

apbose:
Overall, the refit path does not currently forward settings.torch_executed_modules when reconstructing the partition layout.
In _refit.py, it is missing from get_graph_converter_support(), fast_partition() or global_partition()
As a result, a module deliberately kept in Torch during the original compilation may be absorbed into a TensorRT partition during refitting. This can change the partition topology and fail the compatibility checks  or cause the wrong regions to be refitted.

Could we forward settings.torch_executed_modules through all three calls, as the normal compilation path already does, and add a refit regression using a model with an excluded submodule? The test
 should verify that refitting preserves the original Torch/TRT partition structure and produces the expected output.

PR #4641, opened 2026-08-29
Short-circuit index_put with a statically-empty index

apbose:
# Description

An empty index means no positions to scatter into, so index_put is a no-op. Building the scatter path anyway (reshaping an index tensor to a shape containing a 0) trips a TensorRT-internal squeezeDims assertion at build time on DGX Spark (GB10, SM 12.1) -- Error Code 2, "Assertion inputDims != nullptr failed" in validationUtils.cpp. Other platforms don't hit it (verified: passes on H100/A100/GB200-family/ORIN_IGX), but the operation is a no-op regardless of platform, so avoid building it at all.

Related but distinct from #4621 (dynamic-index broadcast bug in the same converter) -- different trigger condition, different TRT error, different fix location.

## Type of change

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/impl/select.py
+++ b/py/torch_tensorrt/dynamo/conversion/impl/select.py
@@ -772,6 +772,13 @@ def index_put_converter(
     else:
         N = 1
 
+    if isinstance(N, int) and N == 0:
+        # Empty index -> no-op. Building the scatter path anyway trips a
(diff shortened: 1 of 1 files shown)


narendasan (approved):
Seems fine @lanluo-nvidia please file a bug with TRT

cehongwang (approved):

PR #4642, opened 2026-08-29
Lift each constant once when exporting a partitioned graph

shoumikhin:
Inlining a partitioned graph leaves one get_attr node per partition that reads a
constant, all carrying the same target. The lifting pass turned each of those into
its own placeholder. fx makes a placeholder's name unique but not its target, and
placeholder codegen emits the target, so the generated forward ended up with the
same argument twice and failed to compile with "SyntaxError: duplicate argument".
The signature also gained one input spec per read, all pointing at the same state
dict key.

Lifting each target once and pointing the later readers at that placeholder fixes
both. One get_attr node with several users was always fine; what breaks is several
get_attr nodes with the same target, which is what the partitions left in PyTorch
produce once the inliner copies them back into one graph.

Test plan:

Added a test that builds that graph directly: three get_attr nodes for one buffer.
It checks the lifted graph has two placeholders rather than four, that their
targets are unique, that the signature carries one buffer spec, that the module
recompiles, and that running it gives the right value.

Without the change all five of those fail, including the recompile with the
duplicate argument error. The value check is there because a version of the fix
that rewires the later readers to the wrong placeholder passes every structural
check while computing the wrong answer, and the buffer holds distinct values so
that a wrong read is visible.

It sits beside the other tests for lift rather than with the exporter's inlining
tests, because it needs no GPU and no TensorRT engine, so it runs in the lane that
every pull request runs. The tests it was written next to are only collected by
suites that run on a schedule, so it would not otherwise have run before merge.

The rest of the exporter inlining file still passes.

### DIFF
--- a/py/torch_tensorrt/dynamo/_exporter.py
+++ b/py/torch_tensorrt/dynamo/_exporter.py
@@ -174,8 +174,25 @@ def lift(
     # At first the user_inputs are only present in the graph_signature.input_specs and hence non_user_input_idx=0
     # The input_specs should be of the form [params, buffers, constant_tensors, custom_obj, user_inputs]
     non_user_input_idx = 0
+    # Inlining the partitions left in PyTorch copies each one's own get_attr back
+    # into the graph, so one constant can arrive as several get_attr nodes sharing a
+    # target. fx uniquifies a placeholder's name but not its target, and placeholder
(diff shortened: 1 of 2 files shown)


narendasan (approved):
LGTM

PR #4643, opened 2026-08-29
fix: TensorRT-RTX Turing (SM 7.5) capability guards and compute-capability targeting

tp5uiuc:
**What** — Adds a `target_compute_capabilities` compilation setting, declares it on the TensorRT-RTX
builder config, and falls back to PyTorch for the ops TensorRT-RTX cannot serve on Turing.

**Why** — TensorRT-RTX runs on SM 7.5 and up, but its support matrix carves Turing out of several
paths: FP32 GEMMs and 3D convolutions are unsupported at compute capability 7.5, and Turing has no
bfloat16 hardware. Torch-TensorRT had no notion of this and handed those ops over anyway:

  - FP32 GEMM, static shapes — `createExecutionContext()` returns null
  - FP32 GEMM, dynamic shapes — the engine builds and runs, returning an **all-zero tensor of the
    correct shape and dtype, with no exception**
  - 3D convolution — null execution context
  - bfloat16 — segmentation fault

The dynamic-shape GEMM is the motivating case, because it fails silently.

Separately, the builder was never told which architecture it was building for. With
`num_compute_capabilities == 0`, Myelin looks for a *precompiled* module instead of JIT-ing one and
the build dies — `Compatible cubin or ptx module for device target '75' not found` — so on Turing
**every refittable build failed**.

**How** — Guards key off the capabilities being *built for*, not the build host, so an
ahead-of-time build for another architecture partitions correctly instead of baking in the build
machine. An undeclared target is resolved rather than skipped, to the current device, matching how
partitioning already resolved it — and by naming `ComputeCapability.CURRENT` rather than looking up
`SM<major><minor>`, because TensorRT-RTX names only a subset of architectures and the implicit path
must not start failing on hosts that build fine today.

The GEMM guard keys on fp32 operands only, so fp16 GEMMs accumulating in fp32 are unaffected. The
convolution guard covers forward 3D only, since transposed 3D works on Turing. bfloat16 is gated at
the partitioner rather than per-converter because the crash is not operator-specific, mirroring the
existing complex-dtype handling. Converter unit tests bypass the partitioner, so a rejected node
raises instead of falling back; those tests skip explicitly.

**Testing** — Confirmation sweep at the series tip, T4 (SM 7.5) and L40S (SM 8.9), driver 595.58.03,
identical stacks: **T4 2668 passed / 17 failed / 226 skipped; L40S 2780 / 13 / 118; 2911 collected on
each.** Declaring the capability alone closes **42 of the 91** original Turing failures. Defaulting
costs nothing off Turing: on an L40S, engines built with the default and with an explicit `[(8, 9)]`
are byte-identical (280148 B and 287100 B for a small conv net; 47148116 B and 24184052 B for
resnet18), with no measurable build-time difference. **On the L40S there is not one status change
across all six test modules** — every guard is inert off Turing.

**Cost / Gotchas**
- The guards switch off **105 tests on Turing**, 97 of them in `conversion/`: the T4 runs 1981 of
  2126 tests where the L40S runs 2079. Green on Turing does not mean equally tested on Turing.
- Also clears two pre-existing lint failures in `aten_ops_converters.py` (an unused
  `type: ignore[assignment]` and a spelling the `typos` hook rejects). Both predate this change, but
  the file is now in the changed set so pre-commit blocks on them.

**Followups** — Three more capability gaps ship as stacked PRs on this branch: 3D convolution behind
the pad-folding pass, the GEMM `cdist` emits internally, and FP32 GEMMs reaching TensorRT-RTX through
`linear` and attention.


tp5uiuc:

py/torch_tensorrt/dynamo/conversion/impl/normalization/ops.py
```diff
@@ -643,6 +643,33 @@ def tri_upper_indices(
...
+    compute_mode: Optional[int],
+    rows: Sequence[Optional[int]],
+) -> bool:
+    """Would :func:`cdist_forward` emit a matrix-multiply layer for these arguments?
+
+    ``rows`` is ``shape[-2]`` of each operand, or ``None`` where it is not statically
+    known; an unknown row count is treated as below the threshold, so callers that
+    cannot see shapes fail open. Shared with the capability validator that rejects this
+    case on Turing and with the converter tests, so none of them can drift from the
+    branch below.
```
tp5uiuc: Remove "Shared with the capability validator that rejects this
    case on Turing and with the converter tests, so none of them can drift from the
    branch below."

py/torch_tensorrt/_utils.py
```diff
@@ -380,3 +380,40 @@ def load_tensorrt_llm_for_nccl() -> bool:
...
+
+# --- TensorRT-RTX architecture targeting -------------------------------------
+#
+# TensorRT-RTX runs on SM 7.5 and up but does not support FP32 GEMMs or 3D convolutions
+# on Turing, which has no bfloat16 hardware either. These helpers let capability
+# validators key off the capabilities the engine is *built for* rather than the build
+# host, so a module compiled on Ampere and shipped to Turing does not retain ops Turing
+# cannot execute.
+
+TURING_COMPUTE_CAPABILITY = (7, 5)
```
tp5uiuc, line 403: Push it to inside the function, no one else consumes it.

py/torch_tensorrt/_utils.py
```diff
@@ -380,3 +380,40 @@ def load_tensorrt_llm_for_nccl() -> bool:
...
+def get_target_compute_capabilities(
+    settings: Optional[Any] = None,
+) -> Tuple[Tuple[int, int], ...]:
+    """Compute capabilities this compilation targets: the declared targets, or the
+    current device when none were declared."""
+    targets = (
+        getattr(settings, "target_compute_capabilities", None) if settings else None
+    )
+    if targets:
+        return tuple((int(major), int(minor)) for major, minor in targets)
```
tp5uiuc, line 416: `tuple(tuple(map(int, major_minor)) for major_minor in targets)`

py/torch_tensorrt/_utils.py
```diff
@@ -380,3 +380,40 @@ def load_tensorrt_llm_for_nccl() -> bool:
...
+
+def get_target_compute_capabilities(
+    settings: Optional[Any] = None,
+) -> Tuple[Tuple[int, int], ...]:
+    """Compute capabilities this compilation targets: the declared targets, or the
+    current device when none were declared."""
+    targets = (
+        getattr(settings, "target_compute_capabilities", None) if settings else None
+    )
+    if targets:
```
tp5uiuc: `if settings and (targets := getattr(settings, "target_compute_capabilities", None)):`

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -385,6 +387,50 @@ def _populate_trt_builder_config(
...
+                    trt.ComputeCapability.CURRENT, 0
+                ):
+                    raise RuntimeError(
+                        "Failed to set the TensorRT-RTX compute capability of the "
+                        "current device"
+                    )
+                _LOGGER.info(
+                    "Targeting the TensorRT-RTX compute capability of the current "
+                    f"device {get_target_compute_capabilities(self.compilation_settings)}"
+                )
```
tp5uiuc: This should be refactored into its own function and unit-tested if possible.

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -385,6 +387,50 @@ def _populate_trt_builder_config(
...
+        # validators, naming ComputeCapability.CURRENT because TensorRT-RTX names only a
+        # subset of architectures.
+        if ENABLED_FEATURES.tensorrt_rtx:
+            declared = self.compilation_settings.target_compute_capabilities
+            if declared:
+                builder_config.num_compute_capabilities = len(declared)
+                for idx, (major, minor) in enumerate(declared):
+                    name = f"SM{major}{minor}"
+                    compute_capability = getattr(trt.ComputeCapability, name, None)
+                    if compute_capability is None:
```
tp5uiuc: transform to use walrus op

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
tp5uiuc: Also the if else path (between declared SM and current) can also be two different hidden functions for ease of readability. so the top level call can be

   set_compute_capabilities(self.build_config, self.compilation_settings.target_compute_capabilities)

and the rest can cascade into own function calls

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -846,12 +847,64 @@ def aten_ops_gelu(
...
+    node: Node, settings: Optional[CompilationSettings] = None
+) -> bool:
+    """Reject FP32 GEMMs on TensorRT-RTX when Turing (SM 7.5) is a build target.
+
+    Static shapes fail loudly -- createExecutionContext() returns nullptr -- but dynamic
+    shapes build and run, returning an all-zero tensor with no exception, which is why
+    this cannot be limited to static shapes. Only operand dtype matters, so fp16 GEMMs
+    accumulating in fp32 (``use_fp32_acc``) keep running on TensorRT. Like every
+    validator here it reads ``meta["val"]`` and fails open when that is absent, which in
+    the converter unit tests depends on whether that test ran the dynamo tracer.
```
tp5uiuc: Unwanted. Just say FP32 GEMMs not supported on TRT-RTX. And you can also add `Only operand dtype matters, so fp16 GEMMs
    accumulating in fp32 (``use_fp32_acc``) keep running on TensorRT.`

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -846,12 +847,64 @@ def aten_ops_gelu(
...
+    """
+    if not trt_rtx_targets_turing(settings):
+        return True
+
+    for arg in node.args[:2]:
```
tp5uiuc: This is horrible to read. Make it declarative and easy to grok. like an example is

inputs = node.args[:2]
def can_run_gemm_on_turing(input) -> bool:
    <better_code>
answer = map(can_run_gemm_on_turing, inputs)
if any(answer):
    <log>
    return False
return True

something liek this.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3219,6 +3272,19 @@ def convolution_capability_validator(
         )
         return False
 
+    # No valid kernel config for 3D ConvFwd on SM 7.5: the engine builds, but
+    # createExecutionContext() then returns nullptr. Transposed 3D is a distinct layer
+    # and is unaffected. aten.convolution input is (N, C, *spatial), so ndim 5 is 3D.
+    if trt_rtx_targets_turing(settings) and not args_bounds_check(node.args, 6):
+        val = node.args[0].meta.get("val") if hasattr(node.args[0], "meta") else None
+        if val is not None and val.ndim == 5:
```
tp5uiuc: Same comment on code readability and walrus operator etc.

py/torch_tensorrt/dynamo/partitioning/_global_partitioner.py
```diff
@@ -166,6 +167,31 @@ def _dtype(n: torch.fx.Node) -> Optional[torch.dtype]:
...
 
+    @staticmethod
+    def _has_bf16_on_turing(
+        node: torch.fx.Node, settings: Optional[object] = None
+    ) -> bool:
+        """Return True if this node touches bfloat16 while targeting Turing (SM 7.5).
+
+        Turing has no bfloat16 hardware -- PyTorch only reports support there because
+        it counts emulation. TensorRT-RTX has no such fallback, and compiling a bf16
+        node for SM 7.5 crashes the process (SIGSEGV) rather than reporting an error,
```
tp5uiuc: leave it at "compiling a bf16 node for SM 7.5 is an error"

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -185,6 +188,7 @@ def cross_compile_for_windows(
         enable_experimental_decompositions (bool): Use the full set of operator decompositions. These decompositions may not be tested but serve to make the graph easier to convert to TensorRT, potentially increasing the amount of graphs run in TensorRT.
         dryrun (bool): Toggle for "Dryrun" mode, running everything except conversion to TRT and logging outputs
         hardware_compatible (bool): Build the TensorRT engines compatible with GPU architectures other than that of the GPU on which the engine was built (currently works for NVIDIA Ampere and newer)
+        target_compute_capabilities (Optional[List[Tuple[int, int]]]): Compute capabilities to build for, e.g. ``[(7, 5)]`` for Turing. Defaults to None, meaning the current device. TensorRT-RTX only. Drives both engine targeting and op partitioning, so ops unsupported on any listed target fall back to PyTorch.
```
tp5uiuc: `cross_compile_for_windows` should not have this flag at all : TRT-RTX does not support it. You can remove.

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -564,6 +572,7 @@ def compile(
...
```
tp5uiuc, line 574: Cab we add a check where we error out if `target_compute_capabilities` is set for standard TensorRT (saying it is not supported)

py/torch_tensorrt/dynamo/_defaults.py
```diff
@@ -63,6 +63,10 @@
 ENABLE_CROSS_COMPILE_FOR_WINDOWS = False
 TILING_OPTIMIZATION_LEVEL = "none"
 L2_LIMIT_FOR_TILING = -1
+# None means "target the current device". Set explicitly to build an engine deployable
+# on other architectures; see torch_tensorrt._utils for why capability validators must
+# consult this rather than the build host's device.
```
tp5uiuc: Remove

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -71,7 +72,10 @@ def _normalize_disabled_constant_fold_exclusions(
         validate_disabled_constant_fold_exclusions,
     )
 
-    return validate_disabled_constant_fold_exclusions(rule_ids)
+    # Bind to a typed local: the imported helper is untyped from mypy's view here,
+    # and returning it directly trips --strict's no-any-return.
```
tp5uiuc: Remove

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -71,7 +72,10 @@ def _normalize_disabled_constant_fold_exclusions(
...
+    normalized: Set[str] = validate_disabled_constant_fold_exclusions(rule_ids)
```
tp5uiuc, line 76: Note : this is a mypy fix.

py/torch_tensorrt/dynamo/_compiler.py
tp5uiuc: This support check can be here or in COmpilationSettings post init, wherever it makes most sense

py/torch_tensorrt/dynamo/_settings.py
```diff
@@ -245,6 +253,10 @@ def __setstate__(self, state: dict[str, Any]) -> None:
     "sparse_weights",
     "engine_capability",
     "hardware_compatible",
+    # An engine built for one set of compute capabilities cannot be reused for a
+    # compile targeting a different set -- doing so would silently reintroduce ops the
```
tp5uiuc: Reemove this comment

tp5uiuc:
### CI summary

| failing check | cause |
|---|---|
| `gate` ×2 | aggregators |

Suites ran to completion: `2 failed, 150 passed, 52 skipped, 4 xpassed` and `1 failed, 52 passed`.

**Why these are not from this PR:** the same `test_cross_runtime_serde` pair fails identically on #4643, #4644, #4645, #4646, #4647 and #4648, whose change sets are disjoint — e.g. #4648 changes `_exporter.py` (the save path) and #4647 changes neither that nor anything else in common, yet both fail the same two tests with byte-identical error text. `executorch-runtime-test` reproduces on `main` nightlies and on unrelated PRs (#4638, #4642). The `gate` entries are aggregators of the above. `test_hf_gqa_model` also fails on #4645 and #4647 and has no overlap with anything this PR changes.

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -568,6 +571,7 @@ def compile(
...
```
narendasan: Might want to list that it defaults to torch.cuda.compute_platform

narendasan (approved):
Mostly looks good, one minor doc change

PR #4644, opened 2026-08-29
fix: guard the 3D convolutions the pad-folding pass hides from the conv guard

tp5uiuc:
**What** — Registers a capability validator on the `tensorrt::conv_asym_pad` custom op so a padded 3D
convolution falls back to PyTorch on Turing.

**Why** — The `fuse_pad_into_convolution` lowering pass rewrites `constant_pad_nd -> convolution`
into `tensorrt::conv_asym_pad`, **erasing the `aten.convolution` node** the 3D-conv guard keys on.
That converter was registered with no `capability_validator`, so a padded conv3d walked straight past
the guard and hit a null execution context (`TRTEngine.cpp:959`). It fires for any non-transposed,
zero-fill, non-negative pad, so symmetric cases fail too. This is a production path, not just tests.

**How** — Share the convolution capability predicate between the aten converter and the fused
custom op, reading rank from the fused node, so the guard survives the rewrite instead of being
erased with the node it keyed on.

**Testing** — Confirmation sweep, both arms: 8 `lowering/` failures closed, 0 Turing-specific
remaining in that module. No status change on the L40S.

**Cost / Gotchas** — Accounts for 8 of the 105 tests the guards switch off on Turing.


tp5uiuc:

py/torch_tensorrt/dynamo/conversion/custom_ops_converters.py
```diff
@@ -27,7 +32,35 @@
...
+    the pass smuggles a 3D convolution past that guard onto Turing (SM 7.5), where it
+    surfaces at runtime as a null execution context. The pass runs in the normal
+    compile path, so this is not confined to tests.
+
+    Spatial rank is read straight off the argument list -- args are
+    (source, weight, bias, stride, pre_padding, post_padding, dilation, groups) -- so
+    the guard holds whether or not the caller traced with dynamo, unlike the
+    meta-based validators. There is no ``transposed`` argument because the pass skips
+    transposed convolutions outright, so the deconvolution half of
+    ``convolution_capability_validator`` has nothing to check here.
```
tp5uiuc: Remove

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3246,6 +3246,34 @@ def aten_ops_le(
...
+    node: Node,
+    spatial_rank: Optional[int],
+    settings: Optional[CompilationSettings] = None,
+) -> bool:
+    """Whether a forward convolution over ``spatial_rank`` spatial dims must fall back.
+
+    No valid kernel config for 3D ConvFwd on SM 7.5: the engine builds, but
+    createExecutionContext() then returns nullptr. Transposed 3D is a distinct layer and
+    is unaffected, so callers must only ask about non-transposed convolutions. Shared by
+    aten.convolution and the fused tensorrt::conv_asym_pad the pad-folding pass emits; a
```
tp5uiuc: No valid kernel config for 3D ConvFwd on SM 7.5 for TRT-RTX, a known gap.

tp5uiuc: Remove the "Shared by...." to the end not needd.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3246,6 +3246,34 @@ def aten_ops_le(
     )
 
 
+_TURING_UNSUPPORTED_CONV_SPATIAL_RANK = 3
```
tp5uiuc: You can hide it inside the function as no one else uses

tp5uiuc:
### CI summary

| failing check | cause |
|---|---|
| `gate` ×2 | aggregators |

Suite ran to completion: `2 failed, 150 passed, 54 skipped, 4 xpassed`.

narendasan (approved):
LGTM

PR #4645, opened 2026-08-29
fix: guard the GEMM the cdist converter emits on Turing

tp5uiuc:
**What** — Adds `cdist_forward_capability_validator`, reproducing the converter's own branch
condition, so an `aten._cdist_forward` that would emit a matrix-multiply layer falls back to PyTorch
on Turing.

**Why** — **This closes zero failing tests, and that is the point.** `impl.normalization.cdist_forward`
computes `p == 2` with a matrix-multiply layer when `compute_mode == 1`, or when it is 0/absent and
either operand has more than 25 rows. The GEMM is emitted **inside** the converter, so the graph holds
a single `_cdist_forward` node and no `mm`/`bmm` for `gemm_capability_validator` to reject.
The covering tests were skipped rather than the converter guarded, so nothing was ever red while the
conversion path stayed open. A skip protects CI and does not protect a caller.

**How** — A validator reproducing the converter's own `p` / `compute_mode` / row-count branch,
so the guard fires exactly when the matmul layer would be emitted. Deliberately **not** keyed on
dtype, unlike `gemm_capability_validator`: measured on a T4, the fused `Matmul_MUL_SUB_SQRT_` pattern
fails for FP16 operands too, under `enabled_precisions` `{f32}`, `{f16}` and `{f16,f32}` alike, even
though a bare FP16 matmul and an FP16 `nn.Linear` both run there. What predicts the failure is whether
the matmul layer is emitted, not its precision.

**Testing** — Confirmation sweep, both arms; `test_cdist_aten.py` and the new guard cases. No status
change on the L40S.

**Cost / Gotchas** — The 25-row threshold is load-bearing, not an optimisation: at or below it no GEMM
is emitted and cdist runs correctly on Turing in both dtypes. Rejecting those anyway would **regress**,
because PyTorch's `cdist_cuda` kernel has no Half implementation, so the fallback would raise where
TensorRT-RTX succeeds today.


tp5uiuc:

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3366,7 +3366,54 @@ def aten_ops_convolution(
...
+# impl.normalization.cdist_forward switches from a broadcast-subtract to a matrix
+# multiply. Kept deliberately in sync with the threshold in that converter.
+def cdist_forward_capability_validator(
+    node: Node, settings: Optional[CompilationSettings] = None
+) -> bool:
+    """Reject the cdist variants whose converter emits a GEMM, on Turing (SM 7.5).
+
+    The GEMM is emitted *inside* the converter, so the graph holds one
+    ``_cdist_forward`` node and no ``mm``/``bmm`` for ``gemm_capability_validator`` to
+    reject, and TensorRT-RTX then fails cuDNN graph compilation with "No valid engine
```
tp5uiuc: REmove everything after fails i.e. from cuDNN graph compilation to the end,

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3366,7 +3366,54 @@ def aten_ops_convolution(
...
+    ``cdist_emits_matmul``, the converter's own predicate, so the two cannot drift.
+    Unlike ``gemm_capability_validator`` this is not dtype-keyed: measured on a T4 the
+    fused pattern fails for FP16 operands too, so what predicts failure is whether the
+    matmul layer is emitted, not its precision.
+    """
+    if not trt_rtx_targets_turing(settings):
+        return True
+
+    rows: List[Optional[int]] = []
+    for arg in node.args[:2]:
```
tp5uiuc: Similar to #4643 name these

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3366,7 +3366,54 @@ def aten_ops_convolution(
...
+        val = arg.meta.get("val") if hasattr(arg, "meta") else None
+        shape = getattr(val, "shape", None)
```
tp5uiuc: Use walrus or other operators to make this code readable

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3366,7 +3366,54 @@ def aten_ops_convolution(
...
+        if shape is None or len(shape) < 2 or not isinstance(shape[-2], int):
```
tp5uiuc:
```
is_valid_shape = shape and len(shape) >= 2 and isinstance(shape[-2], int)
rows.append(shape[-2] if is_valid_shape else None)
```

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -3366,7 +3366,54 @@ def aten_ops_convolution(
...
+            rows.append(None)
+        else:
+            rows.append(shape[-2])
+
+    if not impl.normalization.ops.cdist_emits_matmul(
+        args_bounds_check(node.args, 2, 2.0),
+        args_bounds_check(node.args, 3, None),
```
tp5uiuc: for 2.0 and None, use `replacement` explicitly so its readable

tp5uiuc:
### CI summary

| failing check | cause |
|---|---|
| `gate` ×2 | aggregators |

Suites ran to completion: `2 failed, 150 passed, 61 skipped, 4 xpassed` and `1 failed, 52 passed`.

**Why these are not from this PR:** the same `test_cross_runtime_serde` pair fails identically on #4643, #4644, #4645, #4646, #4647 and #4648, whose change sets are disjoint — e.g. #4648 changes `_exporter.py` (the save path) and #4647 changes neither that nor anything else in common, yet both fail the same two tests with byte-identical error text. `executorch-runtime-test` reproduces on `main` nightlies and on unrelated PRs (#4638, #4642). The `gate` entries are aggregators of the above. `test_hf_gqa_model` also fails on #4643 and #4647 and has no overlap with anything this PR changes.

narendasan (approved):
LGTM

PR #4646, opened 2026-08-29
fix: guard the FP32 GEMMs that reach TensorRT-RTX through linear and attention

tp5uiuc:
**What** — Registers `gemm_capability_validator` on `aten.linear.default` and on the SDPA converters,
and follows that guard through the four test sites it changes.

**Why** — The GEMM guard covered `matmul`/`mm`/`bmm`/`dot`/`mv`/`addmm`, but not `aten.linear.default`
— and Torch-TensorRT deliberately disables the `linear -> addmm` decomposition, so the guarded node
never appears. Attention is converted as a unit, so no `mm`/`bmm` node is ever created either. Both
reach TensorRT-RTX on Turing and fail:

    cuDNN graph compilation failed: No valid engine configs for Matmul_ -> null execution context

and on dynamic shapes the Myelin error is logged at `enqueueV3` but **not raised** — 3 of the 19
affected cases returned silently wrong values.

**How** — Register the validator on `aten.linear.default` and on the three attention converters,
then follow it everywhere it lands. The resource-partitioning test's hard-coded accelerated and
non-accelerated counts no longer hold once the model's trailing FP32 `nn.Linear` is pushed into a
PyTorch block. Three model tests reach an entry point that emits one engine for the whole program and
has no partitioner to fall back to, so the rejection raises instead of producing a PyTorch block —
those run their model in FP16 on Turing, keyed off the same predicate the validator uses so the two
cannot drift.

**Testing** — Confirmation sweep, both arms. 18 failures closed directly; the guard additionally costs
**10 tests** (2 in `partitioning/`, 8 in `models/`) which this PR carries the fixes for. Only 4 of
those 10 showed as `pass -> fail` — the other 6 were already failing under a different cause, so a
status diff alone does not find them. Neutralising the guard at runtime brings back the null execution
context and the silent all-zeros output, confirming it is correct and necessary. No status change on
the L40S.

**Cost / Gotchas**
- **Coverage:** wherever the guard fires, the Turing arm exercises less than the L40S arm — measured
  at ~10% of the streamable weights for the weight-streaming model, which is FP32 `nn.Linear`-heavy.
- The weight-stripping and refit tests now run in **FP16 on Turing and FP32 elsewhere**, so the two
  arms are no longer testing the same dtype. The assertions (engine sizes, stripping, refit) are
  equally meaningful in FP16 and the network is strongly typed, but it is a real asymmetry.

**Followups** — 5 `models/` cosine-similarity failures on Turing (0.9862–0.9872 against a 0.99
threshold) are **not** addressed here. They compile fully to TensorRT with no capability fallback
logged, which rules the guards out, and they were already failing before this series. Not diagnosed.


tp5uiuc:

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -4537,9 +4541,49 @@ def aten_ops_linear(
...
+    node: Node, settings: Optional[CompilationSettings] = None
+) -> bool:
+    """Reject fused attention TensorRT-RTX cannot serve when Turing (SM 7.5) is a target.
+
+    A fused attention op carries the same FP32 GEMMs ``gemm_capability_validator``
+    rejects, but inside the converter, so the graph holds one node and no ``mm``/``bmm``
+    for that guard to see. On Turing the result is a cuDNN graph-compilation failure --
+    a null execution context under static shapes, a silently wrong result under dynamic
+    ones. Only the q/k/v dtypes matter, so FP16 attention keeps running on TensorRT.
+    """
```
tp5uiuc: On Turing TRT-RTX this results in failure. Thats it no need to say more. `Only the q/k/v dtypes matter, so FP16 attention keeps running on TensorRT.` this can be kept.

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -4537,9 +4541,49 @@ def aten_ops_linear(
     )
 
 
+# Operand dtypes TensorRT-RTX cannot serve as a fused attention op on Turing (SM 7.5).
+# Attention is converted as a single fused subgraph, so the matmuls it performs never
+# appear as nodes in the graph and ``gemm_capability_validator`` never sees them; the
+# dtypes that GEMM cannot serve have to be listed again here.
+_TURING_UNSUPPORTED_ATTENTION_DTYPES = (torch.float32,)
```
tp5uiuc: Push it inside the function

py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
```diff
@@ -4537,9 +4541,49 @@ def aten_ops_linear(
...
+    ones. Only the q/k/v dtypes matter, so FP16 attention keeps running on TensorRT.
+    """
+    if not trt_rtx_targets_turing(settings):
+        return True
+
+    for arg in node.args[:3]:
+        val = arg.meta.get("val") if hasattr(arg, "meta") else None
+        if val is not None and (
+            getattr(val, "dtype", None) in _TURING_UNSUPPORTED_ATTENTION_DTYPES
+        ):
```
tp5uiuc: Same readability concernts as #4643 etc.

tp5uiuc:
### CI summary

| failing check | cause |
|---|---|
| `standard / test (dynamo-models, standard, l2)` | `test_cross_runtime_serde::test_save_cpp_load_python`, `::test_save_python_load_cpp`, and `test_weight_stripped_engine::test_torch_compile_with_refittable_weight_stripped_engine` — all Windows only |
| `gate` ×2 | aggregators |

Suite ran to completion: `3 failed, 149 passed, 67 skipped, 4 xpassed`.

The third failure is `PermissionError: [WinError 32] ... timing_cache.bin` — a Windows file lock between parallel xdist workers. This PR does not modify `test_torch_compile_with_refittable_weight_stripped_engine`, and its edits to that file are no-ops off TensorRT-RTX (`_turing_safe_gemm_dtype()` returns `torch.float32` unless `trt_rtx_targets_turing()`, which requires `ENABLED_FEATURES.tensorrt_rtx`).

Note this PR needed `backend: TensorRT` added alongside `backend: TensorRT-RTX` — the RTX label alone resolves `backend=rtx` in `_decide.yml`, so no standard-TensorRT jobs were scheduled. With both labels it resolves to `both` and the standard suite (32 jobs) now runs, including `dynamo-partitioning` and `dynamo-models`, which cover this PR's test changes.

narendasan (approved):
LGTM

PR #3736, opened 2025-07-31
feat: Add support for Groot N1.5 model

peri044:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

Please delete options that are not relevant and/or add your own.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

# Checklist:

- [ ] My code follows the style guidelines of this project (You can use the linters)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #3418, opened 2025-02-27
Added flux demo

cehongwang:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

Please delete options that are not relevant and/or add your own.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

# Checklist:

- [ ] My code follows the style guidelines of this project (You can use the linters)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas and hacks
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests to verify my fix or my feature
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added the relevant labels to my PR in so that relevant reviewers are notified

PR #4647, opened 2026-08-29
fix: keep PyTorch-executed weights on the device when the mutable module offloads

tp5uiuc:
**What** — Gates `deallocate_module()` on the user-facing `offload_module_to_cpu` setting at
the three call sites on ToT that ignores it. Also puts PyTorch-executed submodules back on the target
device after every offload. Adds a test pinning what the setting guarantees for a partially
converted model.

**Why** — A hybrid graph module does not own its weights. `torch.export.export` keeps the source
module's `Parameter` objects, `ExportedProgram.module()` re-registers those same objects, and both
partitioners share attributes by `setattr` — so the compiled module and the module it was compiled
from hold identical `Parameter` instances. A `nn.Module.to()` rebinds `.data` on them and so moving one
moves the other. `deallocate_module()` is a `.to('cpu')`, so any graph the partitioner did not hand
to TensorRT in full had its PyTorch-executed weights dragged off the device while activations stayed
on it, leading to errors like this

```
    RuntimeError: Expected all tensors to be on the same device, but got mat1 is on
    cuda:0, different from other tensors on cpu (wrapper_CUDA_addmm)
```

A fully converted graph does not see this error, because TensorRT submodules keep their weights inside the
engine. Related PRs #3418 added the unconditional deallocate at all four call sites and
#3736 later gated it behind `offload_module_to_cpu` in `compile()` **only**. The other three call sites of  `update_refit_condition()`, `refit_gm()` and the static `load()` kept offloading regardless.

**How** — Do both halves together: gate the three offload sites that ignored the setting, and put
PyTorch-executed submodules back on the target device afterwards. Doing only one of them means the bug is still open.  The `offload_module_to_cpu` test now asserts what the contract actually guarantees for a
partially converted model rather than that the whole source model lands on CPU.

**Testing** — Reproduced on an **L40S (SM 8.9)** with `torch_executed_ops` forcing the fallback and no capability guards involved, so this is neither card- nor backend-specific. Confirmation sweep (T4 SM 7.5 + L40S, driver 595.58.03): 5 `runtime/` and 1 `models/` failures closed, no status change on the L40S.

**Cost / Gotchas** — Behaviour change: `offload_module_to_cpu=True` on a partially converted model
now leaves PyTorch-executed weights on the device. That is what makes the model runnable, but a
caller relying on "everything ends up on CPU" will see a difference.


py/torch_tensorrt/dynamo/runtime/_MutableTorchTensorRTModule.py
```diff
@@ -33,6 +33,33 @@ def _is_modelopt_quantized(model: torch.nn.Module) -> bool:
...
+    Only the TensorRT submodules can spare their weights -- those live inside the engine.
+    Everything the partitioner left in PyTorch still reads its parameters from the graph
+    module, and ``compile_module`` puts those submodules on the target device the same way.
+    """
+    for name, submodule in gm.named_children():
+        if not isinstance(submodule, torch.fx.GraphModule):
+            continue
+        if "_run_on_acc" in name:
+            continue
+        submodule.to(device)
```
tp5uiuc, line 48: This is maybe a behavior change, but it is needed to keep the code from breaking with the `offload_module_to_cpu` flag. It is needed because export share Parameter identity between the original model and the graph module. Having a `.to("cpu")` on the original model yanks the weights that `_run_on_gpu_*` needs to execute in PyTorch. Pinning these submodules back is actually the same placement compile_module already does here 

https://github.com/pytorch/TensorRT/blob/0fad50d6e0e64f730df485eac26cbc8e6e596dae/py/torch_tensorrt/dynamo/_compiler.py#L1428-L1442

and so this restores an invariant. Without it we get errors in the MR description.

tp5uiuc:
### CI summary

| failing check | cause |
|---|---|
| `build / build-windows-arm64-cross`, `build / prepare-upload` | `The operation was canceled` — superseded run; the retry of both passed |
| `gate` ×3 | aggregators |

This PR changes only `_MutableTorchTensorRTModule.py` plus two test files; `test_hf_gqa_model.py` contains no reference to `Mutable`, `offload` or `deallocate`.

PR #4648, opened 2026-08-29
fix: order the placeholders first before computing serde range constraints

tp5uiuc:
**What** — Hoists every placeholder to the front of the node list before `make_constraints()` runs,
preserving their relative order.

**Why** — `torch_tensorrt.save(..., retrace=False)` raises for **any** graph that fell entirely back
to PyTorch, on any device:

    IndexError: list index out of range
    torch/_export/non_strict_utils.py:935, flat_dynamic_shapes[input_index - num_lifted_inputs]

`make_constraints()` walks `enumerate(gm.graph.nodes)` and indexes `flat_dynamic_shapes` by the
*node* index, so it carries an unstated precondition: placeholders must be the leading nodes. For some rewritten graphs, this is not satisifed.

**How** — Hoist placeholders to the front of the node list before the constraints are computed,
keeping their relative order, so the precondition holds for exporter-rewritten graphs too.

**Testing** — Reproduced on an **L40S (SM 8.9)** with `torch_executed_ops` forcing full fallback and
no capability guards involved: a normal compile saves fine, forcing `linear` to PyTorch raises.
Confirmation sweep, both arms: 2 `models/` failures closed, no status change on the L40S.


tp5uiuc:
### CI summary

| failing check | cause |
|---|---|
| `gate` ×2 | aggregators |

Worth stating explicitly for this PR, since it changes `_exporter.py`: the two failing tests are save/load tests, but they fail identically on #4643, #4644, #4645, #4646 and #4647, none of which touch `_exporter.py`.

PR #4651, opened 2026-08-30
fix(fx): drop the dynamic_range argument from the 8 remaining mark_as_int8_layer calls

Anai-Guo:
## Problem

13873bd ["remove dynamic_range in fx" (#4221)](https://github.com/pytorch/TensorRT/commit/13873bd4) changed

```python
# py/torch_tensorrt/fx/converters/converter_utils.py:710
-def mark_as_int8_layer(layer, dynamic_range):
+def mark_as_int8_layer(layer):
```

and cleaned up `adaptive_avgpool.py`, `impl/activation.py`, `maxpool.py`,
`quantization.py`, `transformation.py` and **one of the three** sites in
`linear.py`. Eight call sites were missed and still pass two positional
arguments:

```
TypeError: mark_as_int8_layer() takes 1 positional argument but 2 were given
```

Replaying every `mark_as_int8_layer(...)` call site in `fx/converters/` against
the signature parsed out of `converter_utils.py` on `main`:

```
main signature: (layer)

  TypeError add.py:47   (2 positional) -> mark_as_int8_layer() takes 1 positional argument but 2 were given
  TypeError add.py:64   (2 positional) -> ...
  TypeError add.py:70   (2 positional) -> ...
  TypeError batchnorm.py:21  (2 positional) -> ...
  OK        linear.py:27  (1 positional)
  TypeError linear.py:43  (2 positional) -> ...
  TypeError linear.py:51  (2 positional) -> ...
  TypeError mul.py:46   (2 positional) -> ...
  TypeError impl/convolution.py:130  (2 positional) -> ...

8 broken / 1 correct
```

Each one sits behind an `is_quantized` / `scale is not None` guard, so it is
reached by quantized `add`, `add_relu`, `mul`, `BatchNorm`, `Linear` and
`Conv` conversion through the FX path.

## Fix

Same idiom #4221 used for `linear.py:27`: keep the call, drop the argument.
Where the argument was the only consumer of a `get_dyn_range(...)` result, the
computation and the now-unused import go too.

`get_dyn_range` itself is kept — `quantization.py` still uses it.

Note the two shapes #4221 dealt with: sites guarded by `if input_val.dynamic_range:`
were deleted whole (that attribute is gone in TRT 10), while sites guarded by
`if is_quantized:` kept the call. All eight sites here are the second kind, so
none of them are deleted.

## Checks

`black` (26.3.1, the pinned pre-commit rev) reports all five files unchanged.
`pyflakes` is clean on the patched files — the one remaining note,
`'typing.Any' imported but unused` in `impl/convolution.py`, is pre-existing on
`main` and untouched here.

### DIFF
--- a/py/torch_tensorrt/fx/converters/add.py
+++ b/py/torch_tensorrt/fx/converters/add.py
@@ -6,7 +6,7 @@ import torch
 
 from ..converter_registry import tensorrt_converter
 
-from .converter_utils import get_dyn_range, mark_as_int8_layer
+from .converter_utils import mark_as_int8_layer
 
 
 @tensorrt_converter(operator.add)
@@ -43,8 +43,7 @@ def quantized_add(network, target, args, kwargs, layer_name):
 
     layer = network.add_elementwise(lhs_val, rhs_val, trt.ElementWiseOperation.SUM)
     layer.name = layer_name
-    dyn_range = get_dyn_range(kwargs["scale"], kwargs["zero_point"], torch.quint8)
-    mark_as_int8_layer(layer, dyn_range)
+    mark_as_int8_layer(layer)
 
     return layer.get_output(0)
 
@@ -60,13 +59,12 @@ def quantized_add_relu(network, submod, args, kwargs, layer_name):
 
     layer = network.add_elementwise(lhs_val, rhs_val, trt.ElementWiseOperation.SUM)
     layer.name = f"{layer_name}_add"
-    dyn_range = get_dyn_range(kwargs["scale"], kwargs["zero_point"], torch.quint8)
-    mark_as_int8_layer(layer, dyn_range)
+    mark_as_int8_layer(layer)
 
     layer = network.add_activation(
         input=layer.get_output(0), type=trt.ActivationType.RELU
     )
     layer.name = f"{layer_name}_relu"
-    mark_as_int8_layer(layer, dyn_range)
+    mark_as_int8_layer(layer)
 
     return layer.get_output(0)
--- a/py/torch_tensorrt/fx/converters/batchnorm.py
+++ b/py/torch_tensorrt/fx/converters/batchnorm.py
@@ -6,7 +6,7 @@ import torch
 
 from ..converter_registry import tensorrt_converter
 
-from .converter_utils import get_dyn_range, mark_as_int8_layer, to_numpy
+from .converter_utils import mark_as_int8_layer, to_numpy
 
 
 def common_batchnorm(network, mod, input_val, layer_name, is_quantized):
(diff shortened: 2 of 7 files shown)


micwill755 (changes requested):
After this change, get_dyn_range has no remaining callers. #4221 already removed the last use in quantization.py and left only the import. 

Could you also delete get_dyn_range from converter_utils.py and the unused import in quantization.py? It's leftover from the old ITensor.dynamic_range path which TRT 10 no longer exposes.

Anai-Guo:
Thanks @micwill755 — done in `367eff1`.

- `py/torch_tensorrt/fx/converters/converter_utils.py`: removed `get_dyn_range`.
- `py/torch_tensorrt/fx/converters/quantization.py`: dropped it from the `from .converter_utils import ...` line (the other name on that line, `get_inputs_from_args_and_kwargs`, is still used).

Confirmed there are no other references left in the repo — the only remaining hits for `get_dyn_range` before this commit were the eight `mark_as_int8_layer` call sites this PR already fixed, the definition itself, and that import.

`py/torch_tensorrt/fx` is excluded from both black (`force-exclude`) and isort (`skip`) in `pyproject.toml`, so the deletions are formatter-neutral.

Anai-Guo:
Small note on the red checks here, since they are unrelated to this PR.

`executorch-runtime-test / test / executorch-runtime-test--3.10-cu132` fails on a single assertion in `tests/py/dynamo/executorch/test_api.py` (`1 failed, 220 passed`):

```
FAILED executorch/test_api.py::test_runtime_wheel_pins_cuda_13_native_dependencies
  assert 'TENSORRT_DISTRIBUTION = "tensorrt-cu13"' in setup_source
```

That literal stopped existing on `main` in #4635 ("bring back cu126 support", 2026-09-02), which replaced the hard-coded constant with `TENSORRT_DISTRIBUTION = tensorrt_distribution()` in `py/torch-tensorrt-executorch-runtime/setup.py`. The test still greps for the old string, so it is red for every PR based on current `main`; #4659 already covers it.

This PR only touches `py/torch_tensorrt/fx/converters/{converter_utils,quantization}.py` and the fx tests, none of which the ExecuTorch runtime test imports. The `gate` failure is just the aggregate of that job.

Everything you asked for is in `367eff1`; happy to rebase once #4659 lands if you would like a green run before merging.

Anai-Guo:
Friendly ping @micwill755 — the cleanup you asked for landed in `367eff1`: `get_dyn_range` is deleted from `py/torch_tensorrt/fx/converters/converter_utils.py` and dropped from the `from .converter_utils import ...` line in `quantization.py`. There are no remaining references to it anywhere in the tree.

The branch is `MERGEABLE`; the only red check (`executorch-runtime-test`) is a pre-existing `main` failure unrelated to this PR (#4635 replaced the `TENSORRT_DISTRIBUTION` string literal that the assertion looks for). Could you take another look / dismiss the stale "changes requested" when you get a chance? Happy to rebase if it has gone stale.

PR #4654, opened 2026-09-02
executorch support cu126

lanluo-nvidia:
# Description

Fixes # (issue)

## Type of change

# Checklist:


shoumikhin:
Thanks for picking this up — and apologies for the collision. I opened #4655 against `release/2.14` without noticing this PR was already open (yours predates mine by about 90 minutes), so we independently wrote overlapping fixes. Seven of my nine files touch the same paths as yours, including `.github/scripts/filter-executorch-cuda-arches.py`, which we both *add*.

Having read yours properly, **I think #4654 is the better base and should be the one that lands.** Replacing the build-time metadata reads with declared requirements is a better design than what I did. I patched `installed_version()` / `CUDA_RUNTIME_DISTRIBUTION` to resolve the right distribution per CUDA major; you delete that machinery entirely, which removes the whole failure class instead of fixing one instance of it. Two of my five changes are simply obsolete under yours:

- my `nvidia-cuda-runtime-cu12` fix — no metadata lookup remains to get wrong
- my widening of `tensorrt_distribution` from `12.6` to `12.` — your `get_tensorrt_requirement()` already matches on the major

And you have `version.txt` plus the `artifact-name-prefix` input declaration, which I don't.

## Two things in #4655 that #4654 doesn't cover

Both are in files you already touch, and both are small. Measured against `lluo/executorch_support_cu126` at `ca855f06`.

**1. The arch filter doesn't expand PyTorch's named architecture lists, or split space-separated ones.**

```
                    #4654            #4655
'Pascal'         -> Pascal           6.1+PTX
'5.0 6.0 7.5'    -> 5.0 6.0 7.5      7.5
'Maxwell'        -> Maxwell          (exits non-zero, nothing reaches sm_61)
'5.0,6.0,7.5'    -> 7.5              7.5        # comma case already works
```

Both unfiltered values still reach nvcc as sub-sm_61 targets. Confirmed by asking torch 2.15 directly on an H100:

```
'Pascal'      -> -gencode=arch=compute_60,code=sm_60
'5.0 6.0 7.5' -> -gencode=arch=compute_50,code=sm_50
                 -gencode=arch=compute_60,code=sm_60
```

which is exactly the `__dp4a` failure the filter exists to prevent. `Pascal` is `6.0;6.1+PTX` in torch's own table, and torch normalises spaces to semicolons (`torch/utils/cpp_extension.py` does `_arch_list.replace(' ', ';')`), so a space-separated list is a legitimate input.

My version expands the named list first, mirroring torch's table including its longest-key-first ordering — that ordering matters, since torch applies it by plain substring replacement and `Maxwell+Tegra` has to be replaced before `Maxwell` or a `+Tegra` tail survives. Neither case can fire in current CI, where the container exports a numeric semicolon list, so this is hardening rather than a live break.

**2. `_load_module` imports the loader through `executorch.extension.pybindings.portable_lib`, which is the name `activate()` treats as "stock runtime imported first".**

`__init__.py:15` sets `_WRAPPER_NAME` to that module, and `activate()` raises when it is in `sys.modules`:

```python
if existing is not None or _WRAPPER_NAME in sys.modules:
    raise DelegateCompatibilityError("ExecuTorch's stock runtime was imported first. ...")
```

So loading a program binds that name, and a later `activate()` in the same process refuses. This is what produced the `DelegateCompatibilityError` in `test_activate_twice_is_safe` when you pulled #4655 into this PR — the failure was mine, from that import, not from your changes. Taking the loader off the module `activate()` returns avoids installing the name at all:

```python
get_runtime()          # verifies TensorRTBackend is registered
native = activate()    # the native module it installed
return native._load_for_executorch_from_buffer(data)
```

After that change the full suite is clean on an H100 — `225 passed`, 0 failed, running `pytest -ra -n auto executorch/` from `tests/py/dynamo` so the real `conftest.py` and the whole directory are in play. Worth noting I only found the four regressions you reported because your invocation exercises the whole directory; running single files, as I had been, hides `sys.modules` interactions between them.

## How would you like to resolve it?

Whichever is least work for you:

**A. I rebase #4655 onto this branch and cut it down to just those two items.** You merge as planned; my PR becomes a small follow-up that only touches the filter script and `_load_module`. Nothing blocks you.

**B. You take the two changes directly into #4654 and I close #4655.** Fewest PRs, and the release branch gets one coherent change. I'm happy to hand you the exact diffs, or push them to your branch if you'd rather.

I'd lean toward **B** for a release branch, but A keeps this PR untouched if you're close to merging.

One request either way: if the conflict on `filter-executorch-cuda-arches.py` gets resolved by keeping this branch's copy, the named-arch and space-separator handling is silently lost. That resolution looks correct from the diff, so it's an easy one to make by accident.

PR #4655, opened 2026-09-02
fix(executorch): back device-planned arenas when loading in Python

shoumikhin:
Running a TensorRT delegated `.pte` through the Python runtime wheel fails on the very first operator, as reported by @lanluo-nvidia:

```
[cuda_allocator.cpp:88] cudaMemcpy H2D failed: invalid argument (384 bytes, device 0)
[op__device_copy.cpp:251] Check failed: _h2d_copy: copy_host_to_device failed
[method.cpp:1480] KernelCall failed at instruction 0:0 in operator et_copy::_h2d_copy.out
```

When a program is exported for a GPU delegate, ExecuTorch adds a copy at each delegate boundary and tags the memory-planned arenas those copies write into with the device they live on. A runner has to read those tags and allocate real device memory for the tagged arenas, otherwise the copy is handed a plain host pointer and CUDA rejects it. That rejection is the error above, and 384 bytes is exactly the example input.

ExecuTorch has two ways to load a program from Python and only one reads the tags. The Module API allocates device memory for a tagged arena. The program loader behind `executorch.runtime` plans every arena on the host and never looks at the tags, so every device copy gets a host destination. This wheel used the second one, which is why the C++ reference runner, which allocates the device arenas itself, worked while Python did not.

Load through the Module API instead. Inputs and outputs still cross the Python boundary as CPU tensors and the delegate still runs on the GPU.

## The cu126 build failure

ExecuTorch's CUDA shims call `__dp4a`, which nvcc does not declare before sm_61. The cu126 build environment exports `TORCH_CUDA_ARCH_LIST='5.0;6.0;7.0;7.5;8.0;8.6;9.0'`, so those two passes fail with `identifier "__dp4a" is undefined`, in `int4_plain_mm.cuh`, `int5_plain_mm.cuh`, `int6_plain_mm.cuh` and `int8_plain_mm.cuh`. The CUDA 13 channels start higher and build fine.

The pinned ExecuTorch has no `__CUDA_ARCH__` guard around those calls, so no configuration of 1.4.1 compiles for 5.0 or 6.0. This narrows the requested list on our side instead: nothing in ExecuTorch changes and the pin stays at 1.4.1. Filtering rather than pinning a list here, so a channel that adds a newer architecture still builds for it without editing the filter.

This is also why the wheel went untested. `executorch-runtime-test` is gated on `executorch-runtime-build` succeeding, so the cu126 failure skipped the whole test job and `load_model.py` never ran, which is how the arena bug above reached a release.

## The standalone import

`__init__.py` calls `importlib.util.find_spec` on the strength of a bare `import importlib`, which does not bind the `util` submodule. Anything that imports `torch` first has it bound transitively, so this surfaced only when the runtime wheel was imported on its own, and it is unrelated to the arena bug. The same line is on `main`, so that part wants forward-porting.

## Not #4581

That fixed `kernel 'et_copy::_h2d_copy.out' not found`, where the kernel never ran, and `release/2.14` already contains it. Re-applying it produces `Allocator already registered for device type: 1`, because `EXECUTORCH_BUILD_CUDA=ON` links `cuda_backend.cpp`, whose static initializer already registers `CudaAllocator`.

## Test plan

Measured on the pinned ExecuTorch release with a CUDA delegated program carrying the same device copies, one arena tagged for the GPU: the program loader plans that arena on the host and never asks about devices, while the Module API asks for device memory for it. End to end on a build that registers a device allocator, the Module API returns the expected result where the program loader hits the failure above.

The architecture filter is exercised against the list the cu126 environment actually exports: 5.0 and 6.0 are dropped and the rest kept, a CUDA 13 list comes back byte for byte, a `+PTX` suffix survives with its number, and a request with nothing at or above sm_61 exits non-zero rather than printing an empty list, which would leave nvcc to pick its own default.

For the import, confirmed in a clean interpreter that `import importlib` leaves `hasattr(importlib, "util")` False and that after `import torch` it is True. The added test reports any dotted name the package reads through without importing, restricted to importable modules so an ordinary attribute like `sys.modules` is not flagged.

Not verified here: the end-to-end run used a CUDA delegated program rather than a TensorRT delegated one. The arena machinery is backend independent and carries the identical `et_copy` ops, but the TensorRT path end to end is unproven by me.

### DIFF
--- a/.github/scripts/filter-executorch-cuda-arches.py
+++ b/.github/scripts/filter-executorch-cuda-arches.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3
+"""Drop CUDA architectures ExecuTorch's shims cannot compile for.
+
+The shims call ``__dp4a``, which nvcc does not declare before sm_61, so a list carrying
+5.0 or 6.0 fails with ``identifier "__dp4a" is undefined`` whichever CUDA is in use. The
+cu126 build environment asks for ``5.0;6.0;7.0;7.5;8.0;8.6;9.0`` while the CUDA 13 ones
+start at 7.5, which is why only cu126 broke.
+
+Filtering what the environment asked for rather than pinning a list here, so a channel
+that adds a newer architecture still builds for it without editing this file.
+"""
+
+from __future__ import annotations
+
+import sys
+
+# The first architecture that declares __dp4a. GP100 is 6.0 and does not have it; the
+# rest of Pascal is 6.1 and does.
+MINIMUM = 6.1
+
+# PyTorch accepts named architectures and expands each to a version list before building
+# its gencode flags (torch/utils/cpp_extension.py). Expand them the same way here, or a
+# name would pass through untouched and carry its sub-6.1 members into the build: Pascal
+# alone expands to 6.0;6.1+PTX, and 6.0 is precisely what this filter exists to drop.
+#
+# Ordered longest-key-first, and applied by substring replacement, because that is what
+# torch does: "Maxwell+Tegra" has to be replaced before "Maxwell", or the suffix is left
+# behind as "5.0;5.2+PTX+Tegra".
+NAMED_ARCHES = (
+    ("Kepler+Tesla", "3.7"),
+    ("Kepler", "3.5+PTX"),
+    ("Maxwell+Tegra", "5.3"),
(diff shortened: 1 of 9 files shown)


lanluo-nvidia (approved):
lgtm

PR #4660, opened 2026-09-02
executorch cu126 issue fix

lanluo-nvidia:
# Description

Fixes # (issue)

## Type of change

# Checklist:


shoumikhin:
Heads-up on a test collision from my #4655, which landed on `release/2.14` just before you opened this. This is my mess to flag, not a problem with your change.

`tests/py/dynamo/executorch/test_api.py::test_runtime_wheel_pins_its_native_dependencies_per_cuda_major` will fail here. I rewrote that test in #4655 to *execute* both resolvers rather than grep for package names, because the grep version passed even when the CUDA 12 and CUDA 13 mappings were swapped. The stronger version now correctly detects that this PR changed the shape it was written against.

Three assertions break, checked against `setup.py` at `lluo/executorch_126_0902`:

```
"TENSORRT_DISTRIBUTION = tensorrt_distribution()"        -> not present
"CUDA_RUNTIME_DISTRIBUTION = cuda_runtime_distribution()" -> not present
wanted = ("tensorrt_distribution", "cuda_runtime_distribution")
  module-level functions here: ['cuda_runtime_distribution', 'get_tensorrt_requirement']
  -> the set-equality assert fails on the missing tensorrt_distribution
```

Your restructuring is the better design and I'd keep it — folding the TensorRT choice into `get_tensorrt_requirement()` returning a full requirement string is cleaner than my separate distribution-name resolver, and it removes the build-time metadata reads entirely rather than fixing one of them. The test should follow your code, not the reverse.

Concretely, in that test: `tensorrt_distribution` becomes `get_tensorrt_requirement`, the two `assert ... in setup_source` lines for the module-level assignments go away since there is no `TENSORRT_DISTRIBUTION` constant any more, and the per-version expectations become the requirement strings your function returns:

```python
    wanted = ("get_tensorrt_requirement", "cuda_runtime_distribution")
    ...
    for cuda_version, tensorrt, cuda_runtime in (
        ("12.6", "tensorrt-cu12>=11.1.0,<11.2", "nvidia-cuda-runtime-cu12"),
        ("12.9", "tensorrt-cu12>=11.1.0,<11.2", "nvidia-cuda-runtime-cu12"),
        ("13.0", "tensorrt-cu13>=11.1.0,<11.2", "nvidia-cuda-runtime"),
        ("13.2", "tensorrt-cu13>=11.1.0,<11.2", "nvidia-cuda-runtime"),
    ):
        fake_torch.version.cuda = cuda_version
        assert namespace["get_tensorrt_requirement"]() == tensorrt
        assert namespace["cuda_runtime_distribution"]() == cuda_runtime
```

The rest of the test works unchanged: it parses `setup.py`, executes just those functions with a stubbed `torch.version.cuda`, and also asserts that a CUDA-less torch and an unsupported major both raise. Your `get_tensorrt_requirement` raises in both cases, so that part still holds.

Two small things I noticed while reading, both optional:

- `cuda_runtime_distribution` here is my version verbatim, including a docstring that explains the `nvidia-cuda-runtime` vs `-cu12` split. Since `get_tensorrt_requirement` now splits on the same major, it may read better to say that once at the top rather than in one of the two.
- Worth confirming `>=11.1.0,<11.2` is what you want for both majors. The CUDA 12 and CUDA 13 TensorRT lines do not have to advance in lockstep, so a shared range will need editing whenever either moves.

Thanks for dropping the overlapping files from #4654 — the arena fix and the arch filter are on the branch now, so this rebase is a clean delta. Happy to push the test update to your branch if that is easier than doing it yourself; just say so.

PR #4661, opened 2026-09-02
fix(runtime): support NCCL2 communicator binding

### LINKED ISSUE #4658: Distributed NCCL tests fail: getBackend no longer returns a ProcessGroupNCCL
## What happens

Every multi-device TensorRT test fails as soon as it tries to use NCCL. The error is the same one 20 times:

```
RuntimeError: [Error thrown at core/runtime/TRTEngine.cpp:916]
Expected nccl_pg != nullptr to be true but got false
```

That check is `Backend is not ProcessGroupNCCL`, and it appears 74 times in a single job log. Result on a recent run: **20 failed, 71 passed** in `tests/py/dynamo/distributed/`.

## Which tests

- `distributed/test_native_nccl.py` — 19 of the 20
- `distributed/test_export_save_load.py::test_export_save_load_round_trip` — the remaining one, failing with the same NCCL error rather than an export error

## Where it comes from

`core/runtime/TRTEngine.cpp` asks the process group for its NCCL backend and then casts it:

```cpp
auto backend = pg->getBackend(c10d::ProcessGroup::BackendType::NCCL);
TORCHTRT_CHECK(backend != nullptr, "ProcessGroup '" << this->group_name << "' has no NCCL backend");
auto* nccl_pg = dynamic_cast<c10d::ProcessGroupNCCL*>(backend.get());
TORCHTRT_CHECK(nccl_pg != nullptr, "Backend is not ProcessGroupNCCL");   // line 916, this is what fails
```

The first check passes, so `getBackend` **does** return an object. The second fails, so whatever it returns is **not** a `ProcessGroupNCCL`. The cast yields null and the check trips.

Possibly relevant: the tests build their group with `dist.new_group(ranks=[0])`, a single-rank subgroup. That is the kind of group most likely to be backed by a wrapper rather than the concrete NCCL class.

## It used to pass

Same test file, same Torch-TensorRT code path, two different runs:

| | Torch | Result |
| --- | --- | --- |
| #4464's own merge run, 2026-08-28 | `2.15.0.dev20260826/27+cu130` | **0** `nccl_pg` errors, 91 passed |
| a run today, 2026-09-02 | `2.15.0.dev20260902+cu132` | **74** `nccl_pg` errors, 20 failed |

So this is not a change in `core/runtime/TRTEngine.cpp`. Nothing in that file moved between those runs.

## What I could not determine

**Two things changed between those runs, not one:** the PyTorch nightly date *and* the CUDA channel (`cu130` → `cu132`). I have no run that holds one fixed while moving the other, so I cannot say which is responsible.

This is harder to narrow than it sounds, because the `distributed` suite is `lanes=("nightly",)` in `tests/ci/suites.py`. It does not run on `main` pushes at all — `main`'s latest commit has **zero** distributed rows. It only runs on a pull request carrying the `ci: nightly` label, which is why this surfaces on some PRs and is invisible on others.

I also ruled out one plausible explanation: an ABI mismatch between the Torch that `libtorchtrt` was built against and the one it runs with. A `torch-2.14.0.dev20260728` string does appear in the job log, but it is a Bazel URL for an unrelated config, not the installed Torch.

## Suggested next step

The missing fact is **what type `getBackend` now returns**. Logging that on cast failure would name it instead of leaving us to guess, and it is a small change to the error path:

```cpp
TORCHTRT_CHECK(
    nccl_pg != nullptr,
    "Backend for ProcessGroup '" << this->group_name << "' is not ProcessGroupNCCL, got "
        << c10::demangle(typeid(*backend).name()));
```

With that in one CI run, the fix becomes obvious: either accept the new type, reach through the wrapper, or ask the group differently.

I deliberately have not changed the cast itself. Loosening it would quiet the check while possibly handing TensorRT a communicator that is not NCCL, which is worse than a failing test.

## Reproducing

```
# needs a multi-GPU runner; the suite is nightly-lane only
cd tests/py/dynamo
python -m pytest -ra distributed/test_native_nccl.py
```

On a PR, add the `ci: nightly` label, otherwise the suite is skipped entirely.

apbose:
Fixes https://github.com/pytorch/TensorRT/issues/4658

### DIFF
--- a/toolchains/torch_nccl/defs.bzl
+++ b/toolchains/torch_nccl/defs.bzl
@@ -2,6 +2,7 @@
 
 def _find_nccl_include(repository_ctx, torch_path):
     """Find nccl.h from the pip nvidia-nccl package co-installed with torch."""
+
     # pip's nvidia-nccl package installs nccl.h at <site-packages>/nvidia/nccl/include/
     candidate = torch_path + "/../nvidia/nccl/include"
     result = repository_ctx.execute(["test", "-f", candidate + "/nccl.h"])
@@ -9,16 +10,30 @@ def _find_nccl_include(repository_ctx, torch_path):
         return candidate
     return None
 
+def _has_process_group_nccl2(repository_ctx, torch_path, lib_path):
+    """Check that the installed PyTorch exposes the NCCL2 backend."""
+    header = torch_path + "/include/torch/csrc/distributed/c10d/nccl2/ProcessGroupNCCL.hpp"
+    header_result = repository_ctx.execute(["test", "-f", header])
+    if header_result.return_code != 0:
+        return False
+
+    # The header alone is insufficient: make sure the matching implementation
+    # is present in the PyTorch CUDA library that the detector found.
+    symbol_result = repository_ctx.execute(["grep", "-q", "nccl2", lib_path])
+    return symbol_result.return_code == 0
+
 def _torch_nccl_detect_impl(repository_ctx):
     """Detect if PyTorch was built with NCCL support."""
 
     # Skip detection on non-Linux (NCCL not available)
     os_name = repository_ctx.os.name.lower()
     nccl_include_dir = ""
+    has_process_group_nccl2 = False
     if "linux" not in os_name:
         has_nccl = False
     else:
         # Find libtorch path using the venv python if available, else system python3
+        result = None
         for python_bin in ["python3", "python"]:
             result = repository_ctx.execute([
                 python_bin,
@@ -28,7 +43,7 @@ def _torch_nccl_detect_impl(repository_ctx):
             if result.return_code == 0:
                 break
 
-        if result.return_code != 0:
+        if result == None or result.return_code != 0:
             has_nccl = False
         else:
             torch_path = result.stdout.strip()
@@ -47,6 +62,7 @@ def _torch_nccl_detect_impl(repository_ctx):
                 found = _find_nccl_include(repository_ctx, torch_path)
                 if found:
                     nccl_include_dir = found
+                    has_process_group_nccl2 = _has_process_group_nccl2(repository_ctx, torch_path, lib_path)
                 else:
                     # Cannot find nccl.h — disable to avoid build errors
                     has_nccl = False
@@ -86,8 +102,22 @@ config_setting(
     name = "nccl_enabled",
     flag_values = {{":use_nccl": "True"}},
 )
+
+bool_flag(
+    name = "use_process_group_nccl2",
+    build_setting_default = {has_process_group_nccl2},
+)
+
+config_setting(
+    name = "process_group_nccl2_enabled",
+    flag_values = {{":use_process_group_nccl2": "True"}},
+)
 {nccl_headers_target}
-""".format(has_nccl = has_nccl, nccl_headers_target = nccl_headers_target))
+""".format(
+        has_nccl = has_nccl,
+        has_process_group_nccl2 = has_process_group_nccl2,
+        nccl_headers_target = nccl_headers_target,
+    ))
 
 torch_nccl_detect = repository_rule(
     implementation = _torch_nccl_detect_impl,
@@ -98,5 +128,12 @@ def if_torch_nccl(if_true, if_false = []):
     """Returns if_true if PyTorch has NCCL, else if_false."""
     return select({
         "@torch_nccl//:nccl_enabled": if_true,
+        "//conditions:default": if_false,
+    })
+
+def if_torch_nccl2(if_true, if_false = []):
+    """Returns if_true if PyTorch exposes ProcessGroupNCCL2."""
+    return select({
+        "@torch_nccl//:process_group_nccl2_enabled": if_true,
         "//conditions:default": if_false,
     })
--- a/toolchains/torch_nccl/defs.bzl
+++ b/toolchains/torch_nccl/defs.bzl
@@ -10,17 +10,10 @@ def _find_nccl_include(repository_ctx, torch_path):
         return candidate
     return None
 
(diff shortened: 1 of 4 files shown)


pkisfaludi-nv:
Independent repro that matches this fix, in case it's useful as corroboration.

Building torch-tensorrt from source with `ENABLE_TRT_NCCL_COLLECTIVES` (8× A100, TRT 11.2.1.2, **torch 2.15.0.dev20260902+cu130**, `ENABLED_FEATURES.native_trt_collectives == True`), every test that goes through `distributed_context(group)` dies at exactly the two lines this PR deletes:

```
RuntimeError: [Error thrown at core/runtime/TRTEngine.cpp:916]
Expected nccl_pg != nullptr to be true but got false
Backend is not ProcessGroupNCCL
```

`getBackend(BackendType::NCCL)` returns non-null, so it is specifically the `dynamic_cast<c10d::ProcessGroupNCCL*>` that fails — consistent with the backend now being `c10d::nccl2::ProcessGroupNCCL` on this torch.

Split by whether the binding path is exercised:

| test | result |
|---|---|
| `TestMultirankNccl::test_all_reduce_correctness` (no `distributed_context`) | **passed** |
| `TestMultirankNccl::test_distributed_mode_subgroup` | failed, `TRTEngine.cpp:916` |
| `TestMultirankNccl4GPU::test_two_dimensional_mesh_routing` (from #4380) | failed, `TRTEngine.cpp:916` |

One unrelated gotcha that cost me a while, in case anyone else hits it on a PCIe box: on this node NCCL P2P is broken, and every 2-rank group hung in `dist.barrier()` for 600 s before any of the above surfaced. `NCCL_P2P_DISABLE=1` makes 2-rank collectives work and turns the hangs into the clean error above.

Happy to re-run #4380's 4-GPU mesh-routing test against this branch once it's ready — it's currently the only thing blocking a runtime verdict there.

toolchains/torch_nccl/defs.bzl
```diff
@@ -2,23 +2,38 @@
...
+def _has_process_group_nccl2(repository_ctx, torch_path, lib_path):
+    """Check that the installed PyTorch exposes the NCCL2 backend."""
+    header = torch_path + "/include/torch/csrc/distributed/c10d/nccl2/ProcessGroupNCCL.hpp"
+    header_result = repository_ctx.execute(["test", "-f", header])
+    if header_result.return_code != 0:
+        return False
+
+    # The header alone is insufficient: make sure the matching implementation
+    # is present in the PyTorch CUDA library that the detector found.
+    symbol_result = repository_ctx.execute(["grep", "-q", "nccl2", lib_path])
```
cehongwang: This is generated by AI. I am not sure since I am not very familiar with NCCL but I think it is worth referring to:

This grep matches on PyTorch builds that have no nccl2 backend whatsoever, so the check validates nothing — and because it can only ever subtract from the header result, its sole possible effect is to silently disable nccl2 support in a build that needed it.

I ran the exact command against torch 2.13.0+cu132, which ships no torch/include/torch/csrc/distributed/c10d/nccl2/ directory at all:
```
$ grep -q "nccl2" $TORCH/lib/libtorch_cuda.so; echo $?
0
$ grep -ao "nccl2" $TORCH/lib/libtorch_cuda.so | wc -l
4
All four hits are mangling artifacts, not the namespace you are looking for:

_ZN5torch4cuda4nccl26all2all_single_equal_splitERN2at6TensorES4_iPvRN3c104cuda10CUDAStreamE
  => torch::cuda::nccl::all2all_single_equal_split(at::Tensor&, at::Tensor&, int, void*, c10::cuda::CUDAStream&)
```
4nccl (namespace nccl, 4 chars) is immediately followed by 26, the length prefix of the 26-character all2all_single_equal_split. The substring nccl2 straddles that boundary. So the "make sure the matching implementation is present" comment is not true of the code beneath it.

The check is also redundant. _has_process_group_nccl2 is only reached inside if has_nccl: → if found: (defs.bzl:65), and has_nccl is itself established by grepping ProcessGroupNCCL in the very same libtorch_cuda.so. A PyTorch that ships the nccl2 header ships the nccl2 implementation in the same wheel, so the header test alone already carries the signal. The only outcome the grep can add is a false negative — say libtorch_cuda.so is absent under some SBSA or CPU-only layout — which omits TORCHTRT_HAS_PROCESS_GROUP_NCCL2, compiles the nccl2 branch out, and produces a wheel that throws your new "Unsupported NCCL backend" on the default backend.

Simplest fix is to delete it:
```
def _has_process_group_nccl2(repository_ctx, torch_path, lib_path):
    """Check that the installed PyTorch exposes the NCCL2 backend."""
    header = torch_path + "/include/torch/csrc/distributed/c10d/nccl2/ProcessGroupNCCL.hpp"
    header_result = repository_ctx.execute(["test", "-f", header])
    if header_result.return_code != 0:
        return False
    # The header alone is insufficient: make sure the matching implementation
    # is present in the PyTorch CUDA library that the detector found.
    symbol_result = repository_ctx.execute(["grep", "-q", "nccl2", lib_path])
    return symbol_result.return_code == 0
    return repository_ctx.execute(["test", "-f", header]).return_code == 0
That also drops the now-unused lib_path parameter at the call site. If you would rather keep a symbol check, grep the mangled namespace token instead of the bare string — 5nccl216ProcessGroupNCCL (nccl2 as a 5-char namespace followed by the 16-char ProcessGroupNCCL) cannot collide with torch::cuda::nccl.

```

apbose: removed the binary grep and the now unused lib path parameter. NCCL2 support should be detected from the ProcessNCCL2.hpp header

narendasan (approved):
LGTM, if possible (can be another PR to unblock) would be good to add this stuff to the CMake system

PR #4662, opened 2026-09-02
Raise when TensorRT enqueueV3 fails

### LINKED ISSUE #4481: 🐛 [Bug] Error with Cask conv when using Hardware Compatible flag
##  Bug Description

When compiling on one GPU Arch, and then running on a different GPU Arch, I'm getting the following error:

```sh
ERROR: [Torch-TensorRT] - IExecutionContext::enqueueV3: Error Code 1: Cask (Cask convolution execution In execute at /_src/runtime/gpu/cask/convBaseRunner.cpp:312)
```
I'm using the dynamo path. The flag `hardware_compatible` is set to `True`.
**The error happens at runtime, it doesn't raise exception, it gives dummy outputs.**

[repro_cask_hw_compatible.py](https://github.com/user-attachments/files/30970326/repro_cask_hw_compatible.py)

I compiled on A100 and executed on H100/L4 --> got the error.
I compiled on H100 and executed on A100/B200 --> got the error.

I checked that there is no cuda context conflicts.

## To Reproduce

Attached script: `repro_cask_hw_compatible.py`. It isolates the crash down to
torchaudio's public `WAV2VEC2_BASE` bundle's convolutional feature extractor (no
proprietary model needed) — a stack of `Conv1d` + `GroupNorm`/`LayerNorm` + `GELU`
layers, FP16, `use_explicit_typing=True`.

1. On an Ampere or Hopper GPU (e.g. A100, H100):
   ```
   python repro_cask_hw_compatible.py build --hw-compatible -o engine_hwcompat.pt2
   ```
2. Copy `engine_hwcompat.pt2` to a Blackwell GPU (e.g. RTX 5060), then:
   ```
   python repro_cask_hw_compatible.py run -i engine_hwcompat.pt2
   ```
3. Observe the `Cask (Cask convolution execution ...)` ERROR printed on the first
   inference call (stderr), with no Python exception raised.

Optional control, to see the contrast described above:
```
python repro_cask_hw_compatible.py build -o engine_control.pt2   # hardware_compatible=False
python repro_cask_hw_compatible.py run -i engine_control.pt2     # fails at load, not inference
```

The relevant compile call inside the script:
```python
compiled = torch_tensorrt.dynamo.compile(
    exported,
    inputs=[waveforms],
    use_explicit_typing=True,
    min_block_size=1,
    hardware_compatible=True,
    require_full_compilation=True,
    truncate_double=True,
)
```

## Expected behavior

The engine built with `hardware_compatible=True` on an Hopper GPU should either:
- run correctly on the Blackwell GPU (the contract `hardware_compatible=True` advertises), or
- fail explicitly and loudly at load time, the same way the `hardware_compatible=False`
  control does, if a specific Cask tactic genuinely cannot be made to work cross-architecture.

It should never silently corrupt output at runtime with no raised exception.

## Environment

 - Torch-TensorRT Version (e.g. 1.0.0): 2.11.0
 - PyTorch Version (e.g. 1.0): 2.11.0
 - CPU Architecture: x86-64
 - OS (e.g., Linux): Linux
 - How you installed PyTorch (`conda`, `pip`, `libtorch`, source): `pip`
 - Python version: 3.11.15
 - CUDA version: cuda-toolkit --> 13.0.2
 - GPU models and configuration: H100 / A100-80GB / L4 / B200

## Additional context

The model is explicitly in FP16 (`.half()`)

micwill755:
## Summary
- Check `IExecutionContext::enqueueV3` / `execute_async_v3` and raise if TensorRT returns false, instead of continuing with uninitialized outputs.
- Covers the C++ runtime and the Python TRTEngine path (including CUDA-graph capture).
- Addresses the silent dummy/zero outputs in https://github.com/pytorch/TensorRT/issues/4481. The Cask conv tactic selected under `hardware_compatible` / `kAMPERE_PLUS` is a TensorRT bug and is not fixed here.

## Test plan
- [ ] Rebuild Torch-TRT and rerun the #4481 repro (A30 `hardware_compatible` engine on a different Ampere+ GPU). Confirm a Python exception is raised instead of `abs_mean=0` with no exception.
- [ ] Same-GPU compile + infer still succeeds (enqueueV3 returns true).
- [ ] Python runtime path (`use_python_runtime=True`) still runs a simple conv model.

### DIFF
--- a/core/runtime/execute_engine.cpp
+++ b/core/runtime/execute_engine.cpp
@@ -92,6 +92,15 @@ bool _validate_shapes(std::vector<at::Tensor> inputs, c10::intrusive_ptr<TRTEngi
   return false;
 }
 
+// TensorRT reports enqueue failure via a false return (and its logger), not an
+// exception. Swallowing that leaves output tensors uninitialized / stale.
+void enqueue_v3_checked(nvinfer1::IExecutionContext* ctx, const c10::cuda::CUDAStream& stream) {
+  TORCHTRT_CHECK(
+      ctx->enqueueV3(stream),
+      "IExecutionContext::enqueueV3 failed. The TensorRT engine did not execute; "
+      "output tensors are invalid.");
+}
+
 void setup_input_tensors(
     std::vector<at::Tensor> inputs,
     c10::intrusive_ptr<TRTEngine> compiled_engine,
@@ -471,13 +480,13 @@ std::vector<at::Tensor> execute_engine(std::vector<at::Tensor> inputs, c10::intr
         // Direct execution uses the caller buffers directly. On TRT-RTX with a
         // cuda_graph_strategy set, the engine captures/replays internally during
         // this enqueueV3 call.
-        ctx->enqueueV3(compiled_engine->engine_stream);
+        enqueue_v3_checked(ctx, compiled_engine->engine_stream);
       } else {
         if (need_cudagraphs_record) {
           // If cudagraphs needs to record a graph, capture the enqueueV3 call in a graph
           c10::cuda::CUDAStream recording_stream = compiled_engine->engine_stream;
           compiled_engine->cudagraph.capture_begin();
-          ctx->enqueueV3(recording_stream);
+          enqueue_v3_checked(ctx, recording_stream);
           compiled_engine->cudagraph.capture_end();
           compiled_engine->cudagraph.instantiate();
           if (compiled_engine->profile_execution) {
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
In the C++ cudagraph recording path, `enqueue_v3_checked()` can now throw after `capture_begin()` and before `capture_end()`. That leaves the stream in capture mode, so handling the enqueue failure can poison subsequent CUDA work. Abort/end capture via RAII or cleanup and add a capture-path failure regression.

PR #4664, opened 2026-09-02
fix(dynamo): decide MutableTorchTensorRTModule LIVE from weight snaps…

### LINKED ISSUE #4153: [Bug] MutableTorchTRTModule refit flag stuck at NEEDS_REFIT, never transitions to LIVE on B100/H100 with CUDA 13.x
## Bug Description

`test_resnet18_modify_attribute_no_refit` fails because `update_refit_condition()` does not transition the `RefitFlag` from `NEEDS_REFIT` (2) to `LIVE` (4). This reproduces across all CUDA 13.x versions on both B100-TS2 and H100, and surfaces through multiple dynamo test suites (backend, conversion, models, partitioning).

## Environment

- **GPUs:** B100-TS2, H100
- **Arch:** x86_64
- **CUDA:** 13.2.0 / 13.1.1 / 13.0.2
- **OS:** Ubuntu 24.04
- **cuDNN:** 8.9.7.29
- **Myelin:** 2.17.78+7
- **CASK:** 5.16.17+1
- **Python:** 3.12
- **Package:** qa_tar_py3.12

## Failing Test

```
FAILED runtime/test_mutable_torchtrt_module.py::test_resnet18_modify_attribute_no_refit
  AssertionError: <RefitFlag.NEEDS_REFIT: 2> != <RefitFlag.LIVE: 4> :
  update_refit_condition() failed to set the flag to LIVE.
```

## Reproducible Configurations

| GPU | CUDA | Test Suites Affected | Result |
|-----|------|---------------------|--------|
| B100-TS2/x86_64 | r13.0.2, r13.1.1, r13.2.0 | backend, conversion, models, partitioning | FAILED |
| H100/x86_64 | r13.1.1, r13.2.0 | backend, conversion, models, partitioning | FAILED |

All 20 logged test runs failed — no passing configuration observed.

## Steps to Reproduce

1. Run on B100-TS2 or H100 with CUDA 13.x and the environment above
2. Execute: `pytest runtime/test_mutable_torchtrt_module.py::test_resnet18_modify_attribute_no_refit`

## Expected Behavior

After modifying a model attribute without triggering a refit, `update_refit_condition()` should transition the `RefitFlag` from `NEEDS_REFIT` to `LIVE`, indicating the engine is still valid and does not require a full refit.

## Additional Context

The failure appears in all four dynamo test suite categories (backend, conversion, models, partitioning), suggesting the issue is in the core `MutableTorchTRTModule` runtime logic rather than in any specific converter or partitioning path.

cehongwang:
…hot (#4153)

update_refit_condition compared TRT vs PyTorch outputs to clear UNKNOWN, but H100/B100 engine numerics exceed ATOL/RTOL even when weights are unchanged. Snapshot weight values at compile/refit and treat an exact match as LIVE.

## Problem

`update_refit_condition()` cleared `UNKNOWN` by re-running the cached inputs through the
PyTorch model and comparing against the last TRT output. On H100/B100 the TRT↔eager gap
exceeds the default `ATOL/RTOL = 5e-3` even when the weights are untouched, so the
comparison failed and the flag stayed at its `NEEDS_REFIT` default. An unchanged module
refit on every call. Fixes #4153.

The output compare is also unsound as a LIVE decision independent of any tolerance: it
proves only that the weight change was invisible for *one cached input*, while LIVE applies
to all future inputs. A zero-initialised LoRA adapter (new keys, bit-identical outputs)
returned LIVE, leaving an engine with no LoRA layers in it.

## Design

Decide from the weights themselves, and identify them by digest rather than by a copy.

- At every compile and refit, store one SHA-256 per `state_dict` tensor in
  `_state_dict_digests`, computed as a hash over that tensor's 4 MiB chunk digests.
- `update_refit_condition` then reads: keys or shapes differ → `NEEDS_RECOMPILE`; every
  digest matches → `LIVE`; anything else → `NEEDS_REFIT`.
- Nothing in the decision touches the engine, so there is no tolerance to tune and no
  dependence on which GPU you are on.

Three helpers do the work:

| helper | job |
|---|---|
| `_tensor_chunks` | a tensor's bytes as 4 MiB `memoryview` slices; a CPU tensor is aliased, not copied |
| `_chunk_batches` | generator grouping tensors so at most 256 MiB of GPU→host copies are alive at once |
| `_digest_state_dict` | one `pool.map` per batch, folding chunk digests into one digest per tensor |

Two incidental fixes ride along: `run_info` is gone (it existed only to feed the output
compare, and pinned the last inputs and outputs alive), and the digest baseline is now
refreshed after a *successful* refit, not only in the fallback-recompile path.

## Why a digest

`hashlib` releases the GIL, which is the only reason hashing is affordable here. Serial
SHA-256 runs at 1.6 GiB/s against the 42.9 GiB/s of the `memcmp` it replaces; spread over
32 threads it reaches ~36 GiB/s, i.e. roughly parity. SHA-NI makes SHA-256 3x faster than
BLAKE2b on this hardware.

Chunking (rather than just parallelising over tensors) is what protects LLM- and
diffusion-shaped weights, where most bytes sit in a few huge tensors. On a synthetic
756 MiB / 10-tensor state dict whose largest tensor is 250 MiB:

| | serial | parallel per tensor | chunked |
|---|---|---|---|
| 756 MiB, 10 tensors | 561 ms | 160 ms | **22.6 ms** |
| vit_l_16, 1161 MiB, 296 tensors | 724 ms | 43.7 ms | **38.2 ms** |

Per-tensor parallelism collapses on the first row because the 250 MiB embedding is a single
job whose tail dominates. Benchmarking only vision models would hide this.

## Performance

A40, median of 5, weights in the CPU-offloaded steady state:

| | digest | previous value snapshot |
|---|---|---|
| resident state | 11.6 KiB / 22.9 KiB | 330 MiB / 1161 MiB |
| LIVE check | 17.6 ms / 38.6 ms | 9.9 ms / 34.5 ms |
| peak extra VRAM | 0.0 MiB | 0.0 MiB |
| saved module (`save()`) | 1109 MiB / 3884 MiB | 1439 MiB / 5045 MiB |

(vit_b_16 / vit_l_16.) Resident state drops by up to 51,920x for ~12% more latency on the
LIVE path. The saved module shrinks by exactly one copy of the weights, because the tensor
snapshot was being pickled into it.

Against the output compare this replaces, which cost 139–952 ms *and* required moving the
whole model onto the GPU to run the forward:

| model | digest | output compare |
|---|---|---|
| resnet18 | 10.4 ms | 139 ms |
| vit_l_16 | 38.2 ms | 952 ms |

With weights left GPU-resident (no `offload_module_to_cpu`), the check is 152 ms on
vit_l_16 with 0.0 MiB of host high-water growth and 0.0 MiB of extra VRAM; the
device-to-host transfer dominates, which the snapshot version already paid.

Digest throughput scales with core count in a way `memcmp` does not, so on a small CI
runner hashing with 4–8 threads the LIVE check is slower than the numbers above. Nothing
fails as a result.

## Alternatives measured and rejected

| alternative | why not |
|---|---|
| `(data_ptr, _version)` fingerprint | invisible to `w.data.copy_()`, `w.data += d`, and writes through `.numpy()`, all of which change every value. `.data.copy_()` is how peft fuses a LoRA adapter, so this silently serves a stale engine on the main LoRA flow |
| keep the value snapshot | 1x weight bytes of host RAM, plus the same again in every saved module |
| `zlib.crc32` | 32-bit digest; a collision means silently wrong numerics |
| drop the snapshot, always refit | `refit_gm()` is 9035 ms vs a 38 ms check — 1260x — and it lands on the path that is spurious by construction, since merely *reading* a Parameter marks the module UNKNOWN |
| mmap the snapshot to disk | keeps exactness but costs 15x on snapshot creation, at every compile and refit |

## Tests

`test_update_refit_condition_decides_from_weights` covers all four branches: unchanged
values → LIVE, changed values → NEEDS_REFIT, changed shape → NEEDS_RECOMPILE, and an
in-place `.data.copy_()` → NEEDS_REFIT. The last case asserts `data_ptr` and `_version` are
unchanged *first*, so it fails if the digest is ever swapped for a pointer or version check.

Full file passes (12 tests), including the `test_resnet18_modify_attribute_no_refit` case
from #4153. Byte identity is also stricter than `torch.equal`, which type-promotes, so a
dtype change that preserves values is now correctly not LIVE.

### DIFF
--- a/py/torch_tensorrt/dynamo/runtime/_MutableTorchTensorRTModule.py
+++ b/py/torch_tensorrt/dynamo/runtime/_MutableTorchTensorRTModule.py
@@ -14,7 +14,6 @@ from torch_tensorrt.dynamo import _defaults
 from torch_tensorrt.dynamo._compiler import compile as dynamo_compile
 from torch_tensorrt.dynamo._refit import refit_module_weights
 from torch_tensorrt.dynamo.utils import (
-    check_output_equal,
     deallocate_module,
     to_torch_device,
     to_torch_tensorrt_device,
@@ -153,8 +152,11 @@ class MutableTorchTensorRTModule(object):
         self.arg_dynamic_shapes: Optional[tuple[Any]] = None
         self.kwarg_dynamic_shapes: Optional[dict[Any, Any]] = None
         self.serializable_dynamic_shapes_dims: dict[str, tuple[str, int, int]] = {}
-        self.run_info: Optional[tuple[Any, ...]] = None
         self.state_dict_metadata: dict[str, torch.Size] = {}
+        # CPU snapshot of weight values at last compile/refit. Used by
+        # update_refit_condition so LIVE does not depend on TRT↔PyTorch
+        # numerical agreement (H100/B100 gaps exceed ATOL/RTOL; see #4153).
+        self._state_dict_weights: dict[str, torch.Tensor] = {}
         self._store_state_dict_metadata()
         self.enable_weight_streaming = (
             kwargs["enable_weight_streaming"]
@@ -237,8 +239,25 @@ class MutableTorchTensorRTModule(object):
         return total_dynamic_shape
 
     def _store_state_dict_metadata(self) -> None:
+        self.state_dict_metadata = {}
+        self._state_dict_weights = {}
         for k, v in self.original_model.state_dict().items():
             self.state_dict_metadata[k] = v.shape
+            # Detach + CPU copy so later comparisons do not alias live params.
+            self._state_dict_weights[k] = v.detach().to("cpu", copy=True)
+
+    def _state_dict_values_match(self) -> bool:
+        # Older pickled modules may lack the snapshot; fall back to output compare.
+        weights = getattr(self, "_state_dict_weights", None)
+        if not weights:
+            return False
+        sd = self.original_model.state_dict()
+        if sd.keys() != weights.keys():
+            return False
+        for k, v in sd.items():
+            if not torch.equal(v.detach().cpu(), weights[k]):
+                return False
+        return True
 
     def load_state_dict(
         self, state_dict: Dict[str, Any], strict: bool = True, assign: bool = False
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
After `_state_dict_values_match()` detects that weights changed, this can still mark the module `LIVE` because one cached input happens to produce the same output. A weight update can be inactive for that sample but affect a later input, leaving the TensorRT engine with stale weights. Should an exact state-dict value mismatch always require refit unless the changed weights can be proven unused?

cehongwang:
Yeah I deleted the dummy run entirely

narendasan:
@cehongwang can you explain a bit about the design? Do we hash at each refit/recompile? Is there a perf cost?

PR #4665, opened 2026-09-02
fix(dynamo): preserve zero-length partition boundaries

apbose:
Follow-up to #4631. Stops construct_dynamic_input from clamping zero-capable symbolic extents to one. Affected TensorRT submodules remain in PyTorch, while require_full_compilation fails clearly. Adds an all-false nonzero runtime regression that fails on the parent when setInputShape rejects [0, 1].

### DIFF
--- a/py/torch_tensorrt/dynamo/partitioning/common.py
+++ b/py/torch_tensorrt/dynamo/partitioning/common.py
@@ -22,6 +22,10 @@ logger = logging.getLogger(__name__)
 ProfileSourceBounds = List[Dict[str, Dict[str, int]]]
 
 
+class ZeroLengthInputError(ValueError):
+    """Raised when a TensorRT engine input may contain a zero-length dimension."""
+
+
 def _build_submodule_profiles(
     input_shape: torch.Size,
     union_min: Sequence[int],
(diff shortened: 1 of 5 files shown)


py/torch_tensorrt/dynamo/partitioning/common.py
```diff
@@ -120,15 +127,12 @@ def construct_dynamic_input(
...
                 min_bound = int(min_max_opt["min"])
                 if min_bound < 1:
-                    logger.warning(
-                        "Dynamic input %s (shape: %s) has lower bound %d for dim %d. "
-                        "TensorRT profiles require dimensions >= 1; clamping it to 1.",
-                        name,
-                        input_shape,
-                        min_bound,
-                        d,
+                    raise ZeroLengthInputError(
```
narendasan, line 130: Why do we need to raise? is there a way we can just catch the issue ahead of time?

apbose:
Closing in favor of preserving zero-valued TensorRT profile minima directly in #4631. Standard TensorRT 11.1 retained and executed a min=0 profile successfully, consistent with the TensorRT 11.2.1 IOptimizationProfile contract.

PR #4667, opened 2026-09-02
feat(executorch): let a TensorRT engine write the caller's KV cache in place

Conarnar:
# Description

Let a TensorRT engine write a caller's KV cache in place, instead of routing every update through a staging copy and a copy-back.

An engine with aliased I/O writes its aliased output *through* the aliased input's pointer, so running the engine over the buffer already is the update. Nothing in the ExecuTorch pipeline knows that, so the buffer makes a full round trip on every execution: `PropagateDevicePass` wraps each delegate input in `et_copy._h2d_copy`, so the engine writes a per-call staging copy rather than the caller's buffer, and the aliased output is threaded back out for ExecuTorch to copy in afterwards. On a 30-layer MoE decoder at prefill 2048 / max_len 2560 that is **576.7 MB in and 577.2 MB out every decode step**, at 6.5–7.0 GB/s pageable.

`export(..., zero_copy_kv=True)` opts in. The feature needs two changes at different points in the pipeline, and the export flag performs only the first:

- **before partitioning**, the mutation is repointed at the buffer placeholder, so the aliased output has no user, dies to DCE, and leaves the partition entirely;
- **as a `to_out_var_pass`**, installed by `zero_copy_backend_config()`, the staging is removed so the engine is handed the caller's buffer.

Applying only the first would leave the engine writing a discarded staging copy with the copy-back already gone — the buffer would silently never update. Neither pass is public on its own. `torch_tensorrt.save(output_format="executorch", zero_copy_kv=True)` owns both steps and needs only the one flag: it installs `zero_copy_backend_config()` itself, so the two are alternatives rather than a pair — handing `save()` a config built with that function applies the un-staging twice, and the second run finds the buffers already wired straight to their delegates and changes nothing.

**Nothing enforces that pairing on the two-call path, and that is a stated limitation rather than an oversight.** `to_executorch()` belongs to ExecuTorch's `EdgeProgramManager` and finalizes whatever config it is handed; being present when it runs would mean assigning over the bound method of the manager `export()` returns, which is not an extension point ExecuTorch offers and which `transform()` and `to_backend()` — each of which constructs a *fresh* manager — would silently drop. So a caller who finalizes a zero-copy export owes two calls that nothing checks: `zero_copy_backend_config()` as the config, and `check_zero_copy_kv()` on the program that comes back. Both are documented at every point a caller meets them — `export()`'s docstring, `check_zero_copy_kv()`'s, and the user guide — and both are done for you by `save(..., zero_copy_kv=True)`, which owns finalization. Asking for `zero_copy_kv=True` on a model with no aliased buffer mutation warns and names that check, which is what `save()` refuses on.

`zero_copy_backend_config()` preserves every field of the caller's config, with two exceptions. It *reads* `enable_non_cpu_memory_planning`, off the config it returns and when the pass runs, and refuses every buffer marked for in-place update under host-only planning, which puts every tensor in the one host arena whatever a spec says — that `.pte` exports clean and then fails its first `execute()` on the alias-target guard. And it *refuses* `propagate_device_config.skip_h2d_for_method_inputs` by name rather than clearing it behind the caller's back: that option demands one user per placeholder it un-stages, a rewired cache always has two, so carrying it would return a config that cannot finalize at all.

Nothing is serialized to mark a `.pte` zero-copy: the runtime reads elision off the argument count, short by exactly the engine's aliased-output count. A `.pte` written before this change still takes the threaded-output branch.

## The discriminator, which everything else depends on

An aliased KV mutation and a #4459 copy-back mutation are **identical in the graph**: both are `BUFFER_MUTATION target=X <- getitem(engine, i)` where `X` is also an engine input. They are told apart only by reading the engine's own serialized `aliased_io`. Both ways of getting it wrong are silent — rewiring a copy-back loses a real update, and leaving an aliased cache staged has the engine write scratch that is discarded.

A method may hold both kinds at once, and that combination is supported rather than refused. It is measured on a real engine: a model with `index_copy_` KV caches plus a ring-shifted `conv_state` exports with the caches on their own placeholders and `conv_state` still bound to a delegate output, in one method, with no code change beyond the `aliased_io` read. #4459's `_trt_no_kv_alias` makes that signal cleaner still, by excluding non-KV mutable buffers from aliasing at the converter.

## Measured

A 30-layer MoE decoder, 26B total / 4B active, real weights, int4 MoE + bf16, full 262144 vocab, prefill 2048 / max_len 2560, hybrid TensorRT+CUDA on one 80GB A100, TensorRT 11.1.0.106:

| | staged | zero-copy |
|---|---|---|
| KV staging memcpy per decode step | 171.66 ms | **0.099 ms** |
| decode | 422.5 ms/token | **28.67 ms/token** |
| prefill | 1,934 tok/s | **4,125 tok/s** |

The 1,730× collapse in staging is the mechanism; the rest follows. Host-side gaps fall with it, 114.39 → 4.81 ms, and compute is untouched — the int4 MoE GEMV runs at 298.7 µs against 299.5 before. Generated token ids are **byte-identical** to the staged run.

## Verified

- **334** Python tests in the `-m unit` lane at the tip, all passing; collecting the same directory with no marker gives **361**, all passing. The real-engine cases run on GPU rather than skipping, including the one that finalizes a zero-copy export with ExecuTorch's defaults and shows the resulting program being refused by `check_zero_copy_kv()` — finalization itself succeeds, which is the limitation above, pinned.
- C++ blob-header tests **11 → 39**, all passing: repeated names in every position they can repeat in, blank names, missing and misspelled keys, extents that wrap or overrun, a `device_id` past the int maximum and one exactly at it, and good blobs a stricter scan must still accept. Every behavioural clause was mutated away in turn and the test that bites it named. Two mutants of the metadata extent survive and cannot be killed — rewriting it in addition form, and deleting it altogether: 0 verdict differences over 1,946,720 header-field combinations, against 3,876 and 904 for deleting either of the other two extent checks. The comment there says why it is kept anyway.
- `black` 26.3.1 and `mypy` 1.15.0, both CI-pinned, clean at every commit; `ruff` clean over the changed files and against `origin/main`.
- Real-engine coverage for two previously uncovered multi-delegate shapes: the caches and copy-back split across two TensorRT delegates, and a TensorRT delegate beside an ExecuTorch CUDA delegate. Both bind the right buffer to the right value.
- An earlier revision of this runtime change compiled clean against **TensorRT 11.2.1.2** under CMake (gnu++17) and bazel (C++20), no new warnings against `main`, and `executorch_backend_tests` passed **4/4** on a real device. Both C++ sources have changed since that build. `TensorRTBlobHeader.cpp` has been recompiled standalone (`-Wall -Wextra`, clean) with its 39 gtests passing; `TensorRTBackend.cpp` has not been rebuilt, and the 4/4 suite never covered it — no gtest target depends on `//cpp:tensorrt_executorch_backend`.
- The reference-runner CI lane ran end to end and exited 0; its zero-copy iteration asserts the `aliased_io` discriminator, and its no-sync branch was observed taking the skip path on all three `execute()` calls, against an unguarded mode in the same binary that synchronizes every time. That run predates later changes to the check, the script and the library; the lane has not been re-run since, and it needs the GPU CI to say so.

## Merge note for #4572

#4572 added the coalesced `.pte` to `verify-executorch-reference-runner.sh` as a third positional argument. This stack made the trailing arguments variadic so it can hand the staged and the zero-copy KV `.pte` to one run. Those cannot coexist, so the coalesced model is now named — `--coalesced=PATH` — and the KV models stay variadic. Both sides' behaviour is preserved; only the encoding of #4572's argument changed. The script has one caller, `executorch-test-linux.yml`, changed in the same commit as the script. (#4687 folded `executorch-build-linux.yml` — which called the script with a single positional model — into that workflow and deleted it, so the second caller this stack was written against no longer exists.)

## Known gaps

- **The whole `export()` + `to_executorch()` path is unenforced.** `save(zero_copy_kv=True)` runs `check_zero_copy_kv()` itself and is the only guarded entry point. On the two-call path both halves are the caller's: finalizing without `zero_copy_backend_config()` writes a `.pte` whose caches never update — wrong output, not a crash — and nothing raises. An earlier revision wrapped the returned manager's `to_executorch`; that was dropped, because it sets an instance attribute on a class that is ExecuTorch's, which is not a sanctioned extension point, and because `transform()` and `to_backend()` each build a fresh manager, so the ordinary `export(...).to_backend(p).to_executorch(cfg)` had no guard at all. A guard that holds for one spelling of the call and silently not the next reads as absolute while being best-effort. The obligation is written instead in `export()`'s docstring, `check_zero_copy_kv()`'s, and the user-guide warning box, and the decode example spells both calls out.
- **An arena no planner recorded is refused, and the refusal has a cost.** Only `apply_algo` writes `non_const_buffer_device`, and only with `enable_non_cpu_memory_planning=True`; without it `MethodMeta::memory_planned_buffer_device` answers CPU, so a runner honouring it — this repo's does — backs the cache with host memory and fails the alias-target guard on every `execute()`. `check_zero_copy_kv()` therefore reads absence as a refusal of its own, under a message that names the remedy, rather than as "cannot tell". What that costs is the caller who pairs a device-aware planner of their own with a custom runner that ignores the record and backs every arena on the device: they are correct today, and both entry points now refuse them with no opt-out. A planner brought for the shared prefill/decode cache the guide recommends one for has to go through `apply_algo` to leave the record behind.
- **The write-back pass crosses the mutation map; this works around it at the Edge boundary rather than fixing it.** A zero-copy cache beside an ordinary copy-back buffer used to finalize with the specs crossed, measured as `{'copy__default': 'k_cache', 'b_k_cache': 'v_cache', 'b_v_cache': 'conv_state'}`. The cause is `insert_write_back_for_buffers_pass`, upstream's and reproducible in stock ExecuTorch with `run_reinplace_pass=True` and no TensorRT; it is going upstream as a bug. `export()` reorders the mutations so the ones that still get a copy come first, on the Edge program it returns, splitting on upstream's own `_inplace_lineage` — an order set earlier is discarded, `to_edge_transform_and_lower` re-deriving the signature (measured both ways). The reorder runs on every method of every `export()`, not only the ones zero-copy rewired, because the crossing is not zero-copy's: a plain `nn.Module` writing one buffer from another buffer and a second from a user input comes out of stock `to_edge().to_executorch()` crossed with no TensorRT anywhere (measured), and the reorder returns 0 and touches nothing where the order already holds. One residual: a mutation `run_reinplace_pass` turns in place is out of reach, that pass running inside `to_executorch` after the reorder (measured: crossed either way). The `.pte` is byte-identical; the pairing decides the finalized `graph_signature` and the eager `buffers_to_mutate` path.
- **Neither zero-copy check can separate an exact swap between two KV delegates.** Each stamped delegate is counted against its own compile spec, before planning and again after, which stops one delegate holding both caches while the other reads a discarded staging copy. Two delegates each holding one marked buffer, each the other's, satisfy both counts: the mark records only that *some* engine writes the buffer in place, and the spec lists binding names rather than buffers. Nothing here produces such a method; the multi-delegate test asserts exactly one engine carries the aliasing.
- **The arity predicate has no unit test.** It lives inside `execute()` and needs a real engine; its automated coverage is the CI reference-runner check. The duplicate-`aliased_io` and duplicate-binding-name rejections, the blank-name refusal and the extent bounds all happen when the blob is parsed, and are unit-tested there.
- **Every refusal the blob parser adds is indistinguishable at load time.** `TensorRTBlobHeader::parse` returns a bool and emits no message of its own, so a repeated or blank name, an overrunning extent and a bad magic all surface as `TensorRTBackend::init: failed to parse TensorRT blob`, `Error::InvalidProgram`. Deliberate — over thirty refusal sites share the diagnostic — but a reader with a bad `.pte` and no source cannot tell which defect it has. Giving any one refusal its own message changes runtime behaviour, so it stands.
- **Multi-method zero-copy is not covered end to end.** `_apply_zero_copy_kv` loops over methods and `check_zero_copy_kv()` walks every method, and the real-engine tests do run the check on a genuinely finalized program — but a single-method one. The multi-method behaviour of both rests on fabricated program shapes; no two-method `.pte` was built.
- **A config derived from `zero_copy_backend_config()`'s result with `dataclasses.replace` is not read by the pass it carries.** The planning flag is copied by value, the pass by reference and still bound to the config the function returned; call the function again on the derived config to bind one. `check_zero_copy_kv()` refuses what the mistake produces for any method holding a host tensor to give the shared arena away (measured, pinned by a test), so the cost is a later refusal rather than a wrong `.pte` — wherever that check is run, which is `save()` always and the two-call path only if the caller makes the call. The residual needs two things at once: a method with no host tensor to give the shared arena away, in a program that does record its arena devices — a planner that records nothing is refused rather than passed over. `_host_planned_arenas` documents the first half.
- **A `.pte` produced through `save(..., zero_copy_kv=True)` is not among the CI artefacts.** The reference-runner lane exercises zero-copy through the two-call path, the one with two ways to get it wrong, but the single-flag path most callers will use is covered only by unit tests.
- **`kv_cache_decode_check`'s caller stream is blocking**, so its PASS line proves the guarded and unguarded modes take different branches rather than proving the unguarded one races. Making it non-blocking would turn a deterministic check into one that can surface a race, worth doing on purpose rather than in passing.
- **The Python blob reader collapses a duplicate the C++ parser refuses.** `serialization.py` keys `aliased_io` on the output binding name, so a repeat keeps the last, and it reads a repeated `io_bindings` name into two list entries rather than refusing; the C++ parser refuses both blobs. Only the C++ side is what a runtime loads, so the divergence costs nothing today, but the two do not agree.
- **The method-wide elided-name list on `export()`'s compile spec is inert.** Only the presence of the key is read: `TensorRTPartitioner` drops the spec and re-derives the names per engine. The value is written so the spec reads like the per-partition one, and changing which names it lists changes no delegate.
- Engine metadata is resolved more than once per engine during partitioning. Each read is cheap under `metadata_only=True`; folding them into the existing resolved-record handoff is a follow-up, deliberately not bundled here.

## If `AliasKind.USER` ever gets a producer

`AliasKind.USER` is a placeholder nothing emits, so every `aliased_io` entry a `.pte` carries today is a `kv_cache_update` on a lifted buffer, and neither half of zero-copy reads the kind. Export elides per binding name, the runtime by subtracting the count of *all* `aliased_io` entries, so a `"user"` alias on a plain input beside an elided one is a shape neither the binding check nor `check_zero_copy_kv()` objects to and the runtime cannot execute. `preprocess` refuses that engine outright rather than letting it reach a `.pte`, which is what a producer of the kind would meet first. Supporting the mix instead means giving the runtime the elided names to read rather than an arity to subtract; refusing now leaves that open.

## Type of change

# Checklist:

### DIFF
--- a/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
+++ b/cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
@@ -440,6 +440,12 @@ Result<DelegateHandle*> TensorRTBackend::init(
   // Map each aliased output binding to the index of the input it aliases so
   // execute() can bind it to that input's device pointer (in-place).
   // Non-aliased models have an empty header.aliased_io -> all -1, unchanged path.
+  // The parser has already refused a blob claiming one output twice, so the count
+  // built below is one per distinct output binding, which is what execute()
+  // subtracts on. It has also refused a blob repeating a binding name, so the
+  // first-match scans below reach the only slot carrying that name -- otherwise
+  // the alias would be recorded on the first slot while execute() re-bound the
+  // same TensorRT name for the later one, replacing the caller's buffer address.
   handle->output_aliased_input_idx.assign(handle->num_outputs, -1);
   handle->input_is_alias_target.assign(handle->num_inputs, false);
   for (const auto& ab : header.aliased_io) {
@@ -551,8 +557,10 @@ Result<DelegateHandle*> TensorRTBackend::init(
 // their addresses; no separate output allocation is required.
 //
 // Args layout (mirroring the Python exporter):
-//   args[0 .. num_inputs-1]             – input EValues
-//   args[num_inputs .. num_inputs+num_outputs-1] – output EValues
+//   args[0 .. num_inputs-1]                               -- input EValues
+//   args[num_inputs .. num_inputs+num_delegate_outputs-1] -- output EValues
+// num_delegate_outputs is num_outputs, less the aliased outputs when zero-copy KV
+// elided them from the delegate; see the arity branch at the top of execute().
 // ---------------------------------------------------------------------------
 Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle* handle, Span<EValue*> args) const {
   (void)context;
@@ -561,17 +569,70 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
 
   const size_t num_inputs = engine->num_inputs;
   const size_t num_outputs = engine->num_outputs;
-  // Caller-owned KV: every input is a delegate arg, and each aliased output is
-  // threaded as a delegate output arg (the caller-owned mutable buffer's mutation
-  // slot), so all engine bindings map 1:1 to delegate args.
-  const size_t num_delegate_outputs = num_outputs;
+  // Caller-owned KV comes in two shapes. Either each aliased output is threaded as
+  // a delegate output arg (the caller-owned mutable buffer's mutation slot), so all
+  // engine bindings map 1:1 to delegate args; or -- zero-copy KV -- the aliased
+  // outputs are elided from the delegate entirely, because the engine's in-place
+  // write through the aliased input already IS the buffer update. A .pte written
+  // before zero-copy existed still takes the first branch, and export will not emit
+  // the shorter arity unless zero-copy was asked for, so a short argument list
+  // cannot instead mean "the aliased outputs were never declared".
   const size_t num_delegate_inputs = num_inputs;
-  if (args.size() < num_delegate_inputs + num_delegate_outputs) {
+  const size_t num_aliased_outputs = engine->num_aliased_outputs;
+  // The blob parser refuses a second aliased_io entry for an output already
+  // claimed, and init refuses an entry whose output is not one of the recorded
+  // output bindings, so this holds for any header that reached here. It is
+  // checked anyway because the subtractions below are unsigned: on a header
+  // built some other way one of them wraps, and the length check then accepts
+  // an argument count it should not. Exactly one such count, not any: with the
+  // aliased count above the outputs, the only count that can set the elided
+  // flag is inputs plus outputs minus aliased, and once it is set
+  // num_delegate_outputs is the wrapped outputs-minus-aliased, which adding the
+  // inputs back wraps round to that same count -- so the length check passes
+  // and the call goes on to index past the end of args. Every other count below
+  // inputs plus outputs is still rejected. The two subtractions never wrap in
+  // the same call either: with the aliased count above the outputs but not
+  // above inputs plus outputs only the second wraps, and above both the first
+  // wraps to a value no argument count can match, so the flag is never set and
+  // the second never runs. A duplicate entry is the case this does not catch:
+  // it inflates the count while staying within num_outputs, passes both checks,
+  // and still indexes one past the end of args; the parser's refusal is what
+  // stops that.
+  if (num_aliased_outputs > num_outputs) {
     ET_LOG(
         Error,
-        "TensorRTBackend::execute: expected at least %zu args, got %zu",
-        num_delegate_inputs + num_delegate_outputs,
-        args.size());
+        "TensorRTBackend::execute: %zu aliased output(s) recorded for %zu output binding(s)",
+        num_aliased_outputs,
+        num_outputs);
+    return Error::InvalidProgram;
+  }
+  const bool aliased_outputs_elided =
+      num_aliased_outputs > 0 && args.size() == num_delegate_inputs + num_outputs - num_aliased_outputs;
+  const size_t num_delegate_outputs = num_outputs - (aliased_outputs_elided ? num_aliased_outputs : 0);
+  if (args.size() < num_delegate_inputs + num_delegate_outputs) {
+    // With aliased outputs there are two right answers and only one of them can
+    // ever be num_delegate_outputs here: the elided flag is set only by an
+    // argument count that already fits, so a program that arrives short is
+    // always measured against the threaded count. Reporting that alone tells a
+    // zero-copy .pte to supply the longer list, which is the shape that consumes
+    // its real outputs as mutation slots -- so both counts are named, and the
+    // aliasing that is the reason for the two.
+    if (num_aliased_outputs > 0) {
+      ET_LOG(
+          Error,
+          "TensorRTBackend::execute: expected %zu args with the engine's %zu aliased output(s) threaded, "
+          "or %zu with them elided (zero-copy KV), got %zu",
+          num_delegate_inputs + num_outputs,
+          num_aliased_outputs,
+          num_delegate_inputs + num_outputs - num_aliased_outputs,
+          args.size());
+    } else {
+      ET_LOG(
+          Error,
+          "TensorRTBackend::execute: expected at least %zu args, got %zu",
+          num_delegate_inputs + num_delegate_outputs,
+          args.size());
+    }
     return Error::InvalidArgument;
   }
 
@@ -615,13 +676,18 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   // on the shared exec_ctx. Wait for it before reconfiguring the context below:
   // TensorRT forbids mutating a context while one of its enqueues is in flight, and
   // setInputShape/setTensorAddress run on the host, so this must be a host-side wait.
+  // The flag is cleared only once the wait has actually succeeded: it is the
+  // destructor's only reason to wait, and a failed synchronize is precisely the
+  // case where the enqueue may still be running. Clearing it first would hand
+  // the destructor a handle it believes is idle and let it free the staging
+  // buffers and reset exec_ctx underneath a live enqueue.
   if (engine->inflight_pending) {
     cuda_err = cudaEventSynchronize(engine->inflight_event);
-    engine->inflight_pending = false;
     if (cuda_err != cudaSuccess) {
       ET_LOG(Error, "TensorRTBackend::execute: cudaEventSynchronize failed: %s", cudaGetErrorString(cuda_err));
       return Error::InvalidProgram;
     }
+    engine->inflight_pending = false;
   }
   const auto caller_stream = ::executorch::extension::cuda::getCallerStream();
   const bool caller_stream_set = caller_stream.has_value();
@@ -795,9 +861,15 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
         ET_LOG(Error, "TensorRTBackend::execute: setTensorAddress failed for aliased output '%s'", name.c_str());
         return Error::InvalidState;
       }
-      // The aliased output IS a delegate output arg (the caller-owned mutable
-      // buffer's mutation slot). Consume it and record a reflect so ExecuTorch's
-      // write-back copy_ sees the engine's in-place update.
+      // Elided: setTensorAddress above pointed this output binding at the caller's
+      // buffer, so the engine writes the buffer itself; nothing to reflect into.
+      if (aliased_outputs_elided) {
+        continue;
+      }
+
+      // Otherwise the aliased output IS a delegate output arg (the caller-owned
+      // mutable buffer's mutation slot). Consume it and record a reflect so
+      // ExecuTorch's write-back copy_ sees the engine's in-place update.
       const size_t arg_i = arg_idx++;
       EValue* out_arg = args[arg_i];
       TORCHTRT_ET_CHECK_NOT_NULL(
@@ -945,10 +1017,25 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
   // next execute() and the destructor wait before reusing/freeing exec_ctx. The D2H
   // copies live in the must_sync branch: an output staged to host always sets
   // output_staged_to_host, so outputs_needing_copy is empty on the skip path.
-  // An aliased reflect enqueues the engine's in-place update into the delegate
-  // output EValue on `stream`; ExecuTorch's buffer-mutation copy_ reads that EValue
-  // after execute() returns, so the reflect must complete first. A model with
-  // aliased outputs therefore always syncs here.
+  // A non-elided aliased reflect enqueues the engine's in-place update into the
+  // delegate output EValue on `stream`; ExecuTorch's buffer-mutation copy_ reads
+  // that EValue after execute() returns, so the reflect must complete first, and a
+  // model that threads its aliased outputs as delegate output args syncs here.
+  // Under zero-copy KV skipping the sync is correct, because the aliased buffer
+  // stays device-resident and its next reader is the following engine execute() --
+  // on the same `stream`, provided the runner honours the single-shared-stream
+  // contract every coalesced .pte already depends on. A host reader that
+  // inspected it immediately after execute() returns would see stale data unless
+  // it synchronized `stream` itself; ExecuTorch's KV path never does such a read.
+  // Eliding the reflects drops aliased_reflect_pending, so an execute() with a
+  // caller stream and no host staging no longer syncs at all. What orders the
+  // engine's *other* outputs is then that same contract: the host consumer
+  // ExecuTorch inserts for a device delegate output is an et_copy::_d2h_copy,
+  // whose kernel issues its copy on getCallerStream() -- this stream -- and then
+  // synchronizes it; a device consumer of that output is another delegate,
+  // enqueued on the same stream. That covers the write-back of a copy-back buffer
+  // sitting beside the zero-copy caches. With no caller stream set that kernel
+  // falls back to a blocking cudaMemcpy, and must_sync is true here anyway.
   const bool aliased_reflect_pending = !aliased_reflects.empty();
   const bool must_sync =
       output_staged_to_host || input_staged_from_host || aliased_reflect_pending || !caller_stream_set;
@@ -972,11 +1059,16 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
       }
     }
     cuda_err = cudaStreamSynchronize(stream);
-    engine->inflight_pending = false;
     if (cuda_err != cudaSuccess) {
       ET_LOG(Error, "TensorRTBackend::execute: cudaStreamSynchronize failed: %s", cudaGetErrorString(cuda_err));
       return Error::InvalidProgram;
     }
+    // Same shape as the wait at the top of execute(), and for the same reason,
+    // though nothing here can be armed yet: this branch never records the
+    // event, and the wait at the top has already cleared any flag a previous
+    // call left. Arming happens as the last statement of the other branch, so
+    // the only writer that can make this clear anything is one added later.
+    engine->inflight_pending = false;
     if (copy_err != Error::Ok) {
       return copy_err;
     }
--- a/docsrc/user_guide/runtime_performance/saving_models.rst
+++ b/docsrc/user_guide/runtime_performance/saving_models.rst
@@ -356,6 +356,153 @@ points but does not by itself give them shared mutable state.
     Neither case raises an error or a warning, so treat every shared payload as
     read-only.
 
+.. _executorch_zero_copy_kv:
+
+**Zero-copy KV cache**
+
+When a TensorRT engine has aliased I/O -- a KV cache it updates through an
+aliased binding -- running the engine over the cache already is the update.
+ExecuTorch does not know that, so by default it pays for the update twice per
+execution: it hands the delegate an ``_h2d_copy`` staging copy of the buffer
+instead of the buffer itself, then copies the engine's aliased output back into
+the buffer afterwards. For a KV cache both copies are cache-sized, per token.
+
+``zero_copy_kv=True`` removes them, so the engine writes the caller's buffer
+directly. Through the two-step ``export`` + ``to_executorch`` path it takes two
+calls, one at each end of the Edge boundary:
+
+.. code-block:: python
+
+    from torch_tensorrt.executorch import (
+        check_zero_copy_kv,
+        export,
+        zero_copy_backend_config,
+    )
+
+    edge = export(
+        {"prefill": prefill_program, "decode": decode_program},
+        zero_copy_kv=True,
+    )
+
+    # The argument is optional. Pass your own ExecutorchBackendConfig and
+    # zero_copy_backend_config composes onto it; every other field (memory
+    # planning, passes) is preserved.
+    program = edge.to_executorch(zero_copy_backend_config())
+
+    # Neither call is enforced -- see the warning below.
+    check_zero_copy_kv(program)
+
+Two fields of that config are not merely carried. The engine writes the cache
+wherever memory planning put it, so ``enable_non_cpu_memory_planning=False`` --
+which plans every tensor into a single host arena whatever device its
+``TensorSpec`` asks for -- cannot be combined with zero-copy KV:
+``to_executorch`` raises instead of writing a ``.pte`` whose every ``execute()``
+fails on a host pointer. That refusal reads the config object
+``zero_copy_backend_config`` returned, so two ways of setting the field are
+outside it. Building a new config out of the returned one with
+``dataclasses.replace`` copies the field by value and the pass by reference, so
+the pass goes on reading the config it was built for: call
+``zero_copy_backend_config`` again on the derived config. And a
+``memory_planning_pass`` of your own that does not already have an attribute of
+that name never receives the field at all -- ``to_executorch`` assigns it onto
+the planner rather than passing it -- so where the caches land is that planner's
+own business. Neither has to be left to the runtime to discover:
+``check_zero_copy_kv`` reads the arena planning actually chose -- on this path
+it is yours to call, as the warning below says. That planner does have to record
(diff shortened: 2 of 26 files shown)


shoumikhin (commented):
Some notes inline.

These are about lines the change does not touch, so they could not be attached to one.


In `examples/executorch_reference_runner/kv_cache_decode_check.cpp`, around line 166:

The reference runner check covers the argument count branch, but not the new part that skips the stream sync. The KV decode check never installs a caller stream guard, so the caller stream is absent and the code always takes the synchronizing branch no matter what. The zero copy no sync path never runs in CI.

That is the riskier half. A wrong argument count fails loudly and every time. A missing sync fails sometimes, with numbers that look fine.

Could the KV decode check wrap its decode loop in a caller stream guard on its own stream, the way the main runner already does, and run each model both ways? That would make the branch real coverage instead of a comment.

One related point on wording. Before this change a model with aliased outputs always synchronized at the end of execute, because the reflect forced it. The commit message calls the shared stream contract a source of wrong values. The docs in this repo already say that situation is a race that can surface as wrong results or an illegal memory access. Worth saying it at that strength where the contract is written down.

In `cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp`, around line 518:

In execute, the aliased output count now decides the delegate arity, but init increments it once per aliased_io entry rather than once per aliased output binding. A blob that lists the same entry twice passes every check in init, because both copies resolve to the same output and the same engine alias, and the count still ends up at two.

Two things then go wrong. If the count is above the number of outputs actually flagged, the elision test can be true while the loops consume more arguments than the span holds, and the span index is not bounds checked. If the count is above the number of outputs, the subtraction underflows, the sum in the length check wraps, and the check passes for any arity.

Your own exporter cannot produce a duplicate, so this needs a crafted or corrupt file. It is worth two lines anyway, because before this change the counter only fed a log. Increment only when the slot was still unset, and reject a repeated entry the way a duplicate weight streaming spec is already rejected. Refusing a count larger than the number of output bindings makes the underflow unreachable too.

py/torch_tensorrt/executorch/_zero_copy.py
```diff
@@ -0,0 +1,611 @@
...
+    elided_output_names: List[str] = []
+    output_node = graph_module.graph.output_node()
+    output_args = list(output_node.args[0])
+    output_specs = list(signature.output_specs)
+    for spec_index, mutation in mutations.items():
+        # Marked on the node rather than read back off the engine because the
+        # un-staging pass runs after lowering, where the engine's aliased_io is no
+        # longer reachable from the graph: it has become an opaque delegate blob.
+        mutation.placeholder.meta["_torch_tensorrt_aliased_buffer"] = True
+        output_args[spec_index] = mutation.placeholder
```
shoumikhin, line 359: Repointing the mutation at the buffer placeholder makes ExecuTorch's write-back pass skip the copy for that buffer, which is the point. But that pass then patches the signature with a counter over mutation specs while indexing a list that only holds the copy nodes it created. A mutation with no copy shifts every later one by one.

For a method holding both kinds this crosses them. The KV cache mutation ends up naming the copy that writes the conv state, and the conv state mutation ends up naming the KV placeholder. A KV only method is fine because there is no copy at all, and the same model without zero copy is fine because both get one, so nothing you have measured so far would show it.

The file itself looks like it still runs, because copy carries its own destination and the emitter records outputs by position. What is wrong is the finalized program's mutation map, which the exir execution path and the ETRecord both read.

This was checked against the ExecuTorch version the workflow installs. Moving the rewired mutations behind the un-rewired ones in the output specs and args gives the correct pairing.

docsrc/user_guide/runtime_performance/saving_models.rst
```diff
@@ -404,20 +477,31 @@ illegal memory access.
 The runtime does not impose a shared stream across delegates, so it is the
 **runner's responsibility** to run all delegates on one CUDA stream. Create a
 single stream and, for the duration of execution, direct every backend to use it
-(each backend exposes a caller-stream hook). All GPU work is then enqueued in order
-and every cross-boundary dependency is satisfied, while execution stays
-asynchronous.
+(each backend exposes a caller-stream hook: scope both
+``torch_tensorrt::executorch_backend::CudaStreamGuard`` and
```
shoumikhin: The single stream section now says to scope the Torch-TensorRT CUDA stream guard as well as the ExecuTorch caller stream guard. That first class is gone. The backend README says it was removed on purpose, with no deprecated alias, so this code would not compile. The reason given is also backwards: one caller stream guard already covers every CUDA delegate, because the shared CUDA extension means all of them read the same caller-stream storage, which that README states directly. The old wording, "each backend exposes a caller-stream hook", was correct. Please name only the ExecuTorch guard here.

py/torch_tensorrt/executorch/_zero_copy.py
```diff
@@ -0,0 +1,611 @@
...
+
+    ``config`` is your own configuration -- every field is preserved, and a
+    ``to_out_var_pass`` you already set runs after the un-staging. Omit it to
+    start from ExecuTorch's defaults.
+
+    .. warning::
+        Finalizing a ``zero_copy_kv=True`` program *without* this config does
+        not raise. The engine writes a per-call staging copy that is then
+        discarded and the buffer never updates, which for a KV cache is wrong
+        output rather than a crash. Nothing downstream can detect the omission,
```
shoumikhin: The warning says nothing downstream can detect the omission, so pairing the two calls is the caller's responsibility. The example in this same change detects it: it walks the finalized program for marked buffers that are still staged and refuses to write the file. So the check is possible, and it is already written twice, in the example and in the tests. The single-call path in save holds that finalized program right before it writes, and does not run it, which is the one path that promised the mistake was impossible. Please move that check into the library and call it there, and drop the "nothing downstream can detect" sentence.

py/torch_tensorrt/executorch/_zero_copy.py
```diff
@@ -0,0 +1,611 @@
...
+        if 0 <= output_index < len(names):
+            elided_output_names.append(names[output_index])
+
+    output_node.args = (tuple(output_args),)
+    # Dropping every output of an engine would leave a delegate with no outputs.
+    # Nothing downstream reports that shape: the runtime infers elision from a
+    # single argument count, which a zero-output delegate satisfies, and a
+    # delegate nothing reads is a pure node that a later graph-wide dead-code
+    # elimination can erase, taking the computation with it. Stop here instead.
+    # The eliminate_dead_code() below does not erase this engine node: unlike
```
shoumikhin: In the rewiring pass, the comment above the every output elided check says an execute_engine node is impure to FX and so survives dead code elimination with no users. I do not think that is true. The operator is declared with no write annotation, and the Python fallback registers it with an empty mutates_args, so FX sees a pure op and will erase it once nothing reads it. Nothing in this repository adds it to FX's side effect list either.

That also leaves a small hole in the check above. It only fires when every user of the engine is an elided getitem. If one non elided getitem is still there but is itself dead, dead code elimination erases that getitem first, then the engine, and the computation goes with it.

Either count live users only, or drop the sentence and say plainly that the guard is what keeps the engine alive.

py/torch_tensorrt/executorch/_zero_copy.py
Conarnar: Fixed by ordering the rewired mutations last, though the mechanism is not quite the one described.

The list upstream indexes is not only the copies — `_insert_copy` ends with
`buffer_output_nodes.extend(user_output_nodes)`, so it returns the `copy_` nodes it created
followed by every output it copied nothing for, in graph order. That matters for the fix rather
than being a detail: under the "only the copies" reading a mutation spec past the number of copies
would index off the end, so reordering would not be what saves it. What happens instead is that
ordering the rewired mutations last gives them spec positions `C..M-1` *and* puts them at the head
of that tail in the same order, so the copy-producing specs index the copy prefix one-for-one and
the rewired specs index the tail entries that are those same outputs.

On stock ExecuTorch with `run_reinplace_pass=True` and two mutated buffers where only the first is
reinplaced, the first buffer's mutation names the `copy_` that writes the *second* and the second's
names the `index_put_` that wrote the *first* — a swap, not an index error, which only the tail
explains.

That configuration is also the fix's limit: `reinplace_pass` runs after the ordering and can swap
them again. There is no hook between the two, so it is scoped and stated rather than fixed.

docsrc/user_guide/runtime_performance/saving_models.rst
Conarnar: Fixed.

py/torch_tensorrt/executorch/_zero_copy.py
Conarnar: Fixed. `check_zero_copy_kv()` is public, `save()` runs it before writing, and the example calls it
rather than carrying its own copy.

It walks every method, not just `forward` — the first version took `exported_program()` with no
name, which raised `KeyError` on a multi-method program and skipped the other methods where a
`forward` did exist. "Nothing marked" raises only when no method has marks, matching the warning
`_apply_zero_copy_kv` already emits; a staged buffer raises per-method and names it.

py/torch_tensorrt/executorch/_zero_copy.py
Conarnar: The premise does not hold: `EffectHolder._set_default_effect` gives it `EffectType.ORDERED`, so
`is_impure()` is true and DCE keeps it with no users. The comment was right.

The hole you inferred from it is real, though, and is fixed — the check now counts live users, so a
surviving-but-dead non-elided getitem no longer lets the engine and its computation be erased.

Conarnar:
> In `examples/executorch_reference_runner/kv_cache_decode_check.cpp`, around line 166:
> 
> The reference runner check covers the argument count branch, but not the new part that skips the stream sync. The KV decode check never installs a caller stream guard, so the caller stream is absent and the code always takes the synchronizing branch no matter what. The zero copy no sync path never runs in CI.
> 
> That is the riskier half. A wrong argument count fails loudly and every time. A missing sync fails sometimes, with numbers that look fine.
> 
> Could the KV decode check wrap its decode loop in a caller stream guard on its own stream, the way the main runner already does, and run each model both ways? That would make the branch real coverage instead of a comment.
> 
> One related point on wording. Before this change a model with aliased outputs always synchronized at the end of execute, because the reflect forced it. The commit message calls the shared stream contract a source of wrong values. The docs in this repo already say that situation is a race that can surface as wrong results or an illegal memory access. Worth saying it at that strength where the contract is written down.


Fixed, and instrumented to confirm it: the guarded mode takes the skip path on all three of its
`execute()` calls, against the unguarded mode in the same binary, which synchronizes every time.

Two limits worth stating. The staged `.pte` still takes the synchronizing branch in both modes, so
the new coverage is zero-copy only. And the correctness assertion cannot detect a missing sync here
— `cudaStreamCreate` returns a blocking stream, so the legacy default-stream copy waits anyway.
Deleting the runner's own `cudaStreamSynchronize` still passes 40/40 and `cudaStreamQuery` reads
COMPLETE 200/200. So it covers the branch, not the ordering, and the comment says so.

 
> In `cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp`, around line 518:
> 
> In execute, the aliased output count now decides the delegate arity, but init increments it once per aliased_io entry rather than once per aliased output binding. A blob that lists the same entry twice passes every check in init, because both copies resolve to the same output and the same engine alias, and the count still ends up at two.
> 
> Two things then go wrong. If the count is above the number of outputs actually flagged, the elision test can be true while the loops consume more arguments than the span holds, and the span index is not bounds checked. If the count is above the number of outputs, the subtraction underflows, the sum in the length check wraps, and the check passes for any arity.
> 
> Your own exporter cannot produce a duplicate, so this needs a crafted or corrupt file. It is worth two lines anyway, because before this change the counter only fed a log. Increment only when the slot was still unset, and reject a repeated entry the way a duplicate weight streaming spec is already rejected. Refusing a count larger than the number of output bindings makes the underflow unreachable too.

Fixed at parse rather than in `init`, which makes it unit-testable without a GPU.

Reproducing it turned up a second path the `num_aliased_outputs <= num_outputs` bound does not
cover. With 2 inputs, 3 outputs and one output duplicated, the count reaches 2 while staying within
`num_outputs`, elision is believed, the length check `3 >= 3` passes, and the output loop then reads
`args[3]` on a 3-element span. Confirmed with the argument array against a `PROT_NONE` guard page:
`si_addr == args.data() + args.size()`. No underflow anywhere, so the bound alone would not have
caught it.

It needs no crafted graph — the same engine, argument list and arity come from a correct
two-aliased-output program that returns `Error::Ok`; only one output-name string in the header
differs. The bound stays for the wrap case, and its comment now says that is all it covers.

py/torch_tensorrt/_compile.py
```diff
@@ -879,6 +909,15 @@ def save(
...
         )
+        # For the same reason, the one refusal zero-copy makes on the config
+        # alone. _save_as_executorch reaches it only through
+        # zero_copy_backend_config, which it calls after export() has partitioned
+        # the graph and built every engine -- a whole compile before the caller
+        # is told the combination is not allowed.
+        if executorch_zero_copy_kv and executorch_backend_config is not None:
+            from torch_tensorrt.executorch._zero_copy import _refuse_skip_h2d
+
+            _refuse_skip_h2d(executorch_backend_config)
```
cehongwang, line 923: Is there a reason we make this optional rather than default to zero copy? Is there a case we want to remain the copying?

Conarnar: Zero copy requires two parts `rewire_aliased_mutations_to_buffers` and `unstage_aliased_buffers_pass `. Having the just the first part will result in the buffers silently not mutating.

If using `save`, it would be possible to have both parts on by default. However, when using `export` + `to_edge_transform_and_lower`, that would not be possible since we don't own `to_edge_transform_and_lower`.

cpp/src/torch_tensorrt/executorch/TensorRTBackend.cpp
```diff
@@ -561,17 +569,70 @@ Error TensorRTBackend::execute(BackendExecutionContext& context, DelegateHandle*
...
         Error,
-        "TensorRTBackend::execute: expected at least %zu args, got %zu",
-        num_delegate_inputs + num_delegate_outputs,
-        args.size());
+        "TensorRTBackend::execute: %zu aliased output(s) recorded for %zu output binding(s)",
+        num_aliased_outputs,
+        num_outputs);
+    return Error::InvalidProgram;
+  }
+  const bool aliased_outputs_elided =
```
cehongwang, line 609: **A zero-copy `.pte` carries no record that it is one.** `execute()` re-derives the contract from `args.size()`, which leaves every guard this PR adds on the export side of a wall the artifact never crosses.

```cpp
const bool aliased_outputs_elided =
    num_aliased_outputs > 0 && args.size() == num_delegate_inputs + num_outputs - num_aliased_outputs;
```

The AOT side is thorough — the partitioner derives elided names per engine (`py/torch_tensorrt/executorch/partitioner.py:204`), `preprocess` refuses a partial elision, `unstage_aliased_buffers_pass` cross-checks each stamped delegate (`py/torch_tensorrt/executorch/_zero_copy.py:955`), `check_zero_copy_kv` re-reads the finalized program (`_zero_copy.py:1112`). All four read in-memory state (`_torch_tensorrt_aliased_buffer`, `TensorSpec.mem_id`) that is never serialized, so two things follow:

1. **The artifact is unverifiable.** `check_zero_copy_kv` is public API but only takes a live program. Given a `.pte` there is no way to answer "is this zero-copy, or did it silently fall back?" — and silent fallback is exactly this feature's failure mode.
2. **The safety argument is long and conditional.** The comment above this branch spends twenty lines on unsigned wraparound. I believe it holds, but only while the parser refuses duplicate `aliased_io` entries, the `num_aliased_outputs > num_outputs` check at line 601 runs first, and both subtractions stay `size_t`. Nothing signals if that lapses. The comment names the case it can't close ("a duplicate entry ... still indexes one past the end of `args`"), and `backend.py:318` concedes the same: "the runtime reads elision off a single argument count, so it cannot tell them apart either."

The format already solves this class of problem: `TR02` exists, and this PR taught the parser to hold it to its promise. Extending it lets `init()` assert the arity instead of `execute()` inferring it, and the wraparound reasoning becomes unnecessary:

```diff
 constexpr char TENSORRT_MAGIC_ALIASED_IO[4] = {'T', 'R', '0', '2'};
+// TR03: metadata carries "elided_outputs", the aliased output bindings the
+// delegate does not carry. Parsed into TensorRTBlobHeader::elided_outputs.
+constexpr char TENSORRT_MAGIC_ELIDED_IO[4] = {'T', 'R', '0', '3'};

-  const bool aliased_outputs_elided =
-      num_aliased_outputs > 0 && args.size() == num_delegate_inputs + num_outputs - num_aliased_outputs;
+  // Recorded by the exporter, not inferred: see TR03 in TensorRTBlobHeader.
+  const bool aliased_outputs_elided = engine->aliased_outputs_elided;
+  const size_t expected = num_delegate_inputs + num_outputs -
+      (aliased_outputs_elided ? num_aliased_outputs : 0);
+  if (args.size() != expected) { /* one error, no two-answer branch */ }
```

No zero-copy `.pte` exists yet, so the format change is free today; once they ship, both shapes are readable forever. Deferring is defensible — I'd just like it recorded as a decision with a follow-up issue rather than left implicit.

Either way, one small ask: have `init()` reject `num_aliased_outputs == num_outputs`. Line 601 only rejects strictly greater, so a delegate with zero real outputs reaches the binding loop. The exporter refuses that shape at `py/torch_tensorrt/executorch/backend.py:341`, but that only covers blobs this exporter produced.

cehongwang (approved):
Fix the merge conflicts and it is good

PR #4681, opened 2026-09-04
fix: bind a scalar engine output at rank 0

### LINKED ISSUE #4677: 🐛 [Bug] the execute_engine meta kernel reports rank 0 for a sym_size output the engine returns as rank 1
##  Bug Description

The `execute_engine` meta kernel reports a shape the engine does not produce, for any output
whose FX value is a `SymInt`. The meta kernel says rank 0; the engine returns rank 1. Nothing
compares the two accounts, so the disagreement is silent at compile time and surfaces far
away — in a downstream engine's binding check, or in an AOTInductor buffer sized from the
meta kernel.

The engine's output metadata is recorded from the FX graph *before* TensorRT builds anything.
`py/torch_tensorrt/dynamo/conversion/_conversion.py :: interpret_module_to_result`:

```python
symbolic_shape_expressions = extract_symbolic_shape_expressions(module)   # first statement
...
interpreter_result = interpreter.run()                                    # the engine
```

`py/torch_tensorrt/dynamo/conversion/_symbolic_shape_capture.py :: extract_symbolic_shape_expressions`
reads each output's shape from the node's `meta["val"]`, so a `SymInt` output is recorded as a
rank-0 scalar:

```python
elif isinstance(out_val, (torch.SymInt, torch.SymFloat, int, float, bool)):
    ...
    output_info.append({"shape_exprs": [], "dtype": scalar_dtype, "is_scalar": True})
```

`py/torch_tensorrt/dynamo/conversion/impl/shape.py :: shape`
— the `aten.sym_size.int` converter — gathers with an index built by `get_trt_tensor`, whose
`min_rank` defaults to 1, so the gather result is rank 1:

```python
dim_tensor = get_trt_tensor(ctx, dim, name + "_dim")        # rank 1, not rank 0
gather_layer = ctx.net.add_gather(input_shape, dim_tensor, axis=0)
...
return input_shape                                          # shape [1]
```

That record is the only thing the meta kernel consults
(`dynamo/runtime/meta_ops/register_meta_ops.py :: _apply_symbolic_shape_expressions` builds the
output fake tensors straight from `info["shape_exprs"]`), so every trace, every downstream
consumer, and every AOTInductor buffer is built against an incorrect rank. 

## To Reproduce

```
```

**repro.py**

```
import torch
import torch_tensorrt
from torch._subclasses.fake_tensor import FakeTensorMode
from torch.fx.experimental.symbolic_shapes import ShapeEnv

ROWS = 16


class SymIntEngineOutput(torch.nn.Module):
    """A symbolic dimension read inside the engine and consumed by Torch outside it."""

        y = x * 2.0
        n = y.shape[0]
        return y.sum() + torch.arange(0, n, device=x.device).float().sum()


def build() -> torch.nn.Module:
    """Compiles the model with ``arange`` pinned to Torch so the SymInt leaves the engine."""
    model = SymIntEngineOutput().eval().cuda()
    x = torch.randn((ROWS,), device="cuda")
    rows = torch.export.Dim("rows", min=2, max=64)
    exported = torch.export.export(model, (x,), dynamic_shapes={"x": {0: rows}})
        exported,
        inputs=(x,),
        min_block_size=1,
        # The only scaffolding here: `arange` does have a converter, so pinning it to Torch
        # is what places the engine boundary on the SymInt. The defect is what the boundary
        # records, not the pinning.
        torch_executed_ops={"torch.ops.aten.arange.start_step"},
    )
    return dict(gm.named_children())["_run_on_acc_0"]


def main() -> None:
    mod = build()
    x = torch.randn((ROWS,), device="cuda")

    recorded = [dict(info) for info in mod.symbolic_shape_expressions["outputs"]]
    real = [tuple(t.shape) for t in mod(x)]
    with FakeTensorMode(shape_env=ShapeEnv(), allow_non_fake_inputs=True):
        meta = [
            tuple(t.shape) for t in mod(torch.empty((ROWS,), dtype=torch.float32, device="cuda"))
        ]

    print("engine, as the engine describes itself:")
    print("    " + str(mod.engine).strip().replace("\n", "\n    "))
    print("\nrecorded at compile time (_symbolic_shape_capture):")
    for i, info in enumerate(recorded):
        print(f"    output{i}: {info}")
    print(f"\nmeta kernel reports : {meta}")
    print(f"engine actually returns: {real}")

    reproduced = meta != real
    if reproduced:
        mismatched = [i for i, (m, r) in enumerate(zip(meta, real)) if m != r]
        print(
            f"Output(s) {mismatched}: the meta kernel promises a shape the engine does not "
            "produce, so every consumer downstream is traced against the wrong rank."
        )
    else:
        print("The meta kernel and the engine agree on every output.")


if __name__ == "__main__":
    main()
```

**output**

```
engine, as the engine describes itself:
    Torch-TensorRT TensorRT Engine:
      Name: _run_on_acc_0_engine
      Inputs: [
        id: 0
          name: x
          shape: [16]
          dtype: Float
      ]
      Outputs: [
        id: 0
          name: output0
          shape: [1]
          dtype: Long
        id: 1
          name: output1
          shape: []
          dtype: Float
      ]
      Device: Device(ID: 0, Name: NVIDIA L4, SM Capability: 8.9, Type: GPU)
      Hardware Compatibility: Disabled
      Target Platform: linux_x86_64
      Resource Allocation Strategy: Static
      Multi-Device Engine: 0

recorded at compile time (_symbolic_shape_capture):
    output0: {'shape_exprs': [], 'dtype': torch.int64, 'is_scalar': True}
    output1: {'shape_exprs': [], 'dtype': torch.float32}

meta kernel reports : [(), ()]
engine actually returns: [(1,), ()]

reproduced: True
Output(s) [0]: the meta kernel promises a shape the engine does not produce, so every consumer downstream is traced against the wrong rank.
```

The engine's own `__str__` already disagrees with the record: `output0 shape: [1]` against
`shape_exprs: []`.

### What it does downstream

Sizing a real buffer from the meta kernel is the problem. This model's
second return value is a Python `int`; run through `torch_tensorrt.dynamo.export` and
`aoti_compile_and_package`, the packaged artifact returns a 1-element tensor instead, and
nothing is raised:

```python
class ScalarSecondOutput(torch.nn.Module):
    def forward(self, x):
        y = x * 2.0
        return y.sum(), y.shape[0]
```

```
eager        : (tensor(-0.6317, device='cuda:0'), 16)
recorded _run_on_acc_0: [{'shape_exprs': [], 'dtype': torch.float32}, {'shape_exprs': [], 'dtype': torch.int64, 'is_scalar': True}]
aoti package : (tensor(-0.6317, device='cuda:0'), tensor([16], device='cuda:0'))
```

That a `SymInt` output comes back as a tensor at all is #4614; the point here is the *shape*
— `(1,)` where even the record the artifact was built from says `()`.

On a larger model (14 engines, 21 affected outputs) the same divergence surfaced instead as a
copy into a buffer AOTInductor had sized from the metadata:

```
output with shape [1] doesn't match the broadcast shape [1, 1]
Error: aoti_torch_copy_(buf669, buf657, 0) API call failed
```

Nothing rejects the extra axis in eager, where the operators broadcast over it and the whole
graph runs.

## Expected behavior

The rank the meta kernel reports should be the rank the engine produces. Either side can be
made to move:

1. Make `impl/shape.py :: shape` return the rank-0 value its FX node is typed as. TensorRT 11
   binds rank-0 tensors directly, so nothing else has to change. Verified: appending a reshape
   to `()` onto that converter makes the meta kernel and the engine agree
   (`stock: agree=False` → `patched: agree=True`), with the rest of the pipeline untouched.
2. Record the engine's own binding rank in `_symbolic_shape_capture` after `interpreter.run()`,
   rather than the FX graph's. This is the broader fix — it also covers any other converter
   that promotes a scalar, and it removes the class of defect rather than one instance of it.

Separately: whichever way it is fixed, it would help if the two accounts were compared. The
engine knows its own binding ranks and dtypes at build time; a check against
`symbolic_shape_expressions` before serialization would have named this at compile time
instead of leaving it to be found in an AOTInductor buffer copy.

## Environment

SrivastavaKshitij:
# Description

`extract_symbolic_shape_expressions` runs before the engine is built and records an output
whose FX `meta["val"]` is a `SymInt`/`SymFloat`/scalar as rank 0:

```python
elif isinstance(out_val, (torch.SymInt, torch.SymFloat, int, float, bool)):
    output_info.append({"shape_exprs": [], "dtype": scalar_dtype, "is_scalar": True})
```

TensorRT builds something else. `impl/shape.py :: shape` -- the `aten.sym_size.int`
converter -- gathers with an index built by `get_trt_tensor`, whose `min_rank` defaults to 1,
so the engine binds `[1]`:

```
meta kernel reports    : [(), ()]
engine actually returns: [(), (1,)]
```

That record is the only account of the engine the `execute_engine` meta kernel consults, so
every later trace and every AOTInductor buffer sized from it carries a rank the engine does
not produce, with nothing comparing the two.

This reshapes those outputs back to rank 0 in `TRTInterpreter.output`, after the dtype cast
and before `mark_output`, which makes the two accounts agree without changing the model's
scalar semantics. Only an output the FX graph calls a scalar and that the network produced as
shape `[1]` is reshaped; any other rank is left alone and logged. Aliased outputs are skipped
for the same reason the dtype cast skips them, since a shuffle layer's output no longer shares
storage with the source input. TensorRT 11 binds rank-0 tensors directly, so the runtime is
unchanged.

# Issues

Closes #4677

## Type of change

# Testing

Three cases in `tests/py/dynamo/runtime/test_scalar_engine_outputs.py`: the meta kernel's
output shapes equal the engine's, the engine declares rank 0 for a binding the metadata
records as a scalar, and -- as a control -- a genuine shape-`[1]` tensor output keeps rank 1,
so over-squeezing does not pass. The first two fail without this change. There were no
existing tests covering `symbolic_shape_expressions`.

Run in `nvcr.io/nvidia/pytorch:26.07-py3` (TensorRT 11.1.0.106) with this patch applied to
the container's installed copy: `2 failed, 1 passed` before, `3 passed` after. End to end
through AOTInductor, for a model returning `(y.sum(), y.shape[0])`:

```
eager  : (tensor(-4.5675), 16)
before : (tensor(-4.5675), tensor([16]))   # shape (1,)
after  : (tensor(-4.5675), tensor(16))     # shape ()
```

`conversion/test_sym_size.py`, `conversion/test_sym_numel_aten.py`,
`conversion/test_sym_not_aten.py`, and `models/test_dyn_models.py` are unaffected: 23 passed
before, 23 passed after.

# Checklist:

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
+++ b/py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
@@ -751,6 +751,48 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
         layer.name = f"Cast output {output_name} from {output.dtype} to {output_dtype}"
         return layer.get_output(0)
 
+    def _scalar_output_positions(self) -> Set[int]:
+        """Positions in the FX output node whose value is a scalar, not a tensor.
+
+        These are exactly the positions ``extract_symbolic_shape_expressions``
+        records with ``is_scalar``, so the two stay in step.
+        """
+        if self._cur_node is None:
+            return set()
+
+        fx_outputs = self._cur_node.args[0]
+        if not isinstance(fx_outputs, (tuple, list)):
+            fx_outputs = (fx_outputs,)
+
+        positions = set()
+        for i, fx_output in enumerate(fx_outputs):
+            val = (
+                fx_output.meta.get("val")
+                if isinstance(fx_output, torch.fx.Node)
+                else fx_output
+            )
+            if isinstance(val, (torch.SymInt, torch.SymFloat, int, float, bool)):
+                positions.add(i)
+        return positions
+
+    def _reshape_output_to_scalar(
+        self, output: trt.ITensor, output_name: str
+    ) -> trt.ITensor:
+        if len(output.shape) == 0:
+            return output
+
+        if tuple(output.shape) != (1,):
+            _LOGGER.warning(
+                f"Output {output_name} is a scalar in the graph but the network "
+                f"produced shape {tuple(output.shape)}, leaving its rank alone."
+            )
+            return output
+
+        layer = self.ctx.net.add_shuffle(output)
+        layer.reshape_dims = ()
+        layer.name = f"Reshape output {output_name} to a scalar"
+        return layer.get_output(0)
+
     def output(self, target: str, args: Any, kwargs: Any) -> List[Any]:
         assert len(args) == 1
         if isinstance(args[0], tuple):
@@ -798,6 +840,8 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
             raise RuntimeError(
                 f"Specified output dtypes ({len(self.output_dtypes)}) differ from number of outputs ({len(outputs)})"
             )
+
+        scalar_output_positions = self._scalar_output_positions()
 
         marked_outputs_ids = []
         for i, output in enumerate(outputs):
@@ -854,6 +898,15 @@ class TRTInterpreter(torch.fx.Interpreter):  # type: ignore[misc]
                     output_dtype.to(trt.DataType, use_default=True),
                     name,
                 )
+
+            # TensorRT has no scalar type, so a converter emits a graph scalar as
+            # a rank-1 tensor of length 1. The engine's shape metadata is read off
+            # the FX graph, where the value is rank 0, so reshape the output back
+            # to rank 0 or the meta kernel and the engine report different ranks
+            # for the same binding. Skipped for aliased outputs for the same
+            # reason as the cast above.
+            if alias_info is None and i in scalar_output_positions:
+                output = self._reshape_output_to_scalar(output, name)
 
             output.name = name
             outputs = outputs[:i] + (output,) + outputs[i + 1 :]
--- a/tests/py/dynamo/runtime/test_scalar_engine_outputs.py
+++ b/tests/py/dynamo/runtime/test_scalar_engine_outputs.py
@@ -1,0 +1,103 @@
+import unittest
+
+import torch
+import torch_tensorrt
+from torch._subclasses.fake_tensor import FakeTensorMode
+from torch.fx.experimental.symbolic_shapes import ShapeEnv
+from torch.testing._internal.common_utils import TestCase, run_tests
+from torch_tensorrt.dynamo.runtime import TorchTensorRTModule
+
+import tensorrt as trt  # isort: skip  # imported after torch_tensorrt for RTX alias
+
+ROWS = 16
+
+
+class ScalarSecondOutput(torch.nn.Module):
+    """Returns a tensor and a symbolic int, so the engine has a scalar output."""
+
+    def forward(self, x):
+        y = x * 2.0
+        return y.sum(), y.shape[0]
+
+
+class UnitTensorSecondOutput(torch.nn.Module):
+    """Returns a tensor and a genuine shape-[1] tensor, whose rank must be left alone."""
+
+    def forward(self, x):
+        y = x * 2.0
+        return y.sum(), y[:1]
+
+
+@unittest.skipIf(
+    not torch_tensorrt.ENABLED_FEATURES.torch_tensorrt_runtime,
+    "Torch-TensorRT runtime is not available",
+)
+class TestScalarEngineOutputs(TestCase):
+    def _compile_to_single_engine(self, model):
+        self.addCleanup(torch._dynamo.reset)
+        inputs = (torch.randn((ROWS,), device="cuda"),)
+        rows = torch.export.Dim("rows", min=2, max=4 * ROWS)
+        exported = torch.export.export(
+            model.eval().cuda(), inputs, dynamic_shapes={"x": {0: rows}}
+        )
+        gm = torch_tensorrt.dynamo.compile(
+            exported,
+            inputs=inputs,
+            min_block_size=1,
+            pass_through_build_failures=True,
+            use_python_runtime=False,
+        )
+        engines = [
+            module for module in gm.modules() if isinstance(module, TorchTensorRTModule)
+        ]
+        self.assertEqual(len(engines), 1)
+        return engines[0]
+
+    def _meta_and_real_output_shapes(self, engine_module):
+        real = [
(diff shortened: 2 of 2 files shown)


SrivastavaKshitij:
@apbose @narendasan for viz

apbose:
Lgtm. Ran the CI for testing.

SrivastavaKshitij:
@apbose : what's the next steps to move forward on this PR?

apbose (approved):

tests/py/dynamo/runtime/test_scalar_engine_outputs.py
```diff
@@ -0,0 +1,103 @@
...
+        engines = [
+            module for module in gm.modules() if isinstance(module, TorchTensorRTModule)
+        ]
+        self.assertEqual(len(engines), 1)
+        return engines[0]
+
+    def _meta_and_real_output_shapes(self, engine_module):
+        real = [
+            tuple(output.shape)
+            for output in engine_module(torch.randn((ROWS,), device="cuda"))
```
apbose, line 59: after https://github.com/pytorch/TensorRT/pull/4628 this would fail mostly.
can you update the test to 
```
torch.ops.tensorrt.execute_engine(
   [input_tensor],
  engine_module.engine,
  )
```
since after that engine_module would return .item

py/torch_tensorrt/dynamo/conversion/_TRTInterpreter.py
```diff
@@ -825,6 +825,48 @@ def _cast_output_dtype(
...
+        self, output: trt.ITensor, output_name: str
+    ) -> trt.ITensor:
+        if len(output.shape) == 0:
+            return output
+
+        if tuple(output.shape) != (1,):
+            _LOGGER.warning(
+                f"Output {output_name} is a scalar in the graph but the network "
+                f"produced shape {tuple(output.shape)}, leaving its rank alone."
+            )
```
apbose, line 862: lets keep this as error? this case would mostly mean error in upstream converter, so error I assume would be better

apbose (changes requested):

apbose:
I rebased this PR. Have some review comments

SrivastavaKshitij:
made changes as requested

PR #4682, opened 2026-09-04
fix: name the engine and the binding in execute_engine failures

### LINKED ISSUE #4678: 🐛 [Bug] execute_engine input checks name neither the engine nor the binding that failed
##  Bug Description

Every input check in `execute_engine` reports *what* disagreed and never *which engine* it
disagreed in. On a graph with more than one engine that is not enough to locate the problem,
and a refused engine writes none of its outputs, so what usually reaches the caller is an
unrelated error raised by whatever reads one of them.

`core/runtime/execute_engine.cpp :: setup_input_tensors` validates every input against the
binding it feeds and throws on the first disagreement:

```cpp
TORCHTRT_CHECK(
    inputs[i].dtype() == binding.expected_type,
    "Expected input tensors to have type " << binding.expected_type
        << ", found type " << inputs[i].dtype());
...
TORCHTRT_CHECK(ctx->setInputShape(name.c_str(), dims), "Error while setting the input shape");
```

Neither message names the engine. The shape message additionally names neither the binding nor
the shape that was rejected — TensorRT itself reports both, and that report is discarded and
replaced with a fixed string:

```
IExecutionContext::setInputShape: Error Code 3: API Usage Error (Parameter check failed,
  condition: engineDims.d[i] == dims.d[i]. Static dimension mismatch while setting input shape
  for relu. Set dimensions are [8,17]. Expected dimensions are [8,16].)
```

`torch.ops.tensorrt.set_logging_level(4)` does not close the gap for the exception: the engine
name appears only in the `LOG_DEBUG` line that runs *before* the checks, so it is absent from
any log captured at the default level and absent from the exception in every case.

On a 14-engine model this made every runtime failure a search. These are all the diagnostics
a caller got:

```
Expected input tensors to have type Int, found type long int
Error while setting the input shape
CUDA out of memory. Tried to allocate 15.62 GiB.
```

and, when the refused engine's unwritten outputs were read further on:

```
a Tensor with 0 elements cannot be converted to Scalar
```

## To Reproduce

```
```

**repro.py**

```
import torch
import torch.nn as nn
import torch_tensorrt

ROWS, COLS = 8, 16


class TwoEngines(nn.Module):
    """A graph split in two, so 'which engine?' is a real question."""

        a = x * 2.0
        b = torch.relu(a)  # pinned to Torch below, which splits the graph in two
        return torch.sigmoid(b + 1.0)


def engine_modules(gm: torch.fx.GraphModule) -> list[tuple[str, torch.nn.Module]]:
    """Returns the (name, module) pairs of ``gm``'s TensorRT submodules, in graph order."""
    return [
        (name, mod)
        for name, mod in gm.named_children()
        if getattr(mod, "serialized_engine", None) is not None
    ]


def message_for(call) -> str:
    """Runs ``call`` and returns the exception message it raises."""
    try:
        call()
    except Exception as exc:  # pylint: disable=broad-except
        return str(exc)
    raise RuntimeError("expected the call to fail, but it succeeded")


def main() -> None:
    model = TwoEngines().eval().cuda()
    x = torch.randn(ROWS, COLS, device="cuda")
    exported = torch.export.export(model, (x,))
        exported,
        inputs=(x,),
        min_block_size=1,
        # The only scaffolding here: it places an engine boundary so the graph holds two
        # engines. The diagnostics being reported are the same with one engine.
        torch_executed_ops={"torch.ops.aten.relu.default"},
    )
    engines = engine_modules(gm)
    print(f"engines in the graph: {[name for name, _ in engines]}")
    assert len(engines) > 1, "expected the partitioner to produce more than one engine"

    name, mod = engines[-1]
    mod.setup_engine()
    # The binding names are reachable only through the engine's __str__; the torchbind class
    # exposes no accessor for a binding's name, dtype or shape.
    binding_names = [b for b in str(mod.engine).split() if b.startswith("relu")]
    print(f"failing engine: {name}, input bindings: {binding_names}")

    cases = {
        "dtype check": lambda: torch.ops.tensorrt.execute_engine(
            [torch.zeros(ROWS, COLS, dtype=torch.int32, device="cuda")], mod.engine
        ),
        "shape check": lambda: torch.ops.tensorrt.execute_engine(
            [torch.zeros(ROWS, COLS + 1, device="cuda")], mod.engine
        ),
    }

    reproduced = False
    for label, call in cases.items():
        msg = message_for(call).strip()
        names_engine = name in msg
        names_binding = any(b in msg for b in binding_names)
        print(f"\n{label} raised:\n    " + msg.replace("\n", "\n    "))
        print(f"  names the engine ({name}): {names_engine}")
        print(f"  names a binding {binding_names}: {names_binding}")
        reproduced = reproduced or not names_engine

    if reproduced:
        print(
            "At least one check reports the disagreement without saying which engine it "
            "came from."
        )


if __name__ == "__main__":
    main()
```

**output**

```
engines in the graph: ['_run_on_acc_0', '_run_on_acc_2']
failing engine: _run_on_acc_2, input bindings: ['relu']

dtype check raised:
    [Error thrown at core/runtime/execute_engine.cpp:110] Expected inputs[i].dtype() == expected_type to be true but got false
    Expected input tensors to have type Float, found type int
  names the engine (_run_on_acc_2): False
  names a binding ['relu']: False
ERROR: [Torch-TensorRT] - IExecutionContext::setInputShape: Error Code 3: API Usage Error (Parameter check failed, condition: engineDims.d[i] == dims.d[i]. Static dimension mismatch while setting input shape for relu. Set dimensions are [8,17]. Expected dimensions are [8,16]. In setInputShape at /_src/runtime/api/executionContext.cpp:2295)

shape check raised:
    [Error thrown at core/runtime/execute_engine.cpp:149] Expected compiled_engine->exec_ctx->setInputShape(name.c_str(), dims) to be true but got false
    Error while setting the input shape
  names the engine (_run_on_acc_2): False
  names a binding ['relu']: False

reproduced: True
At least one check reports the disagreement without saying which engine it came from.
```

## Expected behavior

Each check should name the engine and the binding, and the shape check should carry the shape
it rejected alongside the one that was declared. Concretely:

```
[_run_on_acc_2_engine] input 1 ("relu"): expected type Float, found type Int
[_run_on_acc_2_engine] input 1 ("relu"): rejected shape [8, 17], engine declares [8, 16]
```

Two related things would help alongside it:

- The out-of-memory paths in the same function are worth catching for the same reason — a bare
  `CUDA out of memory. Tried to allocate 15.62 GiB.` says nothing about which engine asked for
  how much.
- A binding's name, dtype and shape are currently reachable only by parsing the engine's
  `__str__`; the torchbind class exposes no accessor for them. An accessor would let callers
  diagnose this themselves.

## Environment

SrivastavaKshitij:
# Description

Every runtime check in `execute_engine` says what went wrong and none says where. On a model
with 14 engines that is all the caller gets — and because a refused engine writes none of its
outputs, what usually surfaces first is an unrelated node reading one of them.

Diagnostics only, no behavior change. Each check now carries the engine name and the binding
index and name, the input-shape check reports the shape TensorRT refused next to the one the
engine declares, and both allocations are caught and re-thrown naming the engine and the size
it asked for, appending to the allocator's message rather than replacing it. `TORCHTRT_CHECK`
only evaluates its message on failure, so the success path is unchanged.

Before:

```
Expected input tensors to have device cuda, found device cpu
Expected input tensors to have type Float, found type int
Error while setting the input shape
Error while setting the tensor address for shape inputs
Failed to bind tensor address for x
Error while setting the output tensor address
The shapes of the inputs: [x] cannot be inferred. This could happen if ...
<the allocator's own error, uncaught, for the output tensors>
<the allocator's own error, uncaught, for the activation workspace>
```

After:

```
Input 0 ("x") of engine _run_on_acc_0_engine must be on a CUDA device, found device cpu
Input 0 ("x") of engine _run_on_acc_0_engine expects type Float, found type int
Error while setting the input shape for input 0 ("x") of engine _run_on_acc_0_engine. TensorRT refused shape [8, 17], the engine declares [8, 16]
Error while setting the tensor address for shape input 0 ("x") of engine _run_on_acc_0_engine
Failed to bind the tensor address for input 0 ("x") of engine _run_on_acc_0_engine
Error while setting the tensor address for output "output0" of engine _run_on_acc_0_engine
The shapes of the inputs: [x] of engine _run_on_acc_0_engine cannot be inferred. This could happen if ...
Failed to allocate output "output0" of engine _run_on_acc_0_engine with shape [8, 16] and dtype Float.
Failed to allocate the 134217728-byte activation workspace of engine _run_on_acc_0_engine.
```

`torch.ops.tensorrt.set_logging_level(4)` does not help, because the `LOG_DEBUG` that names the
binding runs *after* the check that throws. TensorRT does report the shapes on the
`setInputShape` path, but only to its own logger, never into the exception.

# Issues

Closes #4678

## Type of change

# Testing

- [x] adds a new unit test `tests/py/dynamo/runtime/test_engine_failure_diagnostics.py` 
- [x] existing unit tests pass

# Checklist:

### DIFF
--- a/core/runtime/execute_engine.cpp
+++ b/core/runtime/execute_engine.cpp
@@ -106,11 +106,14 @@ void setup_input_tensors(
     const auto& name = binding.name;
 
     TORCHTRT_CHECK(
-        inputs[i].is_cuda(), "Expected input tensors to have device cuda, found device " << inputs[i].device());
+        inputs[i].is_cuda(),
+        "Input " << i << " (\"" << name << "\") of engine " << compiled_engine->name
+                 << " must be on a CUDA device, found device " << inputs[i].device());
 
     TORCHTRT_CHECK(
         inputs[i].dtype() == binding.expected_type,
-        "Expected input tensors to have type " << binding.expected_type << ", found type " << inputs[i].dtype());
+        "Input " << i << " (\"" << name << "\") of engine " << compiled_engine->name << " expects type "
+                 << binding.expected_type << ", found type " << inputs[i].dtype());
 
     auto dims = core::util::toDims(inputs[i].sizes());
     auto shape = core::util::toVec(dims);
@@ -126,7 +129,8 @@ void setup_input_tensors(
       compiled_engine->active_shape_tensor_values.emplace_back(std::move(inputs_cpu_vec));
       TORCHTRT_CHECK(
           ctx->setTensorAddress(name.c_str(), compiled_engine->active_shape_tensor_values.back().data()),
-          "Error while setting the tensor address for shape inputs");
+          "Error while setting the tensor address for shape input " << i << " (\"" << name << "\") of engine "
+                                                                    << compiled_engine->name);
 
       if (cudagraphs_enabled) {
         // @peri044 I dont know if this makes sense since they are supposed to be GPU buffers
@@ -134,7 +138,8 @@ void setup_input_tensors(
       }
       TORCHTRT_CHECK(
           ctx->setTensorAddress(name.c_str(), compiled_engine->active_shape_tensor_values.back().data()),
-          "Error while setting the tensor address for shape inputs");
+          "Error while setting the tensor address for shape input " << i << " (\"" << name << "\") of engine "
+                                                                    << compiled_engine->name);
 
     } else {
       compiled_engine->active_input_tensors[i] = inputs[i].view(shape).contiguous();
@@ -155,7 +160,11 @@ void setup_input_tensors(
         compiled_engine->cudagraph_input_staging_buffers[i] = compiled_engine->active_input_tensors[i].clone();
       }
 
-      TORCHTRT_CHECK(ctx->setInputShape(name.c_str(), dims), "Error while setting the input shape");
+      TORCHTRT_CHECK(
+          ctx->setInputShape(name.c_str(), dims),
+          "Error while setting the input shape for input "
+              << i << " (\"" << name << "\") of engine " << compiled_engine->name << ". TensorRT refused shape " << dims
+              << ", the engine declares " << compiled_engine->cuda_engine->getTensorShape(name.c_str()));
 
       at::Tensor final_input;
       if (cudagraphs_enabled && !is_aliased_input) {
@@ -172,7 +181,10 @@ void setup_input_tensors(
       // empty_tensor_placeholder is pre-allocated in TRTEngine constructor
       void* input_addr = final_input.numel() == 0 ? compiled_engine->empty_tensor_placeholder : final_input.data_ptr();
 
-      TORCHTRT_CHECK(ctx->setTensorAddress(name.c_str(), input_addr), "Failed to bind tensor address for " << name);
+      TORCHTRT_CHECK(
+          ctx->setTensorAddress(name.c_str(), input_addr),
+          "Failed to bind the tensor address for input " << i << " (\"" << name << "\") of engine "
+                                                         << compiled_engine->name);
 
       // Record the bound tensor by binding name so the output-binding loop
       // can resolve aliased outputs to their source input's storage.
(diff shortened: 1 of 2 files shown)


SrivastavaKshitij:
@apbose  :  looks like CI is failing on unrelated issue

apbose (approved):

apbose:
approved. waiting on CI run after rebase for merging

PR #4701, opened 2026-09-10
fix: handle empty one-dimensional tensors in concatenation

shoumikhin:
## Problem

Joining an empty tensor with a larger tensor can force a model to run that operation in PyTorch instead of TensorRT. This happens when a model starts with an empty cache and adds values to it.

PyTorch accepts an empty tensor with shape `(0,)` alongside tensors with more dimensions. TensorRT requires matching numbers of dimensions.

## Change

Skip these empty tensors when checking dimensions and building the TensorRT concatenation. Continue rejecting dimension mismatches between the remaining tensors.

An empty tensor can still affect the result's data type. Choose the output type using every input before removing empty ones. Otherwise, later arithmetic could overflow in float16 when ordinary PyTorch would use float32.

The validator's test inputs include the data type it needs.

## Tests

Passed 10/10 focused tests: four validator tests and six tests that build and run TensorRT engines. They cover empty tensors in different positions, positive and negative axes, mixed input types, and a float64 constant with conversion to float32 enabled. The empty-input validator regression fails on the base branch.

After rebasing, the existing nonzero suite passed 17 cases and skipped its RTX-only case, checking the nearby upstream converter changes.

The focused rerun used Linux x86_64, Python 3.12, PyTorch 2.15 nightly, and TensorRT 11.3 with the Python runtime. Earlier native-runtime coverage used TensorRT 11.2; a native TensorRT 11.3 build was not tested.

Windows, aarch64, TensorRT-RTX, TensorRT 10.x, and concatenation dimensions that vary between calls were not tested.

### DIFF
--- a/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
+++ b/py/torch_tensorrt/dynamo/conversion/aten_ops_converters.py
@@ -1,5 +1,6 @@
 # mypy: disallow-untyped-decorators=False
 
+import functools
 import logging
 import operator
 from typing import (
@@ -19,7 +20,7 @@ import numpy as np
 import torch
 from tensorrt import ITensor as TRTTensor
 from torch.fx.node import Argument, Node, Target
-from torch_tensorrt import ENABLED_FEATURES
+from torch_tensorrt import ENABLED_FEATURES, _enums
 from torch_tensorrt._utils import (
     is_tensorrt_rtx_version_supported,
     is_tensorrt_version_supported,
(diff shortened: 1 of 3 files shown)


shoumikhin:
Thanks @cehongwang for the [reproducer and suggested fix](https://github.com/shoumikhin/TensorRT/pull/3#discussion_r4079530638). I applied `statically_known_true` here so checking for an empty tensor does not force a guard on a data-dependent length.

I also added the `nonzero` regression test. It fails before the fix and passes afterward. The 35 concatenation validator and converter tests passed, and the GPU reproducer matched PyTorch for empty and nonempty outputs. These checks used Linux x86_64, PyTorch 2.15 nightly, TensorRT 11.3, and the Python runtime.

PR #4710, opened 2026-09-10
feat: report operators that fall back to PyTorch

shoumikhin:
## Problem

A model can split into TensorRT engines and PyTorch sections without a concise explanation of which operations stayed in PyTorch. The existing support record also omits refused operations with side effects, such as random number generation.

## Change

Record every refused executable operation in both the fast and global partitioners. Use the counts for the dry-run report and emit one summary at `INFO` from the shared compilation path.

The summary names each operation, its count, and the reasons observed during support checks. Reasons cover explicit device transfers, tensor rank limits, complex data types, data-dependent output shapes, caller exclusions through `torch_executed_ops`, and the lack of a converter that passes validation. A registered converter that rejects a node is not described as missing.

Caller-requested fallback is included. Fully supported models remain silent, and setting the log level to `WARNING` hides the summary. Support decisions do not change.

The summary covers graphs that reach partitioning. Small graphs that return earlier keep their existing skipped-compilation message. It does not enumerate supported operations left in PyTorch because a partition is too small.

## Tests

Passed 10 support tests and 12 compilation/logging tests. They cover both partitioners, refusal reasons, caller exclusions by name and target, mixed fallback counts, INFO severity, WARNING-level suppression, and silence for fully supported graphs. Four logging controls fail on the base branch because the summary is absent. The first commit's seven tests also passed independently.

The reporting tests compile models and inspect logs; they do not execute the returned models. Four existing rank-boundary tests separately passed compiled execution and comparisons with PyTorch.

The rerun used Linux x86_64, CUDA, Python 3.12, PyTorch 2.15 nightly, and TensorRT 11.3 with the Python runtime. Earlier coverage used TensorRT 11.2 with the native runtime. A native TensorRT 11.3 build, Windows, aarch64, and TensorRT-RTX were not tested. Formatting, import-order, and lint checks passed.

### DIFF
--- a/py/torch_tensorrt/dynamo/_compiler.py
+++ b/py/torch_tensorrt/dynamo/_compiler.py
@@ -37,6 +37,9 @@ from torch_tensorrt.dynamo.conversion import (
 from torch_tensorrt.dynamo.conversion._ConverterRegistry import (
     DYNAMO_CONVERTERS as CONVERTERS,
 )
+from torch_tensorrt.dynamo.conversion._ConverterRegistry import (
+    ConverterRegistry,
+)
 from torch_tensorrt.dynamo.debug._DebuggerConfig import DebuggerConfig
 from torch_tensorrt.dynamo.debug._supports_debugger import fn_supports_debugger
 from torch_tensorrt.dynamo.lowering import (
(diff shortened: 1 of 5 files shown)


py/torch_tensorrt/dynamo/partitioning/_global_partitioner.py
```diff
@@ -145,6 +146,9 @@ def __init__(
         # Initialize sets of supported/unsupported operators
         self.supported_operators: Dict[str, int] = {}
         self.unsupported_operators: Dict[str, int] = {}
+        # unsupported_operators skips impure nodes, so it cannot answer "did anything
+        # fall back". This one records every refusal.
+        self.fallback_operators: Dict[str, int] = {}
```
narendasan, line 150: Might be nice to include some metadata about reasons for fallback?

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1390,7 +1393,31 @@ def preserve_module_specs(
...
         )
 
-    dryrun_tracker.unsupported_ops = supported_ops.unsupported_operators
+    dryrun_tracker.unsupported_ops = supported_ops.fallback_operators
+
+    # Operators the caller named in torch_executed_ops are left out: that fallback was
+    # asked for. The set can hold either a qualified name string or an operator target
+    # object, and fallback_operators is keyed by name, so normalize to names first or a
+    # target object never matches and the caller is warned about their own choice.
+    excluded_names = {
```
narendasan: I think its fine to report both if we include reasons

py/torch_tensorrt/dynamo/_compiler.py
```diff
@@ -1390,7 +1393,31 @@ def preserve_module_specs(
...
+        node_name: count
+        for node_name, count in supported_ops.fallback_operators.items()
+        if node_name not in excluded_names
+    }
+    if reported_fallbacks:
+        named = ", ".join(
+            f"{node_name} + Operator Count: {count}"
+            for node_name, count in sorted(reported_fallbacks.items())
+        )
+        logger.warning(
```
narendasan: Should this be warning since its expected behavior? Info might be better so it doesnt need to be suppressed

PR #4728, opened 2026-09-22
Unblock the nightly ExecuTorch pin bump

shoumikhin:
The job that moves the ExecuTorch pin forward each night has never once landed a bump. It fails
while working out which packages to install, so the pin sits still while everything else moves
on. Two separate things stop it. Both are fixed here.

## The version is written down twice

The optional install group asks for a range of ExecuTorch versions, built from the pin. Beside
it, the group asks for the companion runtime wheel, and that wheel asks for the one exact
ExecuTorch build it was compiled against.

On an ordinary day the two agree. On the day the pin moves they cannot. The range asks for that
day's build, while every companion published so far still asks for the one before it. Nothing
satisfies both, so the whole thing is rejected, and every failed night adds another day to the
gap.

The companion is now the only place the version is written down. Anyone setting up the project
gets an ExecuTorch and a companion that were built to work together, instead of a pairing that
has never existed.

## The platform list is wider than the platforms we support

One list says any Windows. The list below it says Windows on one processor family. The wheels
everything here depends on are built for that one family, so trying to solve for any other
Windows machine finds nothing to install, and that comes out as a missing wheel rather than as
an unsupported platform. The two lists now agree.

## Nothing gets looser

The companion already turns down a mismatch from its own side, when it is imported, by loading
the delegate and checking that it registers itself. Its own notes say its ExecuTorch lands on an
older date on purpose, so the pairing was never resting on the range.

Two checks move with the first change. They used to require every published install group to
name the range. They now require them to name the companion instead.

## Testing

I reproduced the nightly failure on Linux, moving the pin forward and resolving the way the job
does, and confirmed both reported causes. With these changes the same step succeeds, records the
companion together with the ExecuTorch build it asks for, and leaves the pin file holding the
new version. Putting the range back makes the moved checks fail, so they are doing their job.


lanluo-nvidia (approved):
lgtm

